PR #10, opened 2021-03-24
Adding Metrics

MikeNeilson:
Begining the process of adding metrics. Currently functions for OfficeController. will expand to more with annotations, time permitting.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,3 @@
 {
-    "java.configuration.updateBuildConfiguration": "interactive"
+    "java.configuration.updateBuildConfiguration": "automatic"
 }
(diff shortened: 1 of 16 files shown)


rripken (dismissed):
Looks good to me.  I could never figure out a great use case for the health check registry because everything was either all-up or all-down but maybe this time it will make more sense.

MikeNeilson:
> 
> 
> Looks good to me. I could never figure out a great use case for the health check registry because everything was either all-up or all-down but maybe this time it will make more sense.

Yeah, mostly it's there because the metrics servlet we want breaks without it, but perhaps it will become useful over time.

rripken (approved):

PR #12, opened 2021-03-31
Feature/add standalone

MikeNeilson:
Starting work on making a version that will run standalone

- [x] can run standalone and present data
- [x] swagger-ui works
- [x] docker image
- [x] docker image can push to the appropriate place

I'll work on the metrics after we know it will push to the registry.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Create Release
 on:   
   push:
     branches: 
-      - main
+      - develop
 
 jobs: 
   build:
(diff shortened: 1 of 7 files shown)


rripken (dismissed):

rripken (dismissed):

MikeNeilson:
@brettpalmberg Let me know if I missed anything in the .github/workflows/release.yml, pretty sure it's right but nice to get confirmation that it's probably right before merging.

.github/workflows/release.yml
```diff
@@ -46,3 +46,42 @@ jobs:
...
+      - name: Build, tag, and push image to Amazon ECR (midas-api)
+        if: ${{ success() }}
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: cwms-radar-api
+          IMAGE_TAG: ${{steps.get_version.outputs.VERSION}}
+        run: |
+          ./gradlew prepareDocker
+          cd cwms_radar_standlone/build/docker
+          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG -t $ECR_REGISTRY/$ECR_REPOSITORY:latest .
```
brettpalmberg, line 91: Hi @MikeNeilson  - I think this would probably work, but I typically use a slightly different approach.  I believe as-written, this builds the `.jar` in the actions worker environment (outside of the selected docker container) then copies the resulting `.jar` into the docker container.  Recommend running the build inside a container and using [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) to only copy the `.jar` of interest into the final image and keep image size small.

MikeNeilson: The base image used is the openjdk11 and an uber jar that's 33MB The docker build is only going to add the single 33 MB jar. So while I don't disagree with doing the multi-stage in general, I don't think it's going to do much to help shrink anything. Though if proven wrong we can certainly change it later.

Thanks for the check though. Once someone approves it I'll merge it in and we'll see what happens.

MikeNeilson: Make that adoptopenjdk:11-jre-hotspot, if I'm going to identify a docker image I should identify the one I'm actually using.

brettpalmberg: >  I don't think it's going to do much to help shrink anything.

Agreed.  Main reason for mentioning it was because you could fully containerize the build process and build environment (and ensure build/deploy flavors match exactly).  This would also support identical build process with and without Github Actions as the build tool and eliminate requirement for `jdk` be installed on the build host.

rripken (dismissed):

rripken (approved):

PR #20, opened 2021-05-17
Spelling and file cleanup

DanielTOsborne:
Some mostly small cleanup.
Some spelling corrections.
Renamed gradle example file to avoid inadvertent editing of a versioned file.
Added datetime format to swagger docs.


MikeNeilson (dismissed):
I new I had made some typos but I think realize it was quite that bad. Thanks!

gradle.properties.example
```diff
@@ -1,7 +1,8 @@
 #The following properties are required for the ./gradlew run task to work
-# The actual values should be set in your personal ~/gradle.properties file
+# The actual values should be set in your personal ~/.gradle/gradle.properties file
 # These are presented for example
```
MikeNeilson, line 3: And I should've put here, and so the build can run.

MikeNeilson (approved):
Thanks for the button up and making the build a little more obvious about what it needs.

PR #21, opened 2021-05-17
Catalog

### LINKED ISSUE #22: Implement a Catalog endpoint
it would be good to have a standard endpoint for getting information about what information the system holds.

That said there are a few approaches

1. Get everything by location (including different data types (e.g. timeseries, level, rating) )
2. Get everything by type and have multiple urls required.

Pretty sure number 2 is the simplest.

MikeNeilson:
Still needs work, the two endpoints don't have nearly enough detail and I also need to add Levels and Ratings, but it's at the point where it works and people can comment on the design.

- [x] pagination - went with cursor based for this one.
- [ ] timeseries
- [ ] locations
- [ ] ratings
- [ ] levels
- [ ] catalogs?


.github/workflows/build.yml
```diff
@@ -19,6 +23,112 @@ jobs:
           java-version: '11'
           java-package: jdk             
       - name: build and test     
-        run: ./gradlew test --init-script init.gradle
+        id: thebuild
```
rripken, line 26: "thebuild" seems like a weird name.  I'm still new to gradle so I don't have any suggestions of what would be better.

rripken (dismissed):
Its alot of files but the actual changes aren't that many.

.github/workflows/build.yml
MikeNeilson: that id is the github action name, not the gradle task name. It's certainly awkward but I needed a specific name for the later step were it uses the environment variable this task outputs... I just couldn't think of a better one.

rripken (dismissed):

rripken (approved):

PR #25, opened 2021-05-25
Accept header cleanup

DanielTOsborne:
Cleanup content-type typos, and reference constants where possible.
Few spelling error corrections too.

Some whitespace changes seem to have snuck in, but I'm ok with them.

Note, this changes the wml2 format from application/xml to application/vnd.opengis.waterml+xml.
I believe that's how it should be, but if you don't want to change that, I'll reverse that.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
@@ -1,6 +1,7 @@
 package cwms.radar.api;
 
 import java.sql.SQLException;
+import java.text.Format;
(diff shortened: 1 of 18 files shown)


MikeNeilson (approved):
Thanks!
I hope your editor did most of the formatting work... otherwise, yikes.

DanielTOsborne:
> Thanks!
> I hope your editor did most of the formatting work... otherwise, yikes.

Yeah, this time I didn't intend to make those edits.
However it'll be a good idea to add a linter or something in the future so the editor can enforce a consistent style.

PR #27, opened 2021-06-09
Adding initial group and category support for locations and timeseries.

rripken:
Are the JsonV1 changes approx what you envisioned for how new objects would get added?

### DIFF
--- a/cwms_radar_api/build.gradle
+++ b/cwms_radar_api/build.gradle
@@ -12,14 +12,19 @@ dependencies {
         exclude group: "com.oracle", module: "*"
     }
     implementation "com.googlecode.owasp-java-html-sanitizer:owasp-java-html-sanitizer:$OWASP_SANITIZER_VERSION"
-    
-    compileOnly "io.javalin:javalin:$JAVALIN_VERSION"
+
+    implementation "io.javalin:javalin:$JAVALIN_VERSION"
+
     compileOnly "io.swagger.core.v3:swagger-core:$SWAGGER_CORE_VERSION"
     compileOnly "javax:javaee-web-api:$JAVA_EE_WEB_VERSION"
     compileOnly "io.dropwizard.metrics:metrics-core:$dropwizard_version"
     
     compileOnly project(":annotations")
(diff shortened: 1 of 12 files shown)


cwms_radar_api/build.gradle
```diff
@@ -12,14 +12,19 @@ dependencies {
         exclude group: "com.oracle", module: "*"
     }
     implementation "com.googlecode.owasp-java-html-sanitizer:owasp-java-html-sanitizer:$OWASP_SANITIZER_VERSION"
-    
-    compileOnly "io.javalin:javalin:$JAVALIN_VERSION"
+
+    implementation "io.javalin:javalin:$JAVALIN_VERSION"
```
MikeNeilson, line 16: This may need to say compileOnly here and be added to the _standalone build.gradle. There's a bunch of exclusions required since Javalin by default embeds Jetty. But we can worry about it later.

MikeNeilson:
> 
> 
> Are the JsonV1 changes approx what you envisioned for how new objects would get added?

Yes, and it looks good.

cwms_radar_api/src/main/java/cwms/radar/formatters/JsonV1.java
```diff
@@ -4,37 +4,137 @@
...
-        if( daoList.get(0) instanceof Office ){
-            OfficeFormatV1 fmtv1 = new OfficeFormatV1(
-                daoList.stream().map( office -> (Office)office ).collect(Collectors.toList())
-            );
-            return JavalinJson.toJson(fmtv1);
-        } else {
-			throw new BadRequestResponse(String.format("Format {} not implemented for this data type"));
+	public String format(List<? extends CwmsDTO> daoList)
+	{
+		Object wrapped = buildFormatting(daoList);
```
MikeNeilson, line 59: I like this change, the whole things flows better when reading it. Also a good idea not to overwrite the Javalin root level ObjectMapper since we may have other format choices in the future.

MikeNeilson (approved):

PR #28, opened 2021-06-11
Assigned location

rripken:
Expanded jooq usage, added assigned location info when a single LocationGroup is returned.  Dropped format query arg on new controllers


MikeNeilson (approved):
Looks good to me. However, the lines of code for formats you've commented out, in the feature feel free to just delete them. I'm pretty sure we'll focus on JSON and XML.

PR #33, opened 2021-07-07
Refactored many of the CwmsDataManager methods into individual Dao cl…

rripken:
…asses.


MikeNeilson (approved):
I like it. I was going to go in a slightly different direction that still funneled things through the CwmsDataManager but after reading, this does look cleaning.... though now I have to go rework the catalog... again (note: that was positive sarcasm.)

PR #34, opened 2021-07-08
Feature/clob swagger improvements

MikeNeilson:
Updated the openapi elements to include our or clob wrapper class in the example.

Added pagination to getAll.
Added an includedValues parameter to getAll
Added regular expression search support to getAll

Started adding clobs to the catalog, still deciding if needed

Still working out some minor details, like the total and next-page aren't showing up, but wanted to share now.


rripken (dismissed):

MikeNeilson:
I'm not actually opposed to the clob/like/:query instead of clob/?like=query so let me know y'alls thoughts on the topic.

rripken (approved):
I like the like as a query param instead of its own controller.  Makes the getAll method make alot more sense, especially with includeValues

MikeNeilson:
> 
> 
> I like the like as a query param instead of its own controller. Makes the getAll method make alot more sense, especially with includeValues

Thanks, I thought that made sense but one never really knows.

PR #37, opened 2021-07-20
Cleanup/exception handling

MikeNeilson:
Trying to organize some of the error handling. So far I've created a single json response object to return on all errors, which includes a randomly generated id (I intend to make it hex as some point) and started going through where some of the errors are handled.

Feed back wanted.

@DanielTOsborne @rripken

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
@@ -14,6 +14,7 @@ import cwms.radar.data.dao.ClobDao;
 import cwms.radar.data.dao.JooqDao;
 import cwms.radar.data.dto.Clob;
 import cwms.radar.data.dto.Clobs;
+import cwms.radar.data.dto.CwmsDTOPaginated;
 import cwms.radar.formatters.ContentType;
 import cwms.radar.formatters.Formats;
 import io.javalin.apibuilder.CrudHandler;
@@ -103,7 +104,9 @@ public class ClobController implements CrudHandler {
             String office = ctx.queryParam("office");
             Optional<String> officeOpt = Optional.ofNullable(office);
 
-            String cursor = ctx.queryParam("cursor",String.class,ctx.queryParam("page",String.class,"").getValue()).getValue();
+            String cursor = ctx.queryParam("cursor",String.class,ctx.queryParam("page",String.class,"")
+                .check(CwmsDTOPaginated.CURSOR_CHECK, "Invalid Cursor")
+            .getValue()).getValue();
             int pageSize = ctx.queryParam("pageSize",Integer.class,ctx.queryParam("pagesize",String.class,Integer.toString(defaultPageSize)).getValue()).getValue();
 
             boolean includeValues = ctx.queryParam("includeValues",Boolean.class,"false").getValue().booleanValue();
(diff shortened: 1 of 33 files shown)


cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
```diff
@@ -124,14 +125,23 @@ public void getAll(Context ctx) {
...
-                ctx.result("Invalid Format Options");
+                re = new RadarError("failed to format data", new HashMap<String,String>(){
+                    {
+                        put("cause","Invalid format or format options");
+                    }
+                });
+                ctx.status(HttpServletResponse.SC_NOT_IMPLEMENTED);
+
             }
+            logger.log(Level.SEVERE,String.format("%s: failed to format data",re.getIncidentIdentifier()),fe);
```
DanielTOsborne: Maybe just get the message from the RadarError ctor instead?
Or even just log it inside the ctor instead of a separate statement? Though, if there's a reason to construct a RadarError class that doesn't also need to be logged, then that might not work.

cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
```diff
@@ -89,9 +92,7 @@ protected DSLContext getDslContext(Context ctx)
                                            @OpenApiContent( type = Formats.JSONV2, from = Clobs.class ),
                                            @OpenApiContent( type = Formats.XMLV2, from = Clobs.class )
                                        }
-                      ),
-                      @OpenApiResponse(status="501",description = "The format requested is not implemented"),
```
DanielTOsborne, line 93: Since we are still returning non-200 status codes, should they still be listed here?
HttpServletResponse.SC_NOT_IMPLEMENTED
See: L140 below

cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
MikeNeilson: I think I just missed tweaking that one. I made all of the other instances re.toString() for the log message. And put the formatting of the incident ID and message in that.

cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
MikeNeilson: If you look in RadarAPI.java you'll notice that the codes+format returned by everything are listed in the default confiugration. I just happened to miss this instance.

MikeNeilson: In theory we should be able to do the same for the office, but I didn't want to work that into this PR.

cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
DanielTOsborne: > I think I just missed tweaking that one. I made all of the other instances re.toString() for the log message. And put the formatting of the incident ID and message in that.

Oh, in that case, there's a few more below you missed.

rripken (dismissed):

cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
rripken: Looks like an improvement to me.

DanielTOsborne:
I'm happy to get consistent error handling. 👍

cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
MikeNeilson: Should've looked at the whole code not the one line in the snippet. That was an expirement with adding specific details.... hence the weird log call.

MikeNeilson: nevermind, I apparently can't read after 3pm. the single line was the relevant piece to the comment.

rripken (approved):

PR #38, opened 2021-07-20
Feature/pools controller

rripken:
Initial implementation of the pools controller and dao.  I don't love the way pagination works but maybe we can get someone to revisit the pl/sql for that and provide better procedures


cwms_radar_api/src/main/java/cwms/radar/data/dao/RecentValueDao.java
```diff
@@ -0,0 +1,257 @@
...
+import usace.cwms.db.jooq.codegen.tables.AV_CWMS_TS_ID2;
+import usace.cwms.db.jooq.codegen.tables.AV_TSV;
+import usace.cwms.db.jooq.codegen.tables.AV_TSV_DQU;
+
+import static org.jooq.impl.DSL.asterisk;
+import static org.jooq.impl.DSL.field;
+import static org.jooq.impl.DSL.max;
+import static org.jooq.impl.DSL.partitionBy;
+
+public class RecentValueDao extends JooqDao<TsvDqu>
```
MikeNeilson: This is clearly just timeseries operations, albeit slightly specific ones, does it really make sense to have a separate DAO or should this all just be lumped into the/a TimeSeriesDAO? (I'm not actually sure, just trying to start a discussion.)

cwms_radar_api/src/main/java/cwms/radar/data/dto/Pool.java
```diff
@@ -0,0 +1,123 @@
...
+	{
+		super(p.getPoolName(), p.getProjectId(), p.getBottomLevelId(), p.getTopLevelId(), p.isImplicit());
+	}
+
+	public Number getAttribute()
+	{
+		return attribute;
+	}
+
+	public void setAttribute(Number attribute)
```
MikeNeilson: The DTOs should be read only, use the builder pattern if everything can't be handled in the constructor.

cwms_radar_api/src/main/java/cwms/radar/data/dto/Tsv.java
```diff
@@ -0,0 +1,88 @@
+package cwms.radar.data.dto;
+
+import java.util.Date;
+import javax.persistence.EmbeddedId;
```
MikeNeilson: What led to requiring these imports?

MikeNeilson (dismissed):

cwms_radar_api/src/main/java/cwms/radar/data/dao/RecentValueDao.java
rripken: I just pulled the TimeSeries methods out of CwmsDataManager and combined them with these methods into a TimeSeriesDao object.   These queries were ported over from my cwms mobile queries and I'm not 100% sure I want to expose TsvDqu and Tsv objects but at the moment the recent methods aren't access by the controllers

MikeNeilson: I would say that at the least, that shouldn't be the name. But couldn't disagree with the concept enough to say "no don't do that."

cwms_radar_api/src/main/java/cwms/radar/data/dto/Pool.java
rripken: ok. I reworked Pool and Pools

cwms_radar_api/src/main/java/cwms/radar/data/dto/Tsv.java
rripken: CWMSMobile used similar classes with hibernate - I should have removed them.

cwms_radar_api/src/main/java/cwms/radar/data/dto/Pool.java
```diff
@@ -8,51 +8,33 @@
...
-	}
-
-	public Pool(PoolType p)
-	{
-		super(p.getPoolName(), p.getProjectId(), p.getBottomLevelId(), p.getTopLevelId(), p.isImplicit());
+	private final Number attribute;
+	private final String description;
+	private final String clobText;
+
+	public Pool(Builder b){
```
MikeNeilson: Should this be private?

MikeNeilson (dismissed):
Looks good to me.

MikeNeilson (approved):

PR #41, opened 2021-08-13
Implementing two versions of the Recent value - one takes a list of t…

rripken:
…sIds and the other takes a ts_group and ts_category.

### DIFF
--- a/cwms_radar_standalone/src/main/java/mil/army/usace/hec/RadarAPI.java
+++ b/cwms_radar_standalone/src/main/java/mil/army/usace/hec/RadarAPI.java
@@ -45,6 +45,7 @@ import org.owasp.html.HtmlPolicyBuilder;
 import org.owasp.html.PolicyFactory;
 
 import static io.javalin.apibuilder.ApiBuilder.crud;
+import static io.javalin.apibuilder.ApiBuilder.get;
 
 
 public class RadarAPI {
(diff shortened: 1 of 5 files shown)


cwms_radar_standalone/src/main/java/mil/army/usace/hec/RadarAPI.java
```diff
@@ -77,7 +78,7 @@ public static void main(String[] args){
     public RadarAPI(javax.sql.DataSource ds, int port){
         this.port = port;
         PolicyFactory sanitizer = new HtmlPolicyBuilder().disallowElements("<script>").toFactory();
-        JavalinValidation.register(UnitSystem.class, v -> UnitSystem.systemFor(v) );
+        JavalinValidation.register(UnitSystem.class, UnitSystem::systemFor);
```
MikeNeilson, line 82: Excellent cleanup. I need to stop forgetting about this particular syntax.

MikeNeilson (dismissed):

MikeNeilson (approved):

cwms_radar_standalone/src/main/java/mil/army/usace/hec/RadarAPI.java
rripken: I mostly changed it just so sonarqube stops reminding me about it.  I'm still getting used to the method reference syntax - I wouldn't say its my go-to pattern.

PR #44, opened 2021-08-26
Feature/cwms db basin connectivity

rripken:


cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
```diff
@@ -0,0 +1,189 @@
...
+        {
+            LOGGER.log(Level.SEVERE, "Error retrieving all basins", ex);
+        }
+    }
+
+    @OpenApi(
+            queryParams = {
+                    @OpenApiParam(name="office", required=false, description="Specifies the owning office of the basin whose data is to be included in the response. If this field is not specified, matching basin information from all offices shall be returned."),
+                    @OpenApiParam(name="unit", required=false, description="Specifies the unit or unit system of the response. Valid values for the unit field are:\r\n 1. EN.   Specifies English unit system. Basin values will be in the default English units for their parameters. (This is default if no value is entered)\r\n2. SI.   Specifies the SI unit system. Basin values will be in the default SI units for their parameters."),
+                    @OpenApiParam(name="format", required=false, description="Specifies the encoding format of the response. Valid values for the format field for this URI are: \r\n 1. json (returns a PG-JSON graph. This is default if no value is entered)")
```
MikeNeilson: Don't put a format query parameter in for new endpoints. use the ACCEPT header value.
The content on the swagger-ui is automatically populated from the api responses.

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
```diff
@@ -0,0 +1,189 @@
...
+            },
+            responses = {
+                    @OpenApiResponse(status="200",
+                            content = {
+                                    @OpenApiContent(type = Formats.JSON)
```
MikeNeilson: This should include a DTO object to provide a reference to users and code generators.

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
```diff
@@ -0,0 +1,189 @@
...
+        _getOneRequestTime = _metrics.timer(name(className, "getOne", "time"));
+        _requestResultSize = _metrics.histogram((name(className,"results","size")));
+    }
+    @OpenApi(
+            queryParams = {
+                    @OpenApiParam(name="office", required=false, description="Specifies the owning office of the basin whose data is to be included in the response. If this field is not specified, matching basin information from all offices shall be returned."),
+                    @OpenApiParam(name="unit", required=false, description="Specifies the unit or unit system of the response. Valid values for the unit field are:\r\n 1. EN.   Specifies English unit system. Basin values will be in the default English units for their parameters. (This is default if no value is entered)\r\n2. SI.   Specifies the SI unit system. Basin values will be in the default SI units for their parameters."),
+                    @OpenApiParam(name="format", required=false, description="Specifies the encoding format of the response. Valid values for the format field for this URI are:\r\n 1. json (returns a PG-JSON graph. This is default if no value is entered)")
```
MikeNeilson: see comment down in getOne

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
```diff
@@ -0,0 +1,189 @@
...
+            },
+            responses = {
+                    @OpenApiResponse(status="200",
+                            content = {
+                                    @OpenApiContent(type = Formats.JSON)
```
MikeNeilson: see comment down in getOne

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityGraphBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+
+import cwms.radar.api.graph.basinconnectivity.edges.BasinConnectivityEdge;
+import cwms.radar.api.graph.basinconnectivity.nodes.BasinConnectivityNode;
+import cwms.radar.data.dto.basinconnectivity.Basin;
+import cwms.radar.data.dto.basinconnectivity.Stream;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class BasinConnectivityGraphBuilder
```
MikeNeilson: why not make this an inner class Builder of the main propertyGraph?

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
```diff
@@ -0,0 +1,55 @@
...
+    public Basin getBasin(String basinId, String officeId) throws SQLException
+    {
+        CwmsDbBasinJooq basinJooq = new CwmsDbBasinJooq();
+        String[] pParentBasinId = new String[1];
+        Double[] pSortOrder = new Double[1];
+        String[] pPrimaryStreamId = new String[1];
+        Double[] pTotalDrainageArea = new Double[1];
+        Double[] pContributingDrainageArea = new Double[1];
+        String areaUnit = "km2";
+        Connection c = dsl.configuration().connectionProvider().acquire();
```
rripken: This looks like a way to get a connection from the dsl but it looks like you're missing the release call.  I think going thru dsl.connection(connection -> {...}) is cleaner and frees us from worrying about acquire/release and I'd prefer that way unless there is some downside I'm not seeing.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/nodes/EmptyStreamNode.java
```diff
@@ -0,0 +1,25 @@
+package cwms.radar.api.graph.basinconnectivity.nodes;
+
+import cwms.radar.api.graph.basinconnectivity.nodes.BasinConnectivityNode;
+
+public class EmptyStreamNode extends BasinConnectivityNode//probably needs a better name
```
MikeNeilson: Can you describe the issue concerning why it needs a better name?

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Basin.java
```diff
@@ -0,0 +1,53 @@
...
+        return retval;
+    }
+
+    private JsonObject getDefaultPGJSON()
+    {
+        return Json.createObjectBuilder()
+                .add("nodes", Json.createArrayBuilder())
+                .add("edges", Json.createArrayBuilder())
+                .build();
+    }
```
rripken, line 59: I don't love the javax.json stuff creeping into the DTO object.  I was hoping we could force all the javax.json to happen in the PgJsonFormatter

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamDao.java
```diff
@@ -0,0 +1,128 @@
...
+    public StreamDao(DSLContext dsl)
+    {
+        super(dsl);
+    }
+
+    public Stream getStream(String streamId, String unitSystem, String officeId) throws SQLException
+    {
+        String pStationUnit = "km";
+        String stationUnitOut = "km";
+        if(unitSystem.equals("EN"))
```
MikeNeilson: this would look clear as `String stationUnitOut = "EN".equals(unitSystem) ? "mi" : "km";`

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamDao.java
```diff
@@ -0,0 +1,128 @@
...
+        {
+            stationUnitOut = "mi";
+        }
+        String pOfficeId = null;
+        if (officeId != null)
```
MikeNeilson: same as above comment, would look cleaner with ternary operator.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Stream.java
```diff
@@ -0,0 +1,376 @@
...
+        }
+    }
+
+    private void addEdgesToJSON(List<? extends BasinConnectivityEdge> edges, JsonArrayBuilder builder)
+    {
+        for(BasinConnectivityEdge edge : edges)
+        {
+            builder.add(edge.toPGJSON());
+        }
+    }
```
rripken: I was hoping all the javax.json could be in the PgFormatter.  Maybe that's not possible.  Before you rewrite anything lets see what Mike thinks.

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
```diff
@@ -0,0 +1,111 @@
...
+    {
+        super(dsl);
+    }
+
+    public List<Basin> getAllBasins(String unitSystem, String officeId) throws SQLException
+    {
+        CwmsDbBasinJooq basinJooq = new CwmsDbBasinJooq();
+        String areaUnitIn = "km2";
+        String areaUnitOut = "km2";
+        if(unitSystem.equals("EN"))
```
MikeNeilson: use the ternary operator. also when comparing input string put the constant string first to avoid null pointer issues.
e.g. ` = "EN".equals(unitSystem) ? "mi2" : "km2"`

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
```diff
@@ -0,0 +1,111 @@
...
+        {
+            areaUnitOut = "mi2";
+        }
+        Connection c = dsl.configuration().connectionProvider().acquire();
```
MikeNeilson: @rripken had a comment in another section, use the lamba style if you need a raw connection to avoid forgetting to release the connection.

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
```diff
@@ -0,0 +1,111 @@
...
+    {
+        CwmsDbBasinJooq basinJooq = new CwmsDbBasinJooq();
+        String[] pParentBasinId = new String[1];
+        Double[] pSortOrder = new Double[1];
+        String[] pPrimaryStreamId = new String[1];
+        Double[] pTotalDrainageArea = new Double[1];
+        Double[] pContributingDrainageArea = new Double[1];
+        String areaUnitIn = "km2";
+        String areaUnitOut = "km2";
+        if(unitSystem.equals("EN"))
```
MikeNeilson: switch to ternary and switch order or equals.

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamDao.java
```diff
@@ -0,0 +1,112 @@
...
+    public Stream getStream(String streamId, String officeId) throws SQLException
+    {
+        String pStationUnit = "KM";
+        String pOfficeId = null;
+        if (officeId != null)
+        {
+            pOfficeId = officeId;
+        }
+        CwmsDbStreamJooq streamJooq = new CwmsDbStreamJooq();
+        Connection c = dsl.configuration().connectionProvider().acquire();
```
rripken: I think it would be cleaner to go through dsl.connection(c->{...});

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
```diff
@@ -0,0 +1,111 @@
...
+        {
+            areaUnitOut = "mi2";
+        }
+        Connection c = dsl.configuration().connectionProvider().acquire();
```
MikeNeilson: switch to the lamba style to avoid any issues from not releasing the connection.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Stream.java
```diff
@@ -0,0 +1,376 @@
...
+        _confluenceBank = confluenceBank;
+        _diversionBank = diversionBank;
+        _streamLength = streamLength;
+        _confluenceStation = confluenceStation;
+        _diversionStation = diversionStation;
+        _streamNodes = new ArrayList<>(streamLocations);
+        _tributaries = new ArrayList<>(tributaries);
+        _streamReaches = new ArrayList<>(reachesOnStream);
+        addEmptyNodes(); //adds terminal nodes and tributary junction nodes to _streamNodes, then sorts by station
+    }
```
rripken: This is a massive ctor.  You've converted others to Builder pattern.  I think that might be a good idea here too.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Basin.java
rma-bryson: My latest Basin.java class doesn't look like this. Maybe this didn't pull in the latest?

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamDao.java
```diff
@@ -0,0 +1,128 @@
...
+    {
+        StreamReachDao streamReachDao = new StreamReachDao(dsl);
+        return streamReachDao.getReachesOnStream(streamId, officeId);
+    }
+
+    private Set<Stream> getTributaries(String streamId, String unitSystem, String stationUnitIn, String stationUnitOut, String officeId)  throws SQLException
+    {
+        CwmsDbStreamJooq streamJooq = new CwmsDbStreamJooq();
+        Connection c = dsl.configuration().connectionProvider().acquire();
+        ResultSet rs = streamJooq.catStreams(c, null, "km", null, streamId, null, null, null, null, null, null, null, null, null, null, null, null, officeId);
```
MikeNeilson: "km" should probably be brought out into a constant.

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamReachDao.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public Set<StreamReach> getReachesOnStream(String streamId, String officeId) throws SQLException
+    {
+        String pStationUnitIn = "km";
+        String pOfficeIdMaskIn = null;
+        if (officeId != null)
+        {
+            pOfficeIdMaskIn = officeId;
+        }
+        Connection c = dsl.configuration().connectionProvider().acquire();
```
MikeNeilson: use the appropriate lamba syntax to wrap acquire/release

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamReachDao.java
```diff
@@ -0,0 +1,62 @@
...
+    public StreamReachDao(DSLContext dsl)
+    {
+        super(dsl);
+    }
+
+    public Set<StreamReach> getReachesOnStream(String streamId, String officeId) throws SQLException
+    {
+        String pStationUnitIn = "km";
+        String pOfficeIdMaskIn = null;
+        if (officeId != null)
```
MikeNeilson: use the ternary syntax

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamReachDao.java
```diff
@@ -0,0 +1,62 @@
...
+        return buildReachesFromResultSet(rs);
+    }
+
+    private Set<StreamReach> buildReachesFromResultSet(ResultSet rs) throws SQLException
+    {
+        Set<StreamReach> retVal = new HashSet<>();
+
+        while(rs.next())
+        {
+            String reachId = rs.getString(4);
```
MikeNeilson: Use named columns instead of column index if using result sets directly

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamReachDao.java
```diff
@@ -0,0 +1,62 @@
...
+    {
+        String pStationUnitIn = "km";
+        String pOfficeIdMaskIn = null;
+        if (officeId != null)
+        {
+            pOfficeIdMaskIn = officeId;
+        }
+        Connection c = dsl.configuration().connectionProvider().acquire();
+        CwmsDbStreamJooq streamJooq = new CwmsDbStreamJooq();
+        ResultSet rs = streamJooq.catStreamReaches(c, streamId, null, null, null, pStationUnitIn, pOfficeIdMaskIn);
```
MikeNeilson: consider using jooq to return a well named set of data instead of raw sql.

cwms_radar_api/src/main/java/cwms/radar/data/dto/PgJsonDTO.java
```diff
@@ -0,0 +1,8 @@
+package cwms.radar.data.dto;
+
+import cwms.radar.api.graph.pgjson.PgJsonGraph;
+
+public interface PgJsonDTO extends CwmsDTO
```
MikeNeilson: This appears to be an unnecessary step. the idea behind having all the formatters was that we extract out of the database in into one struct and than adjust as necessary within the formatters.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Basin.java
```diff
@@ -0,0 +1,68 @@
+package cwms.radar.data.dto.basinconnectivity;
+
+import cwms.radar.api.graph.basinconnectivity.BasinConnectivityGraphBuilder;
+import cwms.radar.api.graph.pgjson.PgJsonGraph;
+import cwms.radar.data.dto.PgJsonDTO;
+
+public final class Basin implements PgJsonDTO
+{
+    private final String _basinId;
```
MikeNeilson: What's with the underscores? except for perhaps sort order these appear to be values that should show up in the output to the user.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Basin.java
```diff
@@ -0,0 +1,68 @@
...
+    private final String _officeId;
+    private final Stream _primaryStream;
+    private final Double _sortOrder;
+    private final Double _basinArea;
+    private final Double _contributingArea;
+    private final String _parentBasinId;
+
+    Basin(BasinBuilder basinBuilder)
```
MikeNeilson: this should be private, and the builder a nested class. This forces the user to call Builder.build() to get an actual instance instead of allowing it to be created in an unknown state.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/BasinBuilder.java
```diff
@@ -0,0 +1,101 @@
+package cwms.radar.data.dto.basinconnectivity;
+
+import cwms.radar.data.dto.CwmsDTO;
+import cwms.radar.data.dto.Office;
+import cwms.radar.data.dto.basinconnectivity.buildercontracts.Build;
+
+public class BasinBuilder implements Build<Basin>
```
MikeNeilson: This should be a nested class of Basin. And thus doesn't need any getters. But otherwise looks reasonable.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Stream.java
```diff
@@ -0,0 +1,120 @@
+package cwms.radar.data.dto.basinconnectivity;
+
+import cwms.radar.data.dto.CwmsDTO;
+
+import java.util.*;
+
+public final class Stream implements CwmsDTO
+{
+    private final String _streamId;
```
MikeNeilson: still unsure on the underscores. if the intend is to now render in the json use the `@JsonIgnore` annotation

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamBuilder.java
```diff
@@ -0,0 +1,214 @@
...
+
+import cwms.radar.data.dto.basinconnectivity.buildercontracts.*;
+import cwms.radar.data.dto.basinconnectivity.buildercontracts.stream.*;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+
+public class StreamBuilder implements BuildStream, BuildDiversionStation, BuildConfluenceStation, BuildDiversionBank, BuildConfluenceBank
```
MikeNeilson: Should be a nested class of Stream

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamLocation.java
```diff
@@ -0,0 +1,82 @@
+package cwms.radar.data.dto.basinconnectivity;
+
+import cwms.radar.data.dto.CwmsDTO;
+
+public class StreamLocation implements CwmsDTO
+{
+    private final String _locationId;
```
MikeNeilson: The underscores don't appear to be necessary here.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamLocationBuilder.java
```diff
@@ -0,0 +1,126 @@
+package cwms.radar.data.dto.basinconnectivity;
+
+import cwms.radar.data.dto.basinconnectivity.buildercontracts.Build;
+
+public class StreamLocationBuilder implements Build<StreamLocation>
```
MikeNeilson: This should be a nested class of StreamLocation

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamReach.java
```diff
@@ -0,0 +1,61 @@
+package cwms.radar.data.dto.basinconnectivity;
+
+import cwms.radar.data.dto.CwmsDTO;
+
+public class StreamReach implements CwmsDTO
+{
+    private final String _upstreamLocationId;
```
MikeNeilson: underscores

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamReach.java
```diff
@@ -0,0 +1,61 @@
...
+    private final String _downstreamLocationId;
+    private final String _streamId;
+    private final String _reachId;
```
MikeNeilson: if streamId and reachId are the text name of the stream and reach use "streamName" and "reachName" respectively. 
We're trying to break from some of the bad in database conventions within the name to something a bit friendlier to peoples brains.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamReachBuilder.java
```diff
@@ -0,0 +1,88 @@
+package cwms.radar.data.dto.basinconnectivity;
+
+import cwms.radar.data.dto.basinconnectivity.buildercontracts.Build;
+
+public class StreamReachBuilder implements Build<StreamReach>
```
MikeNeilson: This should be a nested class of StreamReach

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamReachBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+{
+
+    private final String _streamId;
```
MikeNeilson: underscores

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/buildercontracts/Build.java
```diff
@@ -0,0 +1,9 @@
+package cwms.radar.data.dto.basinconnectivity.buildercontracts;
+
+import cwms.radar.data.dto.CwmsDTO;
+import cwms.radar.data.dto.basinconnectivity.Stream;
+
+public interface Build<T>
```
MikeNeilson: I think just using the builder pattern itself is sufficient to identify the need of a "build()" function, also with the Builder be nested in the datatypes are more obvious and better known to the class at hand so I don't think this will be necessary.

Though I'm definately open to disagreement, just means I need to go fix my own if we decide to keep this interface... and move it up to the root cwms.radar.data.dto package level.

cwms_radar_api/src/main/java/cwms/radar/data/util/BasinUnitsConverter.java
```diff
@@ -0,0 +1,170 @@
+package cwms.radar.data.util;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class BasinUnitsConverter
```
MikeNeilson: I believe there are existing unit converters we should just pull in.

Also, if you're doing the queries through jooq (like direct sql and such) you can use the database unit conversion. See the timeseriesdao code for an example.

cwms_radar_api/src/main/java/cwms/radar/formatters/json/PgJsonFormatter.java
```diff
@@ -0,0 +1,94 @@
...
+import cwms.radar.formatters.OutputFormatter;
+import service.annotations.FormatService;
+
+import javax.json.Json;
+import javax.json.JsonArray;
+import javax.json.JsonArrayBuilder;
+import javax.json.JsonObject;
+import java.util.List;
+
+@FormatService(contentType = Formats.JSON, dataTypes = {Basin.class})
```
MikeNeilson: The format here shouldn't be plain JSON but the mime type for the this specific type. 
since one doesn't exist though i suggest the following:
`application/vnd.hec.cwms-basin+json` but that's pretty ugly so I'm open to other ideas.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Basin.java
```diff
@@ -0,0 +1,68 @@
...
+        return _contributingArea;
+    }
+
+    public String getParentBasinId()
+    {
+        return _parentBasinId;
+    }
+
+    @Override
+    public PgJsonGraph getPgJsonGraph()
```
MikeNeilson: The basin appears to be able to build and return a graph. However, should the basin have a graph? Having a member that is the graph would allow openapi/swagger to build a representation of this object to present to the user in the swagger-docs and allow a code generate to make something for clients.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/nodes/BasinConnectivityNode.java
```diff
@@ -0,0 +1,111 @@
...
+        return _station;
+    }
+
+    public String getStreamId()
+    {
+        return _streamId;
+    }
+
+    @Override
+    public JsonObject getProperties()
```
MikeNeilson: It's concern to see a JsonObject at this level. This would preclude adding XML support with relative ease. 

Any reason this can be object or some sort of generic interface?

or even just HashMap<String,Object>?

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/edges/StreamEdge.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    public String getLabel()
+    {
+        return LABEL;
+    }
+
+    @Override
+    public JsonObject getProperties()
```
MikeNeilson: same comment as on BasinConnectivityNode, should probably just be HashMap<String,Object>.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/edges/ReachEdge.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @Override
+    public String getLabel()
+    {
+        return LABEL;
+    }
+
+    @Override
+    public JsonObject getProperties()
```
MikeNeilson: See other comments about HashMap<String,Object> vs something json specific at this level.

MikeNeilson: In addition, if this is a property of the edge or node, it should be a member variable and not a function return.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/edges/BasinConnectivityEdge.java
```diff
@@ -0,0 +1,54 @@
...
+        return _source;
+    }
+
+    public BasinConnectivityNode getTarget()
+    {
+        return _target;
+    }
+
+    @Override
+    public JsonObject toPGJSON()
```
MikeNeilson: This should just be a Hashmap<String,Object> or if it should only ever have those particular parameters a POJO and it should be a member variable of this Edge.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityStream.java
```diff
@@ -0,0 +1,338 @@
...
+import cwms.radar.api.graph.basinconnectivity.nodes.BasinConnectivityNodeComparator;
+import cwms.radar.api.graph.basinconnectivity.nodes.BasinConnectivityStreamLocation;
+import cwms.radar.api.graph.basinconnectivity.nodes.EmptyStreamNode;
+import cwms.radar.data.dto.basinconnectivity.Stream;
+import cwms.radar.data.dto.basinconnectivity.StreamLocation;
+import cwms.radar.data.dto.basinconnectivity.StreamReach;
+
+import java.util.*;
+
+public class BasinConnectivityStream
```
MikeNeilson, line 16: why the underscore on each variable?

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityGraph.java
```diff
@@ -0,0 +1,44 @@
...
+import cwms.radar.api.graph.basinconnectivity.edges.BasinConnectivityEdge;
+import cwms.radar.api.graph.basinconnectivity.nodes.BasinConnectivityNode;
+import cwms.radar.api.graph.pgjson.PgJsonEdge;
+import cwms.radar.api.graph.pgjson.PgJsonGraph;
+import cwms.radar.api.graph.pgjson.PgJsonNode;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class BasinConnectivityGraph implements PgJsonGraph
```
MikeNeilson: should it implement specificially an PGJsonGraph or should it implement a PGGraph that we then encode as json?

The overall structure of the graph generation looks great but it's very specific to one format at a place in the code where everything should be a generic object. Once you get the POJOs setup correctly with the structure and annotations Jackson's json formatter will do a good job of actually encoding the JSON. See the Office controller and DAO and the catalog controllers and DAO, as well as the Pool endpoint that @rripken created for some examples.

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
```diff
@@ -0,0 +1,189 @@
...
+    public void getOne(@NotNull Context ctx, @NotNull String basinId)
+    {
+        _getOneRequest.mark();
+        try(final Timer.Context timeContext = _getAllRequestsTime.time();
+            DSLContext dsl = getDslContext(ctx))
+        {
+            String units = ctx.queryParam("unit");
+            String office = ctx.queryParam("office");
+            String formatParam = ctx.queryParam("format", "");
+            if(units == null)
```
MikeNeilson: instead of this if statement use the queryParam with default construct
`ctx.queryParam("unit","EN")`

or even better
`ctx.queryParam("unit",UnitSystem.EN)` <-- note: i'm doing that from memory but there are constants for the unit systems in the code.

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
```diff
@@ -0,0 +1,189 @@
...
+            {
+                units = "EN";
+            }
+            if(formatParam == null)
```
MikeNeilson: Use the accept header, see the officecontroller for an example of how to check the provided value.)

MikeNeilson: which you already did below. sorry, I started this comment yesterday right be I left. 

You should ditch the format query parameter for this endpoint and only use the accept header.

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
```diff
@@ -0,0 +1,189 @@
...
+            {
+                formatParam = "json";
+            }
+            if(!SUPPORTED_CONTENT_TYPES.contains(formatParam))
+            {
+                throw new FormattingException("content-type " + formatParam + " is not implemented");
```
MikeNeilson: I also need to remove this in places. avoid returning user generated content back to the user. This how cross site scripting works, someone gives a url with an evil format type and suddenly that user is running some javascript and giving away a token to gmail or a bank or something.

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
```diff
@@ -0,0 +1,189 @@
...
+            }
+            String formatHeader = ctx.header(Header.ACCEPT);
+            ContentType contentType = Formats.parseHeaderAndQueryParm(formatHeader, formatParam);
+            ctx.contentType(contentType.toString());
+            BasinDao basinDao = new BasinDao(dsl);
+            Basin basin = basinDao.getBasin(basinId, units, office);
+            if(contentType.getType().equals(Formats.JSON))
```
MikeNeilson: You aren't returning an error to the user in the fail case. This will end up with a 200 OK with no content to the user.

The else case should return 404 and tell the user "you must use "this header or blank"

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
```diff
@@ -0,0 +1,189 @@
...
+            }
+            if(formatParam == null)
+            {
+                formatParam = "json";
+            }
+            if(!SUPPORTED_CONTENT_TYPES.contains(formatParam))
+            {
+                throw new FormattingException("content-type " + formatParam + " is not implemented");
+            }
+            String formatHeader = ctx.header(Header.ACCEPT);
```
MikeNeilson: this should be `ctx.header(Header.ACCEPT) != null ? ctx.header(Header.Accept) : "json"` (or whatever content type we agree on.

cwms_radar_api/build.gradle
```diff
@@ -18,7 +18,7 @@ dependencies {
 
     // https://mvnrepository.com/artifact/de.grundid.opendatalab/geojson-jackson
     implementation group: 'de.grundid.opendatalab', name: 'geojson-jackson', version: '1.14'
-
+    compile 'org.glassfish:javax.json:1.0.4'
```
MikeNeilson: javalin already includes the jackson json library and should be used if something directly requires json/java components. This should not be included.

MikeNeilson (changes requested):
Overall the structure looks good. A major concern is all the underscores in variable names; it looks odd in Java land and since the names of the object variables should come out in the JSON and xml it would look odd there as well.

Also it doesn't appear to allow for swagger to generate an example; this is no-go for any future endpoints, we want people to develop quickly against the api which means using code generators. Those need well defined content types and formats.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/nodes/EmptyStreamNode.java
rma-bryson: Was more if "Empty" was descriptive enough for junction nodes that aren't defined as stream locations, but are needed as either terminal nodes or as a junction between streams where there is no stream location defined. If you are fine with "Empty" then we can leave it.

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
rma-bryson: Removed format query parameter

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
rma-bryson: added Basin DTO object

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
rma-bryson: updated to `ctx.queryParam("unit", UnitSystem.EN.value());`

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
rma-bryson: removed format query parameter

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
rma-bryson: Removed this

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
rma-bryson: added 404 return if using unsupported content type

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Basin.java
rma-bryson: refactored builders into nested class and updated DTO data object constructors to private instead of package private with builder in separate class.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Stream.java
rma-bryson: Removed all underscores for fields. Is an RMA naming convention, that I removed.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamBuilder.java
rma-bryson: refactored into nested class

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamLocation.java
rma-bryson: removed

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamLocationBuilder.java
rma-bryson: refactored into nested class

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamReach.java
rma-bryson: removed

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamReach.java
rma-bryson: Updated to use "Name" instead of "Id"... also updated this for basin and streamLocations

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamReachBuilder.java
rma-bryson: refactored into nested class

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/StreamReachBuilder.java
rma-bryson: removed

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/buildercontracts/Build.java
rma-bryson: Removed this. This is not necessary with the Builders being nested. There is a BuildStream interface used for Streams, but that is to enforce an order of operations specific to building Streams.

cwms_radar_api/src/main/java/cwms/radar/data/util/BasinUnitsConverter.java
rma-bryson: Updated to use database unit conversion. Removed BasinUnitsConverter class

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/BasinBuilder.java
rma-bryson: Refactored in to nested class

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Basin.java
rma-bryson: Removed underscores

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamReachDao.java
rma-bryson: updated to use named columns instead of indexes

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamReachDao.java
rma-bryson: updated to use ternary syntax

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamReachDao.java
rma-bryson: updated to use lambda `dsl.connection(c -> {...})`

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamDao.java
rma-bryson: Brought units strings into an Enum, similar to the UnitsSystem enum.

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamDao.java
rma-bryson: updated to ternary

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamDao.java
rma-bryson: updated to ternary

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
rma-bryson: updated to use `dsl.connection(connection -> {...})`

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
rma-bryson: updated to use` dsl.connection(connection -> {...})`

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
rma-bryson: updated to use ternary and switched equals order

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
rma-bryson: updated to use ternary and switched order. Extracted strings out into Unit enum, and used `UnitSystem.EN.getValue()` instead of "EN"

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityStream.java
rma-bryson: Rma naming convention for field variables. I removed the underscores.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityGraphBuilder.java
rma-bryson: Refactored into an inner class

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
rma-bryson: Removed format query param

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
rma-bryson: Removed format query param

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/edges/BasinConnectivityEdge.java
rma-bryson: Updated to remove all javax.json usages. Removed PG-related stuff from BasinConnectivity base objects, to instead use POJO data objects that wrap the base objects... the PGJsonFormatter uses those POJOs for formatting, to separate the formatting logic from the basin connectivity data objects.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/edges/ReachEdge.java
rma-bryson: Replaced JsonObjects with POJO objects to be used by the formatter.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/edges/StreamEdge.java
rma-bryson: Replaced JsonObjects with POJO objects to be used by the formatter.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/nodes/BasinConnectivityNode.java
rma-bryson: Replaced JsonObjects with POJO objects to be used by the formatter.

cwms_radar_api/src/main/java/cwms/radar/data/dto/PgJsonDTO.java
rma-bryson: Removed this interface.

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Basin.java
rma-bryson: Removed this. The PgJsonFormatter (or whatever other formatter in the future) will handle all the formatting.

cwms_radar_api/build.gradle
rma-bryson: Updated to use jackson json library. Removed javax.json.

cwms_radar_api/src/main/java/cwms/radar/data/dao/StreamDao.java
rma-bryson: Updated to use `dsl.connection(c->{...});`

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
rma-rripken: looks good

cwms_radar_api/src/main/java/cwms/radar/data/dto/basinconnectivity/Basin.java
rma-rripken: maybe I wasn't on latest.  I agree that the latest looks right.

cwms_radar_api/src/main/java/cwms/radar/api/BasinController.java
rma-bryson: Updated to use 

`String formatHeader = ctx.header(Header.ACCEPT) != null ? ctx.header(Header.ACCEPT) : Formats.PGJSON;
            ContentType contentType = Formats.parseHeader(formatHeader);'

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityGraph.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class BasinConnectivityGraph implements Graph
+{
+
+    private final List<BasinConnectivityEdge> edges = new ArrayList<>();
+    private final List<BasinConnectivityNode> nodes = new ArrayList<>();
+    private final String id;
```
MikeNeilson: i feel like we should call this "basinName" @rripken thoughts?

(but otherwise this looks good. Very simple and direct, obvious what it's suppose to be doing. I'm just starting to be really nit-picky about how things are named as it has long lasting effects.)

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityStream.java
MikeNeilson: Thanks.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityGraph.java
rma-rripken: I was wrong, but I thought it had to be id b/c the pg-json format wanted it to be id.  Looking at the pg-json ebnf it looks like the top level graph object is just {nodes:[...], edges:[...]} and there isn't an id field or a properties field.  Hmm,  sort of wish there was a way to name and identify and have properties for the entire graph.  
I could see "id" and "name" attributes being useful in different ways.  "basinName" is more precise about what the field holds.  Not sure how do-able it is but I'd prefer attributes that were basin-connectivity specific end up in properties[].  The structure of the connectivity format should be sort of general purpose.

rma-rripken: ~~Not having an id field at the top level means that our Graph interface isn't right...~~  The Graph and PgGraph dont have to match.

rma-bryson: pg-json doesn't have anything specifying what the name of the graph is really... just {nodes:[...], edges:[...]}. Seems there would need to be some sort of way of identifying what basin we are looking at though, especially if doing a getAll() ?

rma-rripken: We could just add the pg-json graph as an attribute to a top level object that had the name.   {name:"Upper Basin", graph:{"..pg-json here.."}}  The result wouldn't be strictly pg-json but that format isn't exactly widely used as far as I can tell.  Maybe content type goes back to application/json  or we make it application/vnd.basin-pg+json or something basin specific?

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/nodes/BasinConnectivityNode.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.*;
+
+public abstract class BasinConnectivityNode implements Node
+{
+    private final String streamId;
+    private final Double station;
+    private final String bank;
+
+    protected BasinConnectivityNode(String streamId, Double station, String bank)
```
MikeNeilson, line 15: NOT Required: would it make sense to just have two constructors and have this one error on a null bank?

probably doesn't matter much since it's only used by the builder but food for thought.

cwms_radar_api/src/main/java/cwms/radar/data/dao/BasinDao.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.List;
+
+public class BasinDao extends JooqDao<Basin>
+{
+    public BasinDao(DSLContext dsl)
+    {
+        super(dsl);
+    }
+
+    public List<Basin> getAllBasins(String unitSystem, String officeId) throws SQLException
```
MikeNeilson, line 22: We're going to want to add pagination to this. I can imagine this list getting pretty long at some point. 
But I'm okay with dealing with it a little later down the line.

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityGraph.java
rma-bryson: I think its just a matter of naming the content-type to be clear, since including the {name:"Uppder Basin"...} isn't strictly pg-json. @MikeNeilson does application/vnd.basin-pg+json sound good? Or generalize it further to application/vnd.named-pg+json

MikeNeilson: I like that option. The idea is we're returning a Basin, that happens to contain a graph, the graph is just part of the thing returned. But would that mean this BasinConnectivityGraph should be a member of the Basin DTO object, negating the need for an ID/name field here. This class just being "The Graph.".... I hope that made sense.

MikeNeilson: oh and then we have a good disticntion on content

application/vnd.basin+json <- no graph, just basic basin info
application/vnd.basin-pg+json <- basin with graph

assuming we go this way doesn't have to be done now. But it seems flexible to me for the future.

MikeNeilson (dismissed):
While there's still some discussion. I'm happy with the overall intent. If you want to fix something from the comments go ahead and just do so.

if you aren't sure yet but you think we should still consider it, feel free to put in issue in, either here to the other place.

Also if you fundamentally disagree with something I suggested, let me know why and continue the discussion (or I'm just wrong and I'll shut up.)

cwms_radar_api/src/main/java/cwms/radar/api/graph/basinconnectivity/BasinConnectivityGraph.java
rma-bryson: That works. I think it would be cleaner to not have the graph be a member of the Basin DTO object... because its not stored that way in the database. And in the future it gives the option of serializing the Basin object itself... OR serializing the basin as a graph.

MikeNeilson: Ah. that's a discussion Ryan and I have had. _One does not need to match the database in RADAR output_. The purpose of a relation database is to store data in a standardized way to reduce duplication and increase consistency. The purpose of an API like this is to provide that data in a sensible format ready for near immediate use. Perhaps a small transform to application specific format are required; but that's why we label and maintain our output formats.

Our purpose is not to provide the normalized data set, but in fact the denormalized dataset for various aggregated uses.

MikeNeilson (approved):

PR #49, opened 2021-09-03
Adjustment of format used for times in TimeSeries and a test that ver…

rripken:
…ifies json roundtrip serialization.

### DIFF
--- a/cwms_radar_api/src/test/java/cwms/radar/data/dto/TimeSeriesTest.java
+++ b/cwms_radar_api/src/test/java/cwms/radar/data/dto/TimeSeriesTest.java
@@ -1,0 +1,81 @@
+package cwms.radar.data.dto;
+
+import java.time.Duration;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import io.javalin.plugin.json.JavalinJackson;
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.Test;
+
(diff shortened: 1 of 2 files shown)


cwms_radar_api/src/test/java/cwms/radar/data/dto/TimeSeriesTest.java
```diff
@@ -0,0 +1,81 @@
...
+		ZonedDateTime start = ZonedDateTime.parse("2021-06-21T14:00:00-07:00[PST8PDT]");
+		ZonedDateTime end = ZonedDateTime.parse("2021-06-22T14:00:00-07:00[PST8PDT]");
+		TimeSeries ts = new TimeSeries(null, -1, 0, tsId, "LRL", start, end, null, Duration.ZERO);
+
+		ObjectMapper om = buildObjectMapper();
+
+		String tsBody = om.writeValueAsString(ts);
+		assertNotNull(tsBody);
+
+		TimeSeries ts2 = om.readValue(tsBody, TimeSeries.class);
```
DanielTOsborne, line 37: The test fails on this line, with the following message:
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException at TimeSeriesTest.java:37

DanielTOsborne (commented):
I do have a question about how imports should be listed, but aside from that, they appear to not be directly related to the fix (aside from the first hunk), but are the majority of the change content in the TimeSeries.java file.

It's just a nitpick though. The fix is good. 👍

cwms_radar_api/src/test/java/cwms/radar/data/dto/TimeSeriesTest.java
rripken: Oh, I fixed that too. I was trying to figure out why my IDE was listing TimeSeries as changed in two changelists.  Weird.  I'll push that too.

rripken: If my IDE complains about unused imports I usually just let it "optimize imports" or whatever it calls it.  I don't pay too much attention to how imports are ordered.

MikeNeilson (approved):

PR #50, opened 2021-09-21
changes required for javalin v4.

MikeNeilson:


rripken (approved):
Looks good to me.  We may need to revisit whether some things get serialized via ctx.json or whether they get the ObjectMapper from the context and use it themselves.  The .json syntax is way prettier but may not always work if we can't tweak the ObjectMapper javalin uses.

MikeNeilson:
I was thinking the same thing. But I feel like we'll just have to wait and see.

PR #51, opened 2021-09-23
Feature/locations json

rma-bryson:
NOTE: there are updates in the Feature/locations-xml branch not present here. the Feature/locations-xml branch contains the latest version of files. Making a pull request with this branch for preserving history.


cwms_radar_api/src/main/java/cwms/radar/data/dto/Location.java
```diff
@@ -1,24 +1,387 @@
...
+        private Double publishedLatitude;
+        private String verticalDatum;
+        private Double elevation;
+        private String mapLabel;
+        private String boundingOfficeId;
+        private static final String MISSING_NAME_ERROR_MSG = "Location name is required";
+
+        @JsonCreator
+        public Builder(@JsonProperty(value = "name", required = true) String name, @JsonProperty(value = "locationKind", required = true) String locationKind,
+                       @JsonProperty (value = "timezoneId", required = true) ZoneId timeZoneId, @JsonProperty(value = "latitude", required = true) double latitude,
```
MikeNeilson, line 218: If some of these can be not changed (e.g. updating a single parameter of a location) can't most of these just be "required =false" or is that due to the way the JsonCreate is working that I haven't noticed yet?

cwms_radar_api/src/main/java/cwms/radar/api/LocationController.java
```diff
@@ -157,22 +183,209 @@ public void getOne(Context ctx, String locationCode) {
         }
     }
 
-    @OpenApi(ignore = true)
+    @OpenApi(
+        requestBody = @OpenApiRequestBody(
+                content = {
+                          @OpenApiContent(from = Location.class, type = Formats.JSON ),
+
```
MikeNeilson: you can just delete these.

cwms_radar_api/src/main/java/cwms/radar/api/LocationController.java
```diff
@@ -157,22 +183,209 @@ public void getOne(Context ctx, String locationCode) {
...
+            ctx.status(HttpServletResponse.SC_INTERNAL_SERVER_ERROR).json(re);
+        }
     }
 
-    @OpenApi(ignore = true)
+    @OpenApi(
+            requestBody = @OpenApiRequestBody(
+                    content = {
+                          @OpenApiContent(from = Location.class, type = Formats.JSON )
+//                        @OpenApiContent(from = Location.class, type = Formats.TAB ),
```
MikeNeilson: same, you can just delete commented out code like this.

cwms_radar_api/src/main/java/cwms/radar/api/LocationController.java
```diff
@@ -157,22 +183,209 @@ public void getOne(Context ctx, String locationCode) {
...
+            logger.log(Level.SEVERE, re.toString(), ex);
+            ctx.status(HttpServletResponse.SC_INTERNAL_SERVER_ERROR).json(re);
+        }
+
     }
 
-    @OpenApi(ignore = true)
+    @OpenApi(
+            queryParams = {
+                    @OpenApiParam(name = "office", description = "Specifies the owning office of the location whose data is to be deleted. If this field is not specified, matching location information will be deleted from all offices.")
```
MikeNeilson, line 289: Aren't there options in the location delete to decide if the operation should cascade to dependent objects? or are we intentionally leaving that out right now for initial safety?

cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
```diff
@@ -0,0 +1,361 @@
...
+
+    @Override
+    public void storeLocation(Location location) throws IOException
+    {
+        try
+        {
+            dsl.connection(c ->
+            {
+                CwmsDbLoc locJooq = CwmsDbServiceLookup.buildCwmsDb(CwmsDbLoc.class, c);
+                String elevationUnits = "m"; //this should be updated to use Unit enum once merged in
```
MikeNeilson: I like what this comment says, what stopping it from getting done now before merging in?

MikeNeilson (commented):
Overall looks good. Please review and respond to some of the comments before we merge it in.

cwms_radar_api/src/main/java/cwms/radar/data/dto/Location.java
rma-bryson: In the Feature/locations xml branch I updated this so the `required = true` is removed, that way it allows for updating a single parameter of a location, otherwise an error would be thrown. Because of that, now if there is a required field missing, that gets checked on storage, not the creation of the DTO object.

cwms_radar_api/src/main/java/cwms/radar/api/LocationController.java
MikeNeilson: We can add the cascade later when we've better decided how to handle that.

cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
rma-bryson: Wrote this code before we merged in the Unit enum. Updated it to use `Unit.METER.getValue()`

MikeNeilson (approved):
Looks good.

PR #54, opened 2021-09-28
Feature/timeseries write

rripken:
Only handles create/store/delete of ts from json at the moment.


cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -0,0 +1,649 @@
...
+			for(int i = 0; iter.hasNext(); i++)
+			{
+				TimeSeries.Record value = iter.next();
+				timeArray[i] = value.getDateTime().getTime();
+				valueArray[i] = value.getValue();
+				qualityArray[i] = value.getQualityCode();
+			}
+		}
+
+		final Timestamp versionDate = null;  // Where does this come from?
```
MikeNeilson: This would be passed in by the user. Most commonly used for forecast data so the original isn't destroyed. this is a good place holder but I wonder if this logic should just be IN the first call to store that takes the TimeSeries object, as the versionDate will eventually be a non-required parameter of both that, and depending on how we choose to do things, the timeseries value elements. But I also haven't read further than this yet to know if this call is used else where.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -0,0 +1,649 @@
...
+		final boolean createAsLrts = false;
```
MikeNeilson, line 600: Acceptable place holder, Adam or Peter may have some insight as to whether this should be a query parameter or a part of the timeseries object itself.... I think we're still working out some of those details.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -0,0 +1,649 @@
...
+import static org.jooq.impl.DSL.field;
+import static org.jooq.impl.DSL.max;
+import static org.jooq.impl.DSL.partitionBy;
+import static usace.cwms.db.jooq.codegen.tables.AV_CWMS_TS_ID2.AV_CWMS_TS_ID2;
+
+public class TimeSeriesDaoImpl extends JooqDao<TimeSeries> implements TimeSeriesDao
+{
+	private static final Logger logger = Logger.getLogger(TimeSeriesDaoImpl.class.getName());
+
+	public static final String STORE_RULE = "DELETE INSERT";
```
MikeNeilson: Let's make an enum for this and make it a query parameter. We've already had performance issues from overusing this rule in the past best not to accidentally repeat it.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -0,0 +1,649 @@
...
+	{
+		dsl.connection(connection -> {
+			CwmsDbTs tsDao = CwmsDbServiceLookup.buildCwmsDb(CwmsDbTs.class, connection);
+
+			int utcOffsetMinutes = 0;
+			int intervalForward = 0;
+			int intervalBackward = 0;
+			boolean versionedFlag = false;
+			boolean activeFlag = true;
+			tsDao.createTsCodeBigInteger(connection, input.getOfficeId(), input.getName(), utcOffsetMinutes,
```
MikeNeilson: are we sure this function success it's definitely worked? If so perhaps we should catch the exception on error and return our own indication. This is one of those places were a non-generic error message may be handy.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -0,0 +1,649 @@
...
+	{
+		return retrieveTsCode(tsId) != null;
+	}
+
+	protected Duration getDuration(String tsId)
+	{
+		Long numMinutes = 0L;
+		try
+		{
+			BigDecimal intervalVal = CWMS_TS_PACKAGE.call_GET_INTERVAL(dsl.configuration(), tsId);
```
MikeNeilson: Are you getting the Duration or the interval? I know the java api uses the single "Duration" class for both a window of time and a distance of time between element but the name of the function getDuration implies you should be calling call_GET_DURATION not call_GET_INTERVAL here. If it's actually the timeseries interval it should be named getInterval even though it's returning a java "Duration" object.

MikeNeilson (changes requested):
Looks good. Some of those were much cleaner than expected them to be. Only a few minor nitpicks.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
MikeNeilson: Instead of NULL we should probably have a constant, which would still be null, called NON_VERSIONED. It would make the section of code a little more clear.

MikeNeilson (changes requested):
(This may be a duplicate, the interface is acting a little odd this morning.)

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
rripken: Sure. that sounds like a good idea.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
rripken: ok.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
rripken: You are right, should have been getInterval.  I don't think the method was being used. Easiest to just remove it.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
rripken: which part? the createAsLrts or the versionDate or both?

MikeNeilson: the createAsLrts. mostly just commenting that I like that you gave it a same for the functional call; saves the reader having to dig into the called function itself.

cwms_radar_api/src/main/java/cwms/radar/api/TimeSeriesController.java
```diff
@@ -126,7 +207,7 @@ public void getAll(Context ctx) {
                 final Timer.Context timeContext = getAllRequestsTime.time();
                 DSLContext dsl = getDslContext(ctx))
         {
-            TimeSeriesDao dao = new TimeSeriesDao(dsl);
+            TimeSeriesDao dao = getTimeSeriesDao(dsl);
```
MikeNeilson, line 210: we don't need to do this "right now" but would it make more sense to have the DAOs implement AutoClosable and put them up in the try-with-resources block?

MikeNeilson (approved):
That looks good to me.

cwms_radar_api/src/main/java/cwms/radar/api/TimeSeriesController.java
rripken: I'm torn about how the *Controller and *Dao and DSLContext and Connection should work together.  Before Javalin I was coming from cwms mobile and Spring.   In Spring the Controllers, Dao and Services were long-lived objects and used @autowired DataSource or JdbcTemplate to build new Connections as needed.  The Javalin Dao are structured the way it currently is b/c I wanted to see if we could make all the Dao that used Jooq not have to know anything about Connection or Context.   But I hadn't done much development around jOOQ or DSLContext.  Its easier to get a Connection from DSLContext than I first realized and it takes less to construct a DSLContext than I initially realized.  I also had this idea that we'd be able to swap in mocks for the DSLContext and we'd get this cool powerful way to do testing.  But when I tried the DSLContext mocking it didn't work anywhere near the way I wanted it to.  My inclination now would be to have the JooqDao objects not have a private DSLContext field and instead to just have every public dao method take a Connection as one of the arguments.  Then it wouldn't matter if the Dao were short-lived or long-lived.

PR #57, opened 2021-10-04
updated to gradle version 7.2

rma-bryson:


rma-bryson:
Hmm looks like updating to 7.2 is causing some issues

MikeNeilson:
which is... odd, change the build (in the github actions yaml) to be called with --info, perhaps then we'll get more insight into the problem.

rripken:
Might be this:
https://github.com/gradle/gradle/issues/1916#issuecomment-298325759

rripken:
If it is that 1916 issue I think we'd just need to add an explicit version for checkstyle inside the subprojects.  Not sure what the correct version would be.   There is this big chart of versions but I can't see what we'd need to change.
https://checkstyle-addons.thomasjensen.com/checkstyle-compatibility-matrix.html?hl=L[g];L8.12;L6.19;L5.9;L5.7-L5.5

 checkstyle {
        toolVersion "7.6.1"
    }

rripken:
We never did figure out exactly what was causing this to fail.  I'm closing it for now.

PR #59, opened 2021-10-07
Feature/timeseries xml

rripken:
TimeSeries create with xml input wasn't parsing correctly.


MikeNeilson (dismissed):
This looks fine, not sure why the test isn't passing, if you modify the github action to run gradle with --info we should get the full stack trace and better know what's going on.

rripken:
It fails because there is a difference in how the timezone is handled.  By changing my default timezone I can duplicate the error, will look into it in more depth.

cwms_radar_api/src/test/resources/cwms/radar/api/timeseries_create.xml
```diff
@@ -0,0 +1,52 @@
...
+		</column>
+	</valueColumns>
+	<values>
+		<record>
+			<dateTime>1624287600000</dateTime>
+			<value>0.0</value>
+			<qualityCode>0</qualityCode>
+		</record>
+		<record>
+			<dateTime>1624288500000</dateTime>
```
MikeNeilson, line 37: Are these timestamps relative to the UTC Unix Epoch or your local time zone? If you're calling it a Timestamp it SHOULD always be the from the UTC Epoch. We should only allows off UTC Timezones in full ISO8601 text formatted dates.

MikeNeilson:
@DanielTOsborne in looking at the TimeSeries.java class while reviewing this I noticed the TimeSeries.Record has Timestamp instead of ZonedDateTime or plain old long for it's time field. I didn't really think anything of it before but why the java timestamp instead of ZoneDateTime?

DanielTOsborne:
> @DanielTOsborne in looking at the TimeSeries.java class while reviewing this I noticed the TimeSeries.Record has Timestamp instead of ZonedDateTime or plain old long for it's time field. I didn't really think anything of it before but why the java timestamp instead of ZoneDateTime?

Because that always represents UTC time since Unix epoch. It made serialization significantly easier with that design consideration.
It could also be Instant.
It didn't make sense to have a ZonedDateTime, since the zone is always UTC. If anything but UTC was input, it wouldn't round-trip.

MikeNeilson:
That makes sense. Thanks.

rripken:
Looks like fixing the TimeSeries I was using in my test fixed the test.

MikeNeilson (approved):

PR #61, opened 2021-10-25
Feature/location levels json

rma-bryson:
Initial JSON CRUD operations for Levels.


cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
```diff
@@ -37,19 +63,107 @@ public LevelsController(MetricRegistry metrics){
...
-        ctx.status(HttpServletResponse.SC_NOT_IMPLEMENTED).json(RadarError.notImplemented());
+    public void create(@NotNull Context ctx)
+    {
+        createRequest.mark();
+        try(final Timer.Context timeContext = createRequestTime.time();
+            DSLContext dsl = getDslContext(ctx))
+        {
+            LocationLevelsDao levelsDao = new LocationLevelsDaoImpl(dsl);
+            String office = ctx.queryParam("office");
+            String acceptHeader = ctx.req.getContentType();
```
MikeNeilson: Don't use the name acceptHeader here, it's not the user saying I want results in X formats. It's saying this IS the contentType.

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
```diff
@@ -109,22 +224,178 @@ public void getAll(Context ctx) {
...
+            String office = ctx.queryParam("office");
+            String dateString = ctx.queryParam("date");
+            ZonedDateTimeAdapter zonedDateTimeAdapter = new ZonedDateTimeAdapter();
+            ZonedDateTime unmarshalledDateTime = zonedDateTimeAdapter.unmarshal(dateString);
+            ZoneId timezoneId = unmarshalledDateTime.getZone();
+            if(timezoneId == null)
+            {
+                timezoneId = ZoneId.systemDefault();
+            }
+            String acceptHeader = ctx.req.getContentType();
```
MikeNeilson: same as above, it's NOT the acceptHeader but the actual content type of the body.

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
```diff
@@ -109,22 +224,178 @@ public void getAll(Context ctx) {
...
+        {
+            logger.log(Level.SEVERE, "Failed to deserialize level", e);
+            throw new IOException("Failed to deserialize level");
+        }
+        return retVal;
+    }
 
+    private static ObjectMapper getObjectMapperForFormat(String format)
+    {
+        ObjectMapper om = new ObjectMapper();
```
MikeNeilson: this probably shouldn't have a default and should only return a mapper object for a known type and throw and exception on anything else.

cwms_radar_api/src/main/java/cwms/radar/data/dto/LocationLevel.java
```diff
@@ -0,0 +1,342 @@
...
+import java.util.function.Consumer;
+
+@JsonDeserialize
+@JsonPOJOBuilder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public class LocationLevel
+{
+    private String seasonalTimeSeriesId;
+    private List<SeasonalValueBean> seasonalValues;
```
MikeNeilson: has this been tested against TimeSeries based locations levels?

MikeNeilson (changes requested):

cwms_radar_api/src/main/java/cwms/radar/data/dto/LocationLevel.java
rma-bryson: yes, I tested creating a TimeSeries based location level, seasonal, and constant. I'm not sure if we want to keep the attribute properties at this point in the dto. I was told 99.99% of the time those aren't used.

cwms_radar_api/src/main/java/cwms/radar/data/dto/LocationLevel.java
```diff
@@ -0,0 +1,342 @@
...
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.*;
+import java.util.function.Consumer;
+
+@JsonDeserialize
+@JsonPOJOBuilder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public class LocationLevel
```
MikeNeilson: this DTO isn't immutable.

rma-bryson: Updating it to be immutable with a builder to follow suit with basins and locations.

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
rma-bryson: updated name to reqContentType

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
rma-bryson: updated name to reqContentType

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
rma-bryson: updated to only return for supported format types, else an exception is thrown.

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
MikeNeilson: That's a good name. Thanks.

MikeNeilson (approved):

PR #65, opened 2021-10-28
Feature/locations

rma-bryson:
Implemented getOne. Added Deserialization tests for Locations.


MikeNeilson (approved):
This is great. Thanks. nice simple test dataset and a clean way to load it for future tests.

PR #67, opened 2021-10-29
Build fixes

rripken:
Reordering the checkstyle block may fix bryson's previous pull request.  If this one works we can dismiss his.


rripken:
I haven't been able to figure out why this isn't building with a newer gradle and we stopped having the build problem with the current version so I'm aborting.

PR #69, opened 2021-11-01
Begin auth

MikeNeilson:
This is the baseline of what I was thinking for how authorization could work. RADAR wouldn't actually DO authenticate, though the mechanisms it provides are interesting.

This doesn't handle the setting of the CWMS Session key, but I'd like to tackle that separately as long term RADAR will take on more responsibility for how that works.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/helpers/DatabaseHelpers.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/helpers/DatabaseHelpers.java
@@ -1,0 +1,20 @@
+package cwms.radar.helpers;
+
+import java.sql.CallableStatement;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+import javax.servlet.http.HttpServletRequest;
+
+public class DatabaseHelpers {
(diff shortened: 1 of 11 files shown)


cwms_radar_api/src/main/java/cwms/radar/helpers/DatabaseHelpers.java
```diff
@@ -0,0 +1,20 @@
+package cwms.radar.helpers;
+
+import java.sql.CallableStatement;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+import javax.servlet.http.HttpServletRequest;
+
+public class DatabaseHelpers {
```
MikeNeilson, line 9: Actually didn't mean to push this one yet, but anyways. Should probably be the DSL, but the idea is the appropriate info will be present in the users principal object or session and this class would be responsible for knowing how to deal with that and just setup the session on the database side. eventually it would just be a passthrough and we'd handle authorization at the API level here.

rripken (approved):

PR #87, opened 2021-11-05
[all] initial migration of code to only generate a war.

MikeNeilson:
having to maintian the routing in two places is getting non-sensical. Additionally having the split between standalone javalin and the war is clearly going to get problematic if the API depends on anything Jetty we has to be stripped out from the WAR to avoid name collisions with Tomcat.

I'm opening the draft now for comment and demostration, I'll work to get all of the various test working; or at least use the support funds to get them fixes and this is a fairly big change.


MikeNeilson:
Note, there are still some issues with error pages to resolve (namely that tomcat eats the errors and only spits back the error code not the message) but I want to get everyone moved to this before too much time goes on.

rripken (approved):

PR #100, opened 2021-11-11
Feature/expand tests

MikeNeilson:
As much as I want to expand this all the way, better to get small amounts of work into the main branch quicker.


cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
```diff
@@ -56,70 +58,49 @@ public String getLocations(String names,String format, String units, String datu
...
-                locationRef.set(location);
-            });
-        }
-        catch(DataAccessException ex)
-        {
-            throw new IOException("Could not retrieve Location");
-        }
-        return locationRef.get();
+
+        SelectQuery<Record> query = dsl.selectQuery();
```
MikeNeilson, line 60: This one down purely to allow for the Mock. the current version of JOOQ we're using doesn't have a MockDataProvider that handles the appropriate JSON data. Might be worth creating our own, might now. No super worried about it at the moment.

rripken (approved):

Issue #85, opened 2021-11-04
Add filters on Catalog API  GET request


adamkorynta:
The time series ID filter is the most important. Location category and group as well as Time Series category and group filters would be secondary.

adamkorynta:
Similarly for /catalog/locations, filtering based on locationId is necessary. Filtering on categoryId and groupId are secondary.

rripken:
I've added regexp_like filters for id, category and group to the Timeseries and Location catalog endpoint. 
#102

adamkorynta:
It looks like the category and group are for locations only. Additional parameterization of timeseries categories and groups are utilized in the CAVI.

adamkorynta:
Also using the filters causes duplicate entries to show up in the returned catalog.

PR #102, opened 2021-11-11
Bug fixes/timeseries catalog

rripken:
Added id, category and group filters to the Catalog endpoints for Timeseries and Location. 
Ref issue #85

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/api/CatalogController.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/api/CatalogController.java
@@ -138,13 +138,13 @@ public class CatalogController implements CrudHandler{
             String unitSystem = ctx.queryParamAsClass("unitSystem",String.class).getOrDefault(UnitSystem.SI.getValue());
             Optional<String> office = Optional.ofNullable(
                                          ctx.queryParamAsClass("office", String.class).allowNullable()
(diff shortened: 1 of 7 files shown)


cwms_radar_api/src/main/java/cwms/radar/api/CatalogController.java
```diff
@@ -123,18 +138,23 @@ public void getOne(Context ctx, String dataSet) {
             String unitSystem = ctx.queryParamAsClass("unitSystem",String.class).getOrDefault(UnitSystem.SI.getValue());
             Optional<String> office = Optional.ofNullable(
                                          ctx.queryParamAsClass("office", String.class).allowNullable()
-                                            .check( ofc -> Office.validOfficeCanNull(ofc), "Invalid office provided" )
+                                            .check(Office::validOfficeCanNull, "Invalid office provided" )
```
MikeNeilson, line 141: This syntax is still visually odd, but it's definitely growing on me.

MikeNeilson (approved):

cwms_radar_api/src/main/java/cwms/radar/api/CatalogController.java
rripken: Before I submit sonarlint warns me about issues.  I don't really care strongly one way or the other wrt the syntax but after I've looked at the same warning 4-5 times in a day I just give in and change it just to make the warning disappear.

Issue #80, opened 2021-11-03
Timeseries GET request sends error code 500 on time series not found


adamkorynta:
We would like the ORA-20025 SQLException code to send out an error code 404 instead so that the client API's can handle it appropriately.

rripken:
#106  Should address this for version 2 getAll requests.

MikeNeilson:
I think you're probably right, but good idea to verify just in case something got missed.

MikeNeilson:
Can someone verify this actually happened?

krowvin:
@rma-rripken Can we close this?

rma-rripken:
yes

Issue #75, opened 2021-11-02
TimeSeries begin and end without timezone throws IllegalArgumentException exception.


rripken:
The begin and end parameters are documented to default to UTC if no timezone is specified but in testing that results in the server throwing an IllegalArgumentException "Time cannot contain only an offset without the timezone."

I refactored the Dao code around this exception to try and and make sense of how the values were being parsed but I'm not entirely sure why it was being thrown.  Perhaps it can be removed?

MikeNeilson:
@DanielTOsborne 
I assume this block was created when you were implementing your changes to TimeSeries?

https://github.com/USACE/cwms-radar-api/blob/708f04e11454ddb0c07911d2dad5d5f67662d233/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java#L135

DanielTOsborne:
> @DanielTOsborne I assume this block was created when you were implementing your changes to TimeSeries?
> 
> https://github.com/USACE/cwms-radar-api/blob/708f04e11454ddb0c07911d2dad5d5f67662d233/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java#L135

Yes.
If the timezone argument wasn't supplied, it would check to see if the supplied datetime contained an offset+timezone in the string, and use that if so. Makes the formatting/parsing of query parameters and results round-trip and easier to process.

Basically, you can use a ZonedDateTime with ISO_FORMAT to both create the query parameter, and then read the date again later in the result.

MikeNeilson:
I'm not sure it's working correctly:

`curl -X 'GET' \
  'http://localhost:8384/cwms-data/timeseries?name=Alder%20Springs.Precip-Cum.Inst.1Hour.0.Calc-val&office=SPK&unit=in&begin=2021-11-01T13%3A00%3A00&end=2021-11-03T17%3A00%3A00' \
  -H 'accept: application/json;version=2'` (adjust the host/port as required) 

is throwing the error.

DanielTOsborne:
Looks like an edge case, apparently UTC's zone is the same as its offset.
I wonder if it's an OS difference? I don't remember hitting this in my tests, but I also tested on Windows.
Clearly timezones are hard.

DanielTOsborne:
[...]

PR #106, opened 2021-11-12
Making missing TimeSeries return NotFound instead of BadRequest.

rripken:
This should address #80 
I'm introducing a new runtime exception that controllers can catch to better deal with Not Found.  I think it will be an ongoing concern and I'm hoping an exception will simplify implementations.

Also taking out the exception that was being thrown by #75.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/api/TimeSeriesController.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/api/TimeSeriesController.java
@@ -221,31 +221,28 @@ public class TimeSeriesController implements CrudHandler {
                 DSLContext dsl = getDslContext(ctx))
         {
             TimeSeriesDao dao = getTimeSeriesDao(dsl);
-            String format = ctx.queryParamAsClass("format",String.class).getOrDefault("");
+            String format = ctx.queryParamAsClass("format", String.class).getOrDefault("");
             String names = ctx.queryParam("name");
             String office = ctx.queryParam("office");
-            String unit = ctx.queryParamAsClass("unit",String.class).getOrDefault(UnitSystem.EN.getValue());
+            String unit = ctx.queryParamAsClass("unit", String.class).getOrDefault(UnitSystem.EN.getValue());
             String datum = ctx.queryParam("datum");
             String begin = ctx.queryParam("begin");
             String end = ctx.queryParam("end");
             String timezone = ctx.queryParam("timezone");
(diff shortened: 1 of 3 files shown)


cwms_radar_api/src/main/java/cwms/radar/api/TimeSeriesController.java
```diff
@@ -221,31 +221,28 @@ public void getAll(Context ctx) {
...
+                    ctx.queryParamAsClass("pagesize", Integer.class).getOrDefault(defaultPageSize));
 
             String acceptHeader = ctx.header(Header.ACCEPT);
             ContentType contentType = Formats.parseHeaderAndQueryParm(acceptHeader, format);
 
             String results;
             String version = contentType.getParameters().get("version");
-            if(version != null && version.equals("2")) {
+            if(version != null && version.equals("2"))
+            {
```
MikeNeilson, line 245: Does our checkstyle file indicate { on the next line or is your IDE set for a different project?
(I'm not a die hard for either form it just seems odd that your more recent PRs have starting doing this.)

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -232,19 +211,52 @@ protected TimeSeries getTimeseries(String page, int pageSize, String names, Stri
...
+		if(cause instanceof SQLException){
+			SQLException sqlException = (SQLException) cause;
+
+			// There may be better ways to figure out if the exception was b/c Not found.
+
+			// When we type in a garbage tsId the cause is a SQLException.  Its cause is an OracleDatabaseException.
+			String message = sqlException.getMessage();
+
+			// Note: The message I saw looked like this:
+			//ORA-20001: TS_ID_NOT_FOUND: The timeseries identifier "%1" was not found for office "%2"
```
MikeNeilson, line 245: The 20001 is the error code you want to check for. That's specifically "no timeseries with that name found"

MikeNeilson (changes requested):
This otherwise looks good.

cwms_radar_api/src/main/java/cwms/radar/api/TimeSeriesController.java
rripken: I updated an ide plugin.  Must have lost the settings.  I'll look into it.

MikeNeilson (approved):

PR #116, opened 2021-11-22
Bug fixes/assigned ts

rripken:
Fixing the TimeSeries Groups so that they include the assigned time series.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesGroupDao.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesGroupDao.java
@@ -14,9 +14,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jooq.Condition;
 import org.jooq.DSLContext;
 import org.jooq.Record;
-import org.jooq.Record16;
 import org.jooq.RecordMapper;
-import org.jooq.SelectConditionStep;
 import org.jooq.SelectOnConditionStep;
 import org.jooq.SelectOrderByStep;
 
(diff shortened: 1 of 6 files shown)


cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesGroupDao.java
```diff
@@ -27,70 +37,141 @@ public TimeSeriesGroupDao(DSLContext dsl)
...
-		Select select = step;
+		final RecordMapper<Record, Pair<TimeSeriesGroup, AssignedTimeSeries>> mapper = record15 -> {
+			TimeSeriesGroup group = buildTimeSeriesGroup(record15);
+			AssignedTimeSeries loc = buildAssignedTimeSeries(record15);
 
-		if ( officeId != null && !officeId.isEmpty())
+			return new Pair<>(group, loc);
+		};
+
+		SelectOnConditionStep<Record16<String, String, String, String, String, String, String, String, String, String, String, String, BigDecimal, BigDecimal, String, String>>
```
MikeNeilson: FYI, if you're always using the named based getters later you can just do something like `SelectOnConditionStep<?>`

Not a huge deal but I imagine you like typing these out as much as I do.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesGroupDao.java
```diff
@@ -27,70 +37,141 @@ public TimeSeriesGroupDao(DSLContext dsl)
...
+		{
+			List<AssignedTimeSeries> assigned = entry.getValue();
+
+			retval.add(new TimeSeriesGroup(entry.getKey(), assigned));
+		}
+		return retval;
 	}
 
-	public List<TimeSeriesGroup> getTimeSeriesGroups(String officeId, String categoryId, String groupId)
+	private AssignedTimeSeries buildAssignedTimeSeries(Record record15)
```
MikeNeilson: This variable should probably have a better name than record15.

MikeNeilson (changes requested):
It's looks good but while reading I didn't notice anything obvious about what the meaning of "record15" was... though it is 630 in the morning when I'm reading it.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesGroupDao.java
rripken: That makes sense.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesGroupDao.java
rripken: Agreed.

MikeNeilson (approved):

Issue #122, opened 2021-12-03
/locations/{location_code} endpoint not returning Active flag


adamkorynta:
Example cURL:
```
curl -X 'GET' \
  'http://localhost:11524/cwms-data/locations/AARK?office=SWT' \
  -H 'accept: application/json;version=2'
```

Returned data:

```
{
  "name": "AARK",
  "latitude": 37.056418,
  "longitude": -97.0580939,
  "public-name": "ARKANSAS R AT ARKANSAS CITY, KS",
  "long-name": "ARKANSAS R AT ARKANSAS CITY, KS",
  "description": "ARKANSAS R AT ARKANSAS CITY, KS",
  "timezone-name": "CST6CDT",
  "location-kind": "STREAM_LOCATION",
  "nation": "US",
  "state-initial": "KS",
  "county-name": "Unknown County or County N/A",
  "nearest-city": "Arkansas City, KS",
  "horizontal-datum": "NAD83",
  "vertical-datum": "NGVD29",
  "elevation": 1050,
  "bounding-office-id": "SWT",
  "office-id": "SWT"
}
```

MikeNeilson:
well, that seems like a slight oversight there doesn't it.

MikeNeilson:
https://github.com/USACE/cwms-radar-api/blob/b52d735e3250b6273aed13b890ef4c57c7c54afa/cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java#L86

Are you sure you're on the newest version?

adamkorynta:
Yup, grabbed the latest before writing up the ticket. I see the variable in the data object too.

rripken:
#128

PR #128, opened 2021-12-06
Renamed Location.active() to Location.getActive() because the propert…

rripken:
…y wasn't being found.


rripken:
#122

MikeNeilson (approved):
If you want to spend a little more time this is probably a good one to also add an integration test on and get some understanding of that mechanism. 

You can look at the CatalogControllerTestIT for guidance on setting that up.

(except the store/update... still working out whether we want to test as a specific office or just use the "national" db. I'm probably overthinking it.)

PR #135, opened 2021-12-08
Bug fixes/ts needs offset and tz

### LINKED ISSUE #79: Timeseries API needs IntervalOffset and Location Time Zone added
In order to distinguish between Local Regular Time Series (LRTS) and Pseudo Regular Time Series (PRTS), we need the interval offset metadata returned (preferably in seconds), and the location's time zone.

rripken:
Closes #79.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
@@ -202,7 +202,29 @@ public class TimeSeriesDaoImpl extends JooqDao<TimeSeries> implements TimeSeries
 			Field<String> param = DSL.upper(CWMS_UTIL_PACKAGE.call_SPLIT_TEXT(tsId,
 					DSL.val(BigInteger.valueOf(2L)), DSL.val("."),
 					DSL.val(BigInteger.valueOf(6L))));
-			SelectSelectStep<Record8<String, String, String, BigDecimal, String, String, String, Integer>> metadataQuery = dsl.select(
+
+			// What is the syntax for selecting tzName and offsetField from the same subquery?
+			// It works when each field comes from its own subquery.
+			// This didn't work.
+			//			Field<?>[] fields = DSL.select(AV_CWMS_TS_ID2.INTERVAL_UTC_OFFSET.as("INTERVAL_UTC_OFFSET"),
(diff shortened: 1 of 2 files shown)


cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -202,7 +202,29 @@ protected TimeSeries getTimeseries(String page, int pageSize, String names, Stri
...
+			// It works when each field comes from its own subquery.
+			// This didn't work.
+			//			Field<?>[] fields = DSL.select(AV_CWMS_TS_ID2.INTERVAL_UTC_OFFSET.as("INTERVAL_UTC_OFFSET"),
+			//					AV_CWMS_TS_ID2.TIME_ZONE_ID.as("TIME_ZONE_ID"))
+			//					.from(AV_CWMS_TS_ID2).where(AV_CWMS_TS_ID2.CWMS_TS_ID.eq(tsId))
+			//					.fields();
+			//			Field<BigDecimal> offsetField = (Field<BigDecimal>) fields[0];
+			//			Field<String> tzName = (Field<String>) fields[1];
+
+			Field<BigDecimal> offsetField = DSL.select(AV_CWMS_TS_ID2.INTERVAL_UTC_OFFSET.as("INTERVAL_UTC_OFFSET"))
```
rripken, line 216: It seems like these two fields should be fetchable from a single sub-select.  Looking for suggestions of a better way.

rripken:
This was a draft b/c @rma-psmorris was going to rewrite the query but it looks like that isn't going to happen in the immediate future.  Perhaps these changes can still be merged and a new issue opened for @rma-psmorris to come up with a better query?

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
MikeNeilson: what's there seems fine to me. Unless there's a very specific performance issue you see as a result I wouldn't worry about it too much.

MikeNeilson:
That seems reasonable to me. Better to have a reasonable fix in now that wait a random amount of time for the "best" fix.

If you feel it is a strong enough concern open another issue to that specific issue.

MikeNeilson (approved):

PR #138, opened 2021-12-09
The GRP_ASSGN views don't expose the office of the groups.

rripken:
The DB_OFFICE_ID column is the office of the locations, not the office of the groups.  The where clause needs to look at the grp_office from the *cat_grp views.


rripken:
This probably needs an integration test.

MikeNeilson:
> This probably needs an integration test.

No time like the present.

rripken:
I added an integration test that uses a group from CWMS with office specific ts.

MikeNeilson (approved):

PR #152, opened 2022-02-23
Feature/authorize

rripken:


cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
```diff
@@ -103,20 +119,21 @@ public void init() throws ServletException{
...
+        AccessManager accessManager = buildAccessManager();
+
+
         javalin = Javalin.createStandalone(config -> {
             config.defaultContentType = "application/json";
             config.contextPath = context;
             config.registerPlugin(new OpenApiPlugin(getOpenApiOptions()));
             config.enableDevLogging();
             config.requestLogger( (ctx,ms) -> logger.finest(ctx.toString()));
-            //config.addStaticFiles("/static",Location.CLASSPATH);
```
MikeNeilson, line 112: This should stay. This is how the index page gets served... unless that got replaced; going in order.

cwms_radar_api/src/main/java/cwms/radar/security/CwmsAccessManager.java
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	@NotNull
+	private String getFailMessage(@NotNull Context ctx, @NotNull Set<RouteRole> requiredRoles)
+	{
+		Set<RouteRole> specifiedRoles = getRoles(ctx);
+		Set<RouteRole> missing = new LinkedHashSet<>(requiredRoles);
+		missing.removeAll(specifiedRoles);
+
+		return "Request for: " + ctx.req.getRequestURI() + " denied. Missing roles: " + missing;
```
MikeNeilson, line 75: this is technically user input, should we pass it through the sanitizer? at the least I'll make sure to check it before the next release.

cwms_radar_api/src/main/java/cwms/radar/security/CwmsAccessManager.java
```diff
@@ -0,0 +1,204 @@
...
+
+		Principal principal = getPrincipal(ctx);
+
+		if(principal != null){
+			sessionKey = callGetSessionKeyReflectively(principal);
+		}
+		return sessionKey;
+	}
+
+	private String callGetSessionKeyReflectively( Principal principal)
```
MikeNeilson, line 132: I feel like I've failed somewhere, why not cast the Principal to the `CwmsUserPrincipal`?

MikeNeilson: or just totally missunderstood part of how java worked. Either way I'd like to know what led to the reflection.

cwms_radar_api/src/test/java/cwms/radar/api/LocationControllerTest.java
```diff
@@ -77,7 +67,7 @@ public void test_basic_operations() throws Exception {
         HashMap<String,Object> attributes = new HashMap<>();
         attributes.put(ContextUtil.maxRequestSizeKey,Integer.MAX_VALUE);
         attributes.put(JsonMapperKt.JSON_MAPPER_KEY,new JavalinJackson());
-        attributes.put("Authorizer",new CwmsNoAuthorizer());
```
MikeNeilson, line 80: you can just delete this line.

cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
MikeNeilson: nevermind, we comment that out weeks ago.... this is why we should just delete things, like you have.

MikeNeilson (dismissed):

cwms_radar_api/src/main/java/cwms/radar/security/CwmsAccessManager.java
rripken: I also did not expect this.  I believe its a classloader issue.   If I look at the object with a debugger in a test case it believes its a TestCwmsUserPrincipal which extends CwmsUserPrincipal but it throws an exception if I try to cast it to CwmsUserPrincipal.  If the object was created by cwms_aaa it must have a different classloader than radar b/c we could deploy/redeploy them separately.

cwms_radar_api/src/main/java/cwms/radar/security/CwmsAccessManager.java
rripken: At the moment the message only goes into the log.  I don't send it back to the user.

MikeNeilson: ah, okay. Should be fine then. At least until we get some crazy report about slf4j, anyways.

cwms_radar_api/src/main/java/cwms/radar/security/CwmsAccessManager.java
MikeNeilson: hmm. That, unfortunately, makes a lot of sense. the CWMS_AAA code would be in the system classloader and locally here in the app class loader. 

The fix is likely switching this to compile only/provide scope and making sure the api jar is in the system path. I will investigate later. We'll leave it for now.

MikeNeilson (approved):

MikeNeilson:
Well, the initial work is done. Now comes the task of finding someone to really break it.

PR #154, opened 2022-03-10
Features/error report improvements

MikeNeilson:
I'm probably missing most of the places where this should be added but the PR is already getting a bit large.

Primarily moved to a mechanism of controllers and daos and such to be able to say, "yeah this is bad report to user, here's the info" and let a more central place decide just how much to report.

For example a required property can be configured with an annotation and the json library error caught and adjusted.

But a mechanism for more complex logic, like a mutually exclusive set of fields was allow added so that the exact program can be reported to the user vs them having to ask for help and give an incident identifier for well, rather basic issues.


.vscode/launch.json
```diff
@@ -26,22 +26,33 @@
                 "RADAR_DEBUG_LOGGING": "true"
             },
             "sourcePaths": [
-                "build/generated/sources/annotationProcessor/java/main"
+                "**src/main/**",
+                "build/generated/sources/annotationProcessor/java/main",
+                "/home/mike/projects/javalin/**src/main/java/**"
```
rripken: Its a little weird and not portable that /home/mike paths have crept into the repo but I haven't been using vscode so idk.  If its helpful to keep the file in the repo I guess its fine.

cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
```diff
@@ -239,7 +255,7 @@ public static void radarCrud(@NotNull String path, @NotNull CrudHandler crudHand
             throw new IllegalArgumentException("CrudHandler requires a resource base at the beginning of the provided path, e.g. '/users/{user-id}'");
         }
 
-        Map<CrudFunction, Handler> crudFunctions = getHanders(crudHandler, resourceId);
+        Map<CrudFunction, Handler> crudFunctions = CrudHandlerKt.getCrudFunctions(crudHandler, resourceId);//getHanders(crudHandler, resourceId);
```
rripken, line 258: I'm glad you got this working.  I could have sworn I tried to call the function this exact way.

rripken (dismissed):
Its a little weird and not portable that /home/mike paths have crept into the repo but I haven't been using vscode so idk.  If its helpful to keep the file in the repo I guess its fine.

.vscode/launch.json
MikeNeilson: Oops. That was me trying to look at something and those don't need to be there.

rripken (approved):

Issue #153, opened 2022-03-10
endpoint for list of specified levels


MikeNeilson:
While going through and adding documentation to the location levels DTO I realized it wasn't something we had added yet.

PR #165, opened 2022-05-17
Basic Specified Level end-point.

rripken:
Controller to get back a list (just id, office, desc) of available spec levels.  #153


cwms_radar_api/src/main/java/cwms/radar/api/SpecifiedLevelController.java
```diff
@@ -0,0 +1,128 @@
...
+        return Controllers.markAndTime(metrics, getClass().getName(), subject);
+    }
+    
+    
+    @OpenApi(
+        queryParams = {
+            @OpenApiParam(name="office", required=false, description="Specifies the owning office of the Specified Levels whose data is to be included in the response. If this field is not specified, matching rating information from all offices shall be returned."),
+            @OpenApiParam(name="template-id-mask", required=false, description="Mask that specifies the IDs to be included in the response. If this field is not specified, all specified levels shall be returned."),
+        },
+            responses = {
```
DanielTOsborne: Indentation level doesn't match queryParams.
Nitpick, feel free to ignore. :-)

cwms_radar_api/src/main/java/cwms/radar/formatters/json/JsonV1.java
```diff
@@ -42,7 +43,7 @@
 				   Clob.class,
 				   Clobs.class,
 				   TimeSeriesGroup.class,
-				   RecentValue.class
+				   RecentValue.class, SpecifiedLevel.class
```
DanielTOsborne: I thought JSONv1 was specifically for the calls that returned JSON generated by CWMS itself inside of Oracle, whereas v2 was for Radar-generated JSON?
I could be mistaken, as my experience with it was wrt to TimeSeries data only, which does have a CWMS generated version.
Maybe @MikeNeilson can answer that.

cwms_radar_api/src/main/java/cwms/radar/api/SpecifiedLevelController.java
rripken: np. I'll fix this.

cwms_radar_api/src/main/java/cwms/radar/formatters/json/JsonV1.java
rripken: I wasn't aware of that distinction but it sounds plausible.  Its really easy to move these from v1 to v2.  I'll fix it.

MikeNeilson: That was the intent. slightly confusing for new things perhaps but simplifies our design... I hope.

cwms_radar_api/src/main/java/cwms/radar/formatters/json/JsonV1.java
```diff
@@ -43,8 +43,11 @@
 				   Clob.class,
 				   Clobs.class,
 				   TimeSeriesGroup.class,
-				   RecentValue.class, SpecifiedLevel.class
+				   RecentValue.class
 				})
+/**
+ * A Formatter for the calls that returned JSON generated by CWMS itself inside of Oracle.
```
DanielTOsborne, line 49: +1 for the comment. I almost suggested adding one.

DanielTOsborne (approved):

PR #167, opened 2022-05-17
Bug fixes/consistent cursor page

### LINKED ISSUE #83: Cursor/Page parameter is inconsistent between Timeseries and Catalog APIs
The input parameter for the catalog API for the requested cursor is "cursor", while for time series it is "page".

rripken:
Use "page" before "cursor".  Added deprecated path parameters.  Using Controllers method to get metrics.  Closes #83.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
@@ -213,10 +213,10 @@ public class ApiServlet extends HttpServlet {
         get("/", ctx -> ctx.result("Welcome to the CWMS REST API").contentType(Formats.PLAIN));
         radarCrud("/location/category/{category-id}", new LocationCategoryController(metrics), requiredRoles);
         radarCrud("/location/group/{group-id}", new LocationGroupController(metrics), requiredRoles);
-        radarCrud("/locations/{location_code}", new LocationController(metrics), requiredRoles);
+        radarCrud("/locations/{location-id}", new LocationController(metrics), requiredRoles);
         radarCrud("/offices/{office}", new OfficeController(metrics), requiredRoles);
-        radarCrud("/units/{unit_name}", new UnitsController(metrics), requiredRoles);
-        radarCrud("/parameters/{param_name}", new ParametersController(metrics), requiredRoles);
+        radarCrud("/units/{unit-id}", new UnitsController(metrics), requiredRoles);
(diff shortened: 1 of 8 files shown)


cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
```diff
@@ -213,10 +213,10 @@ protected void configureRoutes()
         get("/", ctx -> ctx.result("Welcome to the CWMS REST API").contentType(Formats.PLAIN));
         radarCrud("/location/category/{category-id}", new LocationCategoryController(metrics), requiredRoles);
         radarCrud("/location/group/{group-id}", new LocationGroupController(metrics), requiredRoles);
-        radarCrud("/locations/{location_code}", new LocationController(metrics), requiredRoles);
+        radarCrud("/locations/{location-id}", new LocationController(metrics), requiredRoles);
```
MikeNeilson, line 216: Seems like it would be better to keeps these as `-name` ,or even just `name`.

Since `id` is traditionally used for a numeric value seems like it would reduce confusion.

cwms_radar_api/src/main/java/cwms/radar/api/BlobController.java
```diff
@@ -103,20 +108,17 @@ public void getAll(Context ctx) {
             String office = ctx.queryParam("office");
             Optional<String> officeOpt = Optional.ofNullable(office);
 
+            String cursor = Controllers.queryParamAsClass(ctx, new String[]{"page", "cursor"},
```
MikeNeilson, line 111: well, that's a convenient syntax element they have there.

MikeNeilson: finally scrolled down enough and saw that you actually wrote it. Good idea.

cwms_radar_api/src/main/java/cwms/radar/api/Controllers.java
```diff
@@ -3,16 +3,81 @@
...
+		}
+		else
+		{
+			for(int i = 1; i < names.length; i++)
+			{
+				validator = ctx.queryParamAsClass(names[0], clazz);
+				if(validator.hasValue())
+				{
+					retval = validator.get();
+					metrics.counter(name(className,"deprecated")).inc();
```
MikeNeilson, line 68: excellent idea tracking that metric.

MikeNeilson (approved):

PR #169, opened 2022-06-02
Updated dependency versions where possible. Moved some of the unshare…

rripken:
…d ext named dependency version declarations to the place of use.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,14 +7,10 @@ allprojects {
 
 // versions for shared dependencies
 ext.dropwizard_version="4.1.2"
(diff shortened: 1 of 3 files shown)


rripken:
I'd like to update the tomcat version to 9.0.62 but I wasn't sure. Maybe we are on the version we are b/c it matches servers outside our control?

MikeNeilson:
Bumping the tomcat version is fine. I'll likely be doing that for the server soon anyways (need to spent some time sorting out an RPM update issue.)

MikeNeilson (approved):

PR #172, opened 2022-06-13
Fixed missing array index on loop.

rripken:


cwms_radar_api/src/main/java/cwms/radar/api/Controllers.java
```diff
@@ -9,19 +9,22 @@
 
 public class Controllers
 {
-	private Controllers(){
+
+	private Controllers()
```
MikeNeilson: This should be the `Controllers() {` according to our style guide.

cwms_radar_api/src/main/java/cwms/radar/api/Controllers.java
```diff
@@ -9,19 +9,22 @@
...
-		Meter meter = registry.meter(name(className,subject,"count"));
+		Meter meter = registry.meter(name(className, subject, "count"));
 		meter.mark();
-		Timer timer = registry.timer(name(className,subject,"time"));
+		Timer timer = registry.timer(name(className, subject, "time"));
 		return timer.time();
 	}
 
-	public static <T> T queryParamAsClass(io.javalin.http.Context ctx, String[] names, Class<T> clazz, T defaultValue){
+	public static <T> T queryParamAsClass(io.javalin.http.Context ctx, String[] names, Class<T> clazz, T defaultValue)
```
MikeNeilson: same comment as line12

MikeNeilson (changes requested):
logic looks good, but run checkstyle. or pitch a change thereto.

MikeNeilson (approved):
though it looks like you may have run into that annoying line ending issue. small enough file I'm not super worried about it though.

PR #182, opened 2022-07-08
Bug fixes/checkstyle format api

rripken:
I used my IDE to reformat the cwms.radar.api package to address a large number of checkstyle issues and then manually tweak lines I saw that still didn't look right.


MikeNeilson (approved):
well, I was doing the same but I ain't gonna argue with help.

PR #183, opened 2022-07-08
Fixed order of OutRange and InRange fields.

rripken:
When the builder gets a string like "LINEAR,NEAREST,LINEAR" its splits it and assigns the parts.  According to the pl/sql package documentation I was assigning the parts in the wrong order.


MikeNeilson (commented):
I think the formatting change that merged in broke this. It looks like the whole file was deleted and changed.

rripken:
Its really just a couple of line in the withDateMethods method.  I could do it as two pulls requests to make it cleaner.

PR #189, opened 2022-07-21
added test dataset and rebase lined integration tests.

MikeNeilson:
- [x] expand ts catalog integration test to use the new giant list of data.


MikeNeilson:
The test data set needs a little work (I let a lot of duplicates in so the actual data is smaller than I thought.) but I'll work on that when I handle the location catalog.

rripken (approved):

PR #203, opened 2022-08-10
Feature/location levels

rma-rripken:
Created a LocationLevels class to return pages of LocationLevel data.   
Added a getOne for LocationLevel.  
Added ability to build a LocationLevel from a JDomLocationLevelImpl.  
Added query and result parsing from CwmsLevelJdbcDao but done Jooq style.

### DIFF
--- a/cwms_radar_api/build.gradle
+++ b/cwms_radar_api/build.gradle
@@ -12,8 +12,8 @@ dependencies {
     implementation "javax.xml.bind:jaxb-api:2.3.1"
     implementation "com.sun.xml.bind:jaxb-impl:3.0.2"
     implementation "com.sun.xml.bind:jaxb-core:3.0.2"
-    implementation "mil.army.usace.hec:cwms-db-dao:9.0.2"
-    implementation( "mil.army.usace.hec:cwms-db-jooq:9.0.2"){
+    implementation "mil.army.usace.hec:cwms-db-dao:9.0.4"
+    implementation( "mil.army.usace.hec:cwms-db-jooq:9.0.4"){
         exclude group: "com.oracle", module: "*"
         exclude group: "com.oracle.database.jdbc", module: "*"
         exclude group: "org.jooq.pro", module: "*"
@@ -30,6 +30,7 @@ dependencies {
         //exclude group: "mil.army.usace.hec.swingx"
         exclude group: "*"
     }
+    implementation "mil.army.usace.hec:hec-nucleus-data:1.0.33"
     implementation "mil.army.usace.hec:hec-nucleus-metadata:1.0.33"
     implementation( 'mil.army.usace.hec:hec-cwms-ratings-core:1.0.0' ) {
         //exclude group: "org.python", module: "jython-standalone"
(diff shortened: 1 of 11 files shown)


cwms_radar_api/build.gradle
```diff
@@ -12,8 +12,8 @@ dependencies {
     implementation "javax.xml.bind:jaxb-api:2.3.1"
     implementation "com.sun.xml.bind:jaxb-impl:3.0.2"
     implementation "com.sun.xml.bind:jaxb-core:3.0.2"
-    implementation "mil.army.usace.hec:cwms-db-dao:9.0.2"
-    implementation( "mil.army.usace.hec:cwms-db-jooq:9.0.2"){
+    implementation "mil.army.usace.hec:cwms-db-dao:9.0.4"
```
MikeNeilson, line 15: Has it dawned on any of us much much of an almost circular dependency this is?

Not saying remove it it. Just felt like it needed to be said.

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
```diff
@@ -193,74 +187,150 @@ public void delete(@NotNull Context ctx, String id) {
...
-    @OpenApi(ignore = true)
+    @OpenApi(
+            pathParams = {
+                    @OpenApiParam(name = "id", required = true, description = "Specifies"
+                            + " the requested location level."),
+            },
+            queryParams = {
+                    @OpenApiParam(name = "office", required = true, description = "Specifies the "
+                            + "office of the Location Level to be returned"),
+                    @OpenApiParam(name = "date", required = true, description = "Specifies the "
```
MikeNeilson: should this be "date-time" since it can go down the 2nd?

cwms_radar_api/src/main/java/cwms/radar/formatters/json/JsonV2.java
```diff
@@ -1,7 +1,5 @@
 package cwms.radar.formatters.json;
```
MikeNeilson, line 1: Thanks for fixing the formatting.

MikeNeilson (commented):
Looks okay to me. Just some nitpicks on names to maybe address maybe not.

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
rma-rripken: Maybe?  I didn't think too hard about it b/c the update and delete already had a "date" field.  But maybe all three should be "date-time" or something more explicit like "effective-date" ?

MikeNeilson: I like effective-date more. that's explicitly what we're getting/removing/or updating.

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
```diff
@@ -47,10 +50,15 @@
 
 public class LevelsController implements CrudHandler {
     private static final Logger logger = Logger.getLogger(LevelsController.class.getName());
+    public static final String EFFECTIVE_DATE = "effective-date";
+    public static final String OFFICE = "office";
+    public static final String DATE = "date";
```
MikeNeilson, line 55: good idea. later I'll pull these into a separate class for the other controllers to use.

MikeNeilson (approved):

PR #204, opened 2022-08-11
How the controller is registered in ApiServlet determines the pathPar…

rma-rripken:
…ameter name and needs to match the @pathParameter documentation.

Unfortunately I mixed in lots of cleanup.   The major change was ApiServlet line 232.   LevelsController is registered as 
 radarCrud("/levels/{" + LevelsController.LEVEL_ID + "}",   

The swagger documentation for LevelController   getOne, update, delete were similarly updated with a level-id parameter.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
@@ -1,21 +1,8 @@
 package cwms.radar;
 
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.sql.Connection;
-import java.sql.SQLException;
-import java.util.Arrays;
-import java.util.Map;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-import javax.annotation.Resource;
-import javax.servlet.ServletConfig;
-import javax.servlet.ServletException;
-import javax.servlet.annotation.WebServlet;
-import javax.servlet.http.HttpServlet;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-import javax.sql.DataSource;
+import static io.javalin.apibuilder.ApiBuilder.get;
+import static io.javalin.apibuilder.ApiBuilder.prefixPath;
+import static io.javalin.apibuilder.ApiBuilder.staticInstance;
 
 import com.codahale.metrics.Meter;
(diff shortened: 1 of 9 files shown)


cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
```diff
@@ -196,82 +202,113 @@ public void init() throws ServletException{
...
+    private AccessManager buildAccessManager() {
         return new CwmsAccessManager();
     }
 
 
 
 
-    protected void configureRoutes()
-    {
+    protected void configureRoutes() {
```
MikeNeilson, line 212: Both are fine here, but FYI if it definitely makes more visual sense in a function to have long lines you can put a `@SuppressWarning("linelength")` (might not be exact name) to have checkstyle ignore it there.

I ran into a few of those when I was doing my big push.

cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
```diff
@@ -196,82 +202,113 @@ public void init() throws ServletException{
...
+     * @param roles the accessmanager will require these roles are present to access post, patch
+     *             and delete methods
      */
-    public static void radarCrud(@NotNull String path, @NotNull CrudHandler crudHandler, @NotNull RouteRole... roles) {
+    public static void radarCrud(@NotNull String path, @NotNull CrudHandler crudHandler,
+                                 @NotNull RouteRole... roles) {
         String fullPath = prefixPath(path);
-        String[] subPaths = Arrays.stream(fullPath.split("/")).filter(it -> !it.isEmpty()).toArray(String[]::new);
+        String[] subPaths = Arrays.stream(fullPath.split("/"))
+                .filter(it -> !it.isEmpty()).toArray(String[]::new);
```
MikeNeilson, line 275: not sure if we can codify this in checkstyle (and you don't need to change this) but I've always preferred the visual of the following:

```java 
...  = Array.stream(fullPath.split("/"))
                .filter(it -> !it.isEmpty())
                .toArray(String[]::new);
```

Always seems clearer to me if each operation is on it's own line.

MikeNeilson (approved):
would not have expected that with the use of the "implementation" parameter. 
I'm pretty sure we put that in originally for something to work right.

Did you sanity check the swagger-ui?

PR #215, opened 2022-08-30
Allow timeseries writes.

MikeNeilson:
Modified to use content-type header for check instead of the accept
header.
Uses contentType.toString() get full content type.(if was checking
application/json against application/json;version=2 when both should've
been version=2).

Modified StoreRule Enum to generate appropriate string. The constant
value would only work if used in an actual SQL query.

- [ ] I still need to fix the openapi docs. For some reason it's completely expanding sub times like Duration. Also need to document what is and isn't required for write.


MikeNeilson:
This needs an integration test but I'm also working on some tweaks to the accessmanager. I'll add those in with that PR to also test that.

cwms_radar_api/src/main/java/cwms/radar/data/dao/StoreRule.java
```diff
@@ -9,13 +9,14 @@
...
 	REPLACE_WITH_NON_MISSING(CWMS_UTIL_PACKAGE.REPLACE_WITH_NON_MISSING.toString()),
-	DELETE_INSERT(CWMS_UTIL_PACKAGE.DELETE_INSERT.toString());
-
+	DELETE_INSERT(CWMS_UTIL_PACKAGE.DELETE_INSERT.toString());	
 	private final String rule;
 
 	StoreRule(String rule)
 	{
-		this.rule = rule;
+		System.out.println(rule);
```
rma-rripken, line 17: Remove the System.out

cwms_radar_api/src/main/java/cwms/radar/data/dto/TimeSeries.java
```diff
@@ -52,38 +53,63 @@ public class TimeSeries extends CwmsDTOPaginated {
...
+        schema = @Schema(
+            description = "List of retrieved time-series values",
+            implementation = Record.class
+        )
+    )
     List<Record> values;
 
+    @Schema(
+        accessMode = AccessMode.READ_ONLY,
+        description = "Information about where the measurement takes place in relation to a defined dataum."
```
rma-rripken, line 98: dataum should probably be datum

cwms_radar_api/src/main/java/cwms/radar/data/dto/TimeSeries.java
```diff
@@ -52,38 +53,63 @@ public class TimeSeries extends CwmsDTOPaginated {
...
     // Use the array shape to optimize data transfer to client
     @JsonFormat(shape=JsonFormat.Shape.ARRAY)
-    @Schema(description = "List of retrieved time-series values")
+    @ArraySchema(
+        schema = @Schema(
+            description = "List of retrieved time-series values",
+            implementation = Record.class
+        )
+    )
     List<Record> values;
```
rma-rripken, line 94: I don't appreciate the usefulness of the @ArraySchema annotation.  I'm not saying this sarcastically, I really don't know what it does.  I tried these changes and one of the things that changed was that the $ref field on TimeSeries.items went away.  That $ref field makes the codegen Adam is using map our List to a Record[].  Maybe there are more annotations we can add to make the codegen work once again with the ArraySchema.  I don't know.  It seems like telling swagger that its an array of Record objects could only help but the schema that gets generated doesn't contain anything linking the field to Record.  The generated schema does describe the shape of the array items so it seems like sufficiently smart codegen could still make it work.  Once these changes get merged we should have Adam redo codegen.

Screenshot 2022-09-02 124910

cwms_radar_api/src/main/java/cwms/radar/api/TimeSeriesController.java
```diff
@@ -356,7 +356,10 @@ public void getOne(Context ctx, String id) {
     }
 
     @OpenApi(
-            description = "Update a TimeSeries",
+            description = "Update a TimeSeries with provided values",
+            pathParams = {
+                @OpenApiParam(name="name",description = "Full CWMS Timeseries name")
```
rma-rripken, line 361: We should review the swagger docs and decide on TimeSeries, Timeseries, Time series, Time-series.  I'm definitely guilty of being inconsistent.

cwms_radar_api/src/main/java/cwms/radar/data/dto/TimeSeries.java
```diff
@@ -38,6 +41,7 @@
 @XmlAccessorType(XmlAccessType.FIELD)
 @XmlAccessorOrder(XmlAccessOrder.ALPHABETICAL)
 @JsonPropertyOrder(alphabetic = true)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
```
rma-rripken, line 45: I assume this changed how the JSON is generated?  There is a unit test TimeSeriesTest that does some serialization tests and currently checks some of the xml fields.  This change seems like it would be easy to test in that class

MikeNeilson: Gah, thanks. And yes, I think I updated the json but need to verify again and make sure that's still required. There was some oddness with how the names were generated in the openapi schema vs when the formatters actually rendered it.

cwms_radar_api/src/main/java/cwms/radar/data/dto/TimeSeries.java
MikeNeilson: I'll have to dig into that as well. It was necessary to get the example to show up correctly in the swagger ui, otherwise it doesn't render for the user correctly.
However one can modify the returned schema. so I could probably work something so that it's only rendered that was for the swagger-ui.

But hopefully there's a way to fix the ref, maybe now that I have an example that gets render it's not required.

cwms_radar_api/src/main/java/cwms/radar/data/dto/TimeSeries.java
MikeNeilson: .... oops.

cwms_radar_api/src/main/java/cwms/radar/data/dao/StoreRule.java
MikeNeilson: Thank you. You know the really sad part is that probably would've shown up nicely in the checkstyle report if we didn't still have a bunch of things to fix. (Thanks for slowly fixing those each PR btw.)

rma-rripken (approved):

PR #217, opened 2022-09-07
Feature/use data source with jooq

rma-rripken:


cwms_radar_api/src/main/java/cwms/radar/datasource/ConnectionPreparer.java
```diff
@@ -0,0 +1,8 @@
+package cwms.radar.datasource;
```
MikeNeilson, line 1: That's a neat trick.

MikeNeilson (approved):

PR #222, opened 2022-09-21
Feature/rating metadata

rma-rripken:


cwms_radar_api/src/main/java/cwms/radar/api/RatingMetadataController.java
```diff
@@ -0,0 +1,155 @@
...
+
+        try (final Timer.Context timeContext = markAndTime("getAll");
+             DSLContext dsl = getDslContext(ctx)) {
+            RatingMetadataDao dao = getDao(dsl);
+
+            // I know data
+            RatingMetadataList metadataList = dao.retrieve(cursor, pageSize, office,
+                    ratingIdMask);
+
+            long end = System.nanoTime();
```
MikeNeilson: this might be appropriate to wrap in an additional timer context directory. that way the metrics system can logs the time.

cwms_radar_api/src/main/java/cwms/radar/data/dao/RatingMetadataDao.java
```diff
@@ -0,0 +1,221 @@
...
+        fields.add(specView.OFFICE_ID);
+        fields.add(specView.TEMPLATE_ID);
+        fields.add(specView.RATING_ID);
+        fields.add(ratView.OFFICE_ID);
+        fields.add(ratView.RATING_ID);
+        fields.add(ratView.EFFECTIVE_DATE.desc());
+
+
+        SelectSeekStepN<Record> records = dsl.select(specView.RATING_SPEC_CODE,
+                        specView.RATING_ID,
```
MikeNeilson: I have exactly zero idea what it should/could look like but FYI at this length it's not unreasonable to group a few columns into the same line. It would give the read a change to have the column list within the screen at the same time as the rest of the query.

cwms_radar_api/src/main/java/cwms/radar/data/dao/RatingMetadataDao.java
```diff
@@ -0,0 +1,221 @@
...
+                        specView.DATE_METHODS,
+                        specView.AUTO_MIGRATE_EXT_FLAG,
+                        specView.AUTO_ACTIVATE_FLAG,
+                        specView.AUTO_UPDATE_FLAG,
+                        specView.ACTIVE_FLAG,
+                        specView.SOURCE_AGENCY,
+                        specView.LOC_ALIAS_GROUP,
+                        specView.LOC_ALIAS_CATEGORY,
+                        specView.ALIASED_ITEM,
+                        ratView.OFFICE_ID,
```
MikeNeilson: this this OFFICE_ID ever be different from the spec OFFICE_ID? if not we also need to keep one.

MikeNeilson (dismissed):
This looks reasonable to me. Do we have test data though? Or can we steal some of the work Adam just did on the ratings code?

cwms_radar_api/src/main/java/cwms/radar/data/dao/RatingMetadataDao.java
rma-rripken: I was just about to dig into the office ids and look at them more closely.  Seems like they should be part of the join, right?

MikeNeilson: generally yes, it helps the planner eliminate a lot of possible rows and speeds things up..

cwms_radar_api/src/main/java/cwms/radar/api/RatingMetadataController.java
rma-rripken: Whoops, that log statement was just my curiousity. That wasn't supposed to be there.  On my machine each page takes about 6 seconds.  Not great but its a start.

cwms_radar_api/src/main/java/cwms/radar/data/dao/RatingMetadataDao.java
```diff
@@ -0,0 +1,244 @@
...
+                        ratView.OFFICE_ID, ratView.RATING_ID, ratView.EFFECTIVE_DATE,
+                        ratView.CREATE_DATE, ratView.TRANSITION_DATE,
+                        ratView.VERSION, ratView.DESCRIPTION,
+                        ratView.DATABASE_UNITS, ratView.NATIVE_UNITS,
+                        ratView.ACTIVE_FLAG,
+                        ratView.FORMULA, ratView.ALIASED_ITEM
+                )
+                .from(specView)
+                .leftJoin(ratView)
+                .on(specView.RATING_ID.eq(ratView.RATING_ID)
```
MikeNeilson: put the office ID first. it's super weird but it does tend to matter what order certain columns are in. Gives the planner a hint. otherwise it *may* try to get all the ratings and then eliminate by office. but we already know the by office set of ratings is smaller than the total set of ratings.

cwms_radar_api/src/main/java/cwms/radar/data/dao/RatingMetadataDao.java
rma-rripken: good point.

MikeNeilson (dismissed):

cwms_radar_api/src/main/java/cwms/radar/data/dao/RatingMetadataDao.java
rma-rripken: I got rid of the extra columns and moved the seek columns into the Keyset so they are together with the values.

cwms_radar_api/src/main/java/cwms/radar/api/RatingMetadataController.java
rma-rripken: - [ ] Testing here lacking.  I need to get together with Adam and find out what he's done.

MikeNeilson (approved):

PR #223, opened 2022-09-26
Bugfix/timeseries format

rma-rripken:


cwms_radar_api/src/main/java/cwms/radar/formatters/Formats.java
```diff
@@ -185,7 +185,7 @@ public static String format(ContentType type, List<? extends CwmsDTO> toFormat,
      */
     public static ContentType parseHeaderAndQueryParm(String header, String queryParam) {
         if (queryParam != null && !queryParam.isEmpty()) {
-            if (header != null && !header.isEmpty()) {
+            if (header != null && !header.isEmpty() && !"*/*".equals(header.trim())) {
```
rma-rripken, line 188: This is the only substantive change in this file.

MikeNeilson (approved):

PR #224, opened 2022-09-27
Make the default load method EAGER.

rma-rripken:


MikeNeilson (approved):

adamkorynta:
Should probably specify in the swagger doc what the default is

PR #226, opened 2022-10-03
Feature/rating metadata

rma-rripken:
Based on cwms-database-dev meeting I switch to the pl/sql retrieve and stopped using the rating views.


cwms_radar_api/src/main/java/cwms/radar/data/dao/RatingTemplateDao.java
```diff
@@ -233,7 +235,7 @@ private Set<RatingTemplate> getRatingTemplates(String office, String templateIdM
         }
 
         SelectLimitStep<? extends Record> innerSelect = dsl.select(
-                        OracleDSL.rownum().as("rnum"),
+                        OracleDSL.rownum().as(RNUM),
```
MikeNeilson, line 238: This may break in a few cases. At least I vaguely remember using the row number causing issues with pagination.

MikeNeilson (approved):

MikeNeilson:
Probably good to wait for Adam, but conceptually it all looks okay to me. the Query with the row num is a little iffy, but honestly it if works we can deal with it later. The task at hand is can the client read it, not is it the fastest thing in the world (not that that's unimportant.)

PR #228, opened 2022-10-04
Adding subTypes and @Schema annotations based on QuestionDTO example.

rma-rripken:
Adding these annotations seems to make the codegen happier.


cwms_radar_api/src/main/java/cwms/radar/data/dto/rating/AbstractRatingMetadata.java
```diff
@@ -15,23 +19,31 @@
                 ExpressionRating.class,
                 UsgsStreamRating.class
         },
-        discriminatorProperty = "ratingType",
+        discriminatorProperty = "rating-type",
         discriminatorMapping = {
                 @DiscriminatorMapping(value = TableRating.RATING_TYPE, schema = TableRating.class),
```
MikeNeilson, line 24: I wonder if this will work well for the text timeseries. I suspect so and am glad you found it.

MikeNeilson (approved):

Issue #245, opened 2022-10-14
Develop method to catch Swagger schema vs serialized output mismatch


rma-rripken:
Adam recently pointed out that the Swagger schema for TimeSeriesExtents didn't match the serialized output. 
What happened is that TimeSeriesExtents gets serialized by JSONv2 which sets Kebab case on its object mapper.  Anything that goes thru JSONv2 that doesn't specify its desired naming is going to get kebab-case.
The Swagger schema is generated by Swagger looking at the return type of the method.  In this case the return type includes TimeSeriesExtents object and when Swagger inspects the fields on TimeSeriesExtents they don't have any naming overrides on them so Swagger assumes they are going to be camel-case.

If you code-gen a client from the Swagger schema (which is one of the benefits of publishing the schema) your client is going to expect "earliestTime" but the field you will actually get in the json is "earliest-time".

The fix for this in TimeSeriesExtents is just to annotate the object correctly.  Like this:

@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
public class TimeSeriesExtents {...

I'd like to figure out a way to catch these errors before clients do.  

1)  One way would be to codegen our own clients from the schema and serialization test with them but that seems a little heavy.

2) Maybe there is a swagger config setting somewhere like the flag was set on jackson serializer to use kebab - maybe there is a flag we can set on swagger so that it uses kebab?

3) We could just ensure that we annotate all the DTO reachable classes - this would work but it seems like it would be really easy to miss a class.  This is what we've been doing so far but it doesn't ensure that we don't miss a class.

4) Maybe we could write a util method that serializes an object with an object mapper that has the kebab flag set and then again with another object mapper that doesn't have the flag set and compare the results.  If something changed than the class is missing annotations.  We'd still have to have a test case that calls the util method for each class but that might not be too bad if it was as easy as: object=build();  serializeTwoWays(object);
 
5) If its just about annotations, maybe we could write something that looks at all the DTO classes and inspects their annotations and makes sure there is some sort of @JsonNaming annotation on each class?  Might not be too bad.

6) The real problem is that the output from the controller doesn't match the Swagger schema.  Maybe we could attack the problem head-on and just get the swagger schema and then make sure each field in the serialized output matches the name of the field in the schema.  This could be done a variety of ways.  It might even be possible to have GPT-3 do it...  R&D?

rma-rripken:
OpenApiOptions has several methods that might accomplish approach #2 and do what we need.  
[...]

PR #249, opened 2022-10-18
Change the model converter to generate schema with kebab as default

rma-rripken:
This could partially address #245.  It doesn't catch it but it makes it harder for the mismatch to occur.   This PR should ideally have a test.   Perhaps remove the TimeSeriesExtents annotation   @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)  and verify that it still JsonV2 serializes as "earliest-time" and then also pull the schema and verify that the schema contains "earliest-time".  Just a thought.


MikeNeilson:
that test sounds obnoxious. honestly i might not be too difficult to right a generic processor that checks for camelCase in the elements. Something like the checkstyle matcher for names.

cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
```diff
@@ -325,6 +326,16 @@ private OpenApiOptions getOpenApiOptions() {
                     doc.json("404", RadarError.class);
                 })
                 .activateAnnotationScanningFor("cwms.radar.api");
+
+        // This next section is to make Swagger use KEBAB-CASE by default for
```
MikeNeilson: I swear I tried doing this and it didn't work. I'm glad it did for you.

MikeNeilson (dismissed):

Issue #248, opened 2022-10-18
Add endpoint providing the Major.Minor version of RADAR


adamkorynta:
This will allow clients to enable/disable features that are not supported by instance of RADAR.

MikeNeilson:
We also want this for some performance checks that are in the works in a separate repo.

rma-rripken:
We have a version string.  Its in the info section of the schema.  We just haven't been incrementing it.

rma-rripken:
#250

MikeNeilson:
oh, good point, that actually should just match the gradle version number.

rma-rripken:
It would be nice to be able to version each end-point.  
I don't see a place in the schema to do that.  There are tags, We could tag them with their version but that doesn't seem right.

MikeNeilson:
Yeah that would get a bit much.

adamkorynta:
Seems less useful than overall version number and would only be handy during simultaneous client+server development.

PR #250, opened 2022-10-18
Increment the version number.

rma-rripken:
Increment version string.  #248

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
@@ -102,7 +102,10 @@ public class ApiServlet extends HttpServlet {
     public static final String OFFICE_ID = "office_id";
     public static final String DATA_SOURCE = "data_source";
     public static final String DATABASE = "database";
-    public static final String VERSION = "3.0";
+
+    // The VERSION should match the gradle version but not contain the patch version.
+    // For example 2.4 not 2.4.13
+    public static final String VERSION = "2.4";
 
     private MetricRegistry metrics;
     private Meter totalRequests;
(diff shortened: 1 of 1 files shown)


cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
```diff
@@ -102,6 +102,7 @@ public class ApiServlet extends HttpServlet {
     public static final String OFFICE_ID = "office_id";
     public static final String DATA_SOURCE = "data_source";
     public static final String DATABASE = "database";
+    public static final String VERSION = "3.0";
```
MikeNeilson: why are we jumping to 3?

rma-rripken: It doesn't have to be 3.0.  2.1?

MikeNeilson: shove the gradle version property into a resource file and pull it from there?

rma-rripken: oh. 2.4 to match the gradle.

MikeNeilson: or at least the major/minor. There's a use case to know if it's def newer to run performance checks but I can add a hash laster.

MikeNeilson (approved):

adamkorynta:
Still would like it to grab the version from gradle instead of duplicating....

rma-rripken:
I thought about figuring out how to pull it from gradle or have the build put it in a resource and ... well, maybe next release?

MikeNeilson:
agreed, not hard but this at least works for now.

Issue #243, opened 2022-10-14
catalog/timeseries cursor issue


adamkorynta:
When comparing the rows returned by the JDBC timeseries catalog and the RADAR timeseries catalog for the SWT database. We noticed that the RADAR catalog gets cut off after a certain point. The next-page cursor is returning null, but there is still more data.

I've added two documents showing the difference.

[jdbc_swt.txt](https://github.com/USACE/cwms-radar-api/files/9789121/jdbc_swt.txt)
[leary_swt.txt](https://github.com/USACE/cwms-radar-api/files/9789122/leary_swt.txt)

For reference our page size is set to 1,000 and the only filter we are using is the "SWT" office.

MikeNeilson:
Is is against cwms-data.usace.army.mil or a local instance?

adamkorynta:
Local instance off of the latest in the main branch.

adamkorynta:
I'm seeing the same issue with the location's catalog. I never get a next page cursor

MikeNeilson:
ugh. didn't we deal with something similar barely 2 months ago?

adamkorynta:
Sounds remarkably similar to: https://github.com/USACE/cwms-radar-api/issues/186

MikeNeilson:
well, apparently I need to go through that again with the office id specified.... ugh. Too many broken things.

adamkorynta:
Fixed along with: https://github.com/USACE/cwms-data-api/issues/186

PR #253, opened 2022-10-19
#243 Bugfix/catalog locations - switch to AV_LOC2 for querying

adamkorynta:
Comparing apples to oranges since the original query wasn't giving us back the correct data, but I was able to take the first page of data down from 1 minute to 1 second on my environment.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
@@ -40,6 +40,7 @@ import org.jooq.Record1;
 import org.jooq.SelectConditionStep;
 import org.jooq.SelectLimitPercentStep;
 import org.jooq.SelectSeekStep2;
+import org.jooq.Table;
 import org.jooq.exception.DataAccessException;
 import usace.cwms.db.dao.ifc.loc.CwmsDbLoc;
 import usace.cwms.db.dao.util.services.CwmsDbServiceLookup;
@@ -283,11 +284,10 @@ public class LocationsDaoImpl extends JooqDao<Location> implements LocationsDao 
 
         String locCodeField = "LOCATION_CODE_COL";
         SelectLimitPercentStep<Record1<Long>> forLimit = dsl.select(AV_LOC2.AV_LOC2.LOCATION_CODE.as(locCodeField))
-            .from(AV_LOC2.AV_LOC2)
+            .from(AV_LOC2.AV_LOC2.as("LOCATION_CODES"))
             .where(condition.and(AV_LOC2.AV_LOC2.ALIASED_ITEM.isNull()))
             .orderBy(AV_LOC2.AV_LOC2.LOCATION_ID)
             .limit(pageSize);
-
 
         SelectConditionStep<Record> query = dsl.select(AV_LOC2.AV_LOC2.asterisk())
             .from(AV_LOC2.AV_LOC2)
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
@@ -1,5 +1,6 @@
 package cwms.radar.data.dao;
 
+import static java.util.Comparator.comparing;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
@@ -280,18 +281,18 @@ public class LocationsDaoImpl extends JooqDao<Location> implements LocationsDao 
 
         condition = condition.and(AV_LOC2.AV_LOC2.LOCATION_ID.upper().greaterThan(locCursor));
 
-        SelectLimitPercentStep<Record1<Long>> forLimit = dsl.select(AV_LOC2.AV_LOC2.LOCATION_CODE)
+        String locCodeField = "LOCATION_CODE_COL";
+        SelectLimitPercentStep<Record1<Long>> forLimit = dsl.select(AV_LOC2.AV_LOC2.LOCATION_CODE.as(locCodeField))
             .from(AV_LOC2.AV_LOC2)
             .where(condition.and(AV_LOC2.AV_LOC2.ALIASED_ITEM.isNull()))
             .orderBy(AV_LOC2.AV_LOC2.LOCATION_ID)
             .limit(pageSize);
 
 
-        SelectSeekStep2<Record, String, String> query = dsl.select(AV_LOC2.AV_LOC2.asterisk())
+        SelectConditionStep<Record> query = dsl.select(AV_LOC2.AV_LOC2.asterisk())
             .from(AV_LOC2.AV_LOC2)
             .where(condition)
-            .and(AV_LOC2.AV_LOC2.LOCATION_CODE.in(forLimit))
-            .orderBy(AV_LOC2.AV_LOC2.ALIASED_ITEM.desc(), AV_LOC2.AV_LOC2.LOCATION_ID);
+            .and(AV_LOC2.AV_LOC2.LOCATION_CODE.in(forLimit));
 
         List<? extends CatalogEntry> entries = query.fetch()
             .stream()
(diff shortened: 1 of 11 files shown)


adamkorynta:
This catalog query does not take the office id into account for pagination like the timeseries catalog does. I believe this is because the query is not ordered by office id anywhere. I didn't change this behavior.

rma-rripken (dismissed):

MikeNeilson:
> This catalog query does not take the office id into account for pagination like the timeseries catalog does. I believe this is because the query is not ordered by office id anywhere. I didn't change this behavior.

That's unfortunate, it should. I now pretty certain the main issue here is I worked on timeseries and ignored locations until someone said something.

MikeNeilson:
> Comparing apples to oranges since the original query wasn't giving us back the correct data, but I was able to take the first page of data down from 1 minute to 1 second on my environment.

Sometimes the Orange is just better.

cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
```diff
@@ -248,186 +250,129 @@ public static Feature buildFeatureFromAvLocRecord(Record avLocRecord) {
...
-                        .sorted((left, right) -> (
-                                left.getKey().getLOCATION_ID()
-                                        .compareTo(right.getKey().getBASE_LOCATION_ID())
-                        ))
-                        .map(e -> buildCatalogEntry(e.getKey(), e.getValue())
-                        ).collect(Collectors.toList());
-
+        condition = condition.and(AV_LOC2.AV_LOC2.LOCATION_ID.upper().greaterThan(locCursor));
+
+        SelectLimitPercentStep<Record1<Long>> forLimit = dsl.select(AV_LOC2.AV_LOC2.LOCATION_CODE)
```
MikeNeilson: This should get a fixed name. I've noticed that JOOQ will often build a new query each time and give a random identifier. Reeks havoc on the shared memory for plans.

MikeNeilson: True of any subselect.

cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
```diff
@@ -248,186 +250,129 @@ public static Feature buildFeatureFromAvLocRecord(Record avLocRecord) {
...
+        SelectSeekStep2<Record, String, String> query = dsl.select(AV_LOC2.AV_LOC2.asterisk())
+            .from(AV_LOC2.AV_LOC2)
+            .where(condition)
+            .and(AV_LOC2.AV_LOC2.LOCATION_CODE.in(forLimit))
+            .orderBy(AV_LOC2.AV_LOC2.ALIASED_ITEM.desc(), AV_LOC2.AV_LOC2.LOCATION_ID);
+
+        List<? extends CatalogEntry> entries = query.fetch()
+            .stream()
+            .map(r -> r.into(AV_LOC2.AV_LOC2))
+            .collect(groupingBy(usace.cwms.db.jooq.codegen.tables.records.AV_LOC2::getLOCATION_CODE))
```
MikeNeilson, line 300: As much as I love that particular operation does it exist in java 8? Though I might be getting that mixed up with an that works inline with the stream.

MikeNeilson (commented):
Looks good. just needs a sanity check for my brain on the places I commented on.

cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
adamkorynta: yeah everything here is valid Java 8, I'm running and compiling in JDK 8 for everything RADAR.

cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
adamkorynta: Thanks for the catch, your warning from awhile back is now ringing in my ears.

cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
```diff
@@ -280,18 +281,18 @@ public Catalog getLocationCatalog(String cursor, int pageSize, String unitSystem
 
         condition = condition.and(AV_LOC2.AV_LOC2.LOCATION_ID.upper().greaterThan(locCursor));
 
-        SelectLimitPercentStep<Record1<Long>> forLimit = dsl.select(AV_LOC2.AV_LOC2.LOCATION_CODE)
+        String locCodeField = "LOCATION_CODE_COL";
+        SelectLimitPercentStep<Record1<Long>> forLimit = dsl.select(AV_LOC2.AV_LOC2.LOCATION_CODE.as(locCodeField))
```
MikeNeilson, line 286: forLimit should have an "as name".

MikeNeilson: Then again I might be wrong here. worth printing the query to see what JOOQ does, that might only be for actual tables.

adamkorynta: I didn't see one on `forLimit` itself. The generated query looks like:

```
select "CWMS_20"."AV_LOC2".*
from "CWMS_20"."AV_LOC2"
where (
                  "CWMS_20"."AV_LOC2"."UNIT_SYSTEM" = 'SI'
              and (regexp_like("CWMS_20"."AV_LOC2"."LOCATION_ID", '.*', 'i'))
              and (regexp_like("CWMS_20"."AV_LOC2"."DB_OFFICE_ID", 'SWT', 'i'))
              and upper("CWMS_20"."AV_LOC2"."LOCATION_ID") > '*'
              and "CWMS_20"."AV_LOC2"."LOCATION_CODE" in (
              select "v0" "LOCATION_CODE_COL"
              from (
                       select
                           "x"."v0",
                           rownum "rn"
                       from (
                                select "CWMS_20"."AV_LOC2"."LOCATION_CODE" "v0"
                                from "CWMS_20"."AV_LOC2" "LOCATION_CODES"
                                where (
                                                  "CWMS_20"."AV_LOC2"."UNIT_SYSTEM" = 'SI'
                                              and (regexp_like("CWMS_20"."AV_LOC2"."LOCATION_ID", '.*', 'i'))
                                              and (regexp_like("CWMS_20"."AV_LOC2"."DB_OFFICE_ID", 'SWT', 'i'))
                                              and upper("CWMS_20"."AV_LOC2"."LOCATION_ID") > '*'
                                              and "CWMS_20"."AV_LOC2"."ALIASED_ITEM" is null
                                          )
                                order by "CWMS_20"."AV_LOC2"."LOCATION_ID"
                            ) "x"
                       where rownum <= (0 + 1000)
                   )
              where "rn" > 0
          )
          )
```

Looks to me like `limit` is expanding in a funky way.

rma-rripken: I just think that is how jooq handles limit and offset for oracle.

See the oracle example in this https://www.jooq.org/doc/latest/manual/sql-building/sql-statements/select-statement/limit-clause/

keyset pagination is in someways much better but we'd have to redo how the cursors are generated
https://blog.jooq.org/faster-sql-paging-with-jooq-using-the-seek-method/
I did this briefly for the abstract ratings before it got thrown out.

MikeNeilson: Ryan is correct, oracle doesn't handled the limit quite right so that's what JOOQ does.

I'm certainly not opposed to the keyset pagination... mostly I just keep having issues trying to open the jooq blog on my work machine so I still haven't been able to read it.

But you're right. the name isn't required for the forLimit step here, I don't see a randomly generated "table" name.

MikeNeilson (approved):

cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationsDaoImpl.java
rma-rripken: The tl;dr version of offset vs keyset pagination is that instead of skipping a certain number of rows (offset) you order the rows by a bunch of columns and you pass around the last values of those columns that you saw in the previous page as the cursor.   Jooq made a "seek" function to make writing the queries easier.  The columns to sort by and the values in the seek clause are coupled.  seek is supposed to be more efficient as the number of pages grows b/c the db doesn't have to count the precise number of rows to skip - you may end up needing to add indexes to the sort columns but that's the sort of thing dba's get paid to figure out -right?

DSL.using(configuration)
   .select(T.ID, T.VALUE)
   .from(T)
   .orderBy(T.VALUE, T.ID)
   .seek(2, 533)
   .limit(5);

PR #261, opened 2022-10-26
CwmsDataManager was setting office_id on its own.  Replaced with indi…

rma-rripken:
…vidual dao objects for consistency with all the other controller/dao pairs.


rma-rripken:
This fixes an error that was making the Parameters end-point throw a npe.  I don't understand the exception but this fixes it.

MikeNeilson (approved):
Good consistency.

PR #263, opened 2022-10-28
Can't include DB_OFFICE_ID in the join because its the office id of t…

rma-rripken:
…he location from the timeseries.  The category or group is either going to belong to 'CWMS' or the office and if the office is not CWMS then DB_OFFICE_ID==office

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationGroupDao.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationGroupDao.java
@@ -171,7 +171,7 @@ public class LocationGroupDao extends JooqDao<LocationGroup> {
         if (officeId != null) {
             if (CWMS.equalsIgnoreCase(officeId)) {
                 connectBy = onStep.where(alcg.CAT_DB_OFFICE_ID.eq(CWMS)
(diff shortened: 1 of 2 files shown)


rma-rripken:
My previous fix to the timeseries groups and location groups broken them in a different way.  This PR should make them work correctly.

cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationGroupDao.java
```diff
@@ -26,13 +27,15 @@
...
 import org.jooq.TableField;
-
+import org.jooq.impl.DSL;
 import usace.cwms.db.jooq.codegen.tables.AV_LOC;
 import usace.cwms.db.jooq.codegen.tables.AV_LOC_CAT_GRP;
 import usace.cwms.db.jooq.codegen.tables.AV_LOC_GRP_ASSGN;
 
 public class LocationGroupDao extends JooqDao<LocationGroup> {
 
+    public static final String CWMS = "CWMS";
```
MikeNeilson, line 37: perhaps we should take the time to put these in the Enum section for constants. Not now though, just food for thought, it's already duplicated within this PR though.

MikeNeilson (approved):

Issue #147, opened 2022-01-20
Timeseries datetime format doesn't round-trip, and doesn't follow ISO spec


DanielTOsborne:
This looks like a regression since #29.

The GET /timeseries API begin/end parameters don't allow the colon to be missing from the timezone offset.
The ISO spec allows the colon to be optional.

In addition, the timestamp string in the result output doesn't have the colon, so the value cannot be used as input (doesn't round-trip).

Here's the output if the begin and end times are unspecified (left default):
![image](https://user-images.githubusercontent.com/73709911/150420722-eda11cae-0c3b-4ebd-8785-78b170fb2686.png)

Here's the result if you try to use that same value as input:
![image](https://user-images.githubusercontent.com/73709911/150420814-2e97fd18-ae95-46c9-9dc9-1ca708fd1a01.png)

![image](https://user-images.githubusercontent.com/73709911/150422526-35f39762-48b0-4d8c-8fd2-36e31f68e492.png)

```
20-Jan-2022 21:08:37.017 WARNING [https-jsse-nio-8243-exec-545] cwms.radar.ApiServlet.lambda$init$10 error on request[-182024070399621528]: /spk-data/timeseries
        java.time.format.DateTimeParseException: Text '2022-01-19T20:52:52+0000[UTC]' could not be parsed, unparsed text found at index 19
                at java.time.format.DateTimeFormatter.parseResolved0(DateTimeFormatter.java:1952)
                at java.time.format.DateTimeFormatter.parseBest(DateTimeFormatter.java:1897)
                at cwms.radar.data.dao.TimeSeriesDaoImpl.getZonedDateTime(TimeSeriesDaoImpl.java:139)
                at cwms.radar.data.dao.TimeSeriesDaoImpl.getTimeseries(TimeSeriesDaoImpl.java:120)
                at cwms.radar.api.TimeSeriesController.getAll(TimeSeriesController.java:245)
                at io.javalin.apibuilder.CrudFunction$1.invoke$lambda-0(CrudHandler.kt:30)
```

Adding a colon to the offset allows the query to execute successfully.
E.g., `2022-01-19T20:52:52+00:00[UTC]`

Issue #74, opened 2021-11-02
TimeSeries v2 endpoints date format does not quite match the documentation


rripken:
The documentation for TimeSeries start/end claim ISO 8601 extended with an example.  
But the example is subtly different from what the TimeSeries v2 Dao is expecting.
The documentation gives the example of 2021-06-10T13:00:00-0700[PST8PDT] 
But I believe the v2 Dao is expecting something that will parse with DateTimeFormatter.ISO_DATE_TIME.   I believe the above example would need to be changed so that the -0700 part includes a colon like : -07:00.
Perhaps we need to be more flexible in date parsing and try parsing with a couple of different formats?

It seems a little weird that the API accepts dates in one format but then the results returned use a different format (the ZONED_DATE_TIME_FORMAT inside TimeSeries).

Should the legacy xml/json end-points be adjusted to use a common date format with the newer end-points and then those input values converted to whatever format the pl/sql is expecting?

MikeNeilson:
There is an additional issue where by default a lot of web servers strip the [ and ] characters from URLs. Fortunately there is no specific STIG related requirement for that, but it is a concern.

As for adjusting the legacy formats. Unfortunately no, instead we should plan for their retirement. We must maintain backwards compatibility for a while yet. Once we can provide more training and get people off the hold mechanisms we can remove them.

rma-rripken:
This is related to  #147

rma-rripken:
The with or without colon issue could be solved by the suggestions here: 
https://stackoverflow.com/questions/66812557/how-to-parse-a-date-with-timezone-with-and-without-colon/66812678#66812678

rma-rripken:
I think my suggestion about working with the legacy pl/sql was misunderstood but I think we're on the same page.   I'm not suggesting that Radar modify the legacy pl/sql calls.   I'm suggesting that Radar always parse the start/end date strings with a standard format into a ZonedDateTime and when Radar is required to forward calls into pl/sql b/c of the format query param or similar that Radar then internally formats the ZonedDateTime into the appropriate String format for the pl/sql call that is about to be made.    It would be too difficult to try and document which formats are expected by which combinations of arguments.

MikeNeilson:
Oh yes, that I completely agree with. Sorry.

MikeNeilson:
[...]

PR #296, opened 2022-12-08
Added fallback format to parse offsets that don't have colon and addi…

rma-rripken:
…tional test cases for format that was given in Swagger doc and also example from format TimeSeries uses.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/helpers/DateUtils.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/helpers/DateUtils.java
@@ -1,58 +1,160 @@
 package cwms.radar.helpers;
 
-import java.time.Period;
+import com.google.common.flogger.FluentLogger;
 import java.time.Duration;
+import java.time.Instant;
+import java.time.LocalDate;
 import java.time.LocalDateTime;
+import java.time.Period;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
-import java.util.regex.Matcher;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.DateTimeParseException;
+import java.time.temporal.TemporalAccessor;
 import java.util.regex.Pattern;
 
 
 public class DateUtils {
-    private static final Pattern WITH_TZ_INFO = Pattern.compile(".*([-+][0-9]{2}(:[0-9]{2})?|Z|\\[[a-zA-Z]*(\\/[a-zA-Z]*)?\\])$");
+    private static final FluentLogger logger = FluentLogger.forEnclosingClass();
 
-    public static ZonedDateTime parseUserDate(String date,String timezone){
+    // This pattern matches if a string ends with an offset (like -07:00 or -0700),
+    // or a Z,
+    // or a named timezone (like [UTC] or [US/Pacific])
+    // The pattern is not concerned with the date/time part
+    private static final Pattern WITH_TZ_INFO = Pattern.compile(".*"
+            + "("   // followed by
+            + "[-+]\\d{2}"   // 2 digit hour offset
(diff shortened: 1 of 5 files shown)


rma-rripken:
This is related to #147 and #74

cwms_radar_api/src/main/java/cwms/radar/helpers/DateUtils.java
```diff
@@ -1,58 +1,174 @@
...
+            + "(:?\\d{2})?"  // optional colon and 2 digit minute offset
+            + "|Z" // OR Z
+            + "|\\[" // OR an open square bracket
+            + "[a-zA-Z0-9]*" // any number of letters and numbers
+            + "(/[a-zA-Z_]*)?" // optionally followed by a / and more letters or underbar as in Los_Angeles
+            + "]" // followed by a close square bracket
+            + ")"
+            + "$" // then String must immediately end
+            );
+
```
MikeNeilson, line 38: I don't think I've ever seen anyone layout a regular expression like this... thank you.

cwms_radar_api/src/main/java/cwms/radar/helpers/DateUtils.java
```diff
@@ -1,58 +1,174 @@
...
+     * @return The parsed ZonedDateTime
+     */
+    public static ZonedDateTime parseUserDate(String text, String timezone) {
         ZoneId tz = ZoneId.of(timezone);
-        return parseUserDate(date, tz, ZonedDateTime.now(tz));
+        return parseUserDate(text, tz, ZonedDateTime.now(tz));
     }
 
-    public static ZonedDateTime parseUserDate(String date, ZoneId tz, ZonedDateTime now){
+    public static ZonedDateTime parseUserDate(String date, ZoneId tz, ZonedDateTime now) {
```
MikeNeilson: date here should likely also be "text" to be consistent.

cwms_radar_api/src/main/java/cwms/radar/helpers/DateUtils.java
```diff
@@ -1,58 +1,174 @@
...
+                // Still couldn't parse so might as well throw the original exception.
+                throw e;
+            }
+        }
+
+        return zdt;
     }
 
-    private static ZonedDateTime parserUserPeriod(String date, ZoneId tz, ZonedDateTime now) {
+    private static ZonedDateTime firstMatch(String text, String[] possibleDateFormats) {
```
MikeNeilson: nitpicky, but maybe return optional instead of null?

MikeNeilson (dismissed):

cwms_radar_api/src/main/java/cwms/radar/helpers/DateUtils.java
rma-rripken: agreed

cwms_radar_api/src/main/java/cwms/radar/helpers/DateUtils.java
rma-rripken: Agreed, I forget about Optional.

cwms_radar_api/src/main/java/cwms/radar/helpers/DateUtils.java
rma-rripken: Regex, in general, is missing ergonomics.  I ended up picking apart the regex just to understand it so I figured I'd leave the comments for future readers.  I know there are some Regex builder libraries but we don't ever seem to do enough regex to justify bringing one in.

MikeNeilson (approved):

rma-rripken:
This is a significant change for user input that included zone/offset information.   I think the old behavior was wrong. I encourage reviewers to double check this PR b/c I really don't want to get this wrong.

adamkorynta (approved):

PR #297, opened 2022-12-09
Bugfix/295 npe

rma-rripken:


rma-rripken:
To address the NPE in #295

adamkorynta (approved):

PR #298, opened 2022-12-13
Add a jooq ExecutionListener to the DSLContext so that we can wrap Da…

rma-rripken:
…taAccessException.  Also added a connection method that will catch and wrap exceptions.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
@@ -219,6 +219,7 @@ public class LevelsController implements CrudHandler {
             responses = {
                     @OpenApiResponse(status = "200", content = {
                             @OpenApiContent(type = Formats.JSON),
+                            @OpenApiContent(type = ""),
                             @OpenApiContent(from = LocationLevels.class, type = Formats.JSONV2)
                         }
                     )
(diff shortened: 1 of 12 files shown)


adamkorynta:
LOVE this

cwms_radar_api/src/main/java/cwms/radar/data/dao/JooqDao.java
```diff
@@ -80,4 +94,89 @@ public void accept(org.jooq.Context<?> ctx) {
...
+            if (parts.length > 1) {
+                exception = new NotFoundException(parts[0]);
+            }
+        }
+        return exception;
+    }
+
+
+
+    private static class ExceptionWrappingListener extends DefaultExecuteListener {
```
adamkorynta: styleguide nitpick: inner classes at the bottom of the file

adamkorynta (dismissed):

cwms_radar_api/src/main/java/cwms/radar/data/dao/JooqDao.java
rma-rripken: I'd say this PR is currently only half-baked.  Hopefully discussion in #299 will lead to a more complete approach.

cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
```diff
@@ -356,6 +356,8 @@ String.class, null, metrics, name(LevelsController.class.getName(),
                     UnitSystem.EN.getValue(), unmarshalledDateTime, office);
             ctx.json(locationLevel);
             ctx.status(HttpServletResponse.SC_OK);
+        } catch (NotFoundException e){
```
MikeNeilson, line 359: Arguably we could ditch both of these and just always rethrow. Though if I'm remembering right try{} doesn't supress an exception.

MikeNeilson (approved):

PR #304, opened 2022-12-28
Feature/274 clob endpoint

rma-rripken:


cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
```diff
@@ -76,10 +79,6 @@ private Timer.Context markAndTime(String subject) {
 
     @OpenApi(
             description = "Create new CWMS Location Level",
-            queryParams = {
```
MikeNeilson, line 79: You, me, and others should get together and hash out the office parameter thing. I had to pull the office from the query param (though could be header) for my new authentication stuff to behave correctly... especially if the user had privs in more than one office.

Issue #274, opened 2022-11-08
Add post and delete operations to the clob endpoint


adamkorynta:
Used throughout CWMS clients

MikeNeilson:
club endpoint? to just be clear that is clob right?

adamkorynta:
yeah, not the disco party endpoint

PR #305, opened 2022-12-28
Added Clob create, update, delete.

rma-rripken:
Removed redundant office argument from Clob, Levels and Location controllers. ContentType now handles charset as a field.


rma-rripken:
#274

cwms_radar_api/src/main/java/cwms/radar/api/errors/NotFoundException.java
```diff
@@ -1,4 +1,4 @@
-package cwms.radar.api;
+package cwms.radar.api.errors;
```
MikeNeilson, line 1: would it make sense for this and AlreadyExists to extend from RadarError?

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -164,202 +163,197 @@ protected TimeSeries getTimeseries(String page, int pageSize, String names, Stri
         final String recordCursor = cursor;
         final int recordPageSize = pageSize;
 
-        try {
```
MikeNeilson, line 167: This mostly looks like it was just formatting tweaks. Was the performance of this endpoint verified after all these changes in any way?

cwms_radar_api/src/test/java/cwms/radar/api/ClobControllerTest.java
```diff
@@ -1,51 +1,42 @@
...
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import java.util.HashMap;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.jooq.DSLContext;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
 
 
+@Disabled
```
MikeNeilson: why did these need to be disabled?

cwms_radar_api/src/test/java/cwms/radar/data/dao/ClobDaoTest.java
```diff
@@ -0,0 +1,303 @@
...
+import cwms.radar.data.dto.Clob;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Optional;
+import org.jooq.DSLContext;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+@Disabled
```
MikeNeilson, line 22: why disabled? since it needs to talk to the database we could put it under the IT tests.

MikeNeilson (commented):
Mostly just want some feedback on why certain tests were disabled, otherwise it looks good.

cwms_radar_api/src/main/java/cwms/radar/api/errors/NotFoundException.java
rma-rripken: They could.  I'm still not 100% decided on what should be caught/wrapped/unwrapped/thrown and where.

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
rma-rripken: I should have commented on these changes.  In getTimeseries() there was a try/catch that adapted DataAccessException into a NotFoundException - that now happens in the listener for almost all jooq.  Removing the try/catch change the indent for a bunch of code so it looks like a huge change but it really isn't.  
The create/store/delete previously used dsl.connection(())  - that was changed to use the new JooqDao.connection(dsl, ()) method so that exceptions from create/store/delete will be adapted in a consistent way.

rma-rripken: I manually did a TS retrieve after these changes but I didn't try out all the other methods or the failure cases.

MikeNeilson: That explanation works for me. Always a bit tricky when so much code is getting shifted.

cwms_radar_api/src/main/java/cwms/radar/api/errors/NotFoundException.java
MikeNeilson: fair.

cwms_radar_api/src/test/java/cwms/radar/api/ClobControllerTest.java
rma-rripken: It probably doesn't need to be disabled, that was just me trying to get my IT working.

cwms_radar_api/src/test/java/cwms/radar/data/dao/ClobDaoTest.java
rma-rripken: It was Disabled b/c I didn't think to turn it into an IT.  Should I?

MikeNeilson (approved):

Issue #267, opened 2022-10-31
Add delete types to timeseries endpoint


adamkorynta:
Parameters for "Delete Data", "Delete Key", and "Delete All" are used throughout CWMS.

rma-rripken:
I tried running the integration tests locally because I was going to add some additional tests but I'm having issues getting them to run consistently and in a reasonable amount of time.   I deleted the docker volume and had it build from scratch and after a long wait the IT suite appears to have worked but then when I run it a second time some of the pre-existing tests failed.  When I tried to re-run a specific IT it just sort of hung.  I'm not going to look into it too hard atm b/c I'm thinking that the authorization stuff is going to change again pretty soon.

MikeNeilson:
Yeah, don't worry about that. I'm working on that this week. I'm not entirely sure what's going on but I'm seeing the same behavior in the tests. Though I wasn't too concerned until you just mentioned having the same concern. I'll pivot my focus a bit and see if I can back port quickly; otherwise I'm at the point of actually doing the tests for the auth manager changes and they'll go in with that.

PR #306, opened 2022-12-28
Feature/267 ts delete types

rma-rripken:
#267


MikeNeilson (approved):

adamkorynta:
There was a discussion about having different store rules be gated behind user permissions (DELETE_INSERT specifically). Should that same set of permissions get applied for the types of deletion?

Issue #266, opened 2022-10-31
Add parameters to timeseries post/patch endpoints


adamkorynta:
The CWMS client makes use of various store rules and override protection boolean flags for its various processes. Additionally, the createAsLrts flag for identifying pseudo regular time series intervals vs local regular intervals is also needed.

Currently the defaults are store rule=REPLACE_ALL , override protection=false, and createAsLrts=false https://github.com/USACE/cwms-radar-api/blob/develop/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java#L887

Mike Neilson brought up that special permissions should be investigated for handling destructive store rules such as DELETE_INSERT, especially on national instance.

adamkorynta:
Needs to also support storing the time series version

MikeNeilson:
@DanielTOsborne think you could take on these changes? Inside the TimeSeriesDAOImpl is already placeholders.

DanielTOsborne:
Not sure about the "special permissions" part, but the rest shouldn't be an issue.
I'll take a look at this.

rma-psmorris:
Worked with Adam to re-prioritize issues that are blocking for https://github.com/HydrologicEngineeringCenter/cwms-radar-client

adamkorynta:
Worked with Peter and Ryan to identify blocking issues and priorities for https://github.com/HydrologicEngineeringCenter/cwms-radar-client

rma-rripken:
Worked with Peter and Adam to identify blocking issues and priorities for https://github.com/HydrologicEngineeringCenter/cwms-radar-client

adamkorynta:
Swagger docs don't specify what the store-rule or override-protection parameters behavior is if not present.

The version-data says that a "null" version will be used, but it should probably elaborate on what that means.

adamkorynta:
The override protection and create as LRTS are defined as Strings instead of booleans

adamkorynta:
I am unable to get the version date store to work. The only guess that I have is this line: https://github.com/USACE/cwms-radar-api/blob/develop/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java#L835 where versioned boolean is always false.

MikeNeilson:
That is not yet implemented. It's on the TODO list. actually some sample data or even a test with appropriate data to target would make that faster.

MikeNeilson:
[...]

Issue #268, opened 2022-10-31
Add timeseries rename and update endpoints


adamkorynta:
The rename feature is used to change the time series identifier and the update feature is used set active flags and change the interval offset. (cwms_ts.rename_ts and cwms_ts_update_ts_id procedures respectively)

rma-rripken:
There is already a timeseries update end-point but I believe it serves a different purpose but I'm not sure anyone has been able to use it b/c of authorization.  Could it be reworked?  

We should consider using Json-Patch  ( http://jsonpatch.com/ ).   Instead of sending the objects as we want them to be, we send a series of explicit transformations we want to have happen in the patch.

Even if the existing end-point is being used and can't be reworked we could add a new mime-type for json-patch.

MikeNeilson:
That honestly looks like it could be confusing to the end users. Sure, Adam writing the hec-cwms-data-access code wouldn't care. But we kinda want this to be easy for anyone.

What about something like `.../timeseries/{original-id}/rename-to/{new-id}`?

Those rest endpoints are supposed to be a little sentence like.

rma-rripken:
Summary of work to date:   I'm in the process of creating a new TimeseriesIdentifier end-point.   This will let rename/update/delete happen without having to involve the existing TimeSeries objects.  
I did some experiments with serializing the mil.army.usace.hec.metadata.* classes directly into the CDA output and I got it to work but I believe it would make the java client rely on hec-nucleus-metadata so I stopped that line of work and did the object->dto mapping that we've had to do other places.

PR #318, opened 2023-02-21
Adding utcOffset, intervalForward, intervalBackward, versionFlag, act…

rma-rripken:
…iveFlag, versionDate, createAsLrts, storeRule and overrideProtection to the post and patch end-points

#266

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
@@ -796,12 +796,13 @@ public class TimeSeriesDaoImpl extends JooqDao<TimeSeries> implements TimeSeries
 
     @Override
     public void create(TimeSeries input){
-        create(input,0, 0, 0, false, true, null,
+
+        create(input,0, null, null, false, true, null,
                 true, StoreRule.REPLACE_ALL, true);
     }
 
     @Override
-    public void create(TimeSeries input, int utcOffsetMinutes, int intervalForward, int intervalBackward,
+    public void create(TimeSeries input, int utcOffsetMinutes, Number intervalForward, Number intervalBackward,
                        boolean versionedFlag, boolean activeFlag, Timestamp versionDate,
                        boolean createAsLrts, StoreRule replaceAll, boolean overrideProtection){
         connection(dsl, connection -> {
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
@@ -30,6 +30,7 @@ import java.math.BigInteger;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.Timestamp;
+import java.text.MessageFormat;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDateTime;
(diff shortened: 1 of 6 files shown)


cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -792,47 +794,47 @@ private TsvDquId buildDquId(AV_TSV_DQU tsvView, Record jrecord) {
...
 
-            int utcOffsetMinutes = 0;
-            int intervalForward = 0;
-            int intervalBackward = 0;
-            boolean versionedFlag = false;
-            boolean activeFlag = true;
             BigInteger tsCode = tsDao.createTsCodeBigInteger(connection, input.getOfficeId(),
-                    input.getName(),
-                    utcOffsetMinutes, intervalForward, intervalBackward, versionedFlag, activeFlag);
+                    input.getName(), utcOffsetMinutes, intervalForward, intervalBackward, versionedFlag, activeFlag);
```
MikeNeilson: This screws up data entry. Store_ts itself will create a new time series code and correctly determine the interval offset. We don't want to duplicate that logic.

We especially don't want to default to 0 for utc offset as it screws up a lot of data.

MikeNeilson (changes requested):

cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesDaoImpl.java
rma-rripken: The utcOffsetMinutes was already set to 0 - that isn't new to these changes.   It could still be wrong but it isn't new.  These changes add those method args to the Dao interfaces and add them as optional query params to the controller methods.   

It looks like the TimeSeries object already has an interval-offset field - how about I just forward that along to the create procedure?  That way users can provide null if they want or provide a value if they want.

MikeNeilson: I had to remove it to fix the LRL issue. you can can't a 1day regular timeseries at 6am if the interval-utc-offset is forced to 0. So not it's not new, but it was breaking operations and shouldn't be carried forward.

MikeNeilson: I'm also not entirely sure it's necessary to create the TS first. vs just calling store the first time. Only the location already needs to exist.

adamkorynta: That's what I was going to say, the store_ts routine will create the ts code for you.

adamkorynta: That's why we ended up needing that "create_as_lrts" parameter, it couldn't distinguish between PRTS and LRTS based on the interval identifier alone.

rma-rripken: Ok, do we care about the intervalForward/Backward and the active flag?   I added them because it looked like create could handle it but I can drop them from the TimeSeriesController and if the user wants to tweak those they'd go through the TimeSeriesIdentifierDescriptor that I'm writing for #268

adamkorynta: I don't have a use-case currently for the intervalForward/Backward from the value storage. Makes sense to include them for the TimeSeriesIdentifierDescriptor.

rma-rripken:
Are there tasks that need to be completed on this PR?  I'm not seeing them but I don't always know where to look in the github ui.

MikeNeilson (approved):

MikeNeilson:
> Are there tasks that need to be completed on this PR? I'm not seeing them but I don't always know where to look in the github ui.

No, looks fine now. I must have just gotten distracted.

PR #327, opened 2023-03-03
Make CwmsAccessManager tolerant of null user and/or office.

rma-rripken:
May need some discussion as to how we are going to handle user and office not being provided.


cwms_radar_api/src/main/java/cwms/radar/datasource/SessionOfficePreparer.java
```diff
@@ -20,12 +20,18 @@ public SessionOfficePreparer(String office) {
 
     @Override
     public Connection prepare(Connection conn) {
```
MikeNeilson, line 22: we do have a default office from the ApiServlet.service call. I was thinking the access manager could default to that. That would save the SessionOfficePrepared from needing any additional logic. (I feel like that call *should* be able to just assume it has a valid office.)

MikeNeilson (approved):

PR #328, opened 2023-03-03
Adding {timeseries} to path on PATCH and DELETE methods.

rma-rripken:
Saw that javalin was complaining about the path field.


adamkorynta (approved):

cwms_radar_api/src/main/java/cwms/radar/data/dto/TimeSeriesIdentifierDescriptor.java
```diff
@@ -16,14 +16,14 @@ public class TimeSeriesIdentifierDescriptor implements CwmsDTO {
 
     private final String officeId;
     private final String timeSeriesId;
-    private final ZoneId zoneId;
+    private final String timezoneName;
```
MikeNeilson, line 19: This can be left as ZoneId, you just have to add annotations to tell it how it should be rendered. The TimeSeries.Record class has some examples.

But for something this simple a String and a validator is a fine solution so just keep that in mind for future reference.

PR #332, opened 2023-03-21
initial modifications for a containing access manager.

MikeNeilson:
Getting the start of keycloak integration going. First steps are allowing multiple access managers.

PR open for tracking as some additional reorganizations may happen.


MikeNeilson:
renaming for clarity of current intent.

PR #336, opened 2023-03-23
[RFC] change timeseries delete to require a time window

adamkorynta:
it seems dangerous that a deleteAll could be called due to missing query parameters (especially if they get renamed). The timeseries-identifier endpoint can already handle deleteAll


MikeNeilson (approved):
This seems like a reasonable safety measure to me.

PR #341, opened 2023-03-25
use -1 for default tsItemMask so that delete actually happens

adamkorynta:
add delete_ts test

Maybe someone can explain to me why this is needed? This should be defaulting to `cwms_util.ts_all`: https://bitbucket.hecdev.net/projects/CWMS/repos/cwms_database/browse/schema/src/cwms/cwms_ts_pkg_body.sql#7082

This is defined as `-1`: https://bitbucket.hecdev.net/projects/CWMS/repos/cwms_database/browse/schema/src/cwms/cwms_util_pkg.sql#131 so I wouldn't think this is needed.


rma-rripken (approved):

rma-rripken:
Yeah.  I don't see why its needed either.  It looks like it stays null on the java side.   The jooq code looks like it defaults the parameter to NULL also so it seems like jooq wouldn't be getting it wrong either.

PR #345, opened 2023-03-28
Feature/multi access manager

MikeNeilson:
Allows for multiple access managers to be used.

Time permitting this week (and figuring it out) will be creating a fake Session handler and modifying the builds to use TestTemplates for each possible method. But if not done with that by Thursday will need to rely on manual testing to get this out to LRL for them to continue on working on their usage.


MikeNeilson:
Still a work in progress, but feedback would be good before going too much further. Next step is mimicing the Tomcat Session store so I we "test" CWMSAAA CaC Auth users. Hooks in place, need to sort out details. but this would let us loop over various users at different times.

rma-rripken:
Traefik and authelia are new to me.  Nothing in here seems crazy yet.

MikeNeilson:
Yeah, those are for the docker-compose to mimic a more complete environment for manual testing. They won't affect the automated tests except perhaps to inform some direct... might need authelia in a testcontainer though, OpenID Connect is simple but the flow is difficult to mock simply.

MikeNeilson:
Oh, and since I looks like with this mechanism we don't need to distinguish the users (thank you RequestSpecificationBuilder) I'm going to remove that distinction.

MikeNeilson:
Okay, this gets the basic idea across. We'll need to expand more of the tests to actually use all the possible credential types but I'd like to get a release of new features out and this does properly test the multiple access manager mechanism.

MikeNeilson:
Traefik/authelia will be used in future work for actually Keycloak/Login.gov integration. Original scope was just OpenID but I realize these changes are required first.

MikeNeilson:
I'd like to just merge this in as-is. I've started on the actual keycloak integration but that should just be a separate PR.
This gives us a working multi auth system and a new structure to the tests. Future tests with auth should be parameterized like those in AccessManagerTestIT.java

rma-rripken (approved):

PR #348, opened 2023-04-06
fix location groups returning with null ids

adamkorynta:
the left join was using the group_id and category_id columns from the assigned locations view instead of the location group view this caused the columns to display null when there weren't any assigned locations in for the location group

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationGroupDao.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationGroupDao.java
@@ -10,6 +10,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


cwms_radar_api/src/main/java/cwms/radar/data/dao/LocationGroupDao.java
```diff
@@ -159,35 +171,46 @@ public List<LocationGroup> getLocationGroups(String officeId) {
...
-        connectBy.orderBy(alga.CATEGORY_ID, alga.GROUP_ID, alga.ATTRIBUTE).fetchSize(1000)    //
+        connectBy.orderBy(alcg.LOC_CATEGORY_ID, alcg.LOC_GROUP_ID, alga.ATTRIBUTE).fetchSize(1000)    //
                 // This made the query go from 2 minutes to 10 seconds?
                 .stream().map(mapper::map).forEach(pair -> {
                     LocationGroup locationGroup = pair.component1();
-                    List<AssignedLocation> list = map.computeIfAbsent(locationGroup,
-                            k -> new ArrayList<>());
-                    list.add(pair.component2());
+                    List<AssignedLocation> list = map.computeIfAbsent(locationGroup, k -> new ArrayList<>());
+                    AssignedLocation assignedLocation = pair.component2();
```
rma-rripken, line 210: Probably don't even make the new list if the assignedLocation is null.

adamkorynta: I still need the key in the map and don't want to insert a null collection.

rma-rripken: ok

rma-rripken (approved):

cwms_radar_api/src/test/java/cwms/radar/data/dao/LocationGroupDaoTestIT.java
```diff
@@ -0,0 +1,59 @@
...
+class LocationGroupDaoTestIT extends DataApiTestIT {
+
+    @BeforeAll
+    public static void load_data() throws Exception {
+        loadSqlDataFromResource("cwms/radar/data/sql/store_loc_groups.sql");
+    }
+
+    @AfterAll
+    public static void deload_data() throws Exception {
+        loadSqlDataFromResource("cwms/radar/data/sql/store_loc_groups.sql");
```
MikeNeilson, line 26: FYI, this is a good candidate for creating a template and function inside DataApiTestIT similar to createTS.

That way the locations can be registered and removed at the end of the test sequence. That's what allows for the tests to be cycled constantly without breaking. granted for this one not a huge deal as we can just keep calling create_location for the same name to know ill effect but it's a good practice to start getting into.

PR #357, opened 2023-05-03
Feature/cwbi keycloak

MikeNeilson:
Setup of OpenID Connect workflow.

Given a URL to the appropriate ".well-known/openid-configuration" information it cache the signing keys and setup the appropriate openapi securityscheme definitions.

Currently integrating keycloak itself into the docker-compose so I can make the appropriate mocks within the automated tests. (and the developer test environment has some form of login support also.)


MikeNeilson:
The baseline works correctly within the docker-compose setup.

Next steps are:
[ ] - Export the realm now so it always imports on build. (currently manually setting it up.)
[ ] - Implement simple mock within the automated tests and get the users setup with a bearer token for the test system to use.

MikeNeilson:
quick update. I'm finalizing some changes to the test structure (again) to make it easier to deal with all the different auth mechanisms and users.

after that I just need to include the JWT/OpenID test. That will involve a testcontainer instance of keycloak so that data-api can do the full proper verification sequence. Seems easier than trying to handle statically setting up all the pages.

There's still more work to do like adding to ability to verify things like audience/scopes/roles/etc but I think getting the baseline structure in and making a new 3.0.0 release should be done next week. We have a *lot* of good changes sitting in develop right now.

MikeNeilson:
@DanielTOsborne @krowvin @willbreitkreutz @adamkorynta @adamscarberry @rma-rripken 

So part of this PR is going to need to be settling on "the definitive way" to handle given a data object, how does the user of the API properly identify that they have permissions for it.

Unfortunately with our current API we need to have the session (oracle connection session) set to the appropriate values before storing.

Most endpoints have an "office" parameter, and all the dto's have an office parameter. 

I'm inclined to just always pull the value from the DTO objects. that will involve a little "knowledge" of the data in the authorization components system. OR we could actually have the DTO write objects deal with it? at point of use they need the office ID in hand to write the values. That's small call to make and make it's easier to deal with arrays of data (should that ever come up.)

Alternatively an additional explicit header, like `X-CWMS-Auth-Office`, could be set on each call that requires it separate from the query parameter.

... I may just do that. but I want everyone's thought on the concept. Because a user *could* be working across offices, it's not practical to specify on login; it would involve the user constantly logging in and out for something rather basic.

adamkorynta:
I thought the session office was used for more than just DTO writes. Or is it specifically because CDA is allowing reads through its own special webtest user?

I'm not familiar with the full intent of the session office id as lots of db routines have parameterized offices already. I think I need a little more background information to know which method to take.  What happens, for example, if you call cwms_loc.store_loc_category with a parameterized office different from your session? What happens if we don't have a session office id and we pass in "SPK" to the procedure? I'm guessing triggers play some sort of role in this too...

I thought that it was used as the default if the parameterized office was null. For calls to CDA we are expecting the office to always be provided (at least from my understanding) so my gut instinct is that the header is perfectly acceptable.

Without knowing all the use-cases for how the session office id is used, it seems scary to me to hide it away in the DTO.

MikeNeilson:
The issue is that for writes, even if you specify the office on the write call (in the database API), it checks the session office... sometimes.

I mean arguably it is a little bit of issue with the database code itself. But prudence means working around that and I needed someway to assert the connection session office. The office queryParam was already there on most endpoints because of the need to sometimes use it in a query.

However, since each object *MUST* have the office anyways, it probably does make more sense to not set the office once. However, there is another challenge, if you don't set the office at all things can go haywire changing the user. But if the user has more than one office it can't pick a default. Though for the purposes of CDA here, we could probably just pick the first in a list especially if we move to setting it just before the store calls.

I'm probably going to have to create a PR showing this for it all to make sense though.

MikeNeilson:
I'm flipping this to ready. I still need to add automated tests to cover the openid connect workflow but that does work manually (one can test with the docker-compose file and that instance of keycloak) but there's fixes needed for the other auth methods and downstream tests so I'm going to just make a release after this is merged in so get some current things out and free up some brain space.

cwms_radar_api/src/main/java/cwms/radar/security/CwmsAccessManager.java
```diff
@@ -192,8 +188,7 @@ public String getName() {
 
     @Override
     public boolean canAuth(Context ctx, Set<RouteRole> roles) {
-        return  (ctx.cookie("JSESSIONIDSSO") != null)
-              ||(ctx.cookie("JSESSIONID") != null);
+        return ctx.cookie("JSESSIONIDSSO") != null;
```
rma-rripken: "JSESSIONIDSSO" should be a constant somewhere.  too easy to typo or mix up with "JSESSIONID"

cwms_radar_api/src/test/java/fixtures/TomcatServer.java
```diff
@@ -65,9 +73,34 @@ public TomcatServer(final String baseDir,
...
+                        }
+                    }
+                }
+                Valve next = this.getNext();
+                if (next!=null) {
+                    next.invoke(request, response);
+                }
+            }
+
+        });
```
rma-rripken: delete the commented out code.

cwms_radar_api/src/main/java/cwms/radar/security/CwmsAccessManager.java
rma-rripken: I know its not in your changes but the "office" should probably use the Controllers constant - might make it a little clearer that its hoping that parameter is present.

cwms_radar_api/src/test/java/fixtures/TomcatServer.java
MikeNeilson: oops, thanks.

cwms_radar_api/src/main/java/cwms/radar/security/CwmsAccessManager.java
MikeNeilson: good point, especially until we clear that up.

rma-rripken (approved):

krowvin:
> MN: However, since each object MUST have the office anyways  
> MN: But if the user has more than one office it can't pick a default.  
> MN: if you don't set the office at all things can go haywire changing the user  
> MN: The office queryParam was already there on most endpoints because of the need to sometimes use it in a query.  
> MN: Most endpoints have an "office" parameter, and all the dto's have an office parameter.  
> AK: Without knowing all the use-cases for how the session office id is used, it seems scary to me to hide it away in the DTO.  

> **Alternatively an additional explicit header, like X-CWMS-Auth-Office, could be set on each call that requires it separate from the query parameter.**

I started a discussion on this [here](https://github.com/USACE/cwms-radar-api/discussions/376)

PR #361, opened 2023-05-04
add store, patch (rename), and delete for specified levels

### LINKED ISSUE #273: Add store and delete and rename to specified-levels endpoint
These are used by CWMS clients

adamkorynta:
fixes: #273 

I added a global exception handler for ORA-20019 INVALID_ITEM since that is what the specified level rename routine throws if the original specified level id isn't found. This is routed to 400 error code. I didn't like returning a 404 for INVALID_ITEM. I could make a Jira ticket for the schema if we want the package procedure to throw a _NOT_FOUND error instead.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
@@ -1,3 +1,27 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2023 Hydrologic Engineering Center
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
 package cwms.radar;
 
 import static io.javalin.apibuilder.ApiBuilder.get;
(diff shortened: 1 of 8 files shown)


rma-rripken:
The issue is that the rename cannot happen b/c a level at the original name can't be found?  If you did a get request for the original id it would be not-found -right?  400 is fine, its general all encompassing.  404 seems like a perfect fit to me.  I don't even think you'd need to change the procedure, CDA and the Dao seems like the right place to interpret the database error into http error.   I don't have a problem with leaving it 400.  Maybe I'm reading it wrong but it looked like the test for the rename error didn't actually try to change the name - it tries to patch from BadUpdate42->BadUpdate42

rma-rripken (dismissed):

adamkorynta:
Since we hadn't handled this type of error yet (from what I could fine) I wasn't sure what we expected. It seems correct to me for a 400 error code, but I could see an argument for 404.

adamkorynta:
Yeah my test is a bit of inside baseball since I know the rename doesn't check to see if anything changed first.

cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
```diff
@@ -200,6 +201,11 @@ public void init() {
                     logger.atInfo().withCause(e).log(re.toString(), e);
                     ctx.status(HttpServletResponse.SC_BAD_REQUEST).json(re);
                 })
+                .exception(InvalidItemException.class, (e, ctx) -> {
+                    RadarError re = new RadarError("Not Found.");
```
MikeNeilson: I'm with Ryan on this. This looks semantically odd. You're asserting the text "not found" but not returning the "NOT_FOUND"error code.

But given the other discussion it would make sense to clarify that the "source" level was not found.

cwms_radar_api/src/main/java/cwms/radar/api/SpecifiedLevelController.java
```diff
@@ -27,7 +56,13 @@
 
 
 public class SpecifiedLevelController implements CrudHandler {
+    static final String OFFICE = "office";
+    static final String TEMPLATE_ID_MASK = "template-id-mask";
+    static final String SPECIFIED_LEVEL_ID = "specified-level-id";
+    static final String FAIL_IF_EXISTS = "fail-if-exists";
```
MikeNeilson, line 62: I'm beginning to think we need a class of just query parameter constants. some of these are used everywhere.

NOTE: this is for the record. no change required.

cwms_radar_api/src/test/java/cwms/radar/api/SpecifiedLevelIdIT.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+
+    @Test
+    public void test_update() throws JsonProcessingException {
+        ObjectMapper om = JsonV2.buildObjectMapper();
+        TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
+        long epochSeconds = Instant.now().getEpochSecond();
+        SpecifiedLevel specifiedLevel = new SpecifiedLevel("TestUpdate" + epochSeconds, OFFICE, "CDA Integration Test");
+        String serializedTs = om.writeValueAsString(specifiedLevel);
```
MikeNeilson: serializedLevel? suspect that's just copy paste issues.

MikeNeilson (changes requested):
Mostly minor nitpicks.

cwms_radar_api/src/main/java/cwms/radar/ApiServlet.java
adamkorynta: Oh this was another copy/paste error. Switched to Bad Request. Invalid Item could also work, but that should already be in the oracle error message.

cwms_radar_api/src/main/java/cwms/radar/api/SpecifiedLevelController.java
adamkorynta: Yeah that would help to to keep them consistent.

cwms_radar_api/src/test/java/cwms/radar/api/SpecifiedLevelIdIT.java
adamkorynta: Yup bad copy/paste, I renamed them all

MikeNeilson (approved):

PR #364, opened 2023-05-05
Moved controller constants into Controllers class.  Replaced Meters a…

rma-rripken:
…nd Timers with markAndTime method call.

### DIFF
--- a/cwms_radar_api/src/test/java/cwms/radar/api/SpecifiedLevelIdIT.java
+++ b/cwms_radar_api/src/test/java/cwms/radar/api/SpecifiedLevelIdIT.java
@@ -48,7 +48,7 @@ public class SpecifiedLevelIdIT extends DataApiTestIT {
     public static final String OFFICE = "SPK";
 
     @Test
-    public void test_create_read_delete() throws JsonProcessingException {
+    void test_create_read_delete() throws JsonProcessingException {
         ObjectMapper om = JsonV2.buildObjectMapper();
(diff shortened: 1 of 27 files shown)


cwms_radar_api/src/test/java/cwms/radar/api/SpecifiedLevelIdIT.java
```diff
@@ -48,7 +48,7 @@ public class SpecifiedLevelIdIT extends DataApiTestIT {
     public static final String OFFICE = "SPK";
 
     @Test
-    public void test_create_read_delete() throws JsonProcessingException {
+    void test_create_read_delete() throws JsonProcessingException {
```
MikeNeilson, line 51: Test functions should be public. Otherwise they might not get picked up.

MikeNeilson (changes requested):
Other than one issue on a test this looks great.

cwms_radar_api/src/test/java/cwms/radar/api/SpecifiedLevelIdIT.java
adamkorynta: JUnit5 no longer requires public tests, they just can't be private. JUnit4 also removed the requirement of starting the method name with "test" but I can't kick the habit.

MikeNeilson: fair. and yeah i generally do the test_ as a matter of abundantly clear intent that "this function is testing something specific and not support."

MikeNeilson (approved):

PR #372, opened 2023-05-19
restructure levels patch to not require an effective date for an id rename

### LINKED ISSUE #300: /levels PATCH requires an effective date, makes renaming the level id unclear
The PATCH request for location levels requires an effective date. I understand this purpose when editing effective date data, but this makes the rename functionality difficult to understand.

Investigating the source code the rename function is called if the level id endpoint parameter (/levels/<level-id>) from the level identified within the PATCH body. However, the rename will not occur if the effective date does not exist (or is not provided). Given the level id sits above the individual effective dates, this doesn't make sense.

adamkorynta:
Fixes: #300

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/api/LevelsController.java
@@ -1,29 +1,31 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2023 Hydrologic Engineering Center
+ *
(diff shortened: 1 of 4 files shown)


rma-rripken:
The description of the effective-date argument suggests to me that if I add that argument than my updates will be limited to the level at that date.   Probably should throw an exception if they are updating any general fields ( rename ) but provide effective-date.  If the users do provide effective-date is there anything they can change besides effective-date that won't affect other levels?

rma-rripken (approved):

PR #374, opened 2023-05-23
bundle specified level endpoints with the rest of the levels

adamkorynta:
fix the path parameter so that the OpenAPI isn't adding both "level-id" and "specified-level-id"

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/api/SpecifiedLevelController.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/api/SpecifiedLevelController.java
@@ -24,10 +24,6 @@
 
 package cwms.radar.api;
 
-import static com.codahale.metrics.MetricRegistry.name;
-import static cwms.radar.api.Controllers.*;
-import static cwms.radar.data.dao.JooqDao.getDslContext;
-
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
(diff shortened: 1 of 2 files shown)


cwms_radar_api/src/main/java/cwms/radar/api/SpecifiedLevelController.java
```diff
@@ -42,23 +38,23 @@
...
+
+import static com.codahale.metrics.MetricRegistry.name;
+import static cwms.radar.api.Controllers.*;
+import static cwms.radar.data.dao.JooqDao.getDslContext;
 
 
 public class SpecifiedLevelController implements CrudHandler {
     private static final Logger logger = Logger.getLogger(SpecifiedLevelController.class.getName());
-    private static final String TAG = "Specified Levels";
+    private static final String TAG = "Levels";
```
MikeNeilson, line 57: wait, shouldn't this one actually say Specified Levels?

MikeNeilson: oh wait, I see now. These are intended to be lumped under the same swagger ui box not in separate groupings.

MikeNeilson (dismissed):

cwms_radar_api/src/main/java/cwms/radar/api/SpecifiedLevelController.java
adamkorynta: yeah, in order to keep it consistent with ratings. I was getting annoyed scrolling to the bottom of swagger for the specified level endpoints but the levels endpoints were hidden in the middle.

adamkorynta:
rebased onto develop due to resolve conflicts

rma-rripken (approved):

Issue #321, opened 2023-02-24
Update offices and offices/{office} end points to include an operational data attribute


rma-psmorris:
This is a request to update the CDA offices and offices/{office} end points to include an attribute per office as to whether the office has operational data. Suggested is that the office owning locations be the indicator that an office has operational data.

The reason for the request is that the CWMS client AAA authentication UI displays a selection of offices (sourced from the offices end point) for the user to choose from that is later used in calls against CDA. The addition of this attribute will allow us to limit the set of offices displayed to those offices that have operational data.

rma-rripken:
Instead of adding an attribute to the results could the end-point take an optional filter?  require-locations=true  or something like that.    As far as actually finding the offices that own locations - is that just a matter of " select distinct DB_OFFICE_ID from cwms_20.AV_LOC"   or is there a better way?

MikeNeilson:
I agree with Ryan. It would be cleaner to have a parameter on the request. 

that query is pretty much how I would do it. Though I think it needs to have "where location_code != 0" to avoid the "deleted ts" code. 

FYI, we're going to have Charles work on this particular one because it is simple and gives him some experience but I am going to mark it as approved in case he needs help and Daniel and I are unavailable.

rma-psmorris:
@adamkorynta -- you would be the one coding against this new end feature. Can you please add your thoughts per Ryan's suggestion?

adamkorynta:
I like the parameter `has-data=true` as it hides the location query as an implementation detail for how it is determined. I don't think we'd want this endpoint to balloon into `has-locations=true&has-timeseries=false`etc.

adamkorynta:
Though I guess I'll say the opposite of that if you were to add in attributes instead of a parameter. It may be useful for the 
returned office data to have a list of data types for that office:

```
{
   office-id = '',
   office-name = '',
   has-locations = true,
   has-timeseries = true,
   has-location-levels = true,
}
```

But I don't have a use-case for that kinda of implementation.

MikeNeilson:
let's go with "lack of use-case = no reason to add code."

adamkorynta:
100%

krowvin:
Running `SELECT DISTINCT DB_OFFICE_ID FROM CWMS_20.AV_LOC WHERE LOCATION_CODE != 0 ORDER BY DB_OFFICE_ID` 
against the database returns
```
CWMS
LRB
LRD
LRE
LRH
LRL
LRN
LRP
MVK
[...]

PR #375, opened 2023-05-24
Add has-data parameter to /offices for

krowvin:
Solves issue #321

SQL was tested against national database and output can be seen under the issue chat. 

Testing on local district database using swagger/thunderclient


MikeNeilson (dismissed):
This looks good. Unfortunately we just merged a bunch of other changes in causing a conflict so you're going to need to pull that over and update.

I'm pretty sure it's just changing the literal "office" to the constant in controllers though; so it shouldn't be a huge deal.

config/checkstyle/checkstyle.xml
```diff
@@ -265,5 +265,8 @@
             <property name="exceptionVariableName" value="expected"/>
         </module>
         <module name="CommentsIndentation"/>
+        <module name="ParameterOnNewLine">
```
MikeNeilson: @USACE/cwms-radar-devs thoughts?

This makes autoformatters work better with the OpenAPI definition but it would otherwise tweak our functions.

adamkorynta: Can you link an example of what this looks like? I did not see documentation on this. It also looks like we are using a pretty old version of checkstyle.

MikeNeilson (approved):

PR #377, opened 2023-05-24
Minor corrections to tests.

### LINKED ISSUE #378: TS Groups/Categories not deleting correctly in tests

MikeNeilson:
Went to go make a release, discovered a test wasn't passing.

Looks like simple copy paste errors copying from the location group test.

Did decide to just get a little more drastic cleaning up at the end of the test though.

### DIFF
--- a/cwms_radar_api/src/test/java/cwms/radar/api/LocationGroupControllerTestIT.java
+++ b/cwms_radar_api/src/test/java/cwms/radar/api/LocationGroupControllerTestIT.java
@@ -49,7 +49,7 @@ class LocationGroupControllerTestIT extends DataApiTestIT
 
 	@Test
 	void test_create_read_delete() throws Exception {
-		String officeId = RadarApiSetupCallback.getDatabaseLink().getOfficeId();
+		String officeId = "SPK";
 		String locationId = "LocationGroupTest";
 		createLocation(locationId, true, officeId);
 		TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
@@ -79,6 +79,7 @@ class LocationGroupControllerTestIT extends DataApiTestIT
 		given()
 			.accept(Formats.JSON)
 			.contentType(Formats.JSON)
+			.queryParam(OFFICE, officeId)
 			.body(groupXml)
 			.header("Authorization", user.toHeaderValue())
 			.when()
@@ -157,7 +158,7 @@ class LocationGroupControllerTestIT extends DataApiTestIT
 
 	@Test
 	void test_rename_group() throws Exception {
-		String officeId = RadarApiSetupCallback.getDatabaseLink().getOfficeId();
+		String officeId = "SPK";
 		String locationId = "LocationGroupTest";
 		createLocation(locationId, true, officeId);
 		TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
@@ -188,6 +189,7 @@ class LocationGroupControllerTestIT extends DataApiTestIT
 			.accept(Formats.JSON)
 			.contentType(Formats.JSON)
 			.body(groupXml)
+			.queryParam(OFFICE,officeId)
 			.header("Authorization", user.toHeaderValue())
 			.when()
 			.redirects().follow(true)
@@ -253,7 +255,7 @@ class LocationGroupControllerTestIT extends DataApiTestIT
 
 	@Test
 	void test_add_assigned_locs() throws Exception {
-		String officeId = RadarApiSetupCallback.getDatabaseLink().getOfficeId();
+		String officeId = "SPK";
 		String locationId = "LocationGroupTest";
 		createLocation(locationId, true, officeId);
 		String locationId2 = "LocationGroupTest2";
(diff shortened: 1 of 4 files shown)


adamkorynta:
Caused by not fully fixing the copy/paste in: https://github.com/USACE/cwms-radar-api/pull/371

cwms_radar_api/src/test/java/cwms/radar/api/LocationGroupControllerTestIT.java
```diff
@@ -49,7 +49,7 @@ class LocationGroupControllerTestIT extends DataApiTestIT
 
 	@Test
 	void test_create_read_delete() throws Exception {
-		String officeId = RadarApiSetupCallback.getDatabaseLink().getOfficeId();
```
adamkorynta, line 52: When the automated tests run, they are authorized for all offices and this one is set to HQ?

MikeNeilson: No the account has specific offices in this case just SPK, but the connecting user has the HQ priv. That's basically how the "central writable" system will need to work.

cwms_radar_api/src/test/java/cwms/radar/api/TimeSeriesGroupControllerTestIT.java
```diff
@@ -516,7 +521,7 @@ void test_add_assigned_locs() throws Exception {
             .when()
             .redirects().follow(true)
             .redirects().max(3)
-            .delete("/timeseries/group/" + group.getId())
+            .delete("/timeseries/group/{group}",group.getId())
```
rma-rripken, line 524: I like the curly brace replacement, we should do that everywhere.

cwms_radar_api/src/test/java/cwms/radar/api/LocationGroupControllerTestIT.java
adamkorynta: ok, so I should make sure to always use SPK

cwms_radar_api/src/test/java/cwms/radar/api/TimeSeriesGroupControllerTestIT.java
```diff
@@ -485,9 +489,9 @@ void test_add_assigned_locs() throws Exception {
             .body("office-id", equalTo(group.getOfficeId()))
             .body("id", equalTo(group.getId()))
             .body("description", equalTo(group.getDescription()))
-            .body("assigned-time-series[0].timeseries-id", equalTo(timeSeriesId2))
-            .body("assigned-time-series[0].alias-id", equalTo("AliasId2"))
-            .body("assigned-time-series[0].ref-ts-id", equalTo(timeSeriesId2));
+            .body("assigned-time-series[1].timeseries-id", equalTo(timeSeriesId2))
```
rma-rripken, line 492: changing the index 0->1 makes sense b/c the var is timeSeriesId2.  Should the test also be looking at index 0?

cwms_radar_api/src/test/java/cwms/radar/api/LocationGroupControllerTestIT.java
MikeNeilson: Yeah. Or at the least it should be specific. I'm thinking of making the "TestAccount" users hold a set of offices so it can also be looped over, but I'm not quite sure how that'll work.... yet.

cwms_radar_api/src/test/java/cwms/radar/api/TimeSeriesGroupControllerTestIT.java
MikeNeilson: Yeah, I discovered that a while ago, the primary thing it does is change how the Http Request/Response is pretty printed.

Sadly it doesn't fix the issue with "/" in the actual name like for clobs and blobs. (And that's still broken in Javalin. supposed to support it but the processing code doesn't quite work right. Figured I wouldn't bother about it until we can move to javalin 5 though.)

cwms_radar_api/src/test/java/cwms/radar/api/TimeSeriesGroupControllerTestIT.java
MikeNeilson: Probably to verify it's the same, but I made that change because the asserted clearly showed it should be the 2nd.

MikeNeilson:
FYI, I am going to start a 3.0 release without this fixed. This is clearly an issue with the test themselves; the endpoints themselves are working correctly once the parameters are corrected.
Thought deleting the category doesn't seem to be working due to them not being "empty" so that may not quite be working right, but there are various other needs to get to people.

rma-rripken:
This is still Draft - right?

MikeNeilson:
Yes, if you have time though I was hoping you could just take it over. I'm working on the office id in DTO concept (so far working except minor oddity in a deserializer I'm about to look at.)

MikeNeilson:
except for the change to `.../{id value}", object)` This work was included within the larger office id feature.

PR #382, opened 2023-05-24
Migration to office-id in DTO abstract class.

MikeNeilson:
Currently initial work. A few integration tests don't pass due to the session office issue discussed, but considering this change is relatively clean I'm thinking I'll just fix that here and implement the changes discussed.

After that I'm also going to to the rebranding since the next release is going to be 3.0.0 anyways. Might as well do *all* the breaking things now.

Oh, and this does break the Clob DTO since that's the one that kept "office" vs "office-id". I'm not super worried about it though.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dto/Pool.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dto/Pool.java
@@ -7,8 +7,7 @@ import cwms.radar.api.errors.FieldException;
 import usace.cwms.db.dao.ifc.pool.PoolNameType;
 import usace.cwms.db.dao.ifc.pool.PoolType;
 
-public class Pool extends PoolType implements CwmsDTO
-{
+public class Pool extends PoolType implements CwmsDTOBase {
 	private final Number attribute;
 	private final String description;
 	private final String clobText;
--- a/cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/api/ClobController.java
@@ -253,7 +253,7 @@ public class ClobController implements CrudHandler {
             try {
                 Clob clob = deserialize(ctx.body(), formatHeader);
 
-                if (clob.getOffice() == null) {
+                if (clob.getOfficeId() == null) {
                     throw new FormattingException("An officeId is required when creating a clob");
                 }
 
@@ -341,7 +341,7 @@ public class ClobController implements CrudHandler {
             try {
                 Clob clob = deserialize(ctx.body(), formatHeader);
 
-                if (clob.getOffice() == null) {
+                if (clob.getOfficeId() == null) {
                     throw new HttpResponseException(HttpCode.BAD_REQUEST.getStatus(),
                     "An office is required in the request body when updating a clob");
                 }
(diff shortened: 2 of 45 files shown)


cwms_radar_api/src/main/java/cwms/radar/data/dto/Pool.java
```diff
@@ -7,8 +7,7 @@
 import usace.cwms.db.dao.ifc.pool.PoolNameType;
 import usace.cwms.db.dao.ifc.pool.PoolType;
 
-public class Pool extends PoolType implements CwmsDTO
-{
+public class Pool extends PoolType implements CwmsDTOBase {
```
MikeNeilson, line 10: Should Pools have an office?

adamkorynta: Aren't there CWMS defined pools and then district pools?

MikeNeilson: That's what I thought, which would necessitate the office. since we really just use "CWMS" as a "shared office" of sorts.

adamkorynta: https://bitbucket.hecdev.net/projects/SC/repos/hec-monolith/browse/hec-monolith/src/main/java/hec/data/pool/PoolName.java#22

Our data object has an office id and AV_POOL has an office column

MikeNeilson: Ah, it's not a totally separate DTO. Didn't even notice that when I was reviewing. I'll change that as part of this PR. Last thing we want here is a real dependency on monolith, the DTOs are as much for setting the XML/JSON parameters as much as anything.

adamkorynta:
CWMSVue's CDA usage didn't get added to CWMS 3.3.0 so changing the Clob DTO won't effect that at all.

cwms_radar_api/src/main/java/cwms/radar/data/dto/Pool.java
adamkorynta: Oh I linked the monolith representation. The one linked here is from the cwms-db-dao dependency which isn't based on monolith. Still it's really only a dumb object used to translate between jOOQ and monolith

MikeNeilson:
> CWMSVue's CDA usage didn't get added to CWMS 3.3.0 so changing the Clob DTO won't effect that at all.

Plus there's the added benefit of... very little external uses blob's and clob's at the moment.

adamkorynta:
There's clob usage in the rating editor and it is used pretty extensively in REGI.

MikeNeilson:
Good to know. Blob usage though?

adamkorynta:
The only place that I know about blobs is our script that runs a speed test against the database: https://bitbucket.hecdev.net/projects/CWMS/repos/hec-cwms-data-access/browse/jython-db-api/src/main/resources/time_retrieves.py

but this is to measure performance specifically against JDBC connections so not really applicable for CDA.

MikeNeilson:
Okay, so FYI, the integration tests will fail. Because the CWMS database really isn't setup for a user (or more correctly the connection) to have permissions in multiple offices at the same time. However, that is a fix in the database itself that I will work on later as part of the cloud migration single database effort. 

For what we need this release for now, a single office with single office permissions works just fine. 

Other good news is the errors are fairly obvious about why you don't get permissions, the "next" user be it the api or the user probably doesn't have any. This may be solved by just giving "CWMS User" on every office to the connection user but frankly I've already spent too much time on just gathering the current release.

adamkorynta (approved):

cwms_radar_api/src/main/java/cwms/radar/data/dto/Pool.java
MikeNeilson: NOTE: I decided against fixing this for this PR, work will still need to get done.

PR #384, opened 2023-05-29
remove the filter condition on assigned time series office

adamkorynta:
we want to be able to retrieve empty groups
the documentation specified that the office id is for the group, so we don't care what office the category is in and category names are unique


MikeNeilson (dismissed):
Though i might be good to have the commit comment directly in the source. That's not entirely obvious without all the documentation.

adamkorynta:
Good idea, I added a comment.

MikeNeilson (approved):

PR #390, opened 2023-05-31
return rating metadata even if there are no ratings for the spec

adamkorynta:


rma-rripken:
Is this related to a 13b issue?

adamkorynta:
Trying to get my task 1a tests passing. I didn't create an issue for this.

rma-rripken:
This removes the join with virtual and transitional ratings.  Not saying that those worked perfectly but I was hoping we'd build that out not remove it

rma-rripken (approved):

rma-rripken:
I missed that this switches to av_rating_spec view.  Hopefully that makes the joins I was complaining about unnecessary.

MikeNeilson:
That is worth verifying. I suggest opening an issue with a little more detail about exactly what needs to be tested there.

PR #391, opened 2023-06-01
add hidden cascade delete location endpoint parameter

### LINKED ISSUE #342: add delete location cascade option to locations endpoint
Currently only the `"DELETE LOC"` option is used. We also need the `"DELETE LOC CASCADE"` option.

adamkorynta:
Fixed #342

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/api/LocationController.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/api/LocationController.java
@@ -1,18 +1,28 @@
-package cwms.radar.api;
+/*
+ * MIT License
+ *
+ * Copyright (c) 2023 Hydrologic Engineering Center
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
 
-import static com.codahale.metrics.MetricRegistry.name;
-import static cwms.radar.api.Controllers.CREATE;
-import static cwms.radar.api.Controllers.DATUM;
-import static cwms.radar.api.Controllers.DELETE;
-import static cwms.radar.api.Controllers.FORMAT;
-import static cwms.radar.api.Controllers.GET_ALL;
-import static cwms.radar.api.Controllers.GET_ONE;
-import static cwms.radar.api.Controllers.OFFICE;
-import static cwms.radar.api.Controllers.RESULTS;
-import static cwms.radar.api.Controllers.SIZE;
-import static cwms.radar.api.Controllers.UNIT;
-import static cwms.radar.api.Controllers.UPDATE;
-import static cwms.radar.data.dao.JooqDao.getDslContext;
+package cwms.radar.api;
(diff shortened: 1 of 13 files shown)


adamkorynta:
@MikeNeilson - I'm not sure if you were suggesting that a role check be added here at the CDA level or if that should get added to the CWMS schema. I'm also not sure if that would be a new role or using the PD account.

MikeNeilson:
I'm not really sure where the best place is, but I think this approach is suitable for now. It at least provides an extra barrier to entry to anyone who would currently be authorized to perform this operation.

adamkorynta:
Ok, I'll proceed with adding some tests

adamkorynta:
For a failing location delete (non-cascade) I added a 405 error code. It seemed that or 409 seemed applicable, but I'm definitely interested in thoughts.

cwms_radar_api/src/main/java/cwms/radar/api/LocationController.java
```diff
@@ -352,21 +369,19 @@ public void update(Context ctx, @NotNull String locationId) {
...
-            logger.log(Level.WARNING, re.toString(), e);
-            ctx.status(HttpServletResponse.SC_NOT_FOUND);
-            ctx.json(re);
-        } catch (IOException ex) {
-            RadarError re = new RadarError("Failed to delete location");
-            logger.log(Level.SEVERE, re.toString(), ex);
-            ctx.status(HttpServletResponse.SC_INTERNAL_SERVER_ERROR).json(re);
+        } catch (DataAccessException ex) {
+            SQLException cause = ex.getCause(SQLException.class);
+            if(cause != null && cause.getErrorCode() == 20031) {
```
adamkorynta, line 381: error code 20031 is specific to deleting locations so I didn't put this up in the general ApiServlet handling

MikeNeilson: Makes sense. though it might be worth organizing all of those db error codes into a specific class at some point.

adamkorynta: a mechanism for codgen would be nice

MikeNeilson:
> For a failing location delete (non-cascade) I added a 405 error code. It seemed that or 409 seemed applicable, but I'm definitely interested in thoughts.

Looking at the 405 docs an "allow" header is required in response so I don't think it's quite right. Method in that context does generally mean the HTTP VERB, not our API operation.

409 Conflict makes more sense since desired operation is in conflict with the state of things.

cwms_radar_api/src/main/java/cwms/radar/api/LocationController.java
MikeNeilson: Honestly wouldn't be too hard. or really even need codegen (if you meant JOOQ), the errors are already listed as rows in a table.

adamkorynta: Yeah I was thinking of generating enums for some of the CWMS owned static tables: https://stackoverflow.com/questions/46429671/generate-enum-class-from-table-with-jooq

adamkorynta:
Works for me, I kept seeing 409 as being for POST

cwms_radar_api/src/main/java/cwms/radar/api/LocationController.java
MikeNeilson: Alternatively that code could get moved to the wrapException handler. But in this case this does seem pretty specific to this usage.

cwms_radar_api/src/test/java/cwms/radar/api/LocationControllerTestIT.java
```diff
@@ -0,0 +1,147 @@
...
+                .redirects().max(3)
+                .get("/locations/" + location.getName())
+                .then()
+                .log().body().log().everything(true)
+                .assertThat()
+                .statusCode(is(HttpServletResponse.SC_OK));
+
+        // delete without cascade should fail
+        given()
+                .config(RestAssured.config().jsonConfig(jsonConfig().numberReturnType(JsonPathConfig.NumberReturnType.DOUBLE)))
```
MikeNeilson: This should be the default, was it giving you trouble without it?

cwms_radar_api/src/test/java/cwms/radar/api/LocationControllerTestIT.java
```diff
@@ -0,0 +1,147 @@
...
+                .when()
+                .redirects().follow(true)
+                .redirects().max(3)
+                .delete("/locations/" + location.getName())
+                .then()
+                .log().body().log().everything(true)
+                .assertThat()
+                .statusCode(is(HttpServletResponse.SC_CONFLICT));
+
+        // delete without cascade should fail
```
MikeNeilson: Incorrect comment.

MikeNeilson (changes requested):
A few nitpicks and a question.

MikeNeilson (dismissed):

cwms_radar_api/src/test/java/cwms/radar/api/LocationControllerTestIT.java
adamkorynta: yeah worked without it, I just copy pasted from elsewhere

cwms_radar_api/src/test/java/cwms/radar/api/LocationControllerTestIT.java
adamkorynta: fixed

MikeNeilson (approved):

PR #395, opened 2023-06-08
Feature/dto requires office dao uses

MikeNeilson:
Still some work to do. Also this likely can be done much cleaner but we're a bit in "just get it done mode" so I'm going for as long as it works and we can clean it up over time later.

### DIFF
--- a/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesIdentifierDescriptorDao.java
+++ b/cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesIdentifierDescriptorDao.java
@@ -58,13 +58,18 @@ public class TimeSeriesIdentifierDescriptorDao extends JooqDao<TimeSeriesIdentif
     public void create(TimeSeriesIdentifierDescriptor tsid, boolean versionedFlag,
                        Number intervalForward, Number intervalBackward, boolean failIfExists
     ) {
-        BigDecimal tsCode = CWMS_TS_PACKAGE.call_CREATE_TS_CODE(dsl.configuration(),
+        dsl.connection(c -> {
+            BigDecimal tsCode = CWMS_TS_PACKAGE.call_CREATE_TS_CODE(
+                getDslContext(c,tsid.getOfficeId()).configuration(),
                 tsid.getTimeSeriesId(),
                 tsid.getIntervalOffsetMinutes(), intervalForward, intervalBackward,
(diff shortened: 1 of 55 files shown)


cwms_radar_api/src/main/java/cwms/radar/data/dao/TimeSeriesIdentifierDescriptorDao.java
```diff
@@ -58,13 +58,18 @@ public TimeSeriesIdentifierDescriptorDao(DSLContext dsl) {
     public void create(TimeSeriesIdentifierDescriptor tsid, boolean versionedFlag,
                        Number intervalForward, Number intervalBackward, boolean failIfExists
     ) {
-        BigDecimal tsCode = CWMS_TS_PACKAGE.call_CREATE_TS_CODE(dsl.configuration(),
+        dsl.connection(c -> {
+            BigDecimal tsCode = CWMS_TS_PACKAGE.call_CREATE_TS_CODE(
+                getDslContext(c,tsid.getOfficeId()).configuration(),
```
adamkorynta, line 64: looks like most (all?) `getDslContext` calls wind up just calling `.configuration()` could replace by just returning the `Configuration`

MikeNeilson: That's a good point, mostly I was just trying to keep that original signature but it doesn't really matter. But I think there's larger improvements to be made to clean up the whole thing and keep them all in line.

adamkorynta:
Just to clarify, this PR will move CDA past CWMS Schema version 23.03.16?

MikeNeilson:
> Just to clarify, this PR will move CDA past CWMS Schema version 23.03.16?

On to 23.03.16, *unless* you have have multi office setup in which case it will need a small fix to cwms_env or the next schema update.

MikeNeilson:
Also, the api literally doesn't work in some required contexts without those changes. So I'm a bit between a rock and a hard place.

MikeNeilson:
Just to be clear, this is a smidge rushy and squishy with how it's done. I would actually like to revamp the entire AccessManager system to make a separation between user identification and privileges, especially to make future work easier. But for now we just need certain things to work.

The tests that were disabled were because they were failing for reasons not related to the work at hand and will be investigated later.

adamkorynta (approved):

PR #401, opened 2023-06-09
Feature/rename to cda

MikeNeilson:
Rename things to "dataapi" where it makes sense, otherwise "cda".

Still need to rename all the test time series, that'll be the next commit pushed.


MikeNeilson:
Dear lord, that's a lot of files. But such is a total rename.

MikeNeilson:
No functional changes. other than some minor "if not this new prop/env try the old one"

adamkorynta (approved):

PR #403, opened 2023-06-09
add in method for retrieving levels as time series

### LINKED ISSUE #398: add timeseries retrieve for levels endpoint
The analogous JDBC method: https://bitbucket.hecdev.net/projects/CWMS/repos/hec-cwms-data-access/browse/hec-db-jdbc/src/main/java/wcds/dbi/oracle/cwms/CwmsLevelJdbcDao.java#1027

Given a level id, start and end time, produce a CWMS time series.

This is used in CWMSVue by the RatingEditor's reference tab and RatingCalculator's reference plot.

adamkorynta:
Resolves #398


cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -308,8 +285,10 @@ protected void configureRoutes() {
                 new ParametersController(metrics), requiredRoles);
         cdaCrud("/timezones/{zone}",
                 new TimeZoneController(metrics), requiredRoles);
+        LevelsController levelsController = new LevelsController(metrics);
         cdaCrud("/levels/{" + Controllers.LEVEL_ID + "}",
-                new LevelsController(metrics), requiredRoles);
+                levelsController, requiredRoles);
+        get("/levels/timeseries/{level-id}", levelsController::getLevelAsTimeSeries);
```
MikeNeilson: I feel like this should be "/levels/{level-id}/timeseries/"

MikeNeilson: basic logic there is it's the "time series for the level" not the "level of a timeseries."

MikeNeilson (changes requested):
The amount of timeseries logic dupilcation is a bit unfortunate. Any way to get that rearranged to just use the TimeSeriesDaoImpl class instead of copying it?

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
adamkorynta: That's what I had wanted to do, but I was following the /timeseries/recent/{group-id} pattern

MikeNeilson: Hmm. arguably that timeseries should be `/timeseries/{group-id}/recent` instead of the way it is.

adamkorynta: Sample size of one doesn't really make a pattern, so I'm happy to do levels/{level-id}/timeseries

MikeNeilson: True. Make the change, what you originally intended flows more naturally which is what we should do if at all possible.

adamkorynta:
As far as duplicate logic, are you thinking to re-write this as a SQL query instead? Even if we did that, we'd wind up writing our own interpolation logic (or re-using math classes in monolith?) and wouldn't end up being able to reuse that between the time series dao anyways.

MikeNeilson:
> As far as duplicate logic, are you thinking to re-write this as a SQL query instead? Even if we did that, we'd wind up writing our own interpolation logic (or re-using math classes in monolith?) and wouldn't end up being able to reuse that between the time series dao anyways.

Maybe I'm wrong, but it just looked like you were directly recreating some of the logic from the TimeSeriesDaoImpl control after getting all the required parameters.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -540,5 +555,85 @@ private void getEnAttributeValue(StringValueUnits stringValueUnits, Double oattr
...
+            int quality = 0;
+            if (qualityCode != null) {
+                quality = qualityCode.intValue();
+            }
+            timeSeries.addValue(dateTime, value, quality);
+        }
+        return timeSeries;
+    }
+
+    private ZTSV_ARRAY buildTsvArray(Instant start, Instant end, Interval interval, ZoneId locationTimeZone) {
```
MikeNeilson, line 624: maybe just this part feel duplicated. The above is fairly specific to getting from the location level.

Maybe it's also just the names that are the same.

adamkorynta: The difference here is that the time series dao leverages the AV_TSV view generating the time steps for us. For the levels we generate the time step in CDA and request the values/quality from CWMS.

I could see an iteration on this design where we passed in an array of times in the GET body and skip interval processing at the CDA side altogether. I opted for CDA doing the processing as it would be less of a payload sent from the client and I didn't want to get into the complication of whether we should even be adding a body to a GET request.

MikeNeilson: Oh, I see. that make sense. We can revisit later to see if there is improved design.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
adamkorynta: updated

adamkorynta:
Now added a test

adamkorynta:
image
for some reason the swagger page isn't using the TAG part of the OpenAPI docs to group the call with the reset of the levels. Any advice?

MikeNeilson (approved):

MikeNeilson:
Oops, missed the question about the API before merging. But not at present, there's another doing the name. I'll see if I can figure it out.

MikeNeilson:
This is why `... TimeSeriesController. Please add the `path` and the `method` information to the annotation, so the handler can be matched.`

MikeNeilson:
However, on both the required information is set so it will require more investigation.

adamkorynta:
I added an issue: https://github.com/USACE/cwms-data-api/issues/404

PR #407, opened 2023-06-12
typo fix for xmlelement annotation

adamkorynta:


rma-rripken (approved):

MikeNeilson (approved):
Though, we probably could've just ditched the XML. But perhaps that is an effort for a few years from now.

PR #410, opened 2023-06-13
Feature/allow older db

### LINKED ISSUE #409: Tweak auth behavior to allow install and option on 23.03.16 schema if conditions allow

MikeNeilson:
Fixes #409


MikeNeilson:
FYI, after review it might seem rather obvious that "umm, we really only need one access manager now" and yes, but that work will come later to shift the different auth logic to a before handler or something.

cwms-data-api/src/main/java/cwms/cda/security/OpenIDAccessManager.java
```diff
@@ -68,48 +66,22 @@ public OpenIDAccessManager(String wellKnownUrl, String issuer, int realmKeyTimeo
...
-            // Have it do our extra step last.
-            cpDs.setPreparer(new DelegatingConnectionPreparer(existingPreparer, apiPreparer));
-        } else {
-            ctx.attribute(ApiServlet.DATA_SOURCE,
-                          new ConnectionPreparingDataSource(apiPreparer, dataSource));
-        }
-    }
-
-    private String getUserFromToken(Context ctx) {
+    private Optional<DataApiPrincipal> getUserFromToken(Context ctx) throws CwmsAuthException {
```
adamkorynta: return is never an empty Optional. Should just return DataApiPrincipal instead

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /**
+     * Primary logic to determine if a given principal can perform the desired operation.
+     * Throws exception if not valid, otherwise just returns.
+     * @param ctx
+     * @param p
+     * @param routeRoles
+     */
+    public static void isAuthorized(Context ctx, Optional<DataApiPrincipal> p, Set<RouteRole> routeRoles) throws CwmsAuthException {
```
adamkorynta: Anti-pattern using an Optional parameter

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -0,0 +1,223 @@
...
+                try (ResultSet rs = checkForKey.executeQuery()) {
+                    if (rs.next()) {
+                        return rs.getString(1);
+                    } else {
+                        logger.atInfo().log("No user for key");
+                        throw new CwmsAuthException("Access not authorized.");
+                    }
+                }
+            } catch (SQLException ex) {
+                logger.atWarning().withCause(ex).log("Failed API key check");
```
adamkorynta: The ApiServlet class already logs all thrown exceptions, the log messages in this method don't add any meaningful context.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -0,0 +1,223 @@
...
+        return Optional.of(new DataApiPrincipal(userName,roles));
+    }   
+
+    /**
+     * Setup session environment so we can query the required tables.
+     * @param conn
+     * @throws SQLException
+     */
+    private void setSessionForAuthCheck(Connection conn) throws SQLException {
+        if (hasEnvFix) {
```
adamkorynta: Worth some comments describing what hasEnvFix means.

cwms-data-api/src/main/java/cwms/cda/security/OpenIDAccessManager.java
MikeNeilson: Yeah, I mostly just did that to copy the structure of others. I plan to rearrange all this later to pass the generated DataApiPrincipal to a single AccessManager so all of this should be considered a stopgap.

cwms-data-api/src/main/java/cwms/cda/security/CwmsAccessManager.java
```diff
@@ -37,78 +27,19 @@ public class CwmsAccessManager extends CdaAccessManager {
...
-            buildDataSource(ctx);
-
-            // Let the handler handle the request.
-            handler.handle(ctx);
-        } else {
-            throw new CwmsAuthException("Access not authorized",HttpServletResponse.SC_FORBIDDEN);
-        }
+        Optional<DataApiPrincipal> p = getApiPrincipal(ctx);
+        AuthDao.isAuthorized(ctx,p,requiredRoles);
+        AuthDao.prepareContextWithUser(ctx, p.get());
```
adamkorynta: Calling `get()` on Optional without isPresent()

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: good point.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: ... really? seems better than a null check.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: Hmm, yeah but it doesn't have the details. I'll rework the exception too.

cwms-data-api/src/main/java/cwms/cda/security/OpenIDAccessManager.java
adamkorynta: Sounds good, I got worried about the unchecked `Optional.get()` but then saw there was no nullable `Optional`

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
adamkorynta: Nulls, Optionals , and boolean flag parameters all fall under the same antipattern. It winds up having the function caller needing to know that there are essentially different implementations of the same method that route depending on which parameters were used. This becomes a maintenance issue in public API's as you then have the function callers dictating the implementation support rather than having the API's dictate what they support. 

https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3

adamkorynta (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: Okay, I can buy that. Having an actual function name could often be more readable.

PR #411, opened 2023-06-20
replace unbounded executor with a core pool size of 6

adamkorynta:
current T7 configurations use ~10 connections in the pool. Using a pool size of 6 will give some buffer room for other active connections

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/RatingMetadataDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/RatingMetadataDao.java
@@ -38,17 +38,31 @@ import hec.data.cwmsRating.RatingSet;
 import mil.army.usace.hec.cwms.rating.io.xml.RatingXmlFactory;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.jooq.*;
+import org.jooq.Condition;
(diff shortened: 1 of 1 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/RatingMetadataDao.java
```diff
@@ -64,11 +78,13 @@ public class RatingMetadataDao extends JooqDao<RatingSpec> {
             + "<ratings xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" "
             + "xsi:noNamespaceSchemaLocation=\"https://www.hec.usace.army"
             + ".mil/xmlSchema/cwms/Ratings.xsd\"/>";
-    private final Executor executor = Executors.newCachedThreadPool(r -> {
-        Thread thread = new Thread(r, getClass().getSimpleName());
-        thread.setDaemon(true);
-        return thread;
-    });
+    private final Executor executor = new ThreadPoolExecutor(6, Integer.MAX_VALUE,
```
MikeNeilson, line 81: You know, technically I think since this is a Java EE webapp we should be getting our thread pools from the server itself.

I don't want to deal with it now, but it is something to think about.

MikeNeilson (approved):

PR #413, opened 2023-06-22
Devops/backports

MikeNeilson:
Had to bring the sessionofficepreparer back for handle the guest case.

Also fixed a blinding obvious in hindsight error with not saving the connecting user name correctly.

Some of the other changes were attempting to track down by I was also getting pool exhaustion errors. It seems when we have an exception in certain jooq queries the connection that was checked out is never returned to the pool... really not sure why and not going to try and solve it in this PR.


cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -83,6 +89,8 @@ public static DSLContext getDslContext(Context ctx) {
             retval = DSL.using(dataSource, SQLDialect.ORACLE11G);
         } else {
             // Some tests still use this method
+            logger.atSevere().withStackTrace(StackSize.FULL)
```
adamkorynta, line 92: Good call

adamkorynta (approved):

build.gradle
```diff
@@ -10,7 +10,7 @@ eclipse {
 
 allprojects {
     group = 'mil.army.usace.hec.cwms'
-    version = '3.0.1'   // ApiServlet.VERSION should be updated to match MAJOR.MINOR changes.
+    version = '3.0.2'   // ApiServlet.VERSION should be updated to match MAJOR.MINOR changes.
```
adamkorynta: 3.0.4 now?

MikeNeilson: gah, that should be -SNAPSHOT. this is for going back into develop.

adamkorynta: ah, I see you changed it in a later commit

PR #418, opened 2023-06-29
Explicitly adding DeleteMethod to JavalinValidation registry because …

rma-rripken:
…it was possible for a Controller method that needed a DeleteMethod argument to get called before the Controllers static initialization block which registered the converter was triggered.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
@@ -24,19 +24,17 @@
 
 package cwms.cda.api;
 
+import static com.codahale.metrics.MetricRegistry.name;
+
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import cwms.cda.data.dao.JooqDao;
-import cwms.cda.data.dao.TimeSeriesIdentifierDescriptorDao;
 import io.javalin.core.validation.JavalinValidation;
 import io.javalin.core.validation.Validator;
 
-import static com.codahale.metrics.MetricRegistry.name;
-
 
 public final class Controllers {
-
 
 
     public static final String GET_ONE = "getOne";
@@ -132,8 +130,7 @@ public final class Controllers {
     static final String SPECIFIED_LEVEL_ID = "specified-level-id";
 
     static {
-        JavalinValidation.register(JooqDao.DeleteMethod.class,
-            Controllers::getDeleteMethod);
+        JavalinValidation.register(JooqDao.DeleteMethod.class, Controllers::getDeleteMethod);
     }
 
     private Controllers() {
@@ -234,11 +231,11 @@ public final class Controllers {
         return retval;
     }
 
-    public static TimeSeriesIdentifierDescriptorDao.DeleteMethod getDeleteMethod(String input) {
-        TimeSeriesIdentifierDescriptorDao.DeleteMethod retval = null;
+    public static JooqDao.DeleteMethod getDeleteMethod(String input) {
+        JooqDao.DeleteMethod retval = null;
 
         if (input != null) {
-            retval = TimeSeriesIdentifierDescriptorDao.DeleteMethod.valueOf(input.toUpperCase());
+            retval = JooqDao.DeleteMethod.valueOf(input.toUpperCase());
         }
         return retval;
     }
--- a/cwms-data-api/src/test/java/cwms/cda/api/ControllersTest.java
+++ b/cwms-data-api/src/test/java/cwms/cda/api/ControllersTest.java
@@ -11,9 +11,10 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import com.codahale.metrics.MetricRegistry;
-
+import cwms.cda.data.dao.JooqDao;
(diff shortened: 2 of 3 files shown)


cwms-data-api/src/test/java/cwms/cda/api/ControllersTest.java
```diff
@@ -245,4 +246,67 @@ void testDoubleQueryParamNull() {
         assertNull(intervalForward);
 
     }
+
+    @Test
+    void testDeleteMethod(){
+        JooqDao.DeleteMethod deleteMethod = Controllers.getDeleteMethod(null);
+        assertNull(deleteMethod);
+
+        try {
```
MikeNeilson: Use assertThrows
`assertThrows(IllegalArgumentException.class, () -> { deleteMethod = ...}, "Expected exception not thrown")`

This also makes it more clear in the build run output when the wrong exception gets thrown.

cwms-data-api/src/test/java/cwms/cda/api/ControllersTest.java
```diff
@@ -245,4 +246,67 @@ void testDoubleQueryParamNull() {
...
+        assertEquals(JooqDao.DeleteMethod.DELETE_DATA, deleteMethod);
+
+        deleteMethod = Controllers.getDeleteMethod("delete_key");
+        assertEquals(JooqDao.DeleteMethod.DELETE_KEY, deleteMethod);
+
+        deleteMethod = Controllers.getDeleteMethod("delete_all");
+        assertEquals(JooqDao.DeleteMethod.DELETE_ALL, deleteMethod);
+
+        try {
+            deleteMethod = Controllers.getDeleteMethod("delete-data");
```
MikeNeilson: same as above, use assertThrows

cwms-data-api/src/test/java/cwms/cda/api/ControllersTest.java
```diff
@@ -245,4 +246,67 @@ void testDoubleQueryParamNull() {
...
+        JavalinValidation jv = JavalinValidation.INSTANCE;
+
+        boolean hadConverter = jv.hasConverter(JooqDao.DeleteMethod.class);
+
+        if(!hadConverter){
+            String message = "FYI the converter was not registered.  "
+                    + "This means that the Controllers class static initializer hadn't been called."
+                    + " Triggering the static initializer to run.";
+            // System.out.println(message);
+            try {
```
MikeNeilson: It's acceptable to just had `throws Exception` to test methods.

cwms-data-api/src/test/java/cwms/cda/api/ControllersTest.java
```diff
@@ -245,4 +246,67 @@ void testDoubleQueryParamNull() {
...
+    void testDeleteMethodValidationRegistration(){
+        JavalinValidation jv = JavalinValidation.INSTANCE;
+
+        boolean hadConverter = jv.hasConverter(JooqDao.DeleteMethod.class);
+
+        if(!hadConverter){
+            String message = "FYI the converter was not registered.  "
+                    + "This means that the Controllers class static initializer hadn't been called."
+                    + " Triggering the static initializer to run.";
+            // System.out.println(message);
```
MikeNeilson: just delete or add logger and put a logger.fine or something.

cwms-data-api/src/test/java/cwms/cda/api/ControllersTest.java
```diff
@@ -245,4 +246,67 @@ void testDoubleQueryParamNull() {
...
+            try {
+                Class<?> ignored = Class.forName("cwms.cda.api.Controllers");
+                assertNotNull(ignored);
+            } catch (ClassNotFoundException e) {
+                throw new RuntimeException(e);
+            }
+            // Could also call a static method instead.
```
MikeNeilson: have this execute and assert or just remove.

cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
```diff
@@ -132,8 +130,7 @@ public final class Controllers {
     static final String SPECIFIED_LEVEL_ID = "specified-level-id";
 
     static {
-        JavalinValidation.register(JooqDao.DeleteMethod.class,
-            Controllers::getDeleteMethod);
+        JavalinValidation.register(JooqDao.DeleteMethod.class, Controllers::getDeleteMethod);
```
MikeNeilson, line 133: I take it just doing this once in ApiServlet did not do the trick?

MikeNeilson (changes requested):
Very nitpicky changes. Intent looks good. Honestly probably a good idea for all validation perhaps.

cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
rma-rripken: It should but the tests don't go through ApiServlet.  The registration probably should be its own method, then the test could call it.  Will move the registration test into ApiServletTest which makes more sense.

MikeNeilson: Okay, something like that you can do in a BeforeTest handler but I don't think this will harm anything so we can experiment with it later.

MikeNeilson (approved):

PR #420, opened 2023-07-05
Bugfix/419 location level units

### LINKED ISSUE #419: Location Levels endpoint takes "level" too literally
NAE attempted to retrieve a storage level. This caused a unit conversion error trying to convert ac-ft to feet. Endpoint show allow user to choose the units.

MikeNeilson:
Fixes #419.

Require specification of Units to location level end points. Levels are more than actual "elevation" levels.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
@@ -356,7 +356,7 @@ public class LevelsController implements CrudHandler {
                             + EFFECTIVE_DATE),
                     @OpenApiParam(name = EFFECTIVE_DATE, required = true, description = "Specifies "
                             + "the effective date of Location Level to be returned"),
-                    @OpenApiParam(name = UNIT, required = true, description = "Desired unit for "
+                    @OpenApiParam(name = UNIT, required = false, description = "Desired unit for "
                             + "the values retrieved.")
             },
             responses = {
@@ -645,7 +645,7 @@ public class LevelsController implements CrudHandler {
                             + " response. If this field is not specified, the default time zone "
                             + "of UTC shall be used.\r\nIgnored if begin was specified with "
                             + "offset and timezone."),
-                    @OpenApiParam(name = UNIT, required = true, description = "Desired unit for "
+                    @OpenApiParam(name = UNIT, required = false, description = "Desired unit for "
                             + "the values retrieved."),
             },
             responses = {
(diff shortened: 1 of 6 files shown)


adamkorynta:
Shouldn't the CWMS schema already be defaulting to the parameter's default SI units? What am I missing here?

If the schema isn't doing that for us, CDA should handle that default to match other API's.

MikeNeilson:
Look at the diff. For the single value it was specifically saying "give me the value in feet or meters" regardless of the Parameter. There are more unit options. 

Additionally, anytime data is returned the user should be able to specify the units they want data in. I'm not entirely opposed to a mechanism of looking up a default but it wasn't clear how to do that cleanly here.

adamkorynta:
Yeah my confusion isn't the bug itself, it's why we would need to require the units. I get that some use-cases would want to specify units, but there is also the need to not specify units (for CWMS java applications, we retrieve data assuming in SI and convert at the view layer). Our other API's don't require the units to be specified so I'm not sure why this one is different.

Passing null to the package procedures wasn't working?

MikeNeilson:
hmm, you know I didn't try it. Standby.

MikeNeilson:
Okay, that works.

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
```diff
@@ -55,6 +55,62 @@ public class LevelsControllerTestIT extends DataApiTestIT {
...
+                .build();
+            CwmsDataApiSetupCallback.getDatabaseLink().connection(c -> {
+                DSLContext dsl = dslContext((Connection) c, OFFICE);
+                LocationLevelsDaoImpl dao = new LocationLevelsDaoImpl(dsl);
+                dao.storeLocationLevel(level, level.getLevelDate().getZone());
+            });
+
+        //Read level without unit
+        given()
+            .accept(Formats.JSONV2)
```
MikeNeilson, line 76: @adamkorynta, @rma-rripken FYI for future tests I suggest following this formatting. It makes the given/when/then conditions more obvious when skimming.

Might not matter to a lot but it certainly helps me reading through things.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -675,6 +682,10 @@ public void getLevelAsTimeSeries(Context ctx) {
             String office = ctx.queryParam(OFFICE);
             String begin = ctx.queryParam(BEGIN);
             String end = ctx.queryParam(END);
+            String units = ctx.queryParam(UNIT);
+            if (units == null) {
```
adamkorynta: Can delete this check

adamkorynta (approved):

PR #424, opened 2023-07-14
only retrieve rating metadata for non alias locations

### LINKED ISSUE #422: ratings metadata endpoint returns location aliases
These should be trimmed out of the query or a parameter added to keep them if needed.

adamkorynta:
Fixes #422


adamkorynta:
This should speed up performance quite a bit. On our SWT test db the difference in returned rating ids is ~2k compared to ~11k

MikeNeilson (approved):

PR #429, opened 2023-07-31
Feature/api key endpoint

MikeNeilson:
Decided to move the User API key generation to the API.

This is a user operation after authentication so I think it's reasonable to present here. Most systems, like say BitBucket, work that way so we aren't way out in left field here.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
@@ -42,7 +42,12 @@ import io.javalin.core.util.Header;
 import io.javalin.core.validation.JavalinValidation;
 import io.javalin.http.Context;
 import io.javalin.http.HttpCode;
-import io.javalin.plugin.openapi.annotations.*;
+import io.javalin.plugin.openapi.annotations.HttpMethod;
+import io.javalin.plugin.openapi.annotations.OpenApi;
+import io.javalin.plugin.openapi.annotations.OpenApiContent;
+import io.javalin.plugin.openapi.annotations.OpenApiParam;
+import io.javalin.plugin.openapi.annotations.OpenApiRequestBody;
+import io.javalin.plugin.openapi.annotations.OpenApiResponse;
 import mil.army.usace.hec.cwms.rating.io.xml.RatingXmlFactory;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -108,7 +113,7 @@ public class RatingController implements CrudHandler {
         try (final Timer.Context ignored = markAndTime("create"); DSLContext dsl =
                 getDslContext(ctx)) {
             RatingDao ratingDao = getRatingDao(dsl);
-            RatingSet ratingSet = deserializeRatingSet(ctx);
+            String ratingSet = deserializeRatingSet(ctx);
             boolean storeTemplate = ctx.queryParamAsClass(STORE_TEMPLATE, Boolean.class).getOrDefault(true);
             ratingDao.create(ratingSet, storeTemplate);
             ctx.status(HttpServletResponse.SC_ACCEPTED).json("Created RatingSet");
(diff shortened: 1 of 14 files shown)


MikeNeilson:
Not sure why the rating spec changes got in here; however they are required and allow those tests to actually run so I'm going to just go with it.

adamkorynta:
This would allow us to create an API Key GUI in CWMS clients if we wanted to, as opposed to prompting the user to open a webpage in the browser.

MikeNeilson:
Yep, which could then also be used for longer session. Wouldn't solve the "I want my model to have run in the morning" issue since that's the IT department restarting the computer. But it does make some other things easier.

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
```diff
@@ -134,24 +139,18 @@ private ContentType getContentType(Context ctx) {
...
     }
 
-
-    public RatingSet deserializeRatingSet(String body, ContentType contentType)
-            throws IOException, RatingException {
-        return deserializeRatingSet(body, contentType.getType());
-    }
-
-    public RatingSet deserializeRatingSet(String body, String contentType) throws IOException,
+    String deserializeRatingSet(String body, String contentType) throws IOException,
```
DanielTOsborne, line 146: Shouldn't this be renamed to "serializeRatingSet"?
Seems like the operation was inverted.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -52,7 +61,11 @@ public class AuthDao extends Dao<DataApiPrincipal>{
     private static final String CHECK_API_KEY =
         "select userid from cwms_20.at_api_keys where apikey = ?";
 
-
+    public static final String CREATE_API_KEY = "insert into cwms_20.at_api_keys(userid,key_name,apikey,created,expires) values(UPPER(?),?,?,?,?)";
+    public static final String REMOVE_API_KEY = "delete from cwms_20.at_api_keys where UPPER(userid) = UPPER(?) and key_name = ?";
```
DanielTOsborne, line 67: If AT_API_KEYS table has an index, does it have a function-based index on UPPER(userid)?

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -250,4 +263,116 @@ private static String getFailMessage(@NotNull Context ctx,
...
+     * @return The created ApiKey
+     */
+    public ApiKey createApiKey(DataApiPrincipal p, ApiKey sourceData) throws CwmsAuthException {
+        
+        try {
+            if(!p.getName().equalsIgnoreCase(sourceData.getUserId())) {
+                throw new CwmsAuthException(ONLY_OWN_KEY_MESSAGE, HttpCode.UNAUTHORIZED.getStatus());
+            }
+            SecureRandom randomSource = SecureRandom.getInstanceStrong();
+            String key = randomSource.ints(48,122) // allow a-zA-Z0-9
```
DanielTOsborne: Personally, I'd probably use something like
`randomSource.ints((char)'0',(char)'z')`
to make it clearer to the reader what the range, and exclusions are.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -250,4 +263,116 @@ private static String getFailMessage(@NotNull Context ctx,
...
+                                 .collect(StringBuilder::new,StringBuilder::appendCodePoint, StringBuilder::append)
+                                 .toString();
+            final ApiKey newKey = new ApiKey(sourceData.getUserId().toUpperCase(),sourceData.getKeyName(),key,ZonedDateTime.now(ZoneId.of("UTC")),sourceData.getExpires());
+            dsl.connection(c->{
+                setSessionForAuthCheck(c);
+                try (PreparedStatement createKey = c.prepareStatement(CREATE_API_KEY);) {
+                    createKey.setString(1,newKey.getUserId());
+                    createKey.setString(2,newKey.getKeyName());
+                    createKey.setString(3,newKey.getApiKey());
+                    createKey.setDate(4,new Date(newKey.getCreated().toInstant().toEpochMilli()));
```
DanielTOsborne: The date will be interpreted in the JVM's timezone. Pass a Calendar object for the 3rd argument to enforce UTC interpretation.
For all set/getDate calls.
https://docs.oracle.com/javase/8/docs/api/java/sql/PreparedStatement.html#setDate-int-java.sql.Date-java.util.Calendar-

cwms-data-api/src/test/java/cwms/cda/api/auth/ApiKeyControllerTestIT.java
DanielTOsborne, line 1: Is there a test for using an already expired key?
I couldn't see any actual date range validation code.

Or, a non-existent key.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: Userid is ALWAYS already uppercase. So i might not need the function here but it seemed clearer... and it's possible that will change in the future.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: Thanks. that likely would've bit me a some point.

cwms-data-api/src/test/java/cwms/cda/api/auth/ApiKeyControllerTestIT.java
MikeNeilson: Non-existent yes, though I think it's buried in there. I don't think I checked for an expired key though. I'll add that.

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
MikeNeilson: I'm not going to tweak the name. I think some of the confusion stems from how the libraries we use work.

Plus this just needs to get into `develop`, I was a little surprised to see it in this branch but it is needed for the Rating tests to work.

MikeNeilson:
@DanielTOsborne changes made.

rma-rripken (approved):

PR #430, opened 2023-08-24
Correct date parsing in ApiKey DTO.

MikeNeilson:
Default deserialization could not handle standard formats. Specified a Custom deserializer to use our existing DateUtils that supports more formats.

Cleaned up some of the timeseries.csv data but otherwise allow creation to file. This was due to a recent database schema change that became more picky about the lat/long, county, and nation values.


rma-rripken (approved):
An ApiKey unit test would be a nice addition.

MikeNeilson:
Hmm, fair point and it would make that particular test faster. But I'll consider it in the future.

PR #433, opened 2023-09-14
Bugfix/426 add category filter

### LINKED ISSUE #426: Add category filter parameter to location group and time series group getAll endpoints
This will allow filtering to only groups for specific categories (like "Agency Aliases")

rma-rripken:
Closes #426

### DIFF
--- a/cwms-data-api/src/main/webapp/regexp.html
+++ b/cwms-data-api/src/main/webapp/regexp.html
@@ -1,0 +1,105 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <title>CDA - CWMS Data API</title>
+        <meta name="Description" content="CDA CWMS Data API" />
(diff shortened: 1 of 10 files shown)


cwms-data-api/src/main/webapp/regexp.html
```diff
@@ -0,0 +1,105 @@
+<!DOCTYPE html>
```
MikeNeilson, line 1: Great addition.

MikeNeilson (approved):

cwms-data-api/src/main/webapp/regexp.html
rma-rripken: If users need help figuring out how to write something tricky in Oracle syntax we can add to these examples.

rma-rripken: Actually, maybe we should say that - Request help from HEC if you would like help crafting a regular expression or have suggested examples to add to this doc.  I'm gonna merge this now and let you figure out how to add to or fix that part

MikeNeilson: I would word that as "As for help from HEC or open a new discussion at <link to github discussion>" but yes that's a good idea.

Issue #428, opened 2023-07-17
catalog/locations endpoint paging is suspiciously slow


adamkorynta:
With out test SWT database instance, I am seeing ~90 seconds per page retrieval regardless of page sizing. I am suspicious that this means we are not paging correctly and are somehow querying the entire catalog in each page.

MikeNeilson:
It's possible. But it could just be the planner going wrong. Which would me either the query needs to get changed or some extra indexes or hints added.

But given other reported timing issues I suspect it's just the view itself and it's getting worked on.

rma-rripken:
In working on #427 I found that the Locations catalog query wasn't quite right.  I think that after the first page of results it wasn't adding the various filters back onto the select query for all the other pages.    Its possibly still very slow but if I got the changes right I suspect it will be a little bit better.

rma-rripken:
That this hasn't come up again suggests to me that the mentioned changes did fix the issue

PR #434, opened 2023-09-15
Added a bounding-office-like regex to the Catalog LOCATIONS datatype.…

### LINKED ISSUE #427: Ability to query by "bounding office"
CWMS Locations have a concept of an "owning or database source" office and a bounding office. The bounding office is for the district boundary that a given station actually resides in.

Some districts already require this as all the data is under a division ID, additionally as we centralize some locations will be "owned" by CWMS.

Where relevant we'll need to allow searching by the bounding office instead of the normal office field.

### LINKED ISSUE #437: Bounding office for timeseries catalog
It would be good to search for time series by bounding office.

I suspect the solution is either an additional "limit query"/"with clause" or minor mods to the existing one that already does some searching by location; just not sure if the bounding office is exposed correctly to that query.

rma-rripken:
… and refactored how the Locations catalog query is built.

Despite the branch name this is for #427 but I suspect it will help #428.

Does not include LocationController.getAll JSONv2 support.  That is tbd

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/Catalog.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/Catalog.java
@@ -1,56 +1,71 @@
 package cwms.cda.data.dto;
 
-import java.util.List;
-import java.util.Objects;
-
-import javax.xml.bind.annotation.*;
-
 import cwms.cda.api.errors.FieldException;
 import cwms.cda.data.dto.catalog.CatalogEntry;
 import cwms.cda.data.dto.catalog.LocationCatalogEntry;
 import cwms.cda.data.dto.catalog.TimeseriesCatalogEntry;
 import io.swagger.v3.oas.annotations.media.Schema;
+import java.util.List;
+import java.util.Objects;
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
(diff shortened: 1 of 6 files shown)


MikeNeilson (approved):

cwms-data-api/src/main/java/cwms/cda/data/dto/Catalog.java
```diff
@@ -95,55 +112,90 @@ public CatalogPage(String page){
...
-            StringBuilder sb = new StringBuilder();
-            sb.append(curOffice).append("/").append(cursorId)
-              .append(CwmsDTOPaginated.delimiter).append(searchOffice)
-              .append(CwmsDTOPaginated.delimiter).append(idLike)
-              .append(CwmsDTOPaginated.delimiter).append(locCategoryLike)
-              .append(CwmsDTOPaginated.delimiter).append(locGroupLike)
-              .append(CwmsDTOPaginated.delimiter).append(tsCategoryLike)
-              .append(CwmsDTOPaginated.delimiter).append(tsGroupLike);
-            return sb.toString();
+            return curOffice + "/" + cursorId
```
MikeNeilson, line 191: This is cleaner looking though I wonder if in the future we might want to create a "cursor builder". the delimited section would lend itself fairly easily to a `.items(String delimeter, String ...items)` type call vs doing the `+` everywhere.

PR #436, opened 2023-09-15
Bugfix/362 missing constants

### LINKED ISSUE #362: Migrate query parameters to Constants
considerations: Should this just be the shared by all like "office" or "page-size" or include nested classed for the specific endpoints.

rma-rripken:
I did a couple search/replace across the Controller classes and found and moved all the remaining constants that made sense to move into Controllers. 

closes #362


cwms-data-api/src/main/java/cwms/cda/api/BasinController.java
```diff
@@ -61,13 +65,13 @@ private Timer.Context markAndTime(String subject) {
                             + "their parameters."),
             },
             responses = {
-                    @OpenApiResponse(status = "200",
+                    @OpenApiResponse(status = STATUS_200,
```
MikeNeilson, line 68: In hindsight, Javalin really should've made these integers so the existing constants could be used.

MikeNeilson (approved):
Excellent cleanup. Thanks!

PR #440, opened 2023-09-29
Added bounding-office regex to Timeseries catalog.

### LINKED ISSUE #437: Bounding office for timeseries catalog

rma-rripken:
#427

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/CatalogController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/CatalogController.java
@@ -163,8 +163,8 @@ public class CatalogController implements CrudHandler {
                             description = "Deprecated. Use location-group-like."
                     ),
                     @OpenApiParam(name = BOUNDING_OFFICE_LIKE, description = "Posix <a href=\"regexp.html\">regular expression</a> "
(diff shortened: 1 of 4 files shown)


rma-rripken:
I couldn't find a bounding office on AV_CWMS_TS_ID2 so if that condition is added I have to join with AV_LOC2

MikeNeilson:
A reasonable choice though that av_loc2 view concerns me. Any immediate adverse impact on performance? Considering it's two columns it should be minimal but I verified yesterday that that's the view that killing most performance on the location catalog.

rma-rripken:
I made it so the join only happens if the bounding_office parameter is provided.  I haven't done any performance tests - my database just doesn't have a significant amount of data.

MikeNeilson:
Fair. If is is an issue it'll need to be worked on separately anyways, I suspect.

MikeNeilson (dismissed):

MikeNeilson (approved):

Issue #442, opened 2023-10-05
Timeseries retrieval not properly using binds


MikeNeilson:
https://github.com/USACE/cwms-data-api/blob/f6041742fbc6d16a95863ff71838d0c619823fbd/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java#L216

DSL.val is creating a literal instead of a bind.

PR #443, opened 2023-10-05
Bugfix/442 ts retrieve bind

MikeNeilson:
Dave pointed out that this query wasn't actually using a bind. I suspect is was due to the way that the retrieve_ts_out jooq plus DSL.sql was getting used but it was easy enough to just write it all out.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
@@ -209,19 +209,12 @@ public class TimeSeriesDaoImpl extends JooqDao<TimeSeries> implements TimeSeries
         // Now we're going to call the retrieve_ts_out_tab function to get the data and build an
         // internal table from it so we can manipulate it further
         // This code assumes the database timezone is in UTC (per Oracle recommendation)
-        SQL retrieveSelectData = DSL.sql("table(" +
-                CWMS_TS_PACKAGE.call_RETRIEVE_TS_OUT_TAB(
-                        tsId,
-                        unit,
-                        CWMS_UTIL_PACKAGE.call_TO_TIMESTAMP__2(DSL.val(beginTime.toInstant().toEpochMilli())),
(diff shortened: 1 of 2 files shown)


MikeNeilson:
This will fix #442

rma-rripken (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -209,19 +209,12 @@ protected TimeSeries getTimeseries(String page, int pageSize, String names, Stri
...
-                        CWMS_UTIL_PACKAGE.call_TO_TIMESTAMP__2(DSL.val(beginTime.toInstant().toEpochMilli())),
-                        CWMS_UTIL_PACKAGE.call_TO_TIMESTAMP__2(DSL.val(endTime.toInstant().toEpochMilli())),
-                        DSL.inline("UTC", String.class),
-                        // All times are sent as UTC to the database, regardless of requested
-                        // timezone.
-                        null, null, null, null, null, null, null,
-                        officeId)
-                + ") retrieveTs"
-        );
+        SQL retrieveSelectData = DSL.sql(
```
rma-rripken, line 212: Its too bad that there isn't a cleaner way to do this in jOOQ.  I hate having to drop into DSL.sql and string concat.

rma-rripken: Is the idea on this PR that be replacing those literals with binds the DB will then be able to reuse query plans ?  Sounds right - have you measured performance improvements?  How did Dave spot this opportunity?  Did you collect all the CDA queries and have him manually look at them?  Or was he looking at metrics on the database server and noticed the execution plans weren't being reused?  Curious if we can improve our process or automate things

MikeNeilson: That is correct. No specific measurements but it's well documented that not having queries setup with bind really mucks up the query planner and makes even normally fast queries fast which I have seen.

So the actual performance improvement is hard to measure as it's not specifically about this query but the total system.

PR #444, opened 2023-10-06
Cleanup setting of environment variables for development

ktarbet:


cwms-data-api/src/docker/Dockerfile
```diff
@@ -10,23 +10,14 @@ COPY server.xml /usr/local/tomcat/conf
...
-ENV CDA_JDBC_PASSWORD=
-ENV CDA_POOL_INIT_SIZE=
-ENV CDA_POOL_MAX_ACTIVE=
-ENV CDA_POOL_MAX_IDLE=
-ENV CDA_POOL_MIN_IDLE=
-
 EXPOSE 7000
+
+#  env variables  (set in docker-compose.yml for development)
+# ENV CDA_JDBC_DRIVER=
```
MikeNeilson: These should stay uncommented and with their defaults. They are an indication to users of the image as to what can and should be set for a given usage of the API.

MikeNeilson (changes requested):

MikeNeilson (dismissed):

MikeNeilson (approved):

MikeNeilson (approved):
Sorry, stale view. You did make the requested changes already.

PR #447, opened 2023-10-31
Bugfix/apikey header capitalization

### LINKED ISSUE #448: Apikey auth header requirements aren't clear.
The Swagger-UI doesn't automatically add "apikey" to the front of the auth header value like it would for bearer or basic.

Need to add user documentation so it's at least reasonable clear.

value is also required to be lower case only, which seems unnecessary.

### LINKED ISSUE #449: Auth header doesn't show up in Curl command reference on auth/keys GET handler
That specific GET handler should always require authorization unlike the majority of GET in CDA.

### LINKED ISSUE #450: Need to update Swagger Webjar

### LINKED ISSUE #451: new setClientInfo causing excessive error logging
Appears to be because of using `Context::path` instead of `Context::endpointHandlerPath` as path includes path parameters which can be rather long.

MikeNeilson:
User haven't been:
1. Typing apikey before the actual token when used. So added a note to the description (tried to make it the "scheme" but apparently OpenAPI/Swagger will only accept schemes defined in RFCs) 
2. Uses who did type it mixed case, made check case ingoring compare.
3. Minor cleanup and updates.

the new setClientInfo on the JDBC connections would end up too long and fail.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
@@ -20,6 +20,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 
 import javax.sql.DataSource;
@@ -51,7 +52,7 @@ public class AuthDao extends Dao<DataApiPrincipal>{
                                              + "23.03.16 or later to handle authorization operations.";
     public static final String DATA_API_PRINCIPAL = "DataApiPrincipal";
     // At this level we just care that the user has permissions in *any* office
-    private final String RETRIEVE_GROUPS_OF_USER;
+    private static String RETRIEVE_GROUPS_OF_USER = null;
 
     private static final String SET_API_USER_DIRECT = "begin "
         + "cwms_env.set_session_user_direct(upper(?));"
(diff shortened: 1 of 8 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -69,28 +70,31 @@ public class AuthDao extends Dao<DataApiPrincipal>{
     public static final String GET_SINGLE_KEY = "select userid,key_name,created,expires from cwms_20.at_api_keys where UPPER(userid) = UPPER(?) and key_name = ?";
     public static final String ONLY_OWN_KEY_MESSAGE = "You may not create API keys for any user other than your own.";
 
-    private boolean hasCwmsEnvMultiOfficeAuthFix;
-    private String connectionUser;
-    private String defaultOffice;
+    private static boolean hasCwmsEnvMultiOfficeAuthFix = false;
+    private static String connectionUser = null;
+    private static String defaultOffice = null;
```
rma-rripken, line 75: Making these static and conditionally setting them in the constructor weirds me out.  I can't say exactly why but it feels like asking for trouble. Like if the first request made is for an un-checked get end-point and the request includes office_id will it cause the static value to get assigned to whatever the first request provided?   At the very minimum it seems like assigning to AuthDao.defaultOffice shouldn't happen if the set_session_user_direct failed - right?

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -69,28 +70,31 @@ public class AuthDao extends Dao<DataApiPrincipal>{
...
+                dsl.execute("BEGIN cwms_env.set_session_user_direct(?,?)", connectionUser,defaultOffice);
+                hasCwmsEnvMultiOfficeAuthFix = true;
+            } catch (DataAccessException ex) {
+                if( ex.getLocalizedMessage()
+                    .toLowerCase()
+                    .contains("wrong number or types of arguments in call")) {
+                    hasCwmsEnvMultiOfficeAuthFix = false;
+                }
             }
+            AuthDao.RETRIEVE_GROUPS_OF_USER = ResourceHelper.getResourceAsString("/cwms/data/sql/user_groups.sql",this.getClass());
```
rma-rripken: Is there a reason to do this here vs at the variable declaration?

rma-rripken (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: I'm quite sure there was when I initial did the work.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: This always uses the office id of the context "e.g. cwms -> hq, spk -> spk" not the context of data.

For the 2nd part I think it just won't get used, but I don't think the additional logic is required just for that state.

MikeNeilson: But the short version is that I'm just trying to avoid creating a new instance of this object on every request since it will be used on a LOT of requests.

rma-rripken: ok. the part about the context makes more sense.  Should AuthDao have a static getInstance() and just use it singleton style?

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: Hmm, that's not a bad idea. Let me chew on that one a bit. It is supposed to be usable by all the possible Auth mechanisms.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: I considered the use of a singleton, that's going to need more investigation. Since the clientInfo is different for every request i needs the new context on each run so I think for now we do new to just run new on every request since their could be multiple threads. Or synchronize on the object which seems like a terrible idea.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: Well, that response was to the comment above but I did go ahead and just do this.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: Okay, after realizing it needs the javalin Context information of the current request so that setClientInfo is correct for tracing instead of a pure singleton I made it a singleton per thread.

rma-rripken (approved):

PR #452, opened 2023-11-01
Adding methods to add Cache-Control and ETags

### LINKED ISSUE #417: The generally static endpoints such as states, parameters, and units to have appropriate cache-control headers
This will allow browsers, client apis, and proxies to cache the data and increase performance for those data with minimal efforts.

rma-rripken:
I haven't been able to test this extensively b/c my IDE is suddenly complaining about flogger.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
+++ b/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
@@ -78,6 +78,7 @@ import io.javalin.apibuilder.CrudHandler;
 import io.javalin.apibuilder.CrudHandlerKt;
 import io.javalin.core.JavalinConfig;
 import io.javalin.core.security.RouteRole;
+import io.javalin.core.util.Header;
 import io.javalin.core.validation.JavalinValidation;
 import io.javalin.http.BadRequestResponse;
 import io.javalin.http.Handler;
@@ -90,6 +91,7 @@ import io.swagger.v3.oas.models.Operation;
 import io.swagger.v3.oas.models.PathItem;
 import io.swagger.v3.oas.models.info.Info;
 import io.swagger.v3.oas.models.security.SecurityRequirement;
+import java.util.concurrent.TimeUnit;
 import org.apache.http.entity.ContentType;
 import org.jetbrains.annotations.NotNull;
 import org.owasp.html.HtmlPolicyBuilder;
@@ -200,8 +202,7 @@ public class ApiServlet extends HttpServlet {
                     config.defaultContentType = "application/json";
                     config.contextPath = context;
                     getOpenApiOptions(config);
-                    //config.registerPlugin(new OpenApiPlugin());
-                    //config.enableDevLogging();
+                    config.autogenerateEtags = true;
                     config.requestLogger((ctx, ms) -> logger.atFinest().log(ctx.toString()));
                 })
                 .attribute("PolicyFactory", sanitizer)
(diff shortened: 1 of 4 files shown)


MikeNeilson:
Looks reasonable so far. We should add a check for the header to at least one of the existing integration tests.

cwms-data-api/src/test/java/cwms/cda/api/StateControllerTestIT.java
```diff
@@ -56,4 +57,24 @@ void test_state_catalog()  {
             .body("[0].name", equalTo("Unknown State or State N/A"))
             .body("[0].state-initial", equalTo("00"));
     }
+
+    @Test
+    void test_state_has_ETag_and_Cache_Control()  {
+        String matcher;
+        given()
+            .log().ifValidationFails(LogDetail.ALL,true)
```
MikeNeilson, line 66: Good test, ETag shows up, Cache-Control does not.

FYI had to merge with the apikey fix. one of the things I fixed there was preventing the proper retrieval of the datasource.

rma-rripken: Oh. I think I know whats going on.   At least I hope I do.  In my testing I saw that it adds it as cache-control all lowercase.   according to SO  http/2.0 says that the header field names must be converted to lower case.  
https://stackoverflow.com/a/41169947/249623

rma-rripken: I still haven't been able to get my IDE to run things.  I'm working on something else at the moment but its annoying.

MikeNeilson: The test I ran printed the headers, the cache-control header wasn't present for any casing... which was surprising I didn't see anything wrong with the code you made.

What kind of errors is you're IDE giving you? Karl had some issues with Intellij and Ant due to a version mismatch. Might be something similar since we need to fix the build to allow a java11+ compiler to work even if we are still targetting 8.

MikeNeilson: Also, It's reasonable to put those header expectations in the test_state_catalog. Seems a waste of resources to just make the request again.

Though an interesting test would be to make the request again right after and see if it's cached, but I'm not sure if the restassured client actually does that.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -313,59 +314,98 @@ protected void configureRoutes() {
...
+     */
+    public static void cdaCrudCache(@NotNull String path, @NotNull CrudHandler crudHandler,
+                                    @NotNull RouteRole[] roles, long duration, TimeUnit timeUnit) {
+        cdaCrud(path, crudHandler, roles);
+        addCacheControl(path, duration, timeUnit);
+    }
+
+    private static void addCacheControl(@NotNull String path, long duration, TimeUnit timeUnit) {
+        if(timeUnit != null && duration > 0) {
+            staticInstance().after(path, ctx -> {
```
MikeNeilson, line 405: This is matching on the exact string passed in. in the case of states it's waiting for exactly `/states/{state}`

This worked: `... after(path.replaceAll("(.*)(\\{.*\\})","$1*"), ...`

rma-rripken: States was maybe a bad controller to test this with b/c States doesn't have a getOne method.   Having the replacement string in the path does work for matching requests that include the resourceId but it doesn't match the getAll style requests.  I played around with the regex a bit but I couldnt' get it to work to match both types of url so I just add two after handlers.

MikeNeilson: I would say it was a good controller as it exposed a problem.

MikeNeilson: oh, it should be `(.*)(\\{.*\\})?` so that the path parameter is optional... though that wouldn't match if we include multiple path parameters so maybe a replaceAll with just the `\\{.*\\}` -> `*` would do the trick.

rma-rripken: I added a test where you can play around with the regex approach.  I couldn't get it to work - from what I can see the after handler system doesn't use regex - they do their own custom matching and manipulation.

MikeNeilson: the regex isn't used by the after handler, it just substitutes an ```"{text}"``` with with * so ```"/locations/{location-id}"``` becomes ```"/locations/*"``` as ```"/path/*"``` was the example shown.

Though I can except that that still isn't quite enough, the example was limited and some of our handler path can be a bit more complex.

rma-rripken:
@MikeNeilson I think this is ready

MikeNeilson (approved):
Integration test passes. Nice simple change that should improve things.

rma-rripken:
This relates to contract W912HQ22F0193 Task 8 Topic A.

PR #456, opened 2023-11-07
log the full version name of CDA

adamkorynta:
this can be collected by dashboarding in order to compare metrics between version changes


rma-rripken (approved):

rma-rripken:
This relates to contract W912HQ22F0193 Task 8 Topic B.

PR #461, opened 2023-11-08
Feature/remove deprecated

rma-rripken:
This removes query params that were previously marked as deprecated except for the two places using deprecated "format" params.  I think the format params should be handled in a different way.

This relates to contract W912HQ22F0193 Task 8 Topic A.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ eclipse {
 
 allprojects {
     group = 'mil.army.usace.hec.cwms'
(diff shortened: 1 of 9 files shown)


rma-rripken:
I'm thinking that this should also bump the version from 3.0 to 3.1 ?  thoughts?

adamkorynta:
There should be a review of the CWMS 3.3.0 and CWMS 3.3.1 usage. Should also review that the CDA client repo isn't using the deprecated parameters.

rma-rripken:
The needed CWMS changes have been done.   This PR is ready.

MikeNeilson (approved):

PR #470, opened 2023-11-29
text-timeseries changes.

rma-rripken:


cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -587,7 +587,7 @@ public TimeSeries retrieveLocationLevelAsTimeSeries(ILocationLevelRef levelRef,
         CWMS_ENV_PACKAGE.call_SET_SESSION_OFFICE_ID(dsl.configuration(), officeId);
         ZTSV_ARRAY locLvlValues = CWMS_LEVEL_PACKAGE.call_RETRIEVE_LOC_LVL_VALUES3(dsl.configuration(),
                 specifiedTimes, locationLevelId, levelUnits, attributeId, attributeValue, attributeUnits,
-                "UTC", officeId);
+                "UTC", officeId, "VN");
```
MikeNeilson: what is VN?

cwms-data-api/src/main/java/cwms/cda/data/dao/texttimeseries/RegularTimeSeriesTextDao.java
```diff
@@ -0,0 +1,244 @@
...
+                throw new RuntimeException(e);  // TODO: wrap with something else.
+            }
+        }
+        return rows;
+    }
+
+    @Nullable
+    private static List<RegularTextTimeSeriesRow> buildRows(ResultSet rs) throws SQLException {
+        OracleTypeMap.checkMetaData(rs.getMetaData(), timeSeriesTextColumnsList, TYPE);
+        List<RegularTextTimeSeriesRow> rows = null;
```
MikeNeilson: shouldn't this just allocate the array list or this there a specific reason we're allowing null if rs.next() is empty?

MikeNeilson (commented):
Looks reasonable thus far.

NOTE: This review is mostly so I can use GH's "since last reviewed" feature.

cwms-data-api/src/main/java/cwms/cda/data/dao/texttimeseries/RegularTimeSeriesTextDao.java
rma-rripken: You are probably right.  I will revisit null vs empty.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
rma-rripken: I don't entirely remember making that change but it looks like it was related to updating the jooq codegen.  The method signature now has a p_level_precedence parameter and the procedure def lists "VN" as the default. 

Here's the doc on it:  
 @param p_level_precedence   One or two characters that specify the precedence of virtual and non-virtual location levels in the results.
 *                             The default value 'VN' gives virtual location levels higher precedence, allowing them to hide non-virtual location levels
 *                             where the two types overlap in time. Valid value are:
 * <ul>
 *   <li><b>N</b> specifies results from non-virtual (normal) location levels only
 *   <li><b>V</b> specifies results from virtual location levels only
 *   <li><b>NV</b> specifies results from non-virtual (normal) location levels where they exist, with virtual location levels allowed where non-virtual levels don't exist
 *   <li><b>VN</b> (default) specifies results from virtual location levels where they exist, with non-virtual location levels allowed where virtual levels don't exist
 * </ul>

rma-rripken: I could change it to null - thats probably the more correct thing to do.

rma-rripken: no, passing null doesn't seem to trigger the default.  not sure how to make the call from jooq without specifying the parameter so that the default will take effect.

MikeNeilson: I'd leave it in just put a short comment about the behavior and where to find more information. I doubt I'm the only one that might get confused by that.

cwms-data-api/src/test/java/cwms/cda/data/dao/texttimeseries/RegularTimeSeriesTextDaoTestIT.java
```diff
@@ -210,4 +209,72 @@ private void testDelete(RegularTimeSeriesTextDao dao) {
 
     }
 
+
+    @Test
+    void testStore() throws SQLException {
+        CwmsDatabaseContainer<?> databaseLink = CwmsDataApiSetupCallback.getDatabaseLink();
```
MikeNeilson, line 236: FYI the before all needs to create the location with the base case handler.

Wasn't able to run the test locally due to the location not already existing.

MikeNeilson (commented):
again, mostly set for GH tracking; but wasn't able to run the test locally due to one comment provided.


Oh and you might want to rebase from develop to pick up the gradle changes. (not a huge deal but it does solve a problem for me at least of always having to make sure I have the local environment set to the correct JDK.)

rma-rripken:
Debugging text-timeseries pl/sql errors required https://bitbucket.hecdev.net/projects/CWMS/repos/cwms_database/pull-requests/319

MikeNeilson (commented):
That's looking good. I like the minor changes to organization and function names. Lot more clear.

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+    private Timer.Context markAndTime(String subject) {
+        return Controllers.markAndTime(metrics, getClass().getName(), subject);
+    }
+
+
+    @OpenApi(
+            queryParams = {
+                    @OpenApiParam(name = OFFICE, required = true, description = "Specifies the owning office of "
+                            + "the Text TimeSeries whose data is to be included in the response."),
+                    @OpenApiParam(name = TS_ID, required = true, description = "Specifies the ts-id of the "
```
adamkorynta: numerical time series uses `NAME`

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+                            + "text timeseries"),
+                    @OpenApiParam(name = TIMEZONE,  description = "Specifies "
+                            + "the time zone of the values of the begin and end fields (unless "
+                            + "otherwise specified). If this field is not specified, "
+                            + "the default time zone of UTC shall be used."),
+                    @OpenApiParam(name = START, required = true, description = "The start of the time window"),
```
adamkorynta: numerical time series uses `BEGIN`

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+                    @OpenApiParam(name = END, description = "The end of the time window. If specified the text associated with all times from start to end (inclusive) is deleted."),
+                    @OpenApiParam(name = VERSION_DATE, description = "The version date for the time series.  If not specified, the minimum or maximum version date (depending on p_max_version) is used."),
+                    @OpenApiParam(name = MAX_VERSION, type = Boolean.class, description = "Whether to use the maximum version date if p_version_date is not specified."),
+                    @OpenApiParam(name = Controllers.MIN_ATTRIBUTE, type = Long.class, description = "The minimum attribute value to delete. If not specified, no minimum value is used."),
```
adamkorynta: Confused by deleting an attribute here on a retrieval?

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+                    @OpenApiParam(name = TS_ID, required = true, description = "Specifies the ts-id of the "
+                            + "text timeseries"),
+                    @OpenApiParam(name = TIMEZONE,  description = "Specifies "
+                            + "the time zone of the values of the begin and end fields (unless "
+                            + "otherwise specified). If this field is not specified, "
+                            + "the default time zone of UTC shall be used."),
+                    @OpenApiParam(name = START, required = true, description = "The start of the time window"),
+                    @OpenApiParam(name = END, description = "The end of the time window. If specified the text associated with all times from start to end (inclusive) is deleted."),
+                    @OpenApiParam(name = VERSION_DATE, description = "The version date for the time series.  If not specified, the minimum or maximum version date (depending on p_max_version) is used."),
+                    @OpenApiParam(name = MAX_VERSION, type = Boolean.class, description = "Whether to use the maximum version date if p_version_date is not specified."),
```
adamkorynta: Can we use a null `VERSION_DATE` to stand for `MAX_VERSION`?

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+                    @OpenApiParam(name = OFFICE, required = true, description = "Specifies the owning office of "
+                            + "the Text TimeSeries whose data is to be included in the response."),
+                    @OpenApiParam(name = TS_ID, required = true, description = "Specifies the ts-id of the "
+                            + "text timeseries"),
+                    @OpenApiParam(name = TIMEZONE,  description = "Specifies "
+                            + "the time zone of the values of the begin and end fields (unless "
+                            + "otherwise specified). If this field is not specified, "
+                            + "the default time zone of UTC shall be used."),
+                    @OpenApiParam(name = START, required = true, description = "The start of the time window"),
+                    @OpenApiParam(name = END, description = "The end of the time window. If specified the text associated with all times from start to end (inclusive) is deleted."),
```
adamkorynta: typo "deleted" -> "retrieved"

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
adamkorynta: also, is the default time window of the last 24 hours applicable here (to match numerical time series)? or should the `END` parameter also be required?

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+                    @OpenApiParam(name = VERSION_DATE, description = "The version date for the time series.  If not specified, the minimum or maximum version date (depending on p_max_version) is used."),
+                    @OpenApiParam(name = MAX_VERSION, type = Boolean.class, description = "Whether to use the maximum version date if p_version_date is not specified."),
+                    @OpenApiParam(name = Controllers.MIN_ATTRIBUTE, type = Long.class, description = "The minimum attribute value to delete. If not specified, no minimum value is used."),
+                    @OpenApiParam(name = Controllers.MAX_ATTRIBUTE, type = Long.class, description = "The maximum attribute value to delete. If not specified, no maximum value is used."),
+                    @OpenApiParam(name = "mode", required = true, type = TimeSeriesTextMode.class,
+                            description = "Type of Text TimeSeries to retrieve. Options are:\n"
```
adamkorynta: image

formatting on the Mode renders funky

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+        }
+    }
+
+    @OpenApi(
+        description = "Updates a text timeseries",
+        pathParams = {
+            @OpenApiParam(name = TS_ID, description = "The id of the text timeseries to be updated"),
+        },
+            queryParams = {
+                    @OpenApiParam(name = MAX_VERSION, type = Boolean.class, description = "Whether to use the maximum version date if p_version_date is not specified."),
```
adamkorynta: If we aren't using a p_version_date yet, we should leave off the `MAX_VERSION` (I'm not even sure what max_version means in the case of storage)

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+                    @OpenApiParam(name = REPLACE_ALL, type = Boolean.class)
```
adamkorynta: Given `REPLACE_ALL` isn't required, what is the null behavior?

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+                        + "the default time zone of UTC shall be used."),
+                @OpenApiParam(name = START, required = true, description = "The start of the time"
+                        + " window"),
+                @OpenApiParam(name = END, description = "The end of the time window. If specified"
+                        + " the text associated with all times from start to end (inclusive) is "
+                        + "deleted."),
+                @OpenApiParam(name = VERSION_DATE, description = "The version date for the time "
+                        + "series.  If not specified, the minimum or maximum version date "
+                        + "(depending on p_max_version) is used."),
+                @OpenApiParam(name = MAX_VERSION, type = Boolean.class, description = "Whether to"
```
adamkorynta: `MAX_VERSION` is confusing as I would think the default behavior would be true. I'd recommend leaving this parameter out and always passing in true to the pl/sql if the `VERSION_DATE` is null.

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,357 @@
...
+                        + "the time zone of the values of the begin and end fields (unless "
+                        + "otherwise specified). If this field is not specified, "
+                        + "the default time zone of UTC shall be used."),
+                @OpenApiParam(name = START, required = true, description = "The start of the time"
+                        + " window"),
+                @OpenApiParam(name = END, description = "The end of the time window. If specified"
+                        + " the text associated with all times from start to end (inclusive) is "
+                        + "deleted."),
+                @OpenApiParam(name = VERSION_DATE, description = "The version date for the time "
+                        + "series.  If not specified, the minimum or maximum version date "
```
adamkorynta: let's not support minimum version date at this time

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
adamkorynta: Now I see, it was copied from delete

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
rma-rripken: I added the default value to the docs

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
rma-rripken: yes

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
rma-rripken: I think it means the individual rows.  it the rows don't include a version-date what happens?

adamkorynta: I think that whatever the client is supplying should get stored. There shouldn't be a flag to the server specifying that some of the content is supposed to be ignored.

rma-rripken: I don't think anything got ignored but I'll remove the parameter.

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -0,0 +1,380 @@
...
+            required = true),
+        queryParams = {
+
+            @OpenApiParam(name = REPLACE_ALL, type = Boolean.class, description = "Whether to replace any and all existing text with the specified text. Default is " + DEFAULT_CREATE_REPLACE_ALL)
+                        },
+        method = HttpMethod.POST,
+        tags = {TAG}
+    )
+    @Override
+    public void create(@NotNull Context ctx) {
```
adamkorynta, line 226: @MikeNeilson given the LRTS identifier updates coming down the pipeline. It it acceptable to not support storage of new LRTS text time series? Otherwise we'd need a- flag "store-as-lrts" similar to numerical time series.

I'd think skipping a "store-as-lrts" in both text and binary time series makes sense given we'll get the feature for "free" once the schema update is complete. This would mean we could turn on sessioning flags to require the new identifier and return the new identifier.

cwms-data-api/src/main/java/cwms/cda/data/dao/texttimeseries/StandardTimeSeriesTextDao.java
```diff
@@ -0,0 +1,312 @@
...
+        connection(dsl, c -> {
+            setOffice(c, officeId);
+            CwmsDbText dbText = CwmsDbServiceLookup.buildCwmsDb(CwmsDbText.class, c);
+            dbText.storeStdText(c, stdTextId, stdText, failIfExists, officeId);
+        });
+    }
+
+    public void storeRows(String officeId, String tsId, boolean maxVersion, boolean replaceAll, Collection<StandardTextTimeSeriesRow> stdRows) {
+        for (StandardTextTimeSeriesRow stdRow : stdRows) {
+            store(officeId, tsId, stdRow, maxVersion, replaceAll);
```
adamkorynta: Parameterize in the Connection so that we aren't pulling a new connection from the pool on every row

cwms-data-api/src/main/java/cwms/cda/data/dao/texttimeseries/RegularTimeSeriesTextDao.java
```diff
@@ -0,0 +1,296 @@
...
+
+        return builder.build();
+    }
+
+    public void storeRows(String officeId, String id, boolean maxVersion, boolean replaceAll, Collection<RegularTextTimeSeriesRow> regRows) {
+        // This could be made into a more efficient bulk store.
+        // We'd have to sort the rows by textId and textValue pairs and then build a set of all the matching dates
+        // Then for each set of dates we'd call the appropriate storeTsText or storeTsTextId method.
+
+        for (RegularTextTimeSeriesRow regRow : regRows) {
```
adamkorynta: Another place to parameterize the connection to not tax the pool

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/TextTimeSeries.java
```diff
@@ -0,0 +1,275 @@
...
+
+@JsonDeserialize(builder = TextTimeSeries.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public class TextTimeSeries extends CwmsDTO {
+//  This Monolith version of this class has:
+// https://bitbucket.hecdev.net/projects/SC/repos/hec-monolith/browse/hec-monolith/src/main/java/hec/data/timeSeriesText/TextTimeSeries.java
+//    private ITimeSeriesDescription _timeSeriesDescription;
+    
+    private final String id;
```
adamkorynta: rename to `name` to match numerical time series

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/TextTimeSeries.java
```diff
@@ -0,0 +1,275 @@
...
+import java.util.Map;
+import java.util.NavigableMap;
+import java.util.Set;
+import java.util.TreeMap;
+
+
+@JsonDeserialize(builder = TextTimeSeries.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public class TextTimeSeries extends CwmsDTO {
```
adamkorynta, line 28: Is `CwmsDTOPaginated` going to be implemented in this PR or a follow-up?

adamkorynta, line 28: I recommend adding the location time-zone and the interval offset similar to numerical time series for LRTS.

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/TextTimeSeries.java
rma-rripken: this is where I was wondering if it should use something like the commented out ITimeSeriesDescription.  It would break id/name into its component parts(annoying) but let us specialize the Interval-type.   @adamkorynta  I was hoping you could weigh in on this more.  how should this handle regular, irregular, psuedo-regular etc.

adamkorynta: We need the full id, the interval offset, and the location time zone. Given the LRTS interval rename migration is underway I don't think we need to do anything further as far as breakdown unless CDA started doing interval processing itself. If the timeseries description included ID, Offset, and Loc Time Zone, that would be fine. Otherwise having three separate variables seems ok to me. 

Maybe the non-graphQL-but-kinda-like-graphql updates down the line would see a catalog endpoint with just those three pieces.

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/RegularTextTimeSeriesRow.java
```diff
@@ -0,0 +1,223 @@
...
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public class RegularTextTimeSeriesRow implements TextTimeSeriesRow {
+
+    private final Date dateTime;
+    private final Date versionDate;
+    private final Date dataEntryDate;
+    private final Long attribute;
+    private final String textId;
+    private final String textValue;
+    private final boolean newData;
```
adamkorynta: Pretty sure `newData` is just used by REGI as a modified flag

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/TextTimeSeries.java
```diff
@@ -0,0 +1,275 @@
...
+//  This Monolith version of this class has:
+// https://bitbucket.hecdev.net/projects/SC/repos/hec-monolith/browse/hec-monolith/src/main/java/hec/data/timeSeriesText/TextTimeSeries.java
+//    private ITimeSeriesDescription _timeSeriesDescription;
+    
+    private final String id;
+
+    private final NavigableMap<DateDateKey, RegularTextTimeSeriesRow> regularMap;
+    private final NavigableMap<DateDateKey, StandardTextTimeSeriesRow> stdMap;
+    private Map<String, StandardCatalog> stdCatalog = null;
```
adamkorynta: I'm confused by the usage of the catalog here. My first thought was that this is where the standard text would match the ids to the value in order to reduce the payload sent back to the client, but the Dao looks like it just always sets the text for the standard rows.

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/RegularTextTimeSeriesRow.java
rma-rripken: Thank you! That has been driving me crazy

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/TextTimeSeries.java
rma-rripken: After.

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/TextTimeSeries.java
rma-rripken: Its a little convoluted.   I'll simplify and move it all into the Builder.   I still need to figure out how or when to trigger the catalog behavior - probably another flag.

cwms-data-api/src/main/java/cwms/cda/data/dao/texttimeseries/RegularTimeSeriesTextDao.java
```diff
@@ -0,0 +1,307 @@
...
+            Jooq names this one:   call_STORE_TS_TEXT__2  - this is what we use
+         */
+
+        if(textId != null && textValue != null){
+            // There are two storeTs methods.  You either:
+            // 1.  store a textValue at specific times but you don't care about what the textId is.
+            // 2.  make an existing textId apply at the specified times - you don't care about the current textId to textValue.
+            // This branch is if the user is trying to specify the text_Id and the text_value.
+            // We'll have to make some choices to implement this.
+            throw new UnsupportedOperationException(String.format("TextId:\"%s\" and TextValue:\"%s\" are both specified.  "
```
adamkorynta: This winds up returning error code 500 when I think we want a 400 here

cwms-data-api/src/main/java/cwms/cda/api/TextTimeSeriesController.java
```diff
@@ -226,8 +246,12 @@ public void create(@NotNull Context ctx) {
     @OpenApi(
         description = "Updates a text timeseries",
         pathParams = {
-            @OpenApiParam(name = TS_ID, description = "The id of the text timeseries to be updated"),
+            @OpenApiParam(name = "timeseries", description = "The id of the text timeseries to be updated"),
```
MikeNeilson: Should we make this a constant just for the sake of consistency?

cwms-data-api/src/main/java/cwms/cda/data/dao/texttimeseries/RegularTimeSeriesTextDao.java
MikeNeilson: Yep. Not that "unsupportedOperation" is linguistically wrong.

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/TextTimeSeries.java
MikeNeilson: Given the standard text values are relatively small, the controller/dto/dao, whereever it makes sense (prob the DAO) could cache the standard values after they're retrieved the first time.

MikeNeilson (commented):
Feels like it's starting to come together.
My sense is that it's mostly nitpicky details now, next week we should probably plan, assuming what's in already works for what it is, merging in the current state so we can try to get additional feedback. Seems to me like the primary interface is "there".

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/TextTimeSeries.java
rma-rripken: I added them to the DTO but its gonna take me a bit to figure out how to query for them.  I know the timeseries has a metadata query that fills some of that in...

cwms-data-api/src/main/java/cwms/cda/data/dao/texttimeseries/StandardTimeSeriesTextDao.java
```diff
@@ -0,0 +1,317 @@
...
+        setOffice(c, officeId);
+        CwmsDbText dbText = CwmsDbServiceLookup.buildCwmsDb(CwmsDbText.class, c);
+        dbText.storeStdText(c, stdTextId, stdText, failIfExists, officeId);
+    }
+
+    public void storeRows(String officeId, String tsId, boolean maxVersion, boolean replaceAll, Collection<StandardTextTimeSeriesRow> stdRows) {
+        connection(dsl, connection -> {
+            setOffice(connection, officeId);
+            for (StandardTextTimeSeriesRow stdRow : stdRows) {
+                store(officeId, tsId, stdRow, maxVersion, replaceAll);
```
adamkorynta: connection needs to be passed down into here

cwms-data-api/src/test/resources/cwms/cda/api/spk/text_ts_create_reg.json
```diff
@@ -0,0 +1,16 @@
+{
+  "office-id": "SPK",
+  "id": "TsTextTestLoc.Flow.Inst.1Hour.0.raw",
```
MikeNeilson: name vs id?

cwms-data-api/src/test/resources/cwms/cda/api/spk/text_ts_create_std.json
```diff
@@ -0,0 +1,24 @@
+{
+  "office-id": "SPK",
+  "id": "TsTextTestLoc.Flow.Inst.1Hour.0.raw",
```
MikeNeilson: name vs id?

cwms-data-api/src/test/resources/cwms/cda/api/spk/text_ts_update_reg.json
```diff
@@ -0,0 +1,20 @@
+{
+  "office-id": "SPK",
+  "id": "TsTextTestLoc.Flow.Inst.1Hour.0.raw",
```
MikeNeilson: name vs id?

MikeNeilson (changes requested):
Having a few issues actually running the tests (with the latest-dev) but that looks like a minor issue with the test setup data that I'll fix later.

However once that was solved some of the new tests didn't work due some of the input json not getting updated.

cwms-data-api/src/test/resources/cwms/cda/api/spk/text_ts_create_reg.json
rma-rripken: weird that the test didn't catch that.  now I have to double check what is using that resource.

cwms-data-api/src/test/resources/cwms/cda/api/spk/text_ts_create_std.json
rma-rripken: should have been name.

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/TextTimeSeries.java
rma-rripken: The time-zone and interval offset will be addressed by https://github.com/USACE/cwms-data-api/issues/547

MikeNeilson:
Just updated PR to this PR, new tests still not passing. Overall structure looks good though; appears that data just isn't saving correctly but with no error message that was visible to me.

adamkorynta:
I pulled the latest changes in this PR and the controller integration tests pass for me.

MikeNeilson:
Of course now there's a conflict on one of the files.

MikeNeilson (approved):

PR #474, opened 2023-12-04
Devops/update tomcat

MikeNeilson:
Multiple reasons to update tomcat, but I've also been noticing that the test are running with 8.0.something instead of 9, turns out it was the tomcat8jndi dependency we didn't need anymore.

Additionally migrated docker base image to Alpine. The tomcat base image kept showing various high/medium vulnerability and it was difficult to get that image to update.


rma-rripken (dismissed):
Looks fine to me except there is some sort of merge conflict.

MikeNeilson:
Oh thanks. I think I tweak something in the middle of all that.

MikeNeilson:
Okay, that's better.

rma-rripken (approved):

PR #496, opened 2024-01-08
Update to gradle 8.5.

MikeNeilson:
Move dependencies to libs.version.toml.
Issue with annotation list generation that needs to be resolved.


rma-rripken (approved):

adamkorynta (approved):
Pulled the changes, was able to build and run locally

PR #509, opened 2024-01-11
Feature/501 content type alias

### LINKED ISSUE #501: Update endpoint: states
Create application/json alias

MikeNeilson:
Closes #501 and allows work on the tickets #501 blocks.


MikeNeilson:
I'm inclined to just move literally everything to this new system but wanted to get feedback before I did that.

MikeNeilson:
NOTE: i renamed the annotations folder as I was originally starting to move everything in there, but then released the lookup could just be done at runtime. I plan to just delete the annotation processing that was originally used.

rma-rripken (dismissed):
I don't see any problems.  I'll take whatever improvements we can get.

MikeNeilson:
I'll update this PR to migrate all of them. I won't handle the other work from the tickets but this will at least stage for it, and let me discover any other possible issues.

Probably tomorrow though.

MikeNeilson:
I just realized how much of a slog that will be. I'm going to leave the PR at this much work. As the endpoints get modified to include the appropriate aliases required it makes sense to just do these changes then.

Once it's all done then we can delete the original annotation processor.

rma-rripken (dismissed):

MikeNeilson:
Okay, two things
1. I rebased this on #575 
2. I decided this morning to just go make all the changes required for removing the DTO references from the Formatters.

I'll do the actual alias work in a new PR.

rma-rripken (approved):

PR #510, opened 2024-01-11
Devops/tweak openid for different access url

MikeNeilson:
Was finally able to retrieve an active JWT and see how things worked. CDA can now use the EDIPI provided by the JWT to retrieve CWMS user credentials.

Possible future work involved database changes to allow use of the "sub" which would be appropriate to allow Keycloak Direct  Grant OR Login.gov.

The majority the auth flows are not usable by our users, and some scopes don't work even through their presented. Removing the oauth flows is easy, I'll investigate later setting an appropriate default scope and client id.

Oh and for anyone that ever needs to try TLS/Mutual auth from a termina in windows:

```
curl -v --ca-native --cert "CurrentUser\MY\<Thumbprint from CAC cert>" https://url-requiring-mutual-auth.test" -d "grant_type=password" -d "username=" -d "password=" -d "response_type=code" -d "client_id=<client scope>" -d "scope=openid profile email"
```

Note with the above email did not show up within the JWT, @willbreitkreutz the JWT clearly said email not verified so seems reasonable to ; is there a process for a user to update their information on keycloak, couldn't quite figure out how to log in to my account correctly, assuming it's even configured to do so.


cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -206,6 +208,49 @@ private String checkKey(String key) throws CwmsAuthException {
...
+     * @param edipi
+     * @return
+     * @throws CwmsAuthException
+     */
+    private String userForEdipi(Long edipi) throws CwmsAuthException {
+        try {
+            return dsl.connectionResult(c-> {
+                setSessionForAuthCheck(c);
+                try (PreparedStatement userForEdipi = c.prepareStatement(USER_FOR_EDIPI)) {
+                    userForEdipi.setLong(1, edipi);
```
adamkorynta, line 222: Hard to follow the long boxing/unboxing but this is where a null check should occur, though the Long.parseLong earlier returns a primitive....

MikeNeilson: Oh, that should just be a primitive anyways. Oops.

compose_files/pki/certs/main.conf
```diff
@@ -7,3 +7,6 @@ subjectAltName = @alt_names
 DNS.1 = mike-virtual-machine
 DNS.2 = auth.test
 DNS.3 = cwms-data.test
+DNS.1 = mike-virtual-machine
+DNS.2 = auth.test
+DNS.3 = cwms-data.test
```
ktarbet, line 12: extra lines seem to keep coming back.
```suggestion

```

compose_files/pki/certs/main.crt
ktarbet, line 1: I think several files in pki/certs/* area are re-generated and don't need to be included here?

cwms-data-api/src/main/java/cwms/cda/security/OpenIDConfig.java
```diff
@@ -80,6 +81,15 @@ public OpenIDConfig(URL wellKnown) throws IOException {
         }
     }
 
+    private URL substituteBase(URL endPoint, String altAuthUrl) throws MalformedURLException {
+        if (altAuthUrl == null) {
```
ktarbet:
```suggestion
        if (altAuthUrl == null || altAuthUrl.isEmpty()) {
```

docker-compose.README.md
```diff
@@ -29,11 +29,11 @@ can be verified correctly.
...
 The following users and permissions are available:
 
-| User        | Password    | Office | Permissions    |
-| ----------- | ----------- | ------ | ------------   |
-| l2hectest   | l2hectest   | SPK    | General User   |
-| l1hectest   | l1hectest   | SPL    | No permissions |
-| m5hectest   | m5hectest   | SWT    | General User   |
+| User                  | Password    | Office | Permissions    |
+| --------------------- | ----------- | ------ | ------------   |
+| l2hectest.1234567890  | l2hectest   | SPK    | General User   |
```
ktarbet, line 34: Is username.EDIPI convention in the CWMS database?    It feels like an attribute being attached to the username.

compose_files/pki/certs/main.conf
MikeNeilson: Hmm, maybe I should just delete and add those to the .gitignore. it is intentionally using the local hostname as one of the SANs. But that is an extra user step.

docker-compose.README.md
MikeNeilson: No that's a quirk of keycloak and how it's using the CAC "CN=" portion of the user certificate. That said if it's confusing I can see if I can set it to the normal name but still have the JWT provide the version with the edipi.

ktarbet (approved):

willbreitkreutz:
@MikeNeilson there is a way for users to update their profile, first name, last name and e-maiil anyway.  we're working on getting that workflow documented.  We don't have e-mail verification wired up so they all show not-verified even if there is an e-mail, probably something we need to get incorporated.

adamkorynta (approved):

MikeNeilson:
This all got wrapped into #1007

PR #518, opened 2024-01-25
[WIP] Feature/async sse.  Do not merge. 

rma-rripken:


cwms-data-api/src/main/webapp/sse_demo.html
```diff
@@ -0,0 +1,32 @@
...
+<h1>Getting server updates</h1>
+<div id="result"></div>
+
+<script>
+    // document.cookie = "Authorization=apikey sessionkeyfortesting";
+
+    if (typeof (EventSource) !== "undefined") {
+        var clobId = "/TIME SERIES TEXT/6261044";
+        console.log("clobId: " + clobId + " encodeURIComponent: " + encodeURIComponent(clobId) + " encodeURI: " + encodeURI(clobId));
+        var source = new EventSource("http://localhost:7000/cwms-data/sse/clob"
```
MikeNeilson, line 14: I think I understand part of why it's not as common. Entirely new data type, doesn't seem too difficult to use but it's clear a fallback is required anyways so we might as well do the various long-polling methods.

MikeNeilson: Though we should keep it in mind for things like CWMS Server/Client communication and say, team forecasting. And I can see some value over at OpenDCS/LRGS. But really just in a constrained context where you know it can work.

MikeNeilson:
Need to save this initial idea has ended, we've settled on using an RSS feed for the moment.

PR #524, opened 2024-01-27
Feature/519 text plain

### LINKED ISSUE #519: Add plain/text format to Clob end-point that skips the JSON encapsulation and returns the raw text.
This feature would be used by text-timeseries and async retrieval if it was present.

rma-rripken:
closes #519

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/ClobController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/ClobController.java
@@ -164,6 +164,14 @@ public class ClobController implements CrudHandler {
     @OpenApi(
             queryParams = {
                     @OpenApiParam(name = OFFICE, description = "Specifies the owning office."),
+                    @OpenApiParam(name = CLOB_ID, description = "If this _query_ parameter is provided the id _path_ parameter "
+                            + "is ignored and the value of the query parameter is used.   "
+                            + "Note: this query parameter is necessary for id's that contain '/' or other special "
+                            + "characters.  Because of abuse even properly escaped '/' in url paths are blocked.  "
+                            + "When using this query parameter a valid path parameter must still be provided for the request"
+                            + " to be properly routed.  If your clob id contains '/' you can't specify the clob-id query "
+                            + "parameter and also specify the id path parameter because firewall and/or server rules will "
+                            + "deny the request even though you are specifying this override. \"ignored\" is suggested." )
             },
             responses = {@OpenApiResponse(status = STATUS_200,
                     description = "Returns requested clob.",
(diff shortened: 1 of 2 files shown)


cwms-data-api/src/main/java/cwms/cda/api/ClobController.java
```diff
@@ -156,19 +155,27 @@ public void getAll(@NotNull Context ctx) {
...
 
 
     @OpenApi(
             queryParams = {
                     @OpenApiParam(name = OFFICE, description = "Specifies the owning office."),
+                    @OpenApiParam(name = CLOB_ID, description = "If this _query_ parameter is provided the id _path_ parameter "
+                            + "is ignored and the value of the query parameter is used.   "
+                            + "Note: this query parameter is necessary for id's that contain '/' or other special "
+                            + "characters.  Because of abuse even properly escaped '/' in url paths are blocked.  "
+                            + "When using this query parameter a valid path parameter must still be provided for the request"
```
MikeNeilson, line 169: It's actually a slight failure in Javalin in how it even processes the paths... there's a section that just, doesn't.

But I like this wording.

MikeNeilson: Well, and could be both.

cwms-data-api/src/test/java/cwms/cda/api/ClobControllerTestIT.java
```diff
@@ -69,26 +64,41 @@ void test_create_getOne() throws JsonProcessingException, UnsupportedEncodingExc
             .assertThat()
             .statusCode(is(HttpServletResponse.SC_CREATED));
 
-        /* There is an issue with how javalin handles / in the path that are actually part
-        of the object name (NOTE: good candidate for actually having a GUID or other "code"
-        as part of the path and the actual name as a query parameter.
-        )
+//        /* There is an issue with how javalin handles / in the path that are actually part
```
MikeNeilson, line 67: Odd additional line comments vs block comments.

MikeNeilson (changes requested):
Looks reasonable to me. Just some odd, probably automatic, formatting to fix before it gets merged in.

cwms-data-api/src/main/java/cwms/cda/api/ClobController.java
rma-rripken: what could be both?

cwms-data-api/src/test/java/cwms/cda/api/ClobControllerTestIT.java
rma-rripken: yeah, that is odd.

cwms-data-api/src/main/java/cwms/cda/api/ClobController.java
MikeNeilson: Bad processing by Javalin (I checked)
But also the security issue. I know that encoding in the path can cause issues, but even after I turned off that check it still doesn't work even though the manual says it should.

MikeNeilson (approved):

Issue #520, opened 2024-01-26
Add id queryParameter to Clob end-point 


rma-rripken:
If provided this overrides the id pathParameter.
Needed to work around clobId that contain slashes

MikeNeilson:
This *may* also be a good time to start down the path of a "UUID/integer primary key" key like thing to be able to pass to the endpoint path parameter.

e.g. since it have intentionally *NOT* been returning the database code fields we could start (but not the code or using code) by encoding the office and name of the element similar to the paging.

e.g. office / name -> base64

... maybe not slash though. 

That said the queryParameter id would still be desired so definitely focus on that. If a simplified path parameter can be done that would set the path forward for the other end points.

PR #525, opened 2024-01-27
Feature/521 text plain

### LINKED ISSUE #521: Make Clob and Blob end-points that stream raw data.
End-point should start returning first byte quickly (before 10s idle time).
data is not encapsulated/chunked/paged etc.  Data does not need to be unescaped,decoded etc. 
Just raw data.
Ideally supports range-requests.

rma-rripken:
closes #521

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
@@ -1,7 +1,9 @@
 package cwms.cda.api;
 
 import static com.codahale.metrics.MetricRegistry.name;
+import static cwms.cda.api.Controllers.CREATE;
 import static cwms.cda.api.Controllers.CURSOR;
+import static cwms.cda.api.Controllers.FAIL_IF_EXISTS;
 import static cwms.cda.api.Controllers.GET_ALL;
 import static cwms.cda.api.Controllers.GET_ONE;
 import static cwms.cda.api.Controllers.LIKE;
@@ -16,6 +18,12 @@ import static cwms.cda.api.Controllers.queryParamAsClass;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.xml.JacksonXmlModule;
+import com.fasterxml.jackson.dataformat.xml.XmlMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import com.google.common.flogger.FluentLogger;
 import cwms.cda.api.errors.CdaError;
 import cwms.cda.data.dao.BlobDao;
 import cwms.cda.data.dao.JooqDao;
@@ -24,13 +32,18 @@ import cwms.cda.data.dto.Blobs;
 import cwms.cda.data.dto.CwmsDTOPaginated;
 import cwms.cda.formatters.ContentType;
 import cwms.cda.formatters.Formats;
+import cwms.cda.formatters.FormattingException;
+import cwms.cda.formatters.json.JsonV2;
 import io.javalin.apibuilder.CrudHandler;
 import io.javalin.core.util.Header;
 import io.javalin.http.Context;
 import io.javalin.http.HttpCode;
+import io.javalin.http.HttpResponseException;
+import io.javalin.plugin.openapi.annotations.HttpMethod;
 import io.javalin.plugin.openapi.annotations.OpenApi;
 import io.javalin.plugin.openapi.annotations.OpenApiContent;
 import io.javalin.plugin.openapi.annotations.OpenApiParam;
+import io.javalin.plugin.openapi.annotations.OpenApiRequestBody;
 import io.javalin.plugin.openapi.annotations.OpenApiResponse;
 import java.io.InputStream;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
```diff
@@ -138,50 +156,118 @@ public void getAll(Context ctx) {
...
-
-            if (optAc.isPresent()) {
-                Blob blob = optAc.get();
 
-                ctx.contentType(blob.getMediaTypeId());
-                byte[] value = blob.getValue();
-                ctx.result(value);
-
-                requestResultSize.update(value.length);
+            Consumer<Triple<InputStream, Long, String>> tripleConsumer = triple -> {
```
MikeNeilson, line 170: Neat trick.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -1,14 +1,33 @@
...
+import kotlin.Triple;
+import org.jetbrains.annotations.NotNull;
 import org.jooq.DSLContext;
 import org.jooq.Record;
 import org.jooq.ResultQuery;
+import usace.cwms.db.jooq.codegen.packages.CWMS_TEXT_PACKAGE;
 
 public class BlobDao extends JooqDao<Blob> {
+
+    public static final String BLOB_WITH_OFFICE = "SELECT CWMS_MEDIA_TYPE.MEDIA_TYPE_ID, AT_BLOB.VALUE \n"
```
MikeNeilson, line 21: Does the latest jooq codegen include the at_ tables?

If not, totally fine, if yes, we probably should them. 

Definitely okay with bypassing the view.

cwms-data-api/src/main/java/cwms/cda/api/ClobController.java
```diff
@@ -178,13 +187,39 @@ public void getAll(@NotNull Context ctx) {
...
+                        ctx.status(HttpServletResponse.SC_NOT_FOUND).json(new CdaError("Unable to find "
+                                + "clob based on given parameters"));
+                    } else {
+                        ctx.seekableStream(stream, "text/plain", length);
+                    }
+                };
+
+                dao.getClob(clobId, office.orElse(null), streamConsumer);
+
+                return;
```
MikeNeilson: Would it be better to put the next section in the "else" instead of returning.

@adamkorynta @ktarbet @DanielTOsborne 

NOTE: purely a question of style.

MikeNeilson (commented):
Looks good. a few nitpicks or just general questions.

cwms-data-api/src/main/java/cwms/cda/api/ClobController.java
adamkorynta: yeah an else is a lot more readable to me. Our general rule of thumb is one return per method if possible. Two max with an exception being short-circuiting when throwing IllegalArgumentException or some such at the top of the method

MikeNeilson: yeah, I'm all for direct simple returns of values in short functions, but total shortcut is just a little jarring.

cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
rma-rripken: I don't love it but I'm hoping we can try it for now.   It puts a kotlin class in the interface signature.  But the code is only going to work with Javalin which makes heavy use of kotlin so that seemed the fastest solution.  I didn't see a similar Java class sitting around.  Maybe we can create our own ThreeTuple class or a TriConsumer or something else in the future.  The jOOQ connection has to be open for the streaming to work so passing the consumer is the only way I figured out to make it work.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
rma-rripken: Its a little unclear to me what your comment is asking for.   
Is there a view for Blobs?  Before these changes the BlobDao was already going to AT_BLOB and that table had all the pieces I needed so I didn't look into changing what was queried.
jOOQ doesn't handle Blob and Clob the way I want.  I don't want them mapped to String or byte[]. I want the actual Blob object so I can connect the stream directly into to javalin's result stream.

cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
MikeNeilson: Yeah, it makes sense and that's what callbacks are for. The javalin reference seemed odd but it makes sense to me as we have to pull the library in. Nothing wrong with using something useful.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
MikeNeilson: I thought there was, but I might be wrong. but the JOOQ codegen *should* now also wrap the tables directly.

But if it doesn't behave correctly then don't worry about it. Though maybe a short comment explaining that for the next set of fools.

MikeNeilson (dismissed):

MikeNeilson (dismissed):

MikeNeilson (dismissed):
Perfect.

rma-rripken:
@MikeNeilson  Please review again. I think my change dismissed the previous approval

MikeNeilson (dismissed):

rma-rripken:
Why does it say I dismissed the approval?  All I did was refresh the page.

MikeNeilson:
weird. it's like it thinks you did a rebase again or something.

MikeNeilson (dismissed):

rma-rripken:
Maybe it doesn't like that #519 and #520 have been merged.  I'll try to merge in develop and see if that makes it happy

MikeNeilson:
That likely, I've seen similar behavior on other PR sets were i've branch off and made multiple PRs at the same time for review.

MikeNeilson (approved):

MikeNeilson:
I decided to just merge it in since we've done this back and forth like 3 times now I think; through no fault of either of us.

PR #537, opened 2024-02-09
fixing bug in timezone with location level delete

adamkorynta:
also clean up exception handling so that it bubbles up into the ApiServlet error code processing

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
@@ -215,7 +215,7 @@ public class LocationLevelsDaoImpl extends JooqDao<LocationLevel> implements Loc
         try {
             Timestamp date;
             if (zonedDateTime != null) {
-                date = Timestamp.from(zonedDateTime.toLocalDateTime().atZone(zonedDateTime.getZone()).toInstant());
+                date = Timestamp.from(zonedDateTime.toInstant());
             } else {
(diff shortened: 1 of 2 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -214,10 +214,13 @@ public void deleteLocationLevel(String locationLevelName, ZonedDateTime zonedDat
                                     String officeId, Boolean cascadeDelete) {
         try {
             Timestamp date;
+            String timeZone;
             if (zonedDateTime != null) {
-                date = Timestamp.from(zonedDateTime.toLocalDateTime().atZone(zonedDateTime.getZone()).toInstant());
+                date = Timestamp.from(zonedDateTime.toInstant());
+                timeZone = zonedDateTime.getZone().getId();
```
rma-rripken: This is creeping me out.  Is the timeZone = getZone() thing right?  I guess I'd need to try it in the debugger.  I would have thought it was always utc.

rma-rripken: oh, you changed the date calculation too.   I see it now

adamkorynta: yeah I guess you're right, since we're doing the time zone processing already higher up

rma-rripken (approved):

PR #538, opened 2024-02-10
binary time series

rma-rripken:


cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -373,6 +374,8 @@ protected void configureRoutes() {
         get(recentPath, tsController::getRecent);
         addCacheControl(recentPath, 5, TimeUnit.MINUTES);
 
+        cdaCrudCache("/timeseries/binary/{timeseries}",
```
MikeNeilson, line 377: Remind me if this has already been discuss but just looking at this would it make more sense to do:

`/timeseries/{timeseries}/binary`, and then the same with text, and while the default is "numeric" just adding an, alias for lack for better term, endpoint for numeric?

cwms-data-api/src/test/java/cwms/cda/data/dao/binarytimeseries/TimeSeriesBinaryDaoTest.java
```diff
@@ -0,0 +1,113 @@
+package cwms.cda.data.dao.binarytimeseries;
+
+
+import static cwms.cda.data.dao.DaoTest.getConnection;
```
MikeNeilson: Aren't we trying to move away from this. Needs the CDASEtup... extendswith anyways to make sure it's actually started and you can pull the cwms database testcontainer instance out of that.

MikeNeilson (commented):
Definitely looks like it's all coming together.

cwms-data-api/src/test/java/cwms/cda/data/dao/binarytimeseries/TimeSeriesBinaryDaoTest.java
rma-rripken: Oops.  I didn't mean to submit that.

cwms-data-api/src/main/java/cwms/cda/data/dto/binarytimeseries/BinaryTimeSeriesRow.java
```diff
@@ -0,0 +1,193 @@
...
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public class BinaryTimeSeriesRow {
+    private final Date dateTime;
+    private final Date versionDate;
+    private final Date dataEntryDate;
+    private final String binaryId;
+    private final Long attribute;
+    private final String mediaType;
+    private final String fileExtension;
+    private final byte[] binaryValue;
```
adamkorynta, line 22: Not entirely sure why, but my codegen returns this as a `List<byte[]>`. I don't know what we can do about that

rma-rripken: curious.  Is all the data in the first list item?  It doesn't spread it out or duplicate it

adamkorynta: It won't deserialize because the actual returned JSON only has the byte array. I've just hand-edited the codegenned file as a workaround.

adamkorynta: the schema comes out like:
```
        binary-value:
          type: array
          items:
            type: string
            format: byte
```
Which is what I'd expect

cwms-data-api/src/main/java/cwms/cda/api/BinaryTimeSeriesController.java
```diff
@@ -0,0 +1,344 @@
...
+                    beginZdt == null? null: beginZdt.toInstant(),
+                    endZdt== null? null: endZdt.toInstant(),
+                    versionZdt==null? null: versionZdt.toInstant(),
+                    maxVersion,
+                    retrieveBinary,
+                    minAttr, maxAttr);
+
+            ctx.contentType(contentType.toString());
+
+            String result = Formats.format(contentType, binaryTimeSeries);
```
adamkorynta, line 159: I don't see the corresponding add to JsonV2's format service

rma-rripken: yeah. Its probably missing and needs to be added.

cwms-data-api/src/main/java/cwms/cda/data/dao/binarytimeseries/TimeSeriesBinaryDao.java
```diff
@@ -0,0 +1,323 @@
...
+                versionInstant == null ? null : versionInstant.toInstant(),
+                maxVersion, minAttribute, maxAttribute);
+    }
+
+    private void delete(String officeId, String tsId, String binaryTypeMask,
+                       Instant startTime, Instant endTime, Instant versionInstant, boolean maxVersion,
+                       Long minAttribute, Long maxAttribute) {
+        TimeZone timeZone = OracleTypeMap.GMT_TIME_ZONE;
+
+        CWMS_TEXT_PACKAGE.call_DELETE_TS_BINARY(
```
adamkorynta: Needs a set office id session variable call

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
rma-rripken: I don't think we discussed it in depth.  Might be worth a revisit.  
Every other controller has the {id} part at the end of the path being registered.  Not saying that is ideal but it is consistent.

adamkorynta: I could then see wanting some fancy logic to check the Text/Binary parameter so that you don't need the trailing type. The trailing type would be required specifically for annotated versions. 

Kinda muddies up the API though.....

cwms-data-api/src/main/java/cwms/cda/api/BinaryTimeSeriesController.java
MikeNeilson: I should really get #509 merged in, but don't let that delay you; I'll rebase if needed.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -120,8 +120,17 @@ private static Connection setClientInfo(Context ctx, Connection connection) {
         return connection;
     }
 
-    public static DSLContext getDslContext(Connection database, String officeId) {
-        DSLContext dsl = DSL.using(database, SQLDialect.ORACLE18C);
+    public static DSLContext getDslContext(Connection connection, String officeId) {
```
MikeNeilson, line 123: Good comment.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
MikeNeilson: Fair. we can revisit later.

MikeNeilson (approved):

PR #543, opened 2024-02-14
add gradle run task to the "run" group

adamkorynta:
this prevents it from getting added to the "other" group and matches the gradle Application plugin


rma-rripken (approved):

MikeNeilson (approved):
build failure is from not including the recent tomcat.

PR #545, opened 2024-02-19
fix incorrect method call to get interval minutes for full tsid

adamkorynta:
the other method is for getting the minutes based on just the interval string


adamkorynta:
@devinrei here is the bug fix for the error you were seeing

rma-rripken (approved):

PR #551, opened 2024-02-23
Versioned ts

dreina-gei:
Adding support for versioned time series get endpoint


cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
```diff
@@ -105,6 +105,8 @@ public final class Controllers {
     public static final String CATEGORY_ID = "category-id";
     public static final String EXAMPLE_DATE = "2021-06-10T13:00:00-0700[PST8PDT]";
     public static final String VERSION_DATE = "version-date";
+
+    public static final String VERSION_TYPE = "version-type";
```
adamkorynta: Wondering if "date-version-type" is better as to not confuse with other usages of the name "version" like the freeform strings in a time series id.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -411,6 +378,24 @@ public void getAll(@NotNull Context ctx) {
...
+            String versionType = queryParamAsClass(ctx,
+                    new String[]{VERSION_TYPE, },
+                    String.class, VersionType.MAX_AGGREGATE.getValue(), metrics,
+                    name(TimeSeriesController.class.getName(), GET_ONE));
+
+            // Throw error if max aggregate version type is requested with a version date
+            // or single version version type is requested without a version date
+            if( (Objects.equals(versionType, "MAX_AGGREGATE") && !Objects.equals(versionDate, null))
+                || (Objects.equals(versionType, "SINGLE_VERSION") && Objects.equals(versionDate, null)) ) {
+                throw new IllegalArgumentException();
```
MikeNeilson: Needs explinatory text. Additionally the exception handling in ApiServlet.java, and make sure it's going to map properly to one of the error codes. For example this would be 400 BAD REQUEST ... and then why.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -1,46 +1,7 @@
 package cwms.cda.api;
 
 import static com.codahale.metrics.MetricRegistry.name;
-
-import static cwms.cda.api.Controllers.ACCEPT;
```
MikeNeilson, line 5: Please review the checkstyle.xml file, given the number of elements it would be reasonable to add Controllers as an exception for `import <pkg>.*;` but otherwise we don't do that.

MikeNeilson (commented):
Looks like it's definitely getting there.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -118,19 +121,20 @@ public class TimeSeries extends CwmsDTOPaginated {
...
     }
 
-    public TimeSeries(String page, int pageSize, Integer total, String name, String officeId, ZonedDateTime begin, ZonedDateTime end, String units, Duration interval, VerticalDatumInfo info){
-        this(page, pageSize, total, name, officeId, begin, end, units, interval, info, null, null);
+    public TimeSeries(String page, int pageSize, Integer total, String name, String officeId, ZonedDateTime begin, ZonedDateTime end, String units, Duration interval, VerticalDatumInfo info, Timestamp versionDate){
+        this(page, pageSize, total, name, officeId, begin, end, versionDate, units, interval, info, null, null);
     }
 
-    public TimeSeries(String page, int pageSize, Integer total, String name, String officeId, ZonedDateTime begin, ZonedDateTime end, String units, Duration interval, VerticalDatumInfo info, Long intervalOffset, String timeZone) {
+    public TimeSeries(String page, int pageSize, Integer total, String name, String officeId, ZonedDateTime begin, ZonedDateTime end, Timestamp versionDate, String units, Duration interval, VerticalDatumInfo info, Long intervalOffset, String timeZone) {
```
rma-rripken: If begin and end are ZonedDateTime I'd keep versionDate as ZonedDateTime too.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -56,6 +56,9 @@ public class TimeSeries extends CwmsDTOPaginated {
     @Schema(description = "The units of the time series data",required = true)
     String units;
 
+    @Schema(description = "The version date of the time series trace")
+    Timestamp versionDate;
```
adamkorynta: The version type should also be included given `versionDate` is null for both non-versioned and aggregate use-cases.

cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
dreina-gei: Done

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: By explanatory text do you mean in the error message? I added a comment but I'm assuming you want a message that explains the relationship between versionDate and versionType

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
dreina-gei: Done

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: Done

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
MikeNeilson: Yes, the exception/response mapping system use the message for either the log or direct to user as appropriate to a given error so this should have something appropriate for the user to let them know what they did wrong.

cwms-data-api/build.gradle
```diff
@@ -222,6 +222,7 @@ task run(type: JavaExec) {
     dependsOn generateConfig
```
adamkorynta: Don't push changes to this file

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
adamkorynta: still have some `java.util` in there

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -641,7 +641,7 @@ private static TimeSeries buildTimeSeries(ILocationLevelRef levelRef, Interval i
         TimeSeries timeSeries = new TimeSeries(null, size, size, timeSeriesId,
                 officeId, firstValueTime, lastValueTime, levelUnits,
                 java.time.Duration.ofSeconds(interval.getSeconds()),
-                null, null, locationTimeZone.getId());
+                null, null, locationTimeZone.getId(), null, null);
```
adamkorynta: I'm not a fan of null enums. I think UNVERSOINED is appropriate here

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -214,11 +217,21 @@ protected TimeSeries getTimeseries(String page, int pageSize, String names, Stri
...
         // internal table from it so we can manipulate it further
         // This code assumes the database timezone is in UTC (per Oracle recommendation)
-        SQL retrieveSelectData = DSL.sql(
-            "table(cwms_20.cwms_ts.retrieve_ts_out_tab(?,?,cwms_20.cwms_util.to_timestamp(?),cwms_20.cwms_util.to_timestamp(?),"
-           +"'UTC',?,?,?,?,?,?,?,?) ) retrieveTs",
-            tsId,unit, beginTime.toInstant().toEpochMilli(), endTime.toInstant().toEpochMilli(),
-            null,null,null,null,null,null,null,officeId);
+        SQL retrieveSelectData = null;
+        // Query based on versionDate or query max aggregate
+        if(versionDate != null) {
```
adamkorynta, line 243: cwms_util.to_timestamp doesn't handle nulls so that's why you have this if/else?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -301,7 +331,8 @@ protected TimeSeries getTimeseries(String page, int pageSize, String names, Stri
                             tsMetadata.getValue("interval", Long.class)),
                     verticalDatumInfo,
                     tsMetadata.getValue(AV_CWMS_TS_ID2.INTERVAL_UTC_OFFSET).longValue(),
-                    tsMetadata.getValue(tzName)
+                    tsMetadata.getValue(tzName),
+                    versionDate, dateVersionType
```
adamkorynta: Should AGGREGATE get replaced with UNVERSIONED here if the requested time series isn't versioned?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -880,9 +912,13 @@ private void store(Connection connection, String officeId, String tsId, String u
                 qualityArray[i] = value.getQualityCode();
             }
         }
-
         tsDao.store(connection, officeId, tsId, units, timeArray, valueArray, qualityArray, count,
                 storeRule.getRule(), overrideProtection, versionDate, createAsLrts);
+
+        if(versionDate != null) {
+            CWMS_TS_PACKAGE.call_SET_TSID_VERSIONED(getDslContext(connection,officeId).configuration(),
```
adamkorynta: This call should be above the store, but we don't want to blow up if this throws an error that the tsid couldn't be found as we might be creating it

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -252,12 +266,12 @@ else if(format.equals("xml")) {
...
         schema = @Schema(
             name = "TimeSeries.Record",
-            description = "A representation of a time-series record",
+            description = "A representation of a time-series record in the form [dateTime, value, qualityCode]",
             type="array"
         ),
         arraySchema = @Schema(
             type="array",
-            example = "[123,54.3,0]"
+            example = "[123, 54.3, 0]"
```
adamkorynta: Can you change the `123` timestamp to a timestamp within the last decade?

cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
```diff
@@ -148,7 +148,7 @@ public static void register_users() throws Exception {
                         stmt.setString(3,user.getApikey());
                         stmt.execute();
                     } catch (SQLException ex) {
-                        throw new RuntimeException("Unable to register user",ex);
+                        //throw new RuntimeException("Unable to register user",ex);
```
adamkorynta: don't submit this change

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
dreina-gei: Done

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: Done

cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
dreina-gei: Done

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
dreina-gei: It's for the versionDate.toInstant().toEpochMilli()

adamkorynta: you can use a null Long class instead. Having two different conditions makes that code harder to read and maintain. Also extract variables for each of the nulls for better readability.

.gitignore
```diff
@@ -9,4 +9,5 @@ gradle.properties
 .idea
 kdata
 pki/certs
-compose.env
\ No newline at end of file
+compose.env
+cwms-data-api/build.gradle
```
adamkorynta: This shouldn't be in the .gitignore

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -881,8 +914,18 @@ private void store(Connection connection, String officeId, String tsId, String u
             }
         }
 
+        if(versionDate != null) {
+            try {
+                CWMS_TS_PACKAGE.call_SET_TSID_VERSIONED(getDslContext(connection, officeId).configuration(),
+                        tsId, "T", officeId);
+            } catch(DataAccessException e) {
```
adamkorynta, line 965: Log the exception at a low level (FINER/FINEST). Also this should trap only the exact SQL error code that is thrown if the TSID doesn't exist

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
dreina-gei: It could. If the user passes in null for the versionDate I would have to make a preliminary SQL query to check if the timeseries is versioned right?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
adamkorynta: Couldn't find an easy way to do this with an inline query like this. Added null handling on to_timestamp within the schema: https://bitbucket.hecdev.net/projects/CWMS/repos/cwms_database/pull-requests/324/overview

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
dreina-gei: Unable to perform this update without another db call, waiting on approval to do so

cwms-data-api/build.gradle
adamkorynta, line 1: Can you revert the whitespace changes here?

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -411,6 +411,26 @@ public void getAll(@NotNull Context ctx) {
             String begin = ctx.queryParam(BEGIN);
             String end = ctx.queryParam(END);
             String timezone = ctx.queryParamAsClass(TIMEZONE, String.class).getOrDefault("UTC");
+            String versionDateParam = ctx.queryParam(VERSION_DATE);
+            ZonedDateTime versionDate = null;
+            if (versionDateParam != null) {
+                versionDate = DateUtils.parseUserDate(versionDateParam, timezone);
+            }
+
+            String versionType = queryParamAsClass(ctx,
```
adamkorynta: Map to the enum

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -411,6 +411,26 @@ public void getAll(@NotNull Context ctx) {
...
+                    new String[]{VERSION_TYPE, },
+                    String.class, VersionType.MAX_AGGREGATE.getValue(), metrics,
+                    name(TimeSeriesController.class.getName(), GET_ONE));
+
+            // Throw error if max aggregate version type is requested with a version date
+            // or single version version type is requested without a version date
+            if(Objects.equals(versionType, "MAX_AGGREGATE") && !Objects.equals(versionDate, null)) {
```
adamkorynta: `versionDate != null` is easier to read than `!Objects.equals(versionDate, null)`. Irrelevant when switching to versionType to an enum though

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -411,6 +411,26 @@ public void getAll(@NotNull Context ctx) {
...
+                throw new IllegalArgumentException("Cannot query a max aggregate with a version date.");
+            }
+            if(Objects.equals(versionType, "SINGLE_VERSION") && Objects.equals(versionDate, null)) {
```
adamkorynta: Note to self: There is test coverage of these cases?

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -56,6 +53,14 @@ public class TimeSeries extends CwmsDTOPaginated {
     @Schema(description = "The units of the time series data",required = true)
     String units;
 
+    @Schema(description = "The type of time series query")
+    String dateVersionType;
```
adamkorynta: Use the enum here too instead of the string.

adamkorynta: That schema description needs more information too

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -43,6 +39,7 @@
 @XmlAccessorOrder(XmlAccessOrder.ALPHABETICAL)
 @JsonPropertyOrder(alphabetic = true)
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
```
adamkorynta: What unknown property is getting deserialized?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -881,8 +936,24 @@ private void store(Connection connection, String officeId, String tsId, String u
...
+                SQLException cause = (SQLException)e.getCause();
```
adamkorynta: Need to `instasnceof` check to make sure the cause is actually a `SQLException`

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
adamkorynta: You can call `cwms_ts.is_tsid_versioned`

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,32 +1,709 @@
...
+import static org.hamcrest.Matchers.*;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 @Tag("integration")
 public class TimeseriesControllerTestIT extends DataApiTestIT {
 
+    private static final String OFFICE = "SPK";
+    private static final String locationId = "TsVersionedTestLoc";
+    private static final String tsId = locationId + ".Flow.Inst.1Hour.0.raw";
+    public static final String BEGIN_STR = "2008-05-01T15:00:00Z";
```
adamkorynta: Why public scoped?

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,32 +1,709 @@
...
+    public static final String END_STR = "2008-05-01T17:00:00Z";
+
+    @BeforeAll
+    public static void create() throws Exception {
+        createLocation(locationId, true, OFFICE);
+        createTimeseries(OFFICE, tsId, 0);
+    }
+
+    @Test
+    void test_create_get() throws IOException, java.io.IOException {
```
adamkorynta: method name should specify this is testing for date versioned ts. Also you can just throw `Exception` from tests. The two `IOException`'s are confusing

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,32 +1,709 @@
...
+            .log().ifValidationFails(LogDetail.ALL,true)
+            .accept(Formats.JSONV2)
+            .contentType(Formats.JSONV2)
+            .header("Authorization", user.toHeaderValue())
+            .queryParam("office", OFFICE)
+            .queryParam("name", tsId)
+            .queryParam("begin", BEGIN_STR)
+            .queryParam("end", END_STR)
+            .queryParam("date-version-type", "SINGLE_VERSION")
+            .queryParam("version-date", "2021-06-21T08:00:00-0000[UTC]")
```
adamkorynta: Can you pull this from the deserialized `TimeSeries` object? That way it's only written once.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,32 +1,709 @@
...
+        .when()
+            .redirects().follow(true)
+            .redirects().max(3)
+            .get("/timeseries/")
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+            .assertThat()
+            .statusCode(is(HttpServletResponse.SC_OK))
+            .body("values.size()", equalTo(3))
+            .body("values[0][1]", equalTo(1.0F))
```
adamkorynta: The assertions should check to make sure your new TimeSeries fields are set.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,32 +1,709 @@
...
+        given()
+            .log().ifValidationFails(LogDetail.ALL,true)
+            .accept(Formats.JSONV2)
+            .contentType(Formats.JSONV2)
+            .header("Authorization", user.toHeaderValue())
+            .queryParam("office", OFFICE)
+            .queryParam("name", tsId)
+            .queryParam("begin", BEGIN_STR)
+            .queryParam("end", END_STR)
+            .queryParam("date-version-type", "SINGLE_VERSION")
```
adamkorynta: Use enum here

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,32 +1,709 @@
...
+            .log().ifValidationFails(LogDetail.ALL,true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_OK))
+            .body("values.size()", equalTo(3))
+            .body("values[0][1]", equalTo(1.0F))
+            .body("values[1][1]", equalTo(1.0F))
+            .body("values[2][1]", equalTo(1.0F));
+
+        // Delete all values from timeseries
+        given()
```
adamkorynta: Might as well combine with test up above

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,32 +1,709 @@
...
+            .log().ifValidationFails(LogDetail.ALL,true)
+            .accept(Formats.JSONV2)
+            .contentType(Formats.JSONV2)
+            .header("Authorization", user.toHeaderValue())
+            .queryParam("office", OFFICE)
+            .queryParam("name", tsId)
+            .queryParam("begin", BEGIN_STR)
+            .queryParam("end", END_STR)
+            .queryParam("date-version-type", "MAX_AGGREGATE")
```
adamkorynta: add comment above that these are the two fields in contention

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,32 +1,709 @@
...
+        given()
+            .log().ifValidationFails(LogDetail.ALL,true)
+            .accept(Formats.JSONV2)
+            .contentType(Formats.JSONV2)
+            .header("Authorization", user.toHeaderValue())
+            .queryParam("office", OFFICE)
+            .queryParam("name", tsId)
+            .queryParam("begin", BEGIN_STR)
+            .queryParam("end", END_STR)
+            .queryParam("date-version-type", "SINGLE_VERSION")
```
adamkorynta: add comment above that this field is in contention with not having a version-date

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,32 +1,709 @@
...
+            .statusCode(is(HttpServletResponse.SC_OK))
+            .body("values.size()", equalTo(4))
+            .body("values[0][1]", equalTo(4.0F))
+            .body("values[1][1]", equalTo(4.0F))
+            .body("values[2][1]", equalTo(4.0F))
+            .body("values[3][1]", equalTo(3.0F));
+    }
+
+
+    /*@Test
```
adamkorynta: Uncomment out the existing tests

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
rma-rripken: This was on my list to follow up about too

dreina-gei: This was for versionDateType but I think this can get removed now since my implementation changed. Will remove on next commit

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: yes

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -214,12 +218,39 @@ protected TimeSeries getTimeseries(String page, int pageSize, String names, Stri
...
+
+        if(versionDate != null) {
+            versionDateMilli = versionDate.toInstant().toEpochMilli();
+        } else {
+            maxVersion = "T";
+        }
+
+        // Query based on versionDate or query max aggregate
+        if(versionDate != null) {
+            retrieveSelectData = DSL.sql(
```
MikeNeilson, line 244: this should have a comment that to_timestamp will alllow null in next schema release. looked identical on first read.

cwms-data-api/src/test/java/fixtures/TestAccounts.java
```diff
@@ -8,7 +8,8 @@ public class TestAccounts {
     public enum KeyUser {
         NONE("none",null,null,null,null), // Used for annotations
         GUEST("guest",null,null,null,null), // USED as marker label for tests
-        SPK_NORMAL("l2hectest","l2hectest","1234567890","l2userkey","ATotallyRandomString","CWMS Users", "cac_user"),
+        SPK_NORMAL("l2hectest","l2hectest","1234567890","byp","ATotallyRandomString","CWMS Users", "cac_user"),
```
MikeNeilson: why was this changed?

MikeNeilson: (the adding is fine but we shouldn't existing values without reason)

dreina-gei: I only have access to swt so I needed to create an swt user for my tests

MikeNeilson: Okay, so that's not how these tests are supposed to work. Certainly adding a user is just fine, we want tests that cover multiple users. But you'll need to expand the permissions (newest develop branch should do this automatically now) to allow the test system to create and permit any user it needs.

These automated tests are intentionally allowed to assume total control of the database so the database should never be thought of as an "<office> database" just a database.

Also the specific question, which I'll grant I didn't word very well, was why the key for the l2hectest user was changed?

dreina-gei: Ok good to know on the permissions. 

I'm confused on the second part, my db, which is separate from the docker container since my IT team won't allow docker, doesn't have any l2 users so I needed to create an m5 user that is present in my db.

For the key question, I didn't even notice it was changed. I most likely changed it when trying to get the spk office user to work and forgot to change it back. Will add that to my next pr.

MikeNeilson: Okay, that happens. please revert it back.

Not having docker doesn't affect those permissions. You can just add users to the database; they don't need to be oracle users.

For clarity, are you attempting to do development work in a shared database or does you're IT team at least give you one per dev?

dreina-gei: Ok I can add the user to my db. I tested everything with swt but changed the tests back to spk to maintain uniformity throughout the tests. The db is one that RMA made for me, I know that Adam and Ryan have access to it as they've helped me debug.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Done

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Done

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Done

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Done

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Done

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Done

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Done. I didn't use a Timseries object because it required a lot of parameters I didn't have values for so I created a paramMap object I could use for the different requests

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Done

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Changed to private

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
dreina-gei: Done

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
dreina-gei: Done

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
dreina-gei: Removed

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
dreina-gei: both have been updated

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: done

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: done

cwms-data-api/build.gradle
dreina-gei: done

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
dreina-gei: done

cwms-data-api/src/test/java/fixtures/TestAccounts.java
dreina-gei: api key reverted

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
dreina-gei: done

cwms-data-api/build.gradle
dreina-gei: changes reverted

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: done

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
dreina-gei: done

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
dreina-gei: done

.gitignore
dreina-gei: reverted

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
dreina-gei: done

cwms-data-api/src/test/java/fixtures/TestAccounts.java
rma-rripken: Devin and Zac have their own separate databases that can get trashed and recreated whenever it is convenient for them to do so.  But they do persist outside the lifetime of the test-container

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -103,6 +105,10 @@
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Unmarshaller;
+import javax.xml.ws.Response;
+
+import jdk.nashorn.internal.runtime.Version;
```
adamkorynta: Make sure all the unused dependencies are stripped out, I don't expect the `Java.xml.ws` or `jdk.nashorn` packages to get added here

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -411,6 +411,24 @@ public void getAll(@NotNull Context ctx) {
             String begin = ctx.queryParam(BEGIN);
             String end = ctx.queryParam(END);
             String timezone = ctx.queryParamAsClass(TIMEZONE, String.class).getOrDefault("UTC");
+            String versionDateParam = ctx.queryParam(VERSION_DATE);
+            ZonedDateTime versionDate = null;
+            if (versionDateParam != null) {
+                versionDate = DateUtils.parseUserDate(versionDateParam, timezone);
+            }
+
+            String versionTypeParam = ctx.queryParam(VERSION_TYPE);
```
adamkorynta: Why not use `ctx.ctx.queryParamAsClass()`?

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -411,6 +411,24 @@ public void getAll(@NotNull Context ctx) {
...
+            VersionType versionType = VersionType.versionTypeFor(versionTypeParam);
+
+            // Throw error if max aggregate version type is requested with a version date
+            // or single version version type is requested without a version date
+            if(Objects.equals(versionType, VersionType.MAX_AGGREGATE) && versionDate != null) {
```
adamkorynta: no need for `Objects.equals`, enums can be `==` and that's easier to read

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: done

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: done

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
adamkorynta: Your file `num_ts_create.json` is a `TimeSeries` object, you just have to deserialize it

dreina-gei: Okay I have other parameters that change depending on the call. I could use the Timeseries object for some of the query params but others like end that change from query to query would have to be put in this way or I would need to create a setter for it

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
dreina-gei: done

adamkorynta (dismissed):

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -1,709 +1,28 @@
...
 import static io.restassured.RestAssured.given;
+import static io.restassured.config.JsonConfig.jsonConfig;
 import static org.hamcrest.Matchers.*;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 @Tag("integration")
 public class TimeseriesControllerTestIT extends DataApiTestIT {
-
-    private static final String OFFICE = "SPK";
-    private static final String locationId = "TsVersionedTestLoc";
```
MikeNeilson: Why did all this end up deleted?

MikeNeilson: Just saw the VersionedTimeseriesControllerIT.

Hard to tell from the github GUI, please confirm this view is just everything getting reset back and the new tests are in VersionedTimeseries...IT.

cwms-data-api/src/test/java/cwms/cda/api/VersionedTimeseriesControllerTestIT.java
```diff
@@ -0,0 +1,664 @@
...
+public class VersionedTimeseriesControllerTestIT extends DataApiTestIT {
+
+    private static final String OFFICE = "SWT";
+    private static final String locationId = "TsVersionedTestLoc";
+    private static final String tsId = locationId + ".Flow.Inst.1Hour.0.raw";
+    private static final String BEGIN_STR = "2008-05-01T15:00:00Z";
+    private static final String END_STR = "2008-05-01T17:00:00Z";
+
+    private Map<String, String> createParamMap() {
+        return new HashMap<String, String>()
```
MikeNeilson: Nice trick. The downstream test code looks very clean.

MikeNeilson (changes requested):
Just want a sanity check about the existing TimeSeriesControllerIT. Looks like a lot got deleted but it's probably just the github interface.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: Yes I moved all of my tests to VersionedTimeSeriesIT to separate versioned ts tests from the existing tests

cwms-data-api/src/test/java/cwms/cda/api/VersionedTimeseriesControllerTestIT.java
dreina-gei: Thanks!

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/enums/VersionType.java
```diff
@@ -0,0 +1,48 @@
+package cwms.cda.api.enums;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import jdk.nashorn.internal.runtime.Version;
```
rma-rripken: This is a weird import.

cwms-data-api/src/test/java/cwms/cda/api/VersionedTimeseriesControllerTestIT.java
rma-rripken: I agree it looks clean but I think I prefer seeing all the params in each test.  I don't want to have to spend the mental energy to find where the map is created and what else may have changed it between creation and use.

rma-rripken: I should add that I don't care strongly enough to have you change the pr.  Just pointing out opinions.

cwms-data-api/src/test/java/cwms/cda/api/VersionedTimeseriesControllerTestIT.java
```diff
@@ -0,0 +1,664 @@
...
+                put("office", OFFICE);
+                put("name", tsId);
+                put("begin", BEGIN_STR);
+                put("end", END_STR);
+                put("version-date", "2021-06-21T08:00:00-0000[UTC]");
+            }
+        };
+    }
+
+    private Map<String, String> replaceEndParam(Map<String, String> paramMap, String endDate) {
```
rma-rripken: You shouldn't have to "remove" the old key.  Calling put with the key and newValue will overwrite the old value.  
The method is a little weird - it boils down to just one line so its not shorter to call this than to just put that line of code.  There are other places in the test where you directly modify paramMap so I'd suggest removing this method and just inlining the paramMap.put("end", endDate) call

cwms-data-api/src/test/java/cwms/cda/data/dto/TimeSeriesTest.java
rma-rripken, line 1: Can you add a test or two here that set non-null version-date and date-version-type on a TimeSeries and then serialize/deserialize them?   Maybe also a test that read in a json resource file with an example of the expected format - sort of like you did for the IT but with example of date-version-type b/c I don't see that field in the IT resources.

cwms-data-api/src/main/java/cwms/cda/api/enums/VersionType.java
```diff
@@ -0,0 +1,48 @@
...
+@Schema(
+        name = "Version Type",
+        description = "Version type specifies the type of timeseries response to be received. Can be max aggregate "
+                + "or single version. Max aggregate cannot be ran if version date field is specified."
+)
+public enum VersionType {
+    MAX_AGGREGATE("MAX_AGGREGATE"),
+    SINGLE_VERSION("SINGLE_VERSION"),
+    UNVERSIONED("UNVERSIONED"),
+    UNDEF("UNDEFINED");
```
rma-rripken, line 14: If you rename UNDEF to UNDEFINED you won't need the String ctor argument or versionType field.  or the value() or getValue() methods.

cwms-data-api/src/main/java/cwms/cda/api/enums/VersionType.java
```diff
@@ -0,0 +1,48 @@
...
+import io.swagger.v3.oas.annotations.media.Schema;
+import jdk.nashorn.internal.runtime.Version;
+
+import javax.xml.bind.annotation.adapters.XmlAdapter;
+import java.time.ZonedDateTime;
+
+@Schema(
+        name = "Version Type",
+        description = "Version type specifies the type of timeseries response to be received. Can be max aggregate "
+                + "or single version. Max aggregate cannot be ran if version date field is specified."
```
rma-rripken: "cannot be ran" doesn't sound right.

MikeNeilson: Reddit, at least, agrees: https://www.reddit.com/r/grammar/comments/aefv2u/can_be_ran_vs_can_be_run/

cwms-data-api/src/main/java/cwms/cda/api/enums/VersionType.java
dreina-gei: done

cwms-data-api/src/test/java/cwms/cda/api/VersionedTimeseriesControllerTestIT.java
dreina-gei: done

cwms-data-api/src/main/java/cwms/cda/api/enums/VersionType.java
rma-rripken: I can do this in a followup pr.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
dreina-gei: done

cwms-data-api/src/main/java/cwms/cda/api/enums/VersionType.java
dreina-gei: done

rma-rripken (dismissed):

rma-rripken (approved):

rma-rripken:
I confirmed with Devin before merging.

PR #554, opened 2024-02-27
add healthcheck to integration test tomcat startup

adamkorynta:
sometimes integration tests are executing faster than the war is deployed, resulting in a 404 exception

### DIFF
--- a/cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
+++ b/cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
@@ -26,6 +26,11 @@ import io.restassured.RestAssured;
 import io.restassured.config.JsonConfig;
 import io.restassured.path.json.config.JsonPathConfig;
 
+import javax.servlet.http.HttpServletResponse;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.is;
+
 
 @SuppressWarnings("rawtypes")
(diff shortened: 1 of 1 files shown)


adamkorynta:
@devinrei

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
```diff
@@ -87,6 +89,30 @@ public void beforeAll(ExtensionContext context) throws Exception {
...
+    }
+
+    private static void healthCheck() throws InterruptedException {
+        int attempts = 0;
+        int maxAttempts = 15;
+        for (; attempts < maxAttempts; attempts++) {
+            try {
+                given()
+                .when()
+                    .get("/offices/")
```
MikeNeilson: if you put a specific office at the end of this, like `/offices/SPK` you can check for a 200 OK and the response should also be slightly faster.

MikeNeilson (changes requested):
Looks good. Just a minor suggest. (what I'm suggesting is actually my general goto to see if a given instance is behaving correctly as it requires a database connection.)

rma-rripken (dismissed):

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
adamkorynta: great point, updated

MikeNeilson (approved):

PR #560, opened 2024-03-04
Feature/improve required param support

rma-rripken:
Adding a new exception and some Controllers helpers for required query parameters.  If they aren't found the helpers will throw the new exception and ApiServlet message will make clear to user that a required query parameter was missing.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
@@ -27,10 +27,15 @@ package cwms.cda.api;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import cwms.cda.api.errors.RequiredQueryParameterException;
 import cwms.cda.data.dao.JooqDao;
 import cwms.cda.data.dao.texttimeseries.TimeSeriesTextMode;
+import cwms.cda.helpers.DateUtils;
 import io.javalin.core.validation.JavalinValidation;
 import io.javalin.core.validation.Validator;
+import io.javalin.http.Context;
+import java.time.ZonedDateTime;
+import org.jetbrains.annotations.Nullable;
 
 import static com.codahale.metrics.MetricRegistry.name;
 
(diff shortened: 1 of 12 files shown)


cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
```diff
@@ -260,4 +265,55 @@ public static JooqDao.DeleteMethod getDeleteMethod(String input) {
...
     }
+
+    /**
+     * Returns the first matching query param or throws RequiredQueryParameterException.
+     * @param ctx Request Context
+     * @param name Query parameter name
+     * @return value of the parameter
+     * @throws RequiredQueryParameterException if the parameter is not found
+     */
+    public static String requiredParam(io.javalin.http.Context ctx, String name) {
```
MikeNeilson, line 276: We don't need to this now but might be good if this allowed the caller (not client) to set a more detail message than just the name.

If you agree, open an issue and we'll get to it later.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -272,21 +275,20 @@ protected TimeSeriesDao getTimeSeriesDao(DSLContext dsl) {
     @Override
     public void delete(Context ctx, @NotNull String timeseries) {
 
-        String office = ctx.queryParam(OFFICE);
+        String office = requiredParam(ctx, OFFICE);
```
MikeNeilson, line 278: I don't think this one is required as the office of each timeseries is embedded in the body's JSON or XML.

MikeNeilson: Then again maybe it should be. Something to think about.

MikeNeilson (approved):

cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
rma-rripken: That is a good idea. #561

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
rma-rripken: The delete doesn't take a body but I agree we need to look and think about which params are required and where.  I started to go thru all the params marked required and not all of them actually are (some of the offices we mark as required but then the dao treat null as *).  Its also possible that some of the parameters are required in v1 api but optional in v2.  Not sure exactly how to handle those cases.  I guess we just document them when we find them.

rma-rripken: #562

PR #564, opened 2024-03-05
Feature/558 text ts changes

rma-rripken:


cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/RegularTextTimeSeriesRow.java
```diff
@@ -51,26 +39,23 @@ public String getTextValue() {
...
+
+    public Instant getDataEntryDate() {
         return dataEntryDate;
     }
 
     @JsonIgnore
-    @Override
     public DateDateKey getDateDateKey() {
-        return new DateDateKey(dateTime, dataEntryDate);
+        return new DateDateKey(dateTime == null? null: Date.from(dateTime), dataEntryDate==null ? null: Date.from(dataEntryDate));
```
MikeNeilson: would it be better to have a second constructor that handles this? I think I saw this exact logic in another spot.

rma-rripken: Possibly.  BinaryTimeSeriesRow is probably the other location. That class also transitioned from Date to Instant but kept DateDateKey.  Will revisit this before taking the pr out of draft.  Maybe DateDateKey goes away entirely.

MikeNeilson (commented):
Marking place for later review.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/RegularTextTimeSeriesRow.java
rma-rripken: Turns out this is using DateDateKey from Monolith.  Adding a new constructor isn't worth the effort (in my opinion) so I added a build method.   I don't want to do it right now but I could a follow-up pr for deleting DateDateKey from the text-ts and bin-ts.

MikeNeilson (dismissed):
That looks good.

MikeNeilson (approved):

PR #567, opened 2024-03-07
Bugfix/536 set office

### LINKED ISSUE #536: When are additional setOffice calls needed?
In text-ts development I could not write text-ts unless I added an additional 
setOffice(c, "SPK") call immediately before the write attempt.
Before adding the setOffice call the exception was
Caused by: java.sql.SQLException: ORA-20048: NO_WRITE_PRIVILEGE: User doesnt have write privileges

LocationLevels just hit a similar problem and adding setOffice resolved it.  

#534 

Why are these setOffice calls now needed when they weren't needed before?
Doesn't OfficeSettingConnectionProvider  already set the session office Id for every connection?

Is it that OfficeSettingConnectionProvider is too early ?  If so, it seems like AuthDao should be setting it.

rma-rripken:
Closes #536.
From what I've learned about jOOQ connection management these weren't quite right.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
@@ -45,13 +45,6 @@ public abstract class Dao<T> {
         return cwmsDbVersion;
     }
 
-    /**
-     * This should be called before attempting to write an object to the database.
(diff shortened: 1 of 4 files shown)


MikeNeilson (approved):
Yeah, that drastic change in the 3.14 series that we skipped over made things really confusing.

Issue #556, opened 2024-02-29
Values returning different for Internal vs External CDA instances for a specific Project / Timeseries


krowvin:
**Steps to reproduce the problem**
1.  GET request on `PINE.Elev.Inst.1Hour.0.Ccp-Rev` version 1 or 2 on CDA internal instance

2. GET request on same timeseries on water.usace.army.mil or CWBI PROD
 ```curl -X 'GET' \
  'https://cwms-data.usace.army.mil/cwms-data/timeseries?name=PINE.Elev.Inst.1Hour.0.Ccp-Rev&office=SWT&timezone=US%2FCentral' \
  -H 'accept: application/json;version=2'
  ``` 
  
3. Compare the first value for the same timezone using the default 24 hours of data response against the public endpoints

**What you expected to happen:**
Expect values to match for same request

**What actually happened (or didn't happen)**
~External instance of CDA appears to return data off by 6 hours (TimeZone difference)~

**Internal CDA version is:**
3.1.2 

**Things I tried for diagnosing**
Spot checking other timeseries for this project look correct. 

Checked the same process using version 1 of the API. Same issue.

Confirmed the PINE location metadata is set to (US/Central)

MikeNeilson:
Please paste the exact query/http request in (if using swagger the curl command is the best thing to paste in.)

That said, while the API can take times in different timezone it does, and always will return things in UTC (or at least it should) so given that two different instances have returned different times that appears to be an actual bug you've found.

This may have been fix in some recent PRs as we found some other TZ handling issues. Please put in the exact request to make it easier to diagnose.

krowvin:
Done. Post updated and external curl added.

DanielTOsborne:
I think it might be something else than timezone, and I'm not sure if it's even CDA related.
Here's a query I tested to the public instance: 
https://water.usace.army.mil/cwms-data/timeseries?name=PINE.Elev.Inst.1Hour.0.Ccp-Rev&office=swt&timezone=America/Chicago
Change the query to point to the internal SWT instance (url not posted for obvious reasons), and you get completely different, but still similar data.

![image](https://github.com/USACE/cwms-data-api/assets/73709911/2e5effd3-7dad-446a-9ad7-d27744eac468)

The left side is the public instance data returned, right side is the internal instance.
It _almost_ looks like it's off by one hour, except the data doesn't entirely match, even accounting for floating point rounding errors.

rma-rripken:
I'm at a loss.  
[...]

PR #569, opened 2024-03-07
Slight improvement to ts controller test and doc

rma-rripken:
verifying that we only get one point back and that it has the correct epoch millis.  And trying to better document the timezone parameter

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
@@ -344,7 +344,9 @@ public class TimeSeriesController implements CrudHandler {
                             + DATE_FORMAT + "', e.g., '" + EXAMPLE_DATE + "'."),
                     @OpenApiParam(name = TIMEZONE,  description = "Specifies "
                             + "the time zone of the values of the begin and end fields (unless "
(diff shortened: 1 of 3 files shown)


rma-rripken:
Related to #556

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -344,7 +344,9 @@ public void delete(@NotNull Context ctx, @NotNull String timeseries) {
                             + DATE_FORMAT + "', e.g., '" + EXAMPLE_DATE + "'."),
                     @OpenApiParam(name = TIMEZONE,  description = "Specifies "
                             + "the time zone of the values of the begin and end fields (unless "
-                            + "otherwise specified), as well as the time zone of any times in the"
+                            + "otherwise specified).  "
+                            + "For "+Formats.JSONV2 + " amd " + Formats.XMLV2 + " the results are returned in UTC.  "
```
krowvin: and* (not being petty, we want to look good!)

MikeNeilson (changes requested):
Other than @krowvin 's comment about amd vs and looks good to me.

MikeNeilson (approved):

PR #574, opened 2024-03-11
The code was previously building a new LocationLevel for each level v…

rma-rripken:
…alue.

I was looking into the NAB reported issue with East Sidney.Elev.Inst.0.Top of Conservation and I got CDA to generate a buggy json response.  
[response_1710176111087.json](https://github.com/USACE/cwms-data-api/files/14562431/response_1710176111087.json)

I duplicates the level for each level value in a nasty way.

I think these changes will fix it but I don't have a dataset handy to sufficiently test it locally.


rma-rripken:
https://github.com/USACE/cwms-data-api/issues/586

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -268,6 +298,14 @@ public void getAll(Context ctx) {
 
             if ("2".equals(version)) {
 
+                if (unit == null){
+                    // The dao currently only supports SI.
+                    unit = UnitSystem.SI.getValue();
+                }
+                if(!UnitSystem.SI.getValue().equals(unit)) {
+                    throw new IllegalArgumentException("Levels Version 2 currently only supports SI");
```
MikeNeilson, line 302: Doesn't this also support just setting the unit? I swear I corrected some behavior for Ruth to get %, maybe it's a documentation thing though, the levels operate just like the timeseries as the units can be anything not just a length.

rma-rripken: The getAll is querying against a AV_LOCATION_LEVEL and feeding the rows into the older JDom* classes. 
The Dao code for that method is clearly based on:
https://bitbucket.hecdev.net/projects/CWMS/repos/hec-cwms-data-access/browse/hec-db-jdbc/src/main/java/wcds/dbi/oracle/cwms/CwmsLevelJdbcDao.java

It looks to me like your changes were for calls from getOne for a single Level.  But maybe there are other pl/sql procs we should call instead of hitting the view.

MikeNeilson: Well, that structure of this is fine, so we'll go for it for now. But someone is likely going to complain about not being able to set their preferred units.

MikeNeilson (dismissed):

MikeNeilson (approved):

PR #575, opened 2024-03-11
Devops/fix tests add templating to test data

MikeNeilson:
I discovered this morning that the integration tests were broken again. Mostly due to the addition of a new user (m5hectest), I've simply added that user in as one of the valid users.

Additionally since some of these tests loop over different users and each user may have different permissions I've started work on a template system using freemarker to be able to fill out sections of the test data.

The integration test system should always be run regularly and your tests should pass, *regardless* of the "database" you use. CDA doesn't care, it just needs, and when running the tests will give itself, appropriate permissions to do whatever. 

If that is now working and you're stuck please report the current errors for assistance.

Not ready for review, but certainly ready for comments:
@rripken @adamkorynta @devinrei @DanielTOsborne 

I'll be working today to clean up the new duplicate code, that was a quick test to make sure the concept even works. next step is finding an appropriate way to integrate and make it easier to use.


MikeNeilson:
Also made a minor improvement to the wrapException, they were keeping the cause in some contexts.

rma-rripken (approved):

rma-rripken:
We reported that there were IT issues with having users at multiple offices at the 2/22 CDA Dev meeting.

MikeNeilson:
> We reported that there were IT issues with having users at multiple offices at the 2/22 CDA Dev meeting.

I vaguely remember that now.

PR #578, opened 2024-03-13
Forecast Timeseries DTOs

gei-zchapman:
CDA implementation of forecast data objects
[CWMS-1527](https://jira.hecdev.net/browse/CWMS-1527)

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastInstance.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastInstance.java
@@ -1,0 +1,130 @@
+package cwms.cda.data.dto.forecast;
+
+import cwms.cda.api.errors.FieldException;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+import javax.xml.bind.annotation.*;
+import java.time.Instant;
+import java.util.Map;
+
+@XmlRootElement(name = "forecast-instance")
+@XmlAccessorType(XmlAccessType.FIELD)
+public class ForecastInstance {
+
+  // unique identifier: specId + issueDateTime
+
+  @Schema(description = "Forecast Spec ID")
+  @XmlElement(name = "spec-id")
+  private String specId;
+
+  @XmlAttribute(name = "date-time")
+  private Instant dateTime;
+
+  @XmlAttribute(name = "issue-date-time")
+  private Instant issueDateTime;
+
+  @XmlAttribute(name = "first-date-time")
+  private Instant firstDateTime;
+
+  @XmlAttribute(name = "last-date-time")
+  private Instant lastDateTime;
+
+  @XmlAttribute(name = "max-age")
+  private Integer maxAge;
+
+  @XmlAttribute(name = "time-series-count")
+  private Integer timeSeriesCount;
+
+  @Schema(description = "Forecast Instance Notes")
+  @XmlAttribute
+  private String notes;
+
+  @XmlAttribute(name = "metadata")
+  private Map<String, String> metadata;
+
+  @Schema(description = "Forecast Filename")
+  @XmlAttribute(name = "filename")
+  private String filename;
+
+  @Schema(description = "Description of Forecast File")
+  @XmlAttribute
+  private String fileDescription;
+
+  @Schema(description = "Forecast File binary data")
+  @XmlElement(name = "forecast-data")
+  private byte[] fileData;
+
+  @SuppressWarnings("unused") // required so JAXB can initialize and marshal
+  private ForecastInstance() {}
+
+  public ForecastInstance(String specId, Instant dateTime, Instant issueDateTime,
+                          Instant firstDateTime, Instant lastDateTime, Integer maxAge, Integer timeSeriesCount,
+                          String notes, Map<String, String> metadata, String filename, String fileDescription,
+                          byte[] fileData) {
+    this.specId = specId;
+    this.dateTime = dateTime;
+    this.issueDateTime = issueDateTime;
+    this.firstDateTime = firstDateTime;
+    this.lastDateTime = lastDateTime;
+    this.maxAge = maxAge;
+    this.timeSeriesCount = timeSeriesCount;
+    this.notes = notes;
+    this.metadata = metadata;
+    this.filename = filename;
+    this.fileDescription = fileDescription;
+    this.fileData = fileData;
+  }
+
+  public String getSpecId() {
+    return specId;
+  }
+
+  public Instant getDateTime() {
+    return dateTime;
+  }
+
+  public Instant getIssueDateTime() {
+    return issueDateTime;
+  }
+
+  public Instant getFirstDateTime() {
+    return firstDateTime;
+  }
+
+  public Instant getLastDateTime() {
+    return lastDateTime;
+  }
+
+  public Integer getMaxAge() {
+    return maxAge;
+  }
+
+  public Integer getTimeSeriesCount() {
+    return timeSeriesCount;
+  }
+
+  public String getNotes() {
+    return notes;
+  }
+
+  public Map<String, String> getMetadata() {
+    return metadata;
+  }
+
+  public String getFilename() {
+    return filename;
+  }
+
+  public String getFileDescription() {
+    return fileDescription;
+  }
+
+  public byte[] getFileData() {
+    return fileData;
+  }
+
+  public void validate() throws FieldException {
+    //TODO
+  }
+
+}
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
@@ -1,0 +1,88 @@
+package cwms.cda.data.dto.forecast;
+
+import cwms.cda.api.errors.FieldException;
+import cwms.cda.data.dto.TimeSeriesIdentifierDescriptor;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+import javax.xml.bind.annotation.*;
+import java.util.List;
+
+@XmlRootElement(name = "forecast-spec")
+@XmlAccessorType(XmlAccessType.FIELD)
+public class ForecastSpec {
+
+  @Schema(description = "Forecast Spec ID")
+  @XmlElement(name = "spec-id")
+  private String specId;
+
+  @Schema(description = "Office ID that owns the time-series")
+  @XmlElement(name = "office-id")
+  private String officeId;
+
+  @Schema(description = "Location ID")
+  @XmlElement(name = "location-id")
+  private String locationId;
+
+  @Schema(description = "Source Entity ID")
+  @XmlElement(name = "source-entity-id")
+  private String sourceEntityId;
+
+  @Schema(description = "Forecast Designator")
+  @XmlAttribute
+  private String designator;
+
+  @Schema(description = "Description of Forecast")
+  @XmlAttribute
+  private String description;
+
+  @Schema(description = "List of Time Series IDs belonging to this Forecast Spec")
+  @XmlAttribute(name = "time-series-ids")
+  private List<TimeSeriesIdentifierDescriptor> timeSeriesIds;
+
+  @SuppressWarnings("unused") // required so JAXB can initialize and marshal
+  private ForecastSpec() {}
+
+  public ForecastSpec(String specId, String officeId, String locationId, String sourceEntityId, String designator,
+                      String description, List<TimeSeriesIdentifierDescriptor> timeSeriesIds) {
+    this.specId = specId;
+    this.officeId = officeId;
+    this.locationId = locationId;
+    this.sourceEntityId = sourceEntityId;
+    this.designator = designator;
+    this.description = description;
+    this.timeSeriesIds = timeSeriesIds;
+  }
+
+  public String getSpecId() {
+    return specId;
+  }
+
+  public String getOfficeId() {
+    return officeId;
+  }
+
+  public String getLocationId() {
+    return locationId;
+  }
+
+  public String getSourceEntityId() {
+    return sourceEntityId;
+  }
+
+  public String getDesignator() {
+    return designator;
+  }
+
+  public String getDescription() {
+    return description;
+  }
+
+  public List<TimeSeriesIdentifierDescriptor> getTimeSeriesIds() {
+    return timeSeriesIds;
+  }
+
+  public void validate() throws FieldException {
+    //TODO
+  }
+
+}
--- a/cwms-data-api/src/test/java/cwms/cda/data/dto/forecast/ForecastInstanceTest.java
+++ b/cwms-data-api/src/test/java/cwms/cda/data/dto/forecast/ForecastInstanceTest.java
@@ -1,0 +1,102 @@
+package cwms.cda.data.dto.forecast;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import cwms.cda.formatters.json.JsonV2;
+import org.apache.commons.io.IOUtils;
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.time.Instant;
+import java.util.HashMap;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class ForecastInstanceTest {
+
+  @Test
+  void testRoundTripJson() throws JsonProcessingException
+  {
+    ForecastInstance i1 = buildForecastInstance();
+
+    ObjectMapper om = buildObjectMapper();
+
+    String jsonString = om.writeValueAsString(i1);
+    assertNotNull(jsonString);
+
+    ForecastInstance i2 = om.readValue(jsonString, ForecastInstance.class);
+    assertNotNull(i2);
+
+    assertForecastInstanceEquals(i1, i2);
+  }
+
+  @Test
+  void testRoundTripEmptyJson() throws JsonProcessingException
+  {
+    ForecastInstance i1 = buildEmptyForecastInstance();
+
+    ObjectMapper om = buildObjectMapper();
+
+    String jsonString = om.writeValueAsString(i1);
+    assertNotNull(jsonString);
+
+    ForecastInstance i2 = om.readValue(jsonString, ForecastInstance.class);
+    assertNotNull(i2);
+
+    assertForecastInstanceEquals(i1, i2);
+  }
+
+  @Test
+  void testJsonFile() throws IOException {
+    String json;
+    try (InputStream stream = getClass().getResourceAsStream("forecast_instance_test.json")) {
+      assertNotNull(stream);
+      json = IOUtils.toString(stream, StandardCharsets.UTF_8);
+    }
+
+    ObjectMapper om = buildObjectMapper();
+    ForecastInstance fi = om.readValue(json, ForecastInstance.class);
+
+    assertNotNull(fi);
+    assertForecastInstanceEquals(fi, buildForecastInstance());
+  }
+
+  @NotNull
+  private ForecastInstance buildForecastInstance() {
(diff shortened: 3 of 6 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastTimeSeries.java
```diff
@@ -0,0 +1,4 @@
+package cwms.cda.data.dto.forecast;
+
```
MikeNeilson: I suspect this will likely be better named "ForecastTimeSeriesList" but the roughout here looks reasonable.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastTimeSeries.java
```diff
@@ -0,0 +1,18 @@
+package cwms.cda.data.dto.forecast;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement(name = "alias")
+@XmlAccessorType(XmlAccessType.FIELD)
+public class ForecastTimeSeries {
```
adamkorynta: This object needs to reference the unique identifiers in the forecast (not the code) and then compose in the TimeSeries object (not the code)

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
```diff
@@ -0,0 +1,30 @@
+package cwms.cda.data.dto.forecast;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement(name = "alias")
+@XmlAccessorType(XmlAccessType.FIELD)
+public class ForecastSpec {
```
adamkorynta, line 12: Shouldn't reference the codes. There should be an officeId, forecastSpecId, sourceEntityId, locationId

If the location office can differ from the forecastSpec office, we'd need an additional locationOfficeId, but I think that's likely not necessary.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastInstance.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+import java.time.Instant;
+import java.util.Map;
+
+@XmlRootElement(name = "alias")
+@XmlAccessorType(XmlAccessType.FIELD)
+public class ForecastInstance {
+
+  @XmlAttribute
```
adamkorynta: Remove fcstInstCode/fcstSpecCode and replace with forecastSpecId

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastFile.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlRootElement;
+
+@XmlRootElement(name = "alias")
+@XmlAccessorType(XmlAccessType.FIELD)
+public class ForecastFile {
+
+  @XmlAttribute
```
adamkorynta: Should include the unique identifiers for the forecast instant instead of the fcstInstCode.

blobCode should be replaced with the byte[] itself and the blob id.

adamkorynta (changes requested):
General comment: no need to stick to the `fcst` abbreviation, that was only done to avoid namespace clashing with an older `forecast` API. At some point that old namespace will be removed and the new namespace may be renamed.

adamkorynta:
I think it is worth adding the JSON representation of these objects at this point so that we can review the shape (and you can start unit testing the serialization).

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastTimeSeries.java
MikeNeilson: I'll argue against the actual timeseries object, but agree the codes should be hidden we currently treat those as opaque handles that the user never sees. So generally we expose (office, <further identifying element(s)>). Though it would not be unreasonable to encode that into a "SLUG"/"Key" for applications to use and have an easier time referencing things.

For the timeseries, since that a given forecast spec may have upwards of "we don't actually limit the number" timseries gathering, formatting, and transmitting them all will take too long. Instead a list of URLs to retrieve said timeseries should be provided.

NOTE: for write that's not true so we'll actually still need the time series objects here. Unless we also want to force the downstream users to write to the individual timeseries; which, maybe.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
MikeNeilson: No, I think maybe a forecast could include data from different offices (for example if raw data is always under "CWMS" or something that could happen a lot) but a forecast itself would I think also be per office.

... wait, oh dear, does a need to always have data from one location? wouldn't the CAVI forecasts be multiple locations?

MikeNeilson: I've got some questions out and another comment about in on the schema cwms_fsct PR.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastFile.java
MikeNeilson: Actually I think this either goes away or becomes a generic representation the "binary_file_t" I'm working on for the text timeseries. But in that case it would be a member of the Instance.

adamkorynta: oh right, we revised this to be a single file rather than an unlimited set of files (zip 'em if needed)

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
adamkorynta: yeah, Peter and I weren't sure about how the location code would actually be used. I presume all location's defined in the time series id's would be relevant.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastTimeSeries.java
adamkorynta: I think that muddies up the API necessarily having read/write being such drastically different shapes. I'd argue that the forecast spec (change from forecast instance) would have the list of time series ids. Since we've moved that list away from the instance, a URL is no longer helpful as we don't have the start/end/versionDate to pre-populate the URL parameters with. 

If we think there isn't going to be any extra metadata in the ForecastTimeSeries object, then we should just remove this DTO entirely as the current metadata is only needed on storage and can be set to just query parameters. 

i.e. /forecast/forecast-ts?forecast-spec-id=blah&forecast-date=now would accept and return TimeSeries objects only (if we don't add time window parameters it would default to the forecast instance start/end)

MikeNeilson: True, that's not a good mismatch to have. Just returning the spec/instance and letting that provide the appropriate information seems reasonable. I don't think it unreasonable to add additional support to the time series end point to accept a list of time series, but at the same time and application could just chose to push in parallel.

gei-zchapman: I agree that this DTO is most likely unneeded, I'll remove it for now.

If the Time Series ID list is moving from the Forecast Instance to the Forecast Spec, does that mean Forecast Spec will not have a list of Time Series IDs in itself?

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastFile.java
```diff
@@ -1,23 +1,26 @@
...
-  @XmlAttribute
-  private Integer fcstInstCode;
+  @Schema(description = "Forecast Instance ID")
+  @XmlElement(name = "instance-id")
+  String instanceId;
 
-  @XmlAttribute
-  private Integer blobCode;
+  @Schema(description = "File's Blob ID")
+  @XmlElement(name = "blob-id")
```
MikeNeilson: We're just embedding the "blob" directly or maybe the "file" objects I'm creating for the text timeseries. not entirely sure but this should at at least be "byte forecastData[]" and not a blob Id reference.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastInstance.java
```diff
@@ -1,47 +1,52 @@
...
-  @XmlAttribute
+  @XmlAttribute(name = "max-age")
   private Integer maxAge;
 
-  @XmlAttribute
+  @XmlAttribute(name = "time-series-count")
   private Integer timeSeriesCount;
 
-  @XmlAttribute
+  @XmlAttribute(name = "file-count")
```
MikeNeilson: There's only one file now.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastInstance.java
```diff
@@ -1,47 +1,52 @@
...
   private Integer fileCount;
 
+  @Schema(description = "Forecast Instance Notes")
   @XmlAttribute
   private String notes;
 
   @XmlAttribute
-  private Map<String, String> fcstInfo;
+  private Map<String, String> tags;
```
MikeNeilson: that's go with "metaData". I've always seen tags and a single keyword no key-value pairs.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastFile.java
MikeNeilson: Actually wait shouldn't the forecast data here be per instance?

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
```diff
@@ -1,30 +1,36 @@
...
+  @XmlElement(name = "source-entity-id")
+  String sourceEntityId;
 
+  @Schema(description = "Description of Forecast")
   @XmlAttribute
   private String description;
 
+  @Schema(description = "List of Time Series IDs belonging to this Forecast Spec")
+  @XmlAttribute(name = "time-series-ids")
+  private List<String> timeSeriesIDs;
```
MikeNeilson: Should we do TimeSeriesIdentifier here over a list of strings? (I'm probably splitting hairs though.)

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastFile.java
gei-zchapman: Right this is linked to the instance, so should the blob and file name/description be embedded in the ForecastInstance DTO?

MikeNeilson: Yeah.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
adamkorynta: I like TimeSeriesIdentifier more given that we can determine LRTS vs PRTS (though that'll be moot soon), but we'd also get interval offset and location time zone.

adamkorynta (changes requested):
Can you add a test that reads a JSON text file from resources? That will provide a nice example and ensure that the roundtrip isn't just validating its own invalid state.

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastInstance.java
```diff
@@ -0,0 +1,137 @@
...
+import java.time.Instant;
+import java.util.Map;
+
+@XmlRootElement(name = "forecast-instance")
+@XmlAccessorType(XmlAccessType.FIELD)
+public class ForecastInstance {
+
+  @Schema(description = "Forecast Instance ID")
+  @XmlElement(name = "instance-id")
+  String instanceId;
```
adamkorynta: There is no id for the instance, the uniqueness constraint  s the spec + issue-date-time

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
```diff
@@ -0,0 +1,78 @@
...
+  @XmlElement(name = "source-entity-id")
+  String sourceEntityId;
+
+  @Schema(description = "Description of Forecast")
+  @XmlAttribute
+  private String description;
+
+  @Schema(description = "List of Time Series IDs belonging to this Forecast Spec")
+  @XmlAttribute(name = "time-series-ids")
+  private List<String> timeSeriesIds;
```
adamkorynta: should be list of time series identifiers. We'll likely need to meet to talk about this

cwms-data-api/src/test/resources/cwms/cda/data/dto/forecast/forecast_spec_test.json
```diff
@@ -0,0 +1,12 @@
+{
+  "spec-id": "spec",
```
adamkorynta, line 2: This is missing the "designator" field

gei-zchapman: That's not a field I have in the DTO. Would that be a string?

cwms-data-api/src/test/java/cwms/cda/data/dto/forecast/ForecastSpecTest.java
```diff
@@ -0,0 +1,97 @@
...
+  private ForecastSpec buildEmptyForecastSpec() {
+    return new ForecastSpec(null, null, null, null, null, null);
+  }
+
+  @NotNull
+  public static ObjectMapper buildObjectMapper() {
+    return JsonV2.buildObjectMapper();
+  }
+
+  void assertForecastSpecEquals(ForecastSpec s1, ForecastSpec s2) {
```
adamkorynta: I'd update this to assert that the generated string is equal, that way you don't have to keep adding to this method every time a field is added/edited/removed

cwms-data-api/src/test/java/cwms/cda/data/dto/forecast/ForecastInstanceTest.java
```diff
@@ -0,0 +1,118 @@
...
+            null, null, null, null, null, null,
+            null, null);
+  }
+
+  @NotNull
+  public static ObjectMapper buildObjectMapper() {
+    return JsonV2.buildObjectMapper();
+  }
+
+  void assertForecastInstanceEquals(ForecastInstance i1, ForecastInstance i2) {
```
adamkorynta: Same comment as on the other way. Easier to assert that the serialized strings are equal.

adamkorynta (changes requested):
just a couple comments

cwms-data-api/src/test/resources/cwms/cda/data/dto/forecast/forecast_spec_test.json
adamkorynta: yeah, that's the comment on the schema PR that needs to be added. It will be used to represent the forecast run

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
adamkorynta: instead of `List<String>` use `TimeSeriesIdentifierDescriptors`

gei-zchapman: Looks like the object mapper readValue function throws an error after this update:

com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of cwms.cda.data.dto.TimeSeriesIdentifierDescriptors$Builder (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)

 writeValueAsString works though.

cwms-data-api/src/test/java/cwms/cda/data/dto/forecast/ForecastSpecTest.java
gei-zchapman: Are the JSON objects guaranteed to have their keys in the same order? Also I modified the whitespace in the test files for readability - edit: the comparison reserializes them so the format of the test json file doesn't matter.

gei-zchapman: Oh by serialized/generated string do you not mean the JSON string?

adamkorynta: yeah I mean the JSON string, I think the order is consistent, though not dependable? if it isn't you can always add
```
```
or
```
@JsonPropertyOrder({ "id", "label", "target", "source", "attributes" })
```

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
adamkorynta: if @rripken is still around maybe he knows a tip? otherwise I can help take a look on Monday

cwms-data-api/src/test/java/cwms/cda/data/dto/forecast/ForecastSpecTest.java
gei-zchapman: Looks like it works without needing to define the order!

gei-zchapman:
Need to figure out why `TimeSeriesIdentifierDescriptors` fails to deserialize with JsonV2, maybe merging latest from master would help, currently I'm 106 commits behind

rma-rripken:
> Need to figure out why `TimeSeriesIdentifierDescriptors` fails to deserialize with JsonV2, maybe merging latest from master would help, currently I'm 106 commits behind

cwms.cda.data.dto.TimeSeriesIdentifierDescriptors probably isn't the class you want  - that class is a class to do pagination for one of the Controllers.   You probably want to use your own List or Collection of cwms.cda.data.dto.TimeSeriesIdentifierDescriptor.
You'll serialize the cda version, heads-up there is a similar class here:  usace.cwms.db.dao.ifc.ts.TimeSeriesIdentifierDescriptor 
Somewhere there is a toDto method if you find yourself needed to convert between the two

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
rma-rripken: I'm guessing its just the wrong class.  Use a List<TimeSeriesIdentifierDescriptor> the TimeSeriesIdentifierDescriptors is for something else.

gei-zchapman: Unit tests are passing after changing to a List!

adamkorynta (approved):

adamkorynta:
I think this is good to squash up commits and merge. Controllers/DAO's can follow in separate PR's. 

The controllers can be added without putting in the API Servlet given the DAO's won't be implemented until the schema is a little more ready.

MikeNeilson:
I'm good with that. It's not like we wrote it in stone either. Well, and even if we did we know what water does to stone over time.

PR #579, opened 2024-03-14
Feature/code cleanup

rma-rripken:
Fixed a number of nagging warnings, errors, checkstyle violations etc.


cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
```diff
@@ -146,19 +147,10 @@ private String getFormatted(ContentType type, CwmsDTOBase toFormat) throws Forma
...
-            outputFormatter = contentFormatters.get(klass);
-        }
-        return outputFormatter;
-    }
-
     private String getFormatted(ContentType type, List<? extends CwmsDTOBase> dtos, Class<?
             extends CwmsDTOBase> rootType) throws FormattingException {
         for (ContentType key : formatters.keySet()) {
-            logger.finest(() -> key.toString());
+            logger.finest(key::toString);
```
MikeNeilson: can we use the is level enabled trick with JUL to just shortcut this loop entirely?

cwms-data-api/src/main/java/cwms/cda/formatters/json/JsonV1.java
```diff
@@ -26,145 +21,153 @@
...
-				   Catalog.class,
-				   LocationGroup.class,
-				   LocationCategory.class,
-				   TimeSeriesCategory.class, TimeSeriesGroup.class,
-				   Clob.class,
-				   Clobs.class,
-				   TimeSeriesGroup.class,
-				   RecentValue.class
-				})
+        dataTypes = {
```
MikeNeilson: Review https://github.com/USACE/cwms-data-api/pull/509 again, it cleared after a made a bunch of changes.
(one of us will have to rebase, I don't care which)

cwms-data-api/src/main/java/cwms/cda/security/OpenIDAccessManager.java
```diff
@@ -87,7 +80,16 @@ private DataApiPrincipal getUserFromToken(Context ctx) throws CwmsAuthException
 
     private String getToken(Context ctx) {
         String header = ctx.header("Authorization");
-        return header.split("\\s+")[1];
+        if (header == null) {
```
MikeNeilson, line 84: technically with the way the manager checking works this can't actually be null, though probably still worth the check just would likely be "illegal state" vs argument, at least for the header null check.

MikeNeilson (changes requested):
Looks good. Just some nitpicky details and a staging question.

cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
rma-rripken: oh, I like the old way better.  I'm going to revert that line.

MikeNeilson: Check if it still exists on master though, did remove a fair amount from this class.

MikeNeilson (approved):

PR #580, opened 2024-03-14
Change LineLength to 120

rma-rripken:
Can we extend the LineLength to 120 (or even just 110)? 

I parsed the checkstyle warnings from a recent build and put them in excel.

LineLength

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -34,7 +34,7 @@
 
     <module name="LineLength">
         <property name="fileExtensions" value="java"/>
-        <property name="max" value="100"/>
+        <property name="max" value="120"/>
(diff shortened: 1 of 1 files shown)


adamkorynta (approved):

adamkorynta:
Arguments for 100: easier to review side-by-side diffs in the browser
Arguments for 120: our modern monitors are wide enough to show this with all the other stuff in IDE's

Arguments against 100: incentivizes abbreviations which makes code harder to read
Arguments against >120: long class and variable names are hard to read, but I think we don't get that at 120

MikeNeilson:
Yeah, I'm good with either, we have a lot of lines that are cleaner  if that aren't broken even at that line length.

MikeNeilson (approved):

Issue #582, opened 2024-03-14
CWBI Test /location/group endpoint returning location group without an id


adamkorynta:
```
curl -X 'GET' \
  'https://cwms-data-test.cwbi.us/cwms-data/location/group?location-category-like=RDL_Findley' \
  -H 'accept: application/json'
```

Response
```
[
  {
    "location-category": {
      "office-id": "MVS",
      "id": "RDL_Findley"
    },
    "assigned-locations": []
  }
]
```

The group-id field should be required and the client does not expect it to be missing.

rma-rripken:
Weird.   The Category is in an office but the group has a null group-id and a null group-office-id.  How do those get created and what are they used for?

adamkorynta:
I am also seeing an issue where I filter the category, but not the office and include assigned locations
```
curl -X 'GET' \
  'https://cwms-data-test.cwbi.us/cwms-data/location/group?include-assigned=true&location-category-like=Agency%20Aliases' \
  -H 'accept: application/json'
```

This returns an error code 500, but no body and just these headers:
```
 access-control-allow-credentials: true 
 access-control-allow-headers: x-requested-with,Content-Type,origin,authorization,accept,client-security-token 
 access-control-allow-methods: POST,GET,OPTIONS,DELETE,PUT 
 access-control-max-age: 600 
 connection: close 
 content-length: 0 
 content-security-policy: frame-src 'self' *.usace.army.mil *.sec.usace.army.mil *.ops.usace.army.mil *.cwbi.us *.rsgis.dev; frame-ancestors 'self' *.usace.army.mil *.sec.usace.army.mil *.ops.usace.army.mil *.cwbi.us *.rsgis.dev 
 content-type: application/json 
 date: Fri,15 Mar 2024 18:12:26 GMT 
 server: USA 
 strict-transport-security: max-age=31536000; includeSubDomains 
 vary: Origin 
 x-content-type-options: nosniff 
 x-frame-options: SAMEORIGIN,SAMEORIGIN 
 x-xss-protection: 1; mode=block 
```

I think it might be related?

rma-rripken:
It might be related.  Hard to say.   I think the already merged pr resolved this issue.  If filter the category, but not the office and include assigned locations continues to 500, lets open a new issue.   I just tried your query on my local system and got results (didn't get a 500) but I also don't have ton of data in my db so its possible I just don't have a group that triggers.

rma-rripken:
[...]

PR #584, opened 2024-03-15
If getAll was called with no office-id it was possible for it to retu…

rma-rripken:
…rn location groups that had a null group-id and null group-office.  Adding a loc_group_id.isNotNull condition


rma-rripken:
#582

MikeNeilson (approved):

Issue #587, opened 2024-03-19
Add servers to the swagger docs for use with openapi generators


krowvin:
While working on discussion #351 I had to add the servers to the swagger docs manually. 

This is useful because it will default to these servers on the generators making it so users do *not* need to specify which server to use. 
They can still do this manually and override the server.  

My manual edit on the output is [here](https://github.com/HydrologicEngineeringCenter/cwms-data-api-client-javascript/blob/3d982bc75bd40709a4dc0c2e81ee64cd2ee05f45/cwms-swagger.json#L8)  

And the generated output with the servers in it [here](https://github.com/HydrologicEngineeringCenter/cwms-data-api-client-javascript/blob/3d982bc75bd40709a4dc0c2e81ee64cd2ee05f45/src/src/runtime.ts#L16) 
The generator I used `typescript-fetch` pulled the first in the list in the above output.

PR #588, opened 2024-03-19
Add servers to the swagger docs for use with openapi generators #587

krowvin:
Added servers to the yaml config for swagger. 

I was able to test the json output for the openapi generator. However, I do not have an environment setup for non-production to test these work. 

If the changes look suspect I can attempt to deploy this on our internal CDA instance to confirm all other portions of CDA still work as expected *with* the servers present. 

issue #587
discussion #351

### DIFF
--- a/cwms-data-api/src/main/webapp/swagger-config.yaml
+++ b/cwms-data-api/src/main/webapp/swagger-config.yaml
@@ -8,6 +8,13 @@ info:
     name: Support
     url: https://github.com/usace/cwms-data-api
 host: https://cmws-data.usace.army.mil
+servers: 
+  - url: https://cwms-data.usace.army.mil
+    description: PRIMARY production server
+  - url: https://water.usace.army.mil,
+    description: Alias for the PRIMARY production server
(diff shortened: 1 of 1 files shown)


cwms-data-api/src/main/webapp/swagger-config.yaml
```diff
@@ -8,6 +8,13 @@ info:
     name: Support
     url: https://github.com/usace/cwms-data-api
 host: https://cmws-data.usace.army.mil
+servers: 
+  - url: https://cwms-data.usace.army.mil
+    description: PRIMARY production server
+  - url: https://water.usace.army.mil
+    description: Alias for the PRIMARY production server
+  - url: https://cwms.sec.usace.army.mil
+    description: CWBI Production Server
```
adamkorynta: CWBI is internal only

krowvin:
Added here:

- https://github.com/USACE/cwms-data-api/blob/a61533c46aa431a3c43d1583a8bfcc5b7ca6634d/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java#L887

PR #589, opened 2024-03-19
Add URL's for large CLOB/BLOB retrieval

adamkorynta:
Supersedes: https://github.com/USACE/cwms-data-api/pull/549

Add in two controllers for single Blob/Clob value downloads. URL's for these endpoints are provided by the Binary/Text TimeSeries Controllers if the values are larger than 64 kB (value overridable through system properties).

Integration tests were added to test streaming random text and byte[] values.


MikeNeilson (approved):

rma-rripken:
lgtm

PR #590, opened 2024-03-20
fix swagger UI not displaying non-CRUD endpoints

### LINKED ISSUE #404: Non-crud endpoints not grouped as expected in Swagger (recent time series and levels as time series)
The required OpenAPI annotations seem to be present for both, but they are still being added to the "Default" grouping in Swagger.

adamkorynta:
adding handlers as lambda expressions does not work as the generated anonymous inner class itself isn't annotated

Fixes #404 

Code was moved as-is into new classes.


adamkorynta:
image
image
image
Screen grabs of it showing correctly in swagger now.

MikeNeilson (approved):

PR #598, opened 2024-03-28
Fixed deserialization of ZonedDateTime by removing Json formatter pat…

rma-bryson:
Error received:

com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `java.time.ZonedDateTime` from String "2024-03-28T21:26:00Z": Failed to deserialize java.time.ZonedDateTime: (java.time.format.DateTimeParseException) Text '2024-03-28T21:26:00Z' could not be parsed at index 19
	
Removing the hard-coded formatter resolved this issue.


MikeNeilson:
So then what does the Swagger UI look like with this change?

rma-bryson:
> So then what does the Swagger UI look like with this change?

No change in the Swagger UI from what I can see

MikeNeilson (approved):

PR #591, opened 2024-03-20
Feature/ForecastControllers

dreina-gei:
Controllers for forecast data

PR #603, opened 2024-04-04
Feature/forecast controllers test

rma-rripken:
Added integration test for forecast-spec and forecast-instance.
Also changed forecast-spec and forecast-instance to extend CwmsDTO and added fields to forecast-instance so that it could find its corresponding spec.


rma-rripken:
I don't actually want the gradle file changed - I was just starting from devins pr https://github.com/USACE/cwms-data-api/pull/591

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastInstance.java
```diff
@@ -1,130 +1,283 @@
...
-  }
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class)
+@JsonDeserialize(builder = ForecastInstance.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public class ForecastInstance extends CwmsDTO {
+
+    // unique identifier: spec + issueDateTime
+    // to find spec we need officeId, specId and locationId
+    // we get officeId from CwmsDTO
```
rma-rripken: I'm not sure if this class should have officeId, specId, locationId   
or should it just have the entire ForecastSpec?
or does it actually not need any of those in the actual object and the Controller methods (mostly create and update) should just take extra parameters?

adamkorynta: My vote is to include the spec. That's how I'm doing it in the hec-nucleus models: https://bitbucket.hecdev.net/projects/SC/repos/hec-nucleus/pull-requests/114/overview

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
```diff
@@ -1,88 +1,90 @@
...
-  }
-
-  public List<TimeSeriesIdentifierDescriptor> getTimeSeriesIds() {
-    return timeSeriesIds;
-  }
-
-  public void validate() throws FieldException {
-    //TODO
-  }
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class)
```
adamkorynta, line 28: @rma-bryson was asking if this was enough for the DTO to show up in the OpenAPI? or does it only read the controllers and pick up DTO's referenced in the controller's annotations?

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastInstance.java
adamkorynta: office, spec id, and designator would be the unique identifiers. Still not sure if the location is included in that

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastSpec.java
rma-rripken: I suspect it needs to be referenced but I don't know

adamkorynta (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dto/forecast/ForecastInstance.java
adamkorynta: confirmed, location id is not included in the unique constraint

adamkorynta (approved):

PR #615, opened 2024-04-10
Bugfix/LargeBlobUrl

dreina-gei:
Added '/' in between ctx.path() and tsId for BinaryTimeSeries large blob url


dreina-gei:
This bug is odd. When running integration tests in hec-cwms-data-access the '/' is missing causing a 404 error. However, adding a '/' in cda causing the cda valueUrl test to have '//.' Will have to debug further.

adamkorynta:
I'd expect the same error for the text time series.

MikeNeilson:
Yeah, that's odd. it should automatically add the /, but it does it with a 302 or 301, is the test not following redirects?

dreina-gei:
> I'd expect the same error for the text time series.

Added same logic to text time series

dreina-gei:
> Yeah, that's odd. it should automatically add the /, but it does it with a 302 or 301, is the test not following redirects?

Is there a way I could check this?

MikeNeilson:
Not sure, you could query manually in curl with -i and it'll give you all the headers, but most of the test and I thought I saw yours did to have "followRedirects" set.

adamkorynta (approved):

Issue #607, opened 2024-04-08
Cache the isVersioned status


rma-rripken:
As discussed in https://github.com/USACE/cwms-data-api/pull/602 this ticket is to cache in memory the result of isVersioned

PR #617, opened 2024-04-12
Added a memory-based cache of isVersioned.

rma-rripken:


rma-rripken:
#607

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -230,21 +259,18 @@ public TimeSeries getTimeseries(String page, int pageSize, String names, String
...
-            retrieveSelectData = DSL.sql(
-                    "table(cwms_20.cwms_ts.retrieve_ts_out_tab(?,?,cwms_20.cwms_util.to_timestamp(?),cwms_20.cwms_util.to_timestamp(?),"
-                            + "'UTC',?,?,?,?,?,?,?,?) ) retrieveTs",
-                    tsId, unit, beginTimeMilli, endTimeMilli,
-                    trim, startInclusive, endInclusive, previous, next, versionDateMilli, maxVersion,officeId);
-}
+        // Now we're going to call the retrieve_ts_out_tab function to get the data and build an
+        // internal table from it so we can manipulate it further
+        // This code assumes the database timezone is in UTC (per Oracle recommendation)
+        SQL retrieveSelectData = DSL.sql(
```
MikeNeilson, line 265: this appears to be duplicatated but never used directly..

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-rripken: Which part?  The retrieveSelectData SQL object on 265? It gets used again on 367.   Its not a new variable, I reorganized a little in order to reduce duplication.  Adam, Ryan and I had to recently debug something in here and the three of us had to study the two branches term by term to see if they were really doing the same thing.   Hopefully this change makes it clearer that the only part of the sql that changes is whether we pass a null or call to_timestamp for the versionDate field. 

The sql in this section is pretty complicated.  At some-point we may need to profile it and see where the time is going (or have you done that?).  Maybe some of the pieces could be moved into their own methods or be made optional via flags. Or even moved into their own Controller end-points.  Things like the Extents, Datum, Quality

MikeNeilson: Line 294. The only difference is the ending column name but but otherwise I don't see much difference.

Though I'll grant it's entirely plausible we actually do need to duplicate that just because of that.

MikeNeilson (approved):
Still a comment the I think duplicated code, but I'll let it go if you look at it again and there is a difference.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-rripken: Just to make sure I understand the comment - you are saying that the query starting on 294 in the add_isversioned_cache branch duplicates the query on 266 but uses a different name?  Looking at those two queries, I think one names the table "retrieveTs" and the other names it "retrieveTsTotal".  Is that the duplication?    Its entirely possible that those two queries could be combined together.  I don't see where either table name is used again.  Combining those into one query is a slightly more involved change than I want to do in this PR but it should probably be looked into.  Wouldn't it be great if that improved the runtime?  It would do it on the first page of data too which is probably the most important.  I don't know enough about jOOQ and Oracle to be sure there would be an improvement though.  Seems at least possible that Oracle would recognize that its the same query.   Should I open a new issue to look into that?

rma-rripken: I went ahead and made a new issue #628 so that we don't forget.

PR #627, opened 2024-04-29
Switch dockerfile link from dlcdn.apache.org to archive.apache.org so…

rma-rripken:
… that we can update tomcat versions on our own timeframe.  The dlcdn links apparently go away when new releases are made.


MikeNeilson (approved):
Good idea.

Issue #630, opened 2024-05-06
Bad unit in timeseries post returns 500 instead of reasoanble message.


MikeNeilson:
stack trace:
```
        org.jooq.exception.DataAccessException: Error while running ConnectionCallable
                at org.jooq.impl.DefaultDSLContext.connectionResult(DefaultDSLContext.java:664)
                at org.jooq.impl.DefaultDSLContext.connection(DefaultDSLContext.java:696)
                at cwms.cda.data.dao.JooqDao.connection(JooqDao.java:435)
                at cwms.cda.data.dao.TimeSeriesDaoImpl.create(TimeSeriesDaoImpl.java:903)
                at cwms.cda.api.TimeSeriesController.create(TimeSeriesController.java:213)
                at io.javalin.apibuilder.CrudFunction$3.invoke$lambda-0(CrudHandler.kt:32)
                at io.javalin.apibuilder.CrudFunctionHandler.handle(CrudHandler.kt)
                at cwms.cda.security.CwmsAccessManager.manage(CwmsAccessManager.java:33)
                at cwms.cda.security.MultipleAccessManager.manage(MultipleAccessManager.java:41)
                at io.javalin.http.JavalinServlet.addHandler$lambda-0(JavalinServlet.kt:96)
                at io.javalin.http.JavalinServlet$lifecycle$2$1$1.invoke(JavalinServlet.kt:43)
                at io.javalin.http.JavalinServlet$lifecycle$2$1$1.invoke(JavalinServlet.kt:43)
                at io.javalin.http.JavalinServletHandler.executeNextTask(JavalinServletHandler.kt:99)
                at io.javalin.http.JavalinServletHandler.queueNextTaskOrFinish$lambda-1(JavalinServletHandler.kt:85)
                at java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:995)
                at java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2137)
                at io.javalin.http.JavalinServletHandler.queueNextTaskOrFinish$javalin(JavalinServletHandler.kt:85)
                at io.javalin.http.JavalinServletHandler.executeNextTask$lambda-11$lambda-10(JavalinServletHandler.kt:119)
                at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:616)
                at java.util.concurrent.CompletableFuture.uniApplyStage(CompletableFuture.java:628)
[...]

PR #631, opened 2024-05-07
Bugfix/630 ts units message

rma-rripken:


rma-rripken:
Closes #630

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -434,6 +434,47 @@ private static InvalidItemException buildInvalidItem(RuntimeException input) {
...
+        if (input instanceof DataAccessException) {
+            DataAccessException dae = (DataAccessException) input;
+            cause = dae.getCause();
+        }
+
+        String localizedMessage = cause.getLocalizedMessage();
+        if (localizedMessage != null) {
+            // skip ahead in localizedMessage to "ORA-20102:"
+            int start = localizedMessage.indexOf("ORA-20102:");
+            if (start >= 0) {
```
MikeNeilson, line 471: I say skip past the `ORA-20102:` as well.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -434,6 +434,47 @@ private static InvalidItemException buildInvalidItem(RuntimeException input) {
...
+                localizedMessage = localizedMessage.substring(start);
+                String[] parts = localizedMessage.split("\n");
+                if (parts.length >= 1) {
+                    localizedMessage = parts[0];
+                }
+            }
+        }
+
+        return new InvalidItemException(localizedMessage, cause);
```
MikeNeilson, line 486: Needs the owasp sanitizer used here. The unit itself is user input so it needs to be properly escaped if sent back to the user; which in this case I think it should.

Otherwise the unit `steal user cookie values and do bad things` could cause some problem.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -153,6 +154,47 @@ void test_lrl_1day() throws Exception {
...
+                    .redirects().follow(true)
+                    .redirects().max(3)
+                    .post("/timeseries/")
+                    .then()
+                    .log().ifValidationFails(LogDetail.ALL, true)
+                    .assertThat()
+                    .statusCode(is(HttpServletResponse.SC_BAD_REQUEST))
+                    .body("details.message", containsString("The unit: m is not a recognized CWMS Database unit for the Temp Parameter"));
+
+        } catch( SQLException ex) {
```
MikeNeilson: You can drop this try-catch, the test already throws Exception.

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
rma-rripken: Thats a great point

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
rma-rripken: ok. done

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
rma-rripken: I added this

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
rma-rripken: sure. also changed in test case this was based on

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -422,16 +426,80 @@ private static InvalidItemException buildInvalidItem(RuntimeException input) {
...
+        localizedMessage = sanitizeOrNull(localizedMessage);
+
+        if (localizedMessage == null || localizedMessage.isEmpty()) {
+            localizedMessage = "Invalid Units.";
+        }
+
+        return new InvalidItemException(localizedMessage, cause);
+    }
+
+    private static @Nullable String sanitizeOrNull(@Nullable String localizedMessage) {
```
MikeNeilson, line 489: Couple of things. There is a sanitizer defined in ApiServlet, we should pass that around as it will then contain the project global settings.

Two, if sanitized, you can just pass the data back, `::sanitize` makes the data safe to return to the user. So setting null isn't required.

MikeNeilson: Also to be fair if it's just a pain to get that sanitizer instance here, this is fine. But the sanitized data can be passed back to the user.

MikeNeilson (changes requested):
Generally looks good to me.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
rma-rripken: It may be safe to return but it doesn't seem helpful.    
"The unit: <script>alert()</script> is not a recognized CWMS Database unit for the Temp Parameter"   becomes "The unit: "
Which the user would end up seeing as:
{
"message": "Bad Request.",
"incidentIdentifier": "-12341234123451234521345",
"details": {
"message": "The unit: "
}
}

I guess that is better than nothing and its only nefarious users that would see that message but it sort of looks like CDA is in the middle of a stroke.  Seemed better to just return something we control than to let them figure out where in the input our sanitizer started complaining.

rma-rripken: passing the ctx in isn't convenient b/c the exception get wrapped from jooq listener thing.  Not saying its impossible. I did see the sanitizer in ApiServlet and couldn't think of a clean way to get it.

MikeNeilson: That's fair. Had to ask. Put a note in about that though so we all remain aware.

MikeNeilson: > It may be safe to return but it doesn't seem helpful. "The unit: <script>alert()</script> is not a recognized CWMS Database unit for the Temp Parameter" becomes "The unit: " Which the user would end up seeing as: { "message": "Bad Request.", "incidentIdentifier": "-12341234123451234521345", "details": { "message": "The unit: " } }
> 
> I guess that is better than nothing and its only nefarious users that would see that message but it sort of looks like CDA is in the middle of a stroke. Seemed better to just return something we control than to let them figure out where in the input our sanitizer started complaining.

Hmm, yeah, and it's only checking for `<script>` which is always invalid anyways.

MikeNeilson (approved):

rma-rripken:
I created a new issue for your comment about the sanitizers

Issue #632, opened 2024-05-07
Timeseries Write Payload limit of 20,000


msweier:
There appears to be a write payload limit of 20,000 or so records at the timeseries end point. I was using cwms-python to merge period of record data. The workaround is to break up the store, but it makes clunky code. Not sure if this if a hard and fast limitation of CDA. If it is, maybe cwms-python could include something to handle it. 

![image](https://github.com/USACE/cwms-data-api/assets/9432394/c58ae7fe-b5e4-4d03-abab-a2b60925f62a)

MikeNeilson:
Hmm, we'll have to check the default payload sizes. I'm pretty sure the defaults for upload are rather small. Can you please provide here the raw request? E.g. exactly what should be sent, or at least the body.

msweier:
Well it's in the cwms-python wrapper, but if I limit it to 20,000 records it stores, if it's 25,000 it fails. It is an irregular timeseries.
`combined_data = cwms.timeseries_df_to_json(data = combined_df.head(25000), tsId = 'LockDam_02.Stage.Inst.~15Minutes.0.test', units = units, office_id = OFFICE)`

`cwms.store_timeseries(data=combined_data)`
![image](https://github.com/USACE/cwms-data-api/assets/9432394/8ba268a3-f3f9-4907-a0c6-c4974ceab0d0)

MikeNeilson:
Yeah, the server can be configured to allow more, I just want to see the raw JSON (starting at [29]: but not an image) so we can see the actual size in bytes as that's what the error is based on.

msweier:
I think it's probably a 1 MB limit. I suppose I could do a little rounding on the stage to help.
![image](https://github.com/USACE/cwms-data-api/assets/9432394/e7303cf3-1804-403a-9fb2-02890b61857a)
[Uploading data.json…]()

MikeNeilson:
Yeah, that'll do it. I guess the question would be just how common this volume of data will be?
I'm not opposed to increasing it, I can think of a few rating tables that would need it as well.

msweier:
Probably not that common, but if this is our tool to read/write to the database, it is going to happen here and there. I'm doing a little housekeeping in the DB while testing CDA/cwms-python. 

Is it possible to have the volume increase if we are on the Corps network or a flag? Another option would be wrap the solution in cwms-python.

MikeNeilson:
[...]

PR #644, opened 2024-05-14
Changes to read directly from user provided stream and avoid "Payload…

rma-rripken:
… Too Large"

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
@@ -69,7 +69,7 @@ import io.javalin.plugin.openapi.annotations.OpenApiParam;
 import io.javalin.plugin.openapi.annotations.OpenApiRequestBody;
 import io.javalin.plugin.openapi.annotations.OpenApiResponse;
 import java.io.IOException;
-import java.io.StringReader;
+import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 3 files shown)


rma-rripken:
This is for #632.  
Ideally I'd also add a unit test with an input that exceeds 2MB.   At the moment that seems tedious to create.  I think its just the size of the body that gets POST'd that needs to be big so it probably could be programmatically generated.   Maybe try to do a create with points for 20 years?  I could use a suggestion on the easiest test that demonstrated the old limits

MikeNeilson:
> This is for #632. Ideally I'd also add a unit test with an input that exceeds 2MB. At the moment that seems tedious to create. I think its just the size of the body that gets POST'd that needs to be big so it probably could be programmatically generated. Maybe try to do a create with points for 20 years? I could use a suggestion on the easiest test that demonstrated the old limits

@msweier Got a lot of data to provide us so Ryan can run this test?

MikeNeilson (commented):
Looks good. Let's given Mitch some time to see if he can build that bulk test set for us.

rma-rripken:
I hacked together a test that makes a big input.

MikeNeilson (approved):
Nice trick.

msweier:
Some of our longer period of record TS are about 13 MB or so, but most of that is from the 9 years of 15 minute data.

MikeNeilson:
> Some of our longer period of record TS are about 13 MB or so, but most of that is from the 9 years of 15 minute data.

I think I'd be more worried about CWMS DB being able to save that in a reasonable amount of time than CDA doing the initial processing.

But that'll be a good request to use in some future load testing. I doubt you'd do it often but it could definitely interfere with other real-time ops if done so we should test that.

msweier:
> I think I'd be more worried about CWMS DB being able to save that in a reasonable amount of time than CDA doing the initial processing.

I agree. It takes about 8 minutes to save to a district instance for 13 MB in a loop that breaks it up into 2 MB chunks. It's long but not terrible since it's more of a one off deal. I didn't try CWBI test.

rma-rripken:
I made a test and validated that it failed without these changes so I don't need more in the short term.  These changes won't make the saves happen any faster.  There still might be reasons you want to break bigger timeseries into smaller chunks.   13MB shouldn't cause the size problem you were seeing before.  It will probably still take the same amount of time in total.

PR #648, opened 2024-05-16
In general we don't need to call DSL.upper on the database DB_OFFICE_…

rma-rripken:
…ID fields. They are already in uppercase.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
@@ -19,7 +19,6 @@ import org.jooq.impl.DSL;
 import usace.cwms.db.jooq.codegen.packages.CWMS_TEXT_PACKAGE;
 import usace.cwms.db.jooq.codegen.tables.AV_CLOB;
 import usace.cwms.db.jooq.codegen.tables.AV_OFFICE;
-
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.Reader;
@@ -107,7 +106,7 @@ public class ClobDao extends JooqDao<Clob> {
                             .from(v_clob)
                             .join(v_office).on(v_clob.OFFICE_CODE.eq(v_office.OFFICE_CODE))
                             .where(JooqDao.caseInsensitiveLikeRegex(v_clob.ID, idRegex))
-                            .and(officeLike == null ? noCondition() : DSL.upper(v_office.OFFICE_ID).like(officeLike.toUpperCase()));
+                            .and(officeLike == null ? noCondition() : v_office.OFFICE_ID.like(officeLike.toUpperCase()));
 
             total = count.fetchOne().value1();
         } else {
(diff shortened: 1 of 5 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
```diff
@@ -107,7 +106,7 @@ public Clobs getClobs(String cursor, int pageSize, String officeLike,
                             .from(v_clob)
                             .join(v_office).on(v_clob.OFFICE_CODE.eq(v_office.OFFICE_CODE))
                             .where(JooqDao.caseInsensitiveLikeRegex(v_clob.ID, idRegex))
-                            .and(officeLike == null ? noCondition() : DSL.upper(v_office.OFFICE_ID).like(officeLike.toUpperCase()));
+                            .and(officeLike == null ? noCondition() : v_office.OFFICE_ID.like(officeLike.toUpperCase()));
```
MikeNeilson, line 109: While the column itself is always upper case, the Upper on the column is a hint to the query planner. The indexes are generally all `create index... upper(column)` and the planner doesn't always do a great job of realizing that.

If you can show an explain plan that shows this isn't the case that's a good reason to change. But other wise the hint should stay.

However a change that should be made here is that instead of String::toUpperCase the DSL.upper should also be used on the input, the planner can then decide on if it needs to do the conversion or can just do a case insensitive compare.

Though for the column it should be v_office.OFFICE_ID.upper() I think, DSL.upper has some consequences.

But the TLDR, Query planners often don't make good decisions and need a lot of hints.

cwms-data-api/src/main/java/cwms/cda/data/dao/PoolDao.java
```diff
@@ -44,7 +44,7 @@ public List<Pool> catalogPools(String projectIdMask, String poolNameMask,
 		return dsl.select(DSL.asterisk()).from(view)
 				.where(condition)
 				.orderBy(view.DEFINITION_TYPE,
-						DSL.upper(view.OFFICE_ID), DSL.upper(view.PROJECT_ID), view.ATTRIBUTE, DSL.upper(view.POOL_NAME))
+						view.OFFICE_ID, DSL.upper(view.PROJECT_ID), view.ATTRIBUTE, DSL.upper(view.POOL_NAME))
```
MikeNeilson, line 47: upper on the office here definitely isn't needed. So that's fair.

MikeNeilson (changes requested):
Needs analysis of generated plans. Though DSL.upper the way it's used is likely incorrect.

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
rma-rripken: Ok.  Good point re String::toUpperCase and jOOQ upper.   Looks like Field.upper is deprecated.

PR #653, opened 2024-05-17
initial (working) mechanism for consuming Oracle Queues and sending out through ActiveMQ Artemis

adamkorynta:
utilizes Apache Camel with the AQAPI connection factory in order to hook into Oracle Queues and route to Artemis. 

TODOs:
- [ ] determine artemis connection/port
- [ ] determine which oracle queues to subscribe to
- [ ] determine clientid/durable subscriber name
- [ ] register artemis queues with swagger UI and standardize naming
- [ ] need to add SSL

artemis architecture docs: https://activemq.apache.org/components/artemis/documentation/2.4.0/architecture.html

Requirements:
- [X] connect to Oracle Queues
- [X] Ability to separate high volume from low volume
- [ ] only allow authorized access to messaged - TBD
- [ ] embedded within CDA (Artemis server can be, but not REST API)
- [X] allow access from java applications
- [ ] allow access from browser applications (only through web sockets via STOMP protocol)
- [X] durable subscribers


adamkorynta:
client side proof of concept: https://bitbucket.hecdev.net/projects/CWMS/repos/hec-cwms-data-access/pull-requests/199/overview

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -347,6 +364,55 @@ public void init() {
 
                 .routes(this::configureRoutes)
                 .javalinServlet();
+        setupQueuing();
+    }
+
+    private void setupQueuing() throws ServletException {
+        try {
+            //TODO: determine how the port is configured
+            String activeMqUrl = "tcp://" + InetAddress.getLocalHost().getHostName() + ":61616";
```
MikeNeilson, line 373: 61616 is the default ActiveMQ port, so this should be fairly safe to assume.

MikeNeilson: But I think it would also be reasonable to only use the in-memory version of ActiveMQ in this context... but I haven't read through the rest of the code yet.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -347,6 +364,55 @@ public void init() {
...
+                    .setJournalDirectory("target/data/journal")
+                    //Need to update to verify roles
+                    .setSecurityEnabled(false)
+                    .addAcceptorConfiguration("invm", "vm://0"));
+            ConnectionFactory artemisConnectionFactory = new ActiveMQJMSConnectionFactory("vm://0");
+            camelContext.addComponent("artemis", JmsComponent.jmsComponent(artemisConnectionFactory));
+            camelContext.addRoutes(new RouteBuilder() {
+                public void configure() {
+                    //TODO: configure Oracle Queue name for office
+                    //TODO: determine durable subscription name - should be unique to CDA instance?
```
MikeNeilson, line 402: Or should it be durable per (authorized) client?

MikeNeilson: The likely scenario is multiple parallel CDA instance and the clients my reconnect and be connected to a different instance. HTTP be stateless and most of these message queue systems doing 10 second HTTP long polling at their base operations.

adamkorynta: This implementation is fairly simple and does not take into account connecting to multiple instances of CDA with the intent to capture missed messages from a different instance of CDA. The architecture of ActiveMQ Artemis is designed to allow clustering, but this wouldn't work while being embedded within CDA (which was a requirement identified in the scope) unless we want to teach CDA how to identify discovery groups etc. That seems like we'd really want a dedicated Artemis server instead.

This first simple implementation completely decouples the Oracle Queues from the client as the client is only subscribing to queues/topics that Artemis is broadcasting. The durable subscriber is setup from the client to Artemis with no propagation to the AQ/JMS layer. Artemis itself is setup here to be a durable subscriber to AQ/JMS for the case of CDA outages.

Additionally, this configuration has client <-> Artemis communications go over TCP and not HTTP long polling.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
adamkorynta: https://activemq.apache.org/components/artemis/documentation/1.4.0/configuring-transports.html

TCP or HTTP can be used, I don't know if district T7's would allow TCP connections from the port.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
MikeNeilson: Ah, okay, ultimately it needs to be HTTP only *from* CDA. ActiveMQ does have an HTTP connector in there somewhere.

MikeNeilson: (But the excercise is for various proof of concepts so like, don't waste time hunting that down now if you're on a role with something.)

adamkorynta: HTTP restriction is due to T7 blocking TCP? Documentation for messaging systems seem to advise against HTTP unless TCP is blocked.

MikeNeilson: It's not a critical on the T7 vs future cloud system where plain TCP *will* be difficult to deal with. We'd have to make sure all the login pieces are there (which to be fair to us I did write a CAC auth mechanism for ActiveMQ (pre artemis) almost a decade ago so it's not impossible.)

But another design goal is using the same messaging system from say, a javascript driven web app.

That said it would make sense to expose the non HTTP connector to internal services, but that seemed fairly trivial so my goal here was how do we expose the messaging system to external users, *without* an additional "service" for them to worry about.

adamkorynta: ActiveMQ classic support tunneling over HTTP, but Artemis does not. The design for in-browser support is over WebSockets using a Stomp protocol connector. 

There is also a separate REST API that can be configured, but it is deployed as a separate war and can't extend our Javalin configuration. This would also mean we won't get JMS support.

MikeNeilson: STOMP over WebSockets seems worthwhile to me. There should already be some JMS drivers for it, and even if not should be relatively trivial to do (well, maybe not trivial enough to tack into this particular contract task, but that gives me to some for relative simplicity.)

PR #656, opened 2024-05-21
Feature/camel sse

adamkorynta:
utilizes Apache Camel with the AQAPI connection factory in order to hook into Oracle Queues and SSE's send messages to clients

TODOs:
- [ ] determine clientid/durable subscriber name
- [ ] need to check for authenticated roles
- [ ] add swagger docs, API was not added through Javalin since SSE is only supported by Javalin.

Requirements:
- [X] connect to Oracle Queues
- [X] Ability to separate high volume from low volume - listeners are separated by topics
- [X] only allow authorized access to messaged
- [X] embedded within CDA
- [X] allow access from java applications
- [X] allow access from browser applications
- [ ] durable subscribers - only from CDA to Oracle. Would need to write and maintain a mechanism from the client

Downsides:
- no acknowledgement to the server that the client has received the message
- no client durable subscribers unless we write/maintain our own
- no JMS support
- doesn't hook into javalin


MikeNeilson:
Doesn't look unreasonable. And the JMS support likely wouldn't be too difficult.

Though the lack of report back to server could definitely be an issue. Though not an insurmountable one. Further analysis would need to be done to determine if "at least" once is appropriate to these queues or just "best effort".

Issue #572, opened 2024-03-08
Expand "index page" and incorporate https://usace.github.io/groundwork/#/docs for consistency.


MikeNeilson:
This was brought up during another meeting, as we expand the hosted documentation and perhaps some default behavior this will be useful to maintain the USACE look-and-feel consistency.

adamkorynta:
Should also consider the regex explainer page: https://cwms-data.usace.army.mil/cwms-data/regexp.html

krowvin:
@adamkorynta 

I did have a mockup of the new site in react+groundwork and it had a regex page. 
https://github.com/usace-wm-swd/cwms-data-api/blob/data-query-tool/smart-inputs/cda-gui/src/pages/regexp/index.jsx

Also working on a `data-query` page of sorts. 

But we still have an issue with figuring out the base path for this. Since it is not always `cwms-data`.


@MikeNeilson Should I go ahead and bring this branch over to this repo so it's not so hidden? I kept adding features to this but should probably call it somewhere. (Data query tool)

adamkorynta:
@krowvin - great! It would be good to combine given @rma-rripken is adding more pages. See: https://github.com/USACE/cwms-data-api/pull/1169/files#diff-3403f89d85a1e3c41ccf814f67395f78076323243dae3bd98cbd78969756fea7R1

MikeNeilson:
@krowvin yes, and honestly for now we'll just assume cwms-data, that is the future and so how much do we really want to fight it?

@adamkorynta those docs *probably* should go into the new sphinx/read-the-doc based documentation we've started. I'm just not going to worry about that right now as we set that up right in the middle of Ryan creating those.

adamkorynta:
The work has been completed with the groundwork single page app merged into main.

PR #658, opened 2024-05-21
Index Page to Groundwork/React

krowvin:
# Update Landing Page to Groundwork + React using Vite.js
*This PR is for issue #572*

**NOTE** THIS PR SHOULD NOT BE APPROVED UNTIL CONFIRMED WORKING ON VARIOUS URL PATHS
*See Route Issues* below - Looking for assistance and suggestions on this!

## Major Changes:
* Converted from vanillaJS/CSS to React/Tailwind
* Followed the groundwork guidelines to implement the [Groundwork](https://usace.github.io/groundwork/) components created for WM/USACE
* Converted the static pages to client-side routed `pages` via react-router-dom (see below)
* Requires a build step:
From the `cwms-data-api\src\main\webapp-src` path run `npm run` to see the available commands or open `cwms-data-api/src/main/webapp-src/package.json` and look at `scripts`. 
1. Run `npm run build`
2. Copy the contents that get output into `dist` to the `webapps` directory (there is a script that will do this for windows in the `package.json`)

## Things of Note:
* Groundwork recommends `Redux-Bundler` as the client-side routing framework. However, in testing I attempted to use both this and `react-router-dom` to override the dynamic client-side routing. 
* The `cwms-data-api/src/main/webapp/WEB-INF/web.xml` had to be edited to handle `404 Not Found` errors and pass them *back* to the client to handle the client-side routing. There *may* be a better way to handle this.  

## Route Issues
I was able to get this working by hard coding the `base` path in vite.js and in `cwms-data-api\src\main\webapp-src\src\bundles\route-bundle.js`  

**However**, we need this to be dynamic for the district T7s. 

I took two attempts at this:
* Once with `redux-bundler` -> Initial commits 
* Another try with `react-router-dom` -> Current HEAD state

If you would like to jump to the commit where I had `redux-bundler` setup the commit hash is [7c572b4635d13bbcbd907bab3a47a9b12732c2ce](https://github.com/USACE/cwms-data-api/pull/658/commits/7c572b4635d13bbcbd907bab3a47a9b12732c2ce)


krowvin:
https://github.com/swagger-api/swagger-ui/blob/master/flavors/swagger-ui-react/index.jsx

Provided by @MikeNeilson possible solution to handling swaggerui state in react

krowvin:
I'll submit a new PR with the work wes and I are doing here:

https://github.com/usace-wm-swd/cwms-data-api/tree/data-query-tool/smart-inputs/cda-gui

Issue #360, opened 2023-05-03
implement has data for TimeSeries and Location


MikeNeilson:
This should assume at_ts_extents is up-to-date or the query will take forever.

rma-rripken:
For timeseries could this be achieved by joining with extents and excluding rows where AV_TS_EXTENTS_UTC.VERSION_TIME
AV_TS_EXTENTS_UTC.EARLIEST_TIME
AV_TS_EXTENTS_UTC.LATEST_TIME
AV_TS_EXTENTS_UTC.LAST_UPDATE
are all null ?   
Or is there a better way?
This flag would be on the catalog calls -right?

rma-rripken:
For the Location Catalog does hasData still refer to the timeseries extents?

MikeNeilson:
That's a good question. Technically I think would only be one part of the equation. There's also the location levels.

@perrymanmd is there an equivalent to the extents for the location levels?

perrymanmd:
No - there is no equivalent for location levels.

rma-psmorris:
Mike, you can probably close this one. Please review.

PR #661, opened 2024-05-23
Bugfix/360 implement has data

rma-rripken:
Implementing exclude-empty for timeseries catalog


cwms-data-api/src/main/java/cwms/cda/api/CatalogController.java
```diff
@@ -129,6 +136,15 @@ public void getAll(Context ctx) {
                 @OpenApiParam(name = BOUNDING_OFFICE_LIKE, description = "Posix <a href=\"regexp.html\">regular expression</a> "
                         + "matching against the location bounding office. "
                         + "When this field is used items with no bounding office set will not be present in results."),
+                @OpenApiParam(name = Controllers.INCLUDE_EXTENTS, type = Boolean.class,
+                        description = "Whether the returned catalog entries should include timeseries "
+                                + "extents. Only valid for TIMESERIES. Default is true."),
+                @OpenApiParam(name = Controllers.EXCLUDE_EMPTY, type = Boolean.class,
```
MikeNeilson, line 144: Would it make sense to invert this condition? E.g. only include those that have data?
We do have to join a table so maybe a performance hit, but I know that's cause some confusion in the past.

@krowvin @DanielTOsborne @adamscarberry

rma-rripken: I need a decision on this.   It would make sense to me for the flag to be include-empty or exclude-empty.  It sounds like more than one person expected the results to exclude empty items by default so maybe that is the default behavior.

MikeNeilson: Exclude empty by default, It'll cause less confusion overall with the way publicly available versions are setup.

rma-rripken:
Related to #360

MikeNeilson (approved):

PR #663, opened 2024-05-28
Setting sesssiontimezone to UTC

rma-rripken:
This probably fixes the issue in the integration test.  Need too figure out how to add similar statements to production pools if that is indeed the fix


MikeNeilson:
We could just make it part of the connection setup like the office id/user session. Yeah, it's a bit silly maybe, but it would make sure it was done.

MikeNeilson (approved):

Issue #650, opened 2024-05-17
Location Levels should allow arbitrary units


MikeNeilson:
#596 indicates that the current location levels endpoints assumes users will only required "SI" units. This is an incorrect assumptions.
Location levels can be any unit and the user should be allowed to specify them.

rma-rripken:
#664 removed the SI-only requirement.   But levels does not yet support arbitrary user-supplied units.

PR #664, opened 2024-05-30
Removed use of JDomLocationLevelImpl because it was converting values…

rma-rripken:
Relates to #650


MikeNeilson (commented):
Looking reasonable so far.

rma-rripken:
This doesn't enable user provided units but removes the SI only limitation.

MikeNeilson (approved):
Good start, at least there's some better unit support and hopefully it'll let the next level of work happen easier.

PR #666, opened 2024-05-31
add DTO for CWMS Property

adamkorynta:
includes unit tests for in-memory serialization/deserialization in addition to deserialization from disk

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/Property.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/Property.java
@@ -1,0 +1,149 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2024 Hydrologic Engineering Center
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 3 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dto/Property.java
```diff
@@ -0,0 +1,149 @@
...
+            throw new FieldException("The 'category' field of a Property cannot be null or empty.");
+        }
+        if (this.name == null || this.name.trim().isEmpty()) {
+            throw new FieldException("The 'name' field of a Property cannot be null or empty.");
+        }
+        if (this.office == null || this.office.trim().isEmpty()) {
+            throw new FieldException("The 'office' field of a Property cannot be null or empty.");
+        }
+        if (this.value == null || this.value.trim().isEmpty()) {
+            throw new FieldException("The 'value' field of a Property cannot be null or empty.");
```
MikeNeilson: This seems... odd?

rma-rripken: Which part?  That the validate does anything at all or that value can't be null/empty?

MikeNeilson: I just checked, prop_value and prop_comment are both Nullable in the database.

cwms-data-api/src/test/resources/cwms/cda/data/dto/property.json
```diff
@@ -0,0 +1,6 @@
+{
+  "category": "TestCategory",
+  "name": "TestName",
+  "office": "TestOffice",
```
rma-rripken: lots of cda stuff is office-id and not office

cwms-data-api/src/main/java/cwms/cda/data/dto/Property.java
MikeNeilson: We should also probably have the comment here, maybe hide it behind an option for the endpoint, general get and use value doesn't need it, but a GUI that lists them all for configuration would.

MikeNeilson: > Which part? That the validate does anything at all or that value can't be null/empty?

That value can't be null or empty. Both are a valid state for a property.

adamkorynta: good reminder that I forgot the comment. My understanding is that this method is for validating client submissions, what is the case where we'd accept a null property value? If we want to delete, we'd use the delete function.

cwms-data-api/src/test/resources/cwms/cda/data/dto/property.json
adamkorynta: confusing that the query parameter is office and the DTO is office-id.
What about "name"? I see a similar mismatch and inconsistency among DTO's there

cwms-data-api/src/main/java/cwms/cda/data/dto/Property.java
MikeNeilson: Hmm, fair point, but "Property EXISTS and is set to NULL" does equal "Property does not exist". 
One is far more extreme than the other.

adamkorynta: ok, updated to remove validation on value

MikeNeilson (approved):

PR #667, opened 2024-05-31
add property controller, dao, and integration tests

adamkorynta:
Adds features to handle properties in the Content Data API (CDA). It adds a new PropertyController and PropertyDao to manage access and manipulation of properties in the database layer.

Java client: https://github.com/HydrologicEngineeringCenter/cwms-data-api-client/pull/167

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/PropertyController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/PropertyController.java
@@ -27,14 +27,12 @@ package cwms.cda.api;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.flogger.FluentLogger;
 import cwms.cda.data.dao.PropertyDao;
 import cwms.cda.data.dto.Property;
 import cwms.cda.formatters.ContentType;
 import cwms.cda.formatters.Formats;
 import cwms.cda.formatters.FormattingException;
-import cwms.cda.formatters.json.JsonV2;
 import io.javalin.apibuilder.CrudHandler;
 import io.javalin.core.util.Header;
 import io.javalin.http.Context;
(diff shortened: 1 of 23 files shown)


cwms-data-api/src/test/java/cwms/cda/api/PropertyControllerIT.java
```diff
@@ -0,0 +1,258 @@
...
+        TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
+
+        //Create the property
+        given()
+                .log().ifValidationFails(LogDetail.ALL, true)
+                .accept(Formats.JSONV2)
+                .contentType(Formats.JSONV2)
+                .body(json)
+                .header(AUTH_HEADER, user.toHeaderValue())
+                .when()
```
MikeNeilson: unindent .when and .then. (also looks like it's all indented twice.)

MikeNeilson (changes requested):
Some formatting got away from you, but otherwise looks great.

cwms-data-api/src/main/java/cwms/cda/api/PropertyController.java
```diff
@@ -0,0 +1,296 @@
...
+            retVal = om.readValue(body, Property.class);
+        } catch (Exception e) {
+            throw new IOException("Failed to deserialize property", e);
+        }
+        return retVal;
+    }
+
+    private static ObjectMapper getObjectMapperForFormat(ContentType contentType) {
+        ObjectMapper om;
+        if (ContentType.equivalent(Formats.JSONV2, contentType.toString())) {
```
MikeNeilson: The content-type mapper was design so you can just pass the data type in and with an annotation on the datatype get the right formatter given the content-type string.

This function is duplication of all that logic.

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/PropertyController.java
adamkorynta: Looks like that was only done one-way. I implemented the deserialization and updated my code to use the Formats factory methods. I'll do a follow-up PR removing the redundant code that I had copied this logic from.

cwms-data-api/src/test/java/cwms/cda/api/PropertyControllerIT.java
adamkorynta: oops, missed one, updated

MikeNeilson (approved):
Nice, thanks for the additional improvement for parsing the content as well.

PR #668, opened 2024-06-01
DTO object for Project

rma-rripken:


MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
```diff
@@ -0,0 +1,317 @@
...
+    private final Double federalOAndMCost;
+    private final Double nonFederalOAndMCost;
+    private final String authorizingLaw;
+    private final String projectOwner;
+    private final String hydropowerDesc;
+    private final String sedimentationDesc;
+    private final String downstreamUrbanDesc;
+    private final String bankFullCapacityDesc;
+    private final String pumpBackLocationId;
+    private final String pumpBackOfficeId;
```
rma-rripken: Curious what you think about this being just these two fields.  Or should it be a Location object?

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
```diff
@@ -0,0 +1,317 @@
...
+    private final String nearGageLocationId;
```
rma-rripken: Same with nearGage - should this just be a Location object?   Or should I make a LocationTemplate/LocationRef class that is just office and id and use that ?

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
```diff
@@ -0,0 +1,317 @@
...
+
+@JsonDeserialize(builder = Project.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSON, formatter = JsonV2.class)
+public class Project extends CwmsDTO {
+
+    private final String name;
+    private final Double federalCost;
+    private final Double nonFederalCost;
```
rma-rripken, line 23: What do you think about the cost classes?   Is Double sufficient?  Or should the DTO hold them as BigDecimal?

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
```diff
@@ -0,0 +1,317 @@
...
+    private final Instant costYear;
```
rma-rripken, line 24: Should costYear stay an Instant?  Is it always June 1 ?

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
MikeNeilson: I think double should be fine. Might be worth a review of existing data to see what the actual numbers are but the database type is binary_double so BigDecimal isn't going to get us anything here.

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
MikeNeilson: It's DATE in the database, so it's likely it could vary. I think INSTANT has given us the most flexbility in rendering and reading back.

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
MikeNeilson: hmm, I can see that going either way. But the full Location could be really convenient. Assuming there isn't a bad performance hit, but I wouldn't expect so since the database contains office,location and that query is usually rather quick.

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
MikeNeilson: Oh, I like that that idea. You may not always need the full Location data, but something organized and easier to search with is a good idea.

MikeNeilson (commented):
I like the suggestion you made about locationtemplate/locationref. That would likely be good for the other endpoints with locations as well.

cwms-data-api/src/test/resources/cwms/cda/data/dto/project.json
```diff
@@ -13,10 +13,14 @@
   "sedimentation-desc" : "Sedimentation Description",
   "downstream-urban-desc" : "Downstream Urban Description",
   "bank-full-capacity-desc" : "Bank Full Capacity Description",
-  "pump-back-location-id" : "Pumpback Location Id",
-  "pump-back-office-id" : "SPK",
-  "near-gage-location-id" : "Near Gage Location Id",
-  "near-gage-office-id" : "SPK",
+  "pump-back" : {
```
MikeNeilson: I feel like this should still say "-location", but that might be just me. @adamkorynta @DanielTOsborne @krowvin 

(Don't wait forever for comment, just tagging for possible thoughts now.)

MikeNeilson (commented):
I have one generally minor concern but let's give people a hour or so to check in.

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
adamkorynta: I think location ref is a good idea if you don't want to go full Location, for the Embankments that I'm working on, I have two fields projectId and projectOfficeId that could be modified to projection-location-ref.

base-location-id
sub-location-id
location-id(?)
office-id

cwms-data-api/src/test/resources/cwms/cda/data/dto/project.json
adamkorynta: I agree with `pump-back-location` to be absolutely clear it isn't operational data

rma-rripken: Adding -location Makes sense to me.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
rma-rripken: I'm going to see how far I can get with Location.  I may still switch to some sort of LocationRef.  Surprised we haven't needed a LocationRef until now

MikeNeilson (approved):

cwms-data-api/src/main/java/cwms/cda/data/dto/Project.java
MikeNeilson: Sounds reasonable, if we go to the point of "LocationRef" we'll want to revisit talks about single unique identifier and how that gets looked up. e.g. `/locations/<LocationRef>`  where location ref is something that can be in the path correctly.

PR #669, opened 2024-06-03
standardize deserialization through Formats API

adamkorynta:


adamkorynta:
Need to determine if XML serialization should use JAXB vs Jackson. The TimeSeries DTO doesn't work with Jackson (at least with the current configuration). Need to determine which is more widely used for XML.

adamkorynta:
| DTO | Current Lib | Works with Jackson | Works with JAXB |
| ---- | -------| :------: | :-------: |
| Catalog | N/A | N/A | N/A | 
| LocationLevel | Jackson | No | No |
| Office | Jackson | N/A | N/A |
| Clob | Jackson | Yes | Yes |
| Clobs | Jackson | Yes | Yes |
| TimeSeries | Jackson | No | Yes |

MikeNeilson:
Probably just needs a few hints honestly, If we can make it work simply I'm certainly in favor of moving away from JAXB. The JSON ObjectMapper concept seems to make more sense. Maybe that's just me though.

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
```diff
@@ -15,43 +15,75 @@
...
     private final Double publishedLongitude;
+    @XmlElement(name = "published-latitude")
     private final Double publishedLatitude;
+    @XmlElement(name = "vertical-datum")
     private final String verticalDatum;
+    @XmlElement(name = "elevation")
     private final Double elevation;
+    @XmlElement(name = "map-label")
     private final String mapLabel;
+    @XmlElement(name = "bounding-office-id")
```
MikeNeilson: Okay, there has to be a way to do that at a more global level? That looks like it was super annoying to do.

adamkorynta: yeah, all the documentation says that I DON'T have to do this.....

Now that I've got the JAXB working and everything routing through the standard API, I'm going to try and switch to Jackson (which will definitely let me remove all these annotations)

adamkorynta:
> DTO	Current Lib	Works with Jackson	Works with JAXB
> Catalog	N/A	N/A	N/A
> LocationLevel	Jackson	No	No
> Office	Jackson	N/A	N/A
> Clob	Jackson	Yes	Yes
> Clobs	Jackson	Yes	Yes
> TimeSeries	Jackson	No	Yes

was able to get all of the data types working with Jackson XML parsing. Needed to add mixins for behaviors that differ between JSON and XML for VerticalDatumInfo and TimeSeries.

rma-rripken (approved):

PR #670, opened 2024-06-04
add Embankment API

adamkorynta:
will replace LookupType with Bryson's DTO when available


cwms-data-api/src/test/resources/cwms/cda/data/dto/location/kind/embankment.json
rma-rripken, line 1: Are the fields in this order b/c the java variables are in that order? Not a deal breaker but I'd prefer office and id up near the top

rma-rripken (approved):

PR #671, opened 2024-06-04
Adding content type alias mapping process

### LINKED ISSUE #500: Update endpoint: counties
create application/json alias.

RyanM-RMA:
Adding content type alias map to county, so we can map from generic to more specific content types.
Fixes #500


cwms-data-api/src/main/java/cwms/cda/api/CountyController.java
```diff
@@ -58,6 +59,13 @@
 public class CountyController implements CrudHandler {
     private final MetricRegistry metrics;
     private final Histogram requestResultSize;
+    private static final ContentTypeAliasMap CONTENT_TYPE_ALIAS_MAP = new ContentTypeAliasMap();
+
+    static
+    {
```
MikeNeilson: This kinda makes sense, the but it requires extra dev time configuration.
There's an annotation for the supported formatters that could have "default = true" added to it and the Formats processor can pick it up. 

Then we can set the default format at the DTO level, which is were we also establish the allow formats.
(Granting we have to do it again anyways at the moment due to how OpenAPI works.)

MikeNeilson (commented):
The alias map itself looks great, just mildly concerned about the way it was implemented within the CountyController vs the DTO for County.

cwms-data-api/src/main/java/cwms/cda/api/CountyController.java
RyanM-RMA: Oh, annotations are a much better idea for this, especially at the DTO level.  I didn't notice that part of the design, thank you for bringing this up Mike.

cwms-data-api/src/main/java/cwms/cda/api/CountyController.java
```diff
@@ -100,7 +93,7 @@ public void getAll(@NotNull Context ctx) {
             CountyDao dao = new CountyDao(dsl);
             List<County> counties = dao.getCounties();
             String formatHeader = ctx.header(Header.ACCEPT);
-            ContentType contentType = Formats.parseHeader(formatHeader, CONTENT_TYPE_ALIAS_MAP);
+            ContentType contentType = Formats.parseHeader(formatHeader, ContentTypeAliasMap.forDtoClass(County.class));
```
rma-rripken: Can Formats do the ContentTypeAliasMap.forDtoClass call internally?  Not saying you have to but I'm curious if there are reasons it should be passed in - if the Formats method was parseHeader(formatHeader, klass.class)  then ContentTypeAliasMap would be mostly a private implementation detail -right?

RyanM-RMA: Hah, check the latest changes.  I was already ahead of you on that.  ;-)

cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
```diff
@@ -226,14 +227,29 @@ public static ContentType parseQueryParam(String queryParam) {
 
 
     public static ContentType parseHeader(String header) {
+        return parseHeader(header, null);
+    }
+
+    public static ContentType parseHeader(String header, Class<? extends CwmsDTOBase> klass) {
```
rma-rripken, line 233: I know you have the County IT but Formats.parseHeader is public static - would be real easy to add a test of the Format.parseHeader mapping for County in either a CountyTest or in a FormatsTest unit test.

cwms-data-api/src/test/java/cwms/cda/api/CountyControllerTestIT.java
```diff
@@ -61,4 +61,58 @@ void test_county_catalog() {
             .body("[0].county-id", equalTo("000"))
             .body("[0].state-initial", equalTo("00"));
     }
+
+    @Test
+    void test_county_catalog_default_content_type() {
+        given()
+            .log().ifValidationFails(LogDetail.ALL,true)
+            .accept(Formats.DEFAULT)
+            .contentType(Formats.DEFAULT)
```
rma-rripken: Is contentType needed here?   I would think that is only needed on end-points that involve the client sending a request body.  I wouldn't think its needed on GET.

rma-rripken: We probably aren't great about doing this right in the all the IT.  You probably copied this from somewhere else.  I'm wondering if we should go thru all the tests and make sure there isn't extra cruft

MikeNeilson: Yeah, if I remember right we were having some minor issue and it was easier to just add both, but content-type *should* only be for the POST body and accept for the desired response.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesIdentifierDescriptorController.java
```diff
@@ -155,7 +155,7 @@ public void getAll(Context ctx) {
                     dao.getTimeSeriesIdentifiers(cursor, pageSize, office, idRegex);
 
             String formatHeader = ctx.header(Header.ACCEPT);
-            if ("*/*".equals(formatHeader)) {
```
MikeNeilson, line 158: I'm not entirely sure how we want to handle this but Firefox actually sends the following `Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8` as the default.

So this might not be the best location to determine if default.

MikeNeilson: Actually looks like Formats.java:231 already handles this case for us.

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesIdentifierDescriptorController.java
RyanM-RMA: The comment below also points out that it's defaulting to v1 - I think that's not quite right, it looks like it switches to just `application/json`.  This looks like an attempt like the aliasing stuff.  Does this need to be resolved at this point, or should this be added to the list of classes that need the aliasing?

cwms-data-api/src/test/java/cwms/cda/api/CountyControllerTestIT.java
RyanM-RMA: I removed the contenttype part, and re-tested, it works fine for our use case.
As for where I copied it from, the existing test there `test_county_catalog()` is what I ended up using as the framework.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesIdentifierDescriptorController.java
MikeNeilson: That's a good question.

rma-rripken (approved):

PR #672, opened 2024-06-05
Feature/cto 104 reservoir project

rma-rripken:


cwms-data-api/src/main/java/cwms/cda/data/dao/ProjectDao.java
```diff
@@ -0,0 +1,624 @@
...
+import usace.cwms.db.jooq.codegen.udt.records.PROJECT_OBJ_T;
+import usace.cwms.db.jooq.dao.util.LocationTypeUtil;
+
+public class ProjectDao extends JooqDao<Project> {
+    private static final Logger logger = Logger.getLogger(ProjectDao.class.getName());
+
+    private final Calendar UTC_CALENDAR = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+
+    public static final String SELECT_PART = "select project.office_id,\n"
+            + "       project.location_id as project_id,\n"
```
MikeNeilson: For want of Java 17 that allows the python style """ strings.

cwms-data-api/src/main/java/cwms/cda/data/dao/ProjectDao.java
```diff
@@ -0,0 +1,624 @@
...
+
+
+    private static Project buildProject(usace.cwms.db.jooq.codegen.tables.records.AV_PROJECT r) {
+        Project.Builder builder = new Project.Builder();
+        builder.withOfficeId(r.getOFFICE_ID());
+        builder.withName(r.getPROJECT_ID());
+        builder.withPumpBackLocation(new Location.Builder(r.getOFFICE_ID(), r.getPUMP_BACK_LOCATION_ID())
+                .withActive(null)
+                .build()
+        ); // Can we assume same office?
```
MikeNeilson, line 193: Hmm, that's a good question. And one for the districts, but I suspect it's cleaner here later if we just don't.
The query above does include the given office_id for each and should probably be used.

MikeNeilson (commented):
Do have a comment on the pumpback/etc location office.

cwms-data-api/src/main/java/cwms/cda/data/dao/ProjectDao.java
rma-rripken: It will be cleaner when I move the query into a resource file

cwms-data-api/src/main/java/cwms/cda/data/dao/ProjectDao.java
rma-rripken: Just removed the method that used the view with the missing office-ids.   We'll use the table join for now.

cwms-data-api/src/main/java/cwms/cda/data/dao/ProjectDao.java
rma-rripken: done

MikeNeilson (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/ProjectDao.java
MikeNeilson: A reasonable substitute.

PR #673, opened 2024-06-05
Adding aliases to State content type

### LINKED ISSUE #501: Update endpoint: states

RyanM-RMA:
Adding integration tests for aliased content types 
Adding base case tests to FormatsTest, so we can confirm that the non-aliased values work.
Fixes #501

### DIFF
--- a/cwms-data-api/src/test/java/cwms/cda/api/StateControllerTestIT.java
+++ b/cwms-data-api/src/test/java/cwms/cda/api/StateControllerTestIT.java
@@ -35,6 +35,8 @@ import cwms.cda.formatters.Formats;
 import io.javalin.core.util.Header;
 import io.restassured.filter.log.LogDetail;
 import javax.servlet.http.HttpServletResponse;
+
+import io.restassured.response.ValidatableResponse;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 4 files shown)


MikeNeilson (dismissed):

cwms-data-api/src/test/java/cwms/cda/api/StateControllerTestIT.java
rma-rripken, line 1: Its not necessary to refactor the restassured tests this much.     
The restassured tests should read:   given...when...then...assertThat

RyanM-RMA: Too much DRY for unit tests?  Do you want me to adjust it here, or leave the refactoring out going forward?

rma-rripken: I see how its cleaner but sometimes more is more.  I don't want to have to think about the tests - just want to read them.    I can see how for your three test cases having the refactored method makes it obvious what is duplicated/changed between the tests.  But then I probably wouldn't want theh ETag and cache-control test to get refactored

rma-rripken: Fix it now while its fresh and easy

MikeNeilson: Was an interesting idea, but it's good Ryan R spoke up, one does need to be careful how far you take DRY within test code.

MikeNeilson (approved):

PR #674, opened 2024-06-05
Implemented DTO For LookupType

rma-bryson:


cwms-data-api/src/main/java/cwms/cda/data/dto/LookupType.java
```diff
@@ -0,0 +1,142 @@
...
+import cwms.cda.formatters.Formats;
+import cwms.cda.formatters.annotations.FormattableWith;
+import cwms.cda.formatters.json.JsonV1;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlRootElement;
+import java.util.Objects;
+
+@XmlRootElement(name = "lookupType")
```
adamkorynta: You only called out support for JSON, so need for the JAXB annotations (I'm also trying to get rid of JAXB and only use Jackson for xml)

MikeNeilson: Extra note. For brand new data types our idea is XML *only* if the data type really calls for it. For example the Ratings data's primary format is XML so there's less translation, but something like this will likely always be JSON only.

rma-bryson: Removed JAXB

adamkorynta (approved):

rma-psmorris (commented):
No feedback, looks good

PR #681, opened 2024-06-07
Moved every ResultSet into a try-with-resources block.  Removed Atomi…

rma-rripken:
…cReferences that were used where connectionResult should have been used instead.

### DIFF
--- a/cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
+++ b/cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
@@ -26,7 +26,6 @@ import java.time.ZonedDateTime;
 import javax.servlet.http.HttpServletResponse;
 import mil.army.usace.hec.test.database.CwmsDatabaseContainer;
 import org.apache.commons.io.IOUtils;
-import org.junit.jupiter.api.Disabled;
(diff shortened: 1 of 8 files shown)


cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -563,7 +564,7 @@ void test_lrl_trim() throws Exception {
         }
     }
 
-
+@Disabled("This test is disabled because it is slow")
```
MikeNeilson: Probably good to use EnableIfProperty so it's easier to run when desired.

MikeNeilson (commented):
That works too. but if it is that slow nothing wrong with having it off by default. the EnableIf just means the developer doesn't need to remember to comment/uncomment code.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
rma-rripken: Removed the @Disable.  I was not familiar with EnableIf and DisableIf.  They do look nice just need some thought into hold I'd want it to work

MikeNeilson (approved):

PR #682, opened 2024-06-07
Blob was missing update and delete support.

rma-rripken:
This still needs IT to verify the update and delete.


MikeNeilson (commented):
Looks reasonable so far.

PR #683, opened 2024-06-07
add embankment controller and integration tests

adamkorynta:


adamkorynta:
Still working on getting the controller integration test to pass.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
rma-rripken, line 1: I have no idea what these changes are doing.  Is this not needed anymore b/c of your previous jaxb to jackson changes?

adamkorynta: yeah not needed anymore and when I added jacoco, it was throwing an NPE because jacoco adds an invisible field and the null check in there wasn't quite right

rma-rripken (approved):

PR #690, opened 2024-06-10
Bugfix/content type basins

### LINKED ISSUE #506: Update endpoint: basins
create alias application/json alias for aplication/vnd.named+pg+json

This seems like a good idea since client might decide to just always send "application/json" or whatever, but maybe because it's so specific it's not required.

RyanM-RMA:
Added aliases for default `\*/\*` and `application/json` to the `application/vnd.named+pg+json` and updated the `application/vnd.pg+json` to use the right correct formatter.  Added unit tests and integration tests to validate expected behaviors.

One note for @MikeNeilson, after talking with @rripken, and a few attempts to resolve this, I wasn't able to validate the BasinControllerTestIT behaviors.  The getOne and getAll always return a no data found error that has a contentType of html.  This has to do with the unimplemented Create method for basins, and also there are no default basins, and the only databases we have with basins are SWT databases.  I couldn't get my IT CDA to connect to a SWT db we have at RMA.  I was also unable to create basins using REGI against my docker database instance.  At this point, I'm confident it works because of the FormatsTest updates to verify the alias conversion, but I can't confirm it works end to end.  If you have any recommendations, I'm happy to hear them.

Fixes #506


rma-rripken (approved):

MikeNeilson:
Re the data, I think once the appropriate creation mechanisms for basins is in place that'll be the time to actually setup some test data. 

It is also appropriate to make up some arbitrary data to load in, sometimes we use a resource file that's just a SQL query to store it all, but I can see how difficult that could be with this particular dataset.

rma-rripken:
I went ahead and merged the pr for RyanM.  We can open another to revisit once its easier to create basin or if we decide we need to add the resource file.

MikeNeilson:
Yeah, we'll definitely want it at some point, but for this I'm not super worried since it's just detailing with the content-types. Bit ridiculous to expect the entire endpoint to be fixed at the same time and that's already on the docket for additional things.

PR #691, opened 2024-06-10
Bugfix/content type time series

### LINKED ISSUE #505: Update endpoint: Timeseries::getAll
Add deprecated header to original formats.

create application/json alias to version=2

RyanM-RMA:
Added deprecated notification to the getAll method for TimeseriesController, and aliased `application/json` and `\*/\*` as aliases to `application/json;version=2`.  Added `application/xml` as an alias for `application/xml;version=2`.
Fixes #505


rma-rripken (approved):

MikeNeilson:
So, did anyone check that the old format is still accessible?

RyanM-RMA:
If the old format was accessible before, it shouldn't have an issue with the new code, however I didn't write any code to verify the older formats still work, unless the version was tagged as a FormattableWith on the DTO.

RyanM-RMA:
@MikeNeilson do you want me to add an integration test for the older format for Time Series as part of #505 ?  It'll probably take 30 mins.

PR #694, opened 2024-06-11
Adding location level content type aliasing for application/json and xml

### LINKED ISSUE #504: Update endpoint: Level::getAll
Add deprecate header to original formats.

Create appropriate application/json -> application/json;version=2 alias.

RyanM-RMA:
Added integration tests for older behaviors
Fixes #504

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
@@ -48,6 +48,7 @@ import static cwms.cda.api.Controllers.STATUS_200;
 import static cwms.cda.api.Controllers.TIMEZONE;
 import static cwms.cda.api.Controllers.UNIT;
 import static cwms.cda.api.Controllers.UPDATE;
+import static cwms.cda.api.Controllers.VERSION;
 import static cwms.cda.api.Controllers.addDeprecatedContentTypeWarning;
 import static cwms.cda.api.Controllers.queryParamAsClass;
 import static cwms.cda.api.Controllers.requiredParam;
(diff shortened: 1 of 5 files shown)


rma-rripken (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -287,8 +282,13 @@ public void getAll(@NotNull Context ctx) {
             String timezone = ctx.queryParamAsClass(TIMEZONE, String.class)
                     .getOrDefault("UTC");
 
-            if ("2".equals(version)) {
+            String format = ctx.queryParamAsClass(FORMAT, String.class).getOrDefault("");
 
+            if (format.isEmpty())
```
rma-rripken: I'm not sure if we decided what should happen if there isn't a format parameter but the accept-header is for application/json;version=1.

rma-rripken: its possible that format.isEmpty check isn't enough

rma-rripken: you can fix that here or in a follow-up

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
rma-rripken, line 1: Haven't done things quite like this before. I'm thinking about it

rma-rripken: Can you re-arrange the code a little more so that its a single if/else?   Like maybe introduce a boolean and figure out ahead of time whether you are going legacy or java and then switch on that.
On the legacy vs java methods the parameters seem a little weird. Do they both take all the same options? Can you put them in the same order?  Ideally an order that makes sense somehow.  I'm not sure the methods are needed but I guess it could make things clear

RyanM-RMA: I reorganized it, and removed the methods.  the methods had 10 parameters, which is a bit excessive.  I hadn't put enough thought into the methods and their input parameters, just extracted it.  I don't think the methods are useful if I'm only calling them once, aside from making things a bit more focused and shorter - I only extracted them because of the need for calling the legacy twice.

rma-rripken (approved):

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
MikeNeilson: If the except header is `applciation/json;version=1` AND it's one of the endpoints that would've gone to the database or been formatted that way the system should treat it like `format=json` was provided.

PR #700, opened 2024-06-14
ts profile json

AndreasChristmann:
suggested json format for TS Profile definition and TS Profile Instance


cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofile.json
```diff
@@ -0,0 +1,8 @@
+{
+  "description": "Description",
+  "parameter-list": ["Temperature","Depth"],
+  "location-id": "Location",
```
MikeNeilson, line 4: would want more input but for the Project DTO we went with an embedded location object with a minimum of office,location-id, but could include everything

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofileinstance.json
```diff
@@ -0,0 +1,41 @@
...
+      "times" : ["09/09/2019,13:16:01", "09/09/2019,13:17:20"],
+      "value":
+      {
+        "parameter": "Temp-Water",
+        "unit": "F",
+        "values": ["1", "2"]
+      }
+    },
+    {
+      "value":
```
MikeNeilson: Times should be either within each "value" or outside the value list.

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofileinstance.json
```diff
@@ -0,0 +1,41 @@
...
+    "parameter-list": [ "Temp-Water", "Pres", "%-DO",  "Conc-DO", "Cond", "pH", "TurbN",  "Depth","Volt-Battery"],
+    "description": "Demonstration Profile for Couger Lake",
+    "office-id": "SWT",
+    "ref-ts-id": "DET.Elev-Forebay.Inst.0.0.Mixed-Rev"
+  },
+
+  "value-list" :  [
+    {
+      "time-zone": "US/Pacific",
+      "times" : ["09/09/2019,13:16:01", "09/09/2019,13:17:20"],
```
MikeNeilson: CDA *always* does either UNIX Epoch UTC or Full ISO 8601 Date-Time-TZ, but we still prefer to render everything as UTC.

AndreasChristmann: we can use the full iso and get rid of the time-zone tag

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofileinstance.json
AndreasChristmann: when they are part of the value, does that mean the profile timeseries can have different time arrays?

MikeNeilson: I don't believe so but we generally try to avoid any usage side ambiguity with CDA input/output.

AndreasChristmann: value-list, maybe a better name would be timeseries-list. I had the times inside the list because the timearray corresponds to all value arrays.

MikeNeilson: Oh, I didn't see the nesting correctly. It's okay at that level. When I first looked at it I thought the times were only in the first "value" which mean the user always has to parse the first one if they want it or not.

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofileinstance.json
AndreasChristmann: removed the timezone, added timezone to the ISO8601 times

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofileinstance.json
AndreasChristmann: changed values to be time-value pairs

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofileinstance.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+
+  "timeseries-list" :  [
+    {
+         "parameter": "Temp-Water",
+         "unit": "F",
+         "time-value-pairs":
+        [
+           {
+              "time": "09/09/2019,13:16:01Z",
```
MikeNeilson, line 19: As a TODO: this should be in the ISO8601 format which is YYYY-MM-DDTHH:MM:SS.sss[Timezone in some format].

This is fine for demonstration purposes though.

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofileinstance.json
rma-rripken: Generally your DTO will use (preferable) Instants or (second choice) ZonedDateTime and then jackson will just do the right with regard to format.   In the text-timeseries work we found that current version of jackson accepts epoch millis or ISO8601 when deserializing Instant which is user friendly.   Serializing Instant into epoch millis isn't very sexy but its short, efficient and rarely confused.

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofile.json
rma-psmorris: Agreed, this should be re-using the location definition that is being done in related tickets. Also, the Java classes that are generating this JSON should be added to this PR.

In the database the parameters have a position, what is the reason for dropping that here?

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofile.json
```diff
@@ -0,0 +1,8 @@
...
+  "office-id": "Office",
+  "ref-ts-id": "TimeSeries",
```
rma-psmorris, line 6: Use a correctly formatted time series id for examples

rma-psmorris (commented):
Feedback on the profile example JSON

cwms-data-api/src/test/resources/cwms/cda/data/dto/timeseriesprofile/timeseriesprofile.json
AndreasChristmann: the position of the parameter refers to the location of the timeseries of the specified parameter inside an input data block. In this format without the datablock the position is irrelevant.

rma-psmorris: The position is not exposed by the API and is only used for DB parameter ordering. We do not need the int value outside of the database.

PR #703, opened 2024-06-16
DTO and serialization for Time Zones

### LINKED ISSUE #499: Update Endpoint: TimeZones
Add deprecated header

RyanM-RMA:
Adding DTO and serialization process for time zones, so we can use the newer java json/xml serialization/deserialization

Fixes #499


rma-rripken:
Can we get  a unit test of just the TimeZone DTO serialization by itself?   Include a json or xml example file.

rma-rripken:
I'd like to see a resource file with an example of the serialized format.

RyanM-RMA:
oh, actually I should also get a test for the List<TimeZone> in there too, since that's what we're serializing with getAll.  I totally meant to do that.

MikeNeilson (approved):

rma-rripken (approved):

PR #705, opened 2024-06-18
CTO-67 - Implemented Stream, StreamLocation and StreamJunctionIdentifier DTOs

rma-bryson:


cwms-data-api/src/main/java/cwms/cda/data/dto/Stream.java
```diff
@@ -0,0 +1,243 @@
...
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public final class Stream implements CwmsDTOBase {
+
+    private final Boolean startsDownstream;
+    private final String flowsIntoStream;
+    private final Double flowsIntoStation;
+    private final String flowsIntoBank;
+    private final String divertsFromStream;
+    private final Double divertsFromStation;
+    private final String divertsFromBank;
```
adamkorynta: From a JSON perspective, it is hard to determine why some of these are String (ids) vs Double (flows). Recommending to rename for clarity. Also `divertsFromBank` can likely be a L/R (left/right) enum

cwms-data-api/src/main/java/cwms/cda/data/dto/Stream.java
```diff
@@ -0,0 +1,243 @@
...
+        result = 31 * result + Objects.hashCode(getDivertsFromStation());
+        result = 31 * result + Objects.hashCode(getDivertsFromBank());
+        result = 31 * result + Objects.hashCode(getLength());
+        result = 31 * result + Objects.hashCode(getSlope());
+        result = 31 * result + Objects.hashCode(getComment());
+        result = 31 * result + Objects.hashCode(getLocationIdentifier());
+        return result;
+    }
+
+    public static class Builder {
```
adamkorynta: should also be final

cwms-data-api/src/main/java/cwms/cda/data/dto/Stream.java
rma-bryson: This has been re-done to use a new stream junction identifier object

cwms-data-api/src/main/java/cwms/cda/data/dto/Stream.java
rma-bryson: made final

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamJunctionIdentifier.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private StreamJunctionIdentifier(Builder builder) {
+        this.streamId = builder.streamId;
+        this.bank = builder.bank;
+        this.station = builder.station;
+    }
+
+    @Override
+    public void validate() throws FieldException {
+        if (this.streamId == null) {
```
adamkorynta, line 61: bank isn't required?

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamJunctionIdentifier.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Objects;
+
+@FormattableWith(contentType = Formats.JSON, formatter = JsonV1.class)
+@JsonDeserialize(builder = StreamJunctionIdentifier.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+/**
+ * Reference object for a stream location containing the location identifier, bank, and station
+ */
+public final class StreamJunctionIdentifier implements CwmsDTOBase {
```
adamkorynta, line 47: `station` isn't part of the uniqueness identifier, maybe just call it `StreamJunction`?

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamJunctionIdentifier.java
```diff
@@ -0,0 +1,122 @@
...
+    private final Double station;
+
+    private StreamJunctionIdentifier(Builder builder) {
+        this.streamId = builder.streamId;
+        this.bank = builder.bank;
+        this.station = builder.station;
+    }
+
+    @Override
+    public void validate() throws FieldException {
```
adamkorynta, line 60: validate should call `streamId.validate()` after the null check

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamLocation.java
```diff
@@ -0,0 +1,171 @@
...
+        this.publishedStation = builder.publishedStation;
+        this.navigationStation = builder.navigationStation;
+        this.lowestMeasurableStage = builder.lowestMeasurableStage;
+        this.totalDrainageArea = builder.totalDrainageArea;
+        this.ungagedDrainageArea = builder.ungagedDrainageArea;
+    }
+
+    @Override
+    public void validate() throws FieldException {
+        if (this.streamLocationId == null) {
```
adamkorynta, line 66: Should also call `streamLocationId.validate()`

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamJunctionIdentifier.java
adamkorynta: or `StreamJunctionRef`? if that doesn't make sense then no need to rename

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamJunctionIdentifier.java
rma-bryson: Doesn't look required in the db. Seems pretty useless without that information though, other than to say this stream intersects this other stream someplace on some side.

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamJunctionIdentifier.java
rma-bryson: added validate call

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamLocation.java
rma-bryson: added validate call

adamkorynta (approved):

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/Stream.java
```diff
@@ -0,0 +1,161 @@
...
+import cwms.cda.formatters.annotations.FormattableWith;
+import cwms.cda.formatters.json.JsonV1;
+
+import java.util.Objects;
+
+@FormattableWith(contentType = Formats.JSON, formatter = JsonV1.class)
+@JsonDeserialize(builder = Stream.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public final class Stream implements CwmsDTOBase {
```
rma-psmorris, line 21: Why is the physical location data omitted from these implementations?

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/Stream.java
```diff
@@ -0,0 +1,161 @@
...
+    public String getComment() {
+        return comment;
+    }
+
+    public LocationIdentifier getStreamId() {
+        return streamId;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
rma-psmorris, line 83: I have questions on the addition of equals/hashcode in the DTOs in general.
1. They do not follow the business rules defined for this data's uniqueness.
2. They add more data than what is used for this data's uniqueness.
3. Is there a usage of hashcode/equals outside of unit tests?

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamJunctionIdentifier.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Objects;
+
+@FormattableWith(contentType = Formats.JSON, formatter = JsonV1.class)
+@JsonDeserialize(builder = StreamJunctionIdentifier.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+/**
+ * Reference object for a stream location containing the location identifier, bank, and station
+ */
+public final class StreamJunctionIdentifier implements CwmsDTOBase {
```
rma-psmorris, line 47: This class is confusing, a junction is where two streams meet. This seems to be a stream node. Also, what is the purpose of this class?

cwms-data-api/src/test/resources/cwms/cda/data/dto/stream-junction-identifier.json
```diff
@@ -0,0 +1,8 @@
+{
+  "stream-id": {
+    "office-id": "Office123",
+    "location-id": "Stream123"
```
rma-psmorris, line 4: these examples are confusing as I think these should be stream locations and not stream identifiers

rma-psmorris (commented):
I'd like to go over this code and JSON in a meeting with you.

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/Stream.java
rma-bryson: A Location isn't returned from the db like in other cases such as Embankments. I think we want to use the minimum number of queries here, and let the client make the Location query if needed. Unless you think that a fully-qualified Location is always going to be needed here.

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/Stream.java
rma-bryson: Removed these. This is addressed in this PR: https://github.com/USACE/cwms-data-api/pull/721

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/StreamJunctionIdentifier.java
rma-bryson: Renamed to StreamNode here: https://github.com/USACE/cwms-data-api/pull/721

cwms-data-api/src/test/resources/cwms/cda/data/dto/stream-junction-identifier.json
rma-bryson: Looking at this again, this is the CwmsIdentifier that represents the stream that this stream-node is on.

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/Stream.java
rma-psmorris: I think your reasoning is fine, end application can query the web service for the location physical data if it needs it

MikeNeilson: Is it common that the application *doesn't need it* or that it already has it.

While minimizing database queries is good, so is minimizing CDA queries. For CDA we de-normalize as appropriate (which maybe not here, hence the follow up question.)

PR #706, opened 2024-06-18
Lock and LockRevokerRights classes.

rma-rripken:
I'm including the dao here so that everything compiles.  Additional dao tests will be included in followup pr


cwms-data-api/src/main/java/cwms/cda/data/dto/project/Lock.java
```diff
@@ -1,7 +1,46 @@
...
+import cwms.cda.data.dto.CwmsDTO;
+import cwms.cda.formatters.Formats;
+import cwms.cda.formatters.annotations.FormattableWith;
+import cwms.cda.formatters.json.JsonV1;
+
+@JsonDeserialize(builder = Lock.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSON, formatter = JsonV1.class)
+public class Lock extends CwmsDTO {
```
adamkorynta: while the packages would be different, I think we'd want this class to be named `ProjectLock` to differentiate from the physical location `Lock` that will get added.

cwms-data-api/src/main/java/cwms/cda/data/dto/project/Id.java
```diff
@@ -0,0 +1,55 @@
...
+import cwms.cda.api.errors.FieldException;
+import cwms.cda.data.dto.CwmsDTOBase;
+import cwms.cda.formatters.Formats;
+import cwms.cda.formatters.annotations.FormattableWith;
+import cwms.cda.formatters.json.JsonV1;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSON, formatter = JsonV1.class)
+public class Id implements CwmsDTOBase {
```
adamkorynta: I don't see where this is referenced. What is the intended usage?

rma-rripken: Its a class to serialize the lock id back from a lock request.  {"id":"guid-thing-here"}

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
```diff
@@ -1,67 +1,99 @@
...
     }
 
-    public String requestLock(String projectId, String appId,
-                              boolean revokeExisting, int revokeTimeout, String office) {
+    public String requestLock(String office, String projectId, String appId,
+                              boolean revokeExisting, int revokeTimeout) {
         BigInteger revokeTimeoutBI = toBigInteger(revokeTimeout);
+        String pRevokeExisting = OracleTypeMap.formatBool(revokeExisting);
         return connectionResult(dsl,
                 c -> CWMS_PROJECT_PACKAGE.call_REQUEST_LOCK(getDslContext(c, office).configuration(),
```
adamkorynta, line 65: Need to discuss the usage of this procedure as it internally it grabs session state for the machine, osuser, and program: https://bitbucket.hecdev.net/projects/CWMS/repos/cwms_database/browse/schema/src/cwms/cwms_project_pkg_body.sql#910 to populate the lock data. REGI uses that to identify whether the current user and computer has the project locked: https://bitbucket.hecdev.net/projects/CL/repos/cwms-regi-tools/browse/regi-cache/regi-dao/src/main/java/usace/rowcps/regi/status/AtProjectLockManager.java#378

cwms-data-api/src/main/java/cwms/cda/data/dto/project/Id.java
adamkorynta: Can it get renamed to `LockId`? the packaging makes it seem like it is the project id.

adamkorynta: or `ProjectLockId` would be even better

cwms-data-api/src/main/java/cwms/cda/data/dto/project/Lock.java
rma-rripken: As you noted, its in a different package but I'm not opposed to renaming.  To be consistent do I rename the other classes so they are ProjectLock*** ?   It seemed to get out of hand.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
rma-rripken: I think you are saying you'd like the CDA user to be able to collect those themselves and pass it into the method call

adamkorynta: There needs to be some sort of client identification. REGI doesn't persist the project lock id on the client between sessions and there have been times when REGI isn't able to cleanup active locks before a shutdown (process force killed). When the happens, REGI reports a third lock state so that the user is made aware that they have to cleanup their locks. Currently, the client can't just use the JDBC username since CAC authentication uses the CWMS service account user for all access, so osuser and machine name are used instead.

MikeNeilson: The session user will be set to the authenticated user before the SQL call, if that helps.

MikeNeilson: But to be blunt, CDA shouldn't have to care about any distributed transactions... or it it does it needs to be a larger system that works with all the other endpoints (or at least could work time permitting.)

rma-rripken: Couldn't we duplicate/refactor the REQUEST_LOCK call into one that pulls details out of the session and then calls an inner proc that does the work.  If the CDA call provides all the details CDA could just call the inner call directly.  Then REGI could pass-in whatever it wants for machine-name or os-user.  Those fields got set in the env used by the REGI calls somehow - surely REGI can figure out how to query for them directly and pass them in?

cwms-data-api/src/main/java/cwms/cda/data/dto/project/Lock.java
rma-rripken: Renamed

cwms-data-api/src/main/java/cwms/cda/data/dto/project/Id.java
rma-rripken: Renamed

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
adamkorynta: That seems like the best solution to me. The way it currently coded, we'd be leaking the CDA server's machine name and osuser back to the client.

rma-rripken: So the extra fields would be required and CDA only calls the inner proc.   If its just machine-name and os-user that are the problem we could (probably) alternately just set those in the env to the cac-username and the CDA hostname

MikeNeilson: Thankfully there is only one host right now, but consider the future were there could be multiple instances of CDA talking to the database and the applications wouldn't know from where.

adamkorynta: Also, you'd need a `/whoami` endpoint for the client to determine who they are. The user provided apikey doesn't have that information and neither does the X509 certificate.

Also I think you still want the session information to always be correct so that audits in the database are accurate.

rma-rripken: The apikey doesn't contain the user but the CDA auth is able to map it to a specific db user.  At least that is how I thought it worked.  As long as getting a lock and listing the locks determine the username the same way does it matter if its the os-user or db-user acct?

adamkorynta: Yeah, the server knows who the client is, but client doesn't know who they are in order to compare against

MikeNeilson: Would it make sense for the client to generate a unique transaction id to track?

MikeNeilson: a UUID type 1, (with the mac address) or version 7 seem the most useful. Or type 3 or 5 to track the user. You wouldn't be able to use it but would be useful to see if one user has consistent usage.

In the case of 7,3, or 5, it would make the most sense to create from CDA, as it has access to the user... and the same clock; which seems pretty important.

MikeNeilson: So basically, client calls a generic `GET transaction` end point to get the UUID and then `DELETE transaction/uuid`.

I am focused on CDA behavior and possible future usage, so not sure how that tracks within the database but doesn't seem like a change that's overly large since you were already encoding information.

rma-rripken: I don't comprehend your transaction system suggestion.  Maybe it could work.  Would need more than just my input.  I can open an issue for the os-user and extra pieces.

rma-rripken: https://github.com/USACE/cwms-data-api/issues/711

rma-rripken:
@adamkorynta  @MikeNeilson  What other changes do you need to see in here?   The Dao issues will get re-litigated when Controller/Handler IT are further along.

adamkorynta:
I'll let Peter review the JSON/DTO, I don't have anymore comments on the rest.

adamkorynta (approved):

PR #712, opened 2024-06-19
Feature/cto 136 rights controller

rma-rripken:


cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockCatalogHandlerIT.java
```diff
@@ -0,0 +1,141 @@
...
+            String userName = user.getName();
+            String officeMask = OFFICE;
+            Project testProject = buildTestProject(OFFICE, projId);
+            prjDao.create(testProject);
+            try {
+                int revokeTimeout = 10;
+
+                lockDao.updateLockRevokerRights(OFFICE, userName, projId, appId + "_1", officeMask, true);
+                lockDao.updateLockRevokerRights(OFFICE, userName, projId, appId + "_2", officeMask, true);
+                String lock1 = lockDao.requestLock(OFFICE, projId, appId + "_1", true, revokeTimeout);
```
MikeNeilson: I see two locks that appear to need to be done/undone in the same order. For clarity is this testing testing the case of two different individuals attempting to lock or something else?

I ask because it implies that we should support asking for multiple objects within a single lock request to avoid race conditions.

MikeNeilson:
In the realm of naming is hard. I took me I think 5-10 minutes let my brain settle that this is an "Object lock" in the computer science sense and not the operations of a lock.

Any thoughts on a naming scheme we can use like maybe just "ObjectLock" or "DatabaseObjectLock<Type>" to make this more clear to the reader. After reviewing the initial Project code it dawned on me, as you saw in another comment, that it may be good to implement a more general distributed transaction system in the future, better names now would make that more practical.

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockCatalogHandlerIT.java
rma-rripken: I just wanted multiple locks so that the catalog type call would return more than one results.  In my test I made them have different applicationId - that was bad and unrealistic.  A more realistic test would have created multiple projects with the same applicationId.  REGI makes use of these Locks and LockRevokerRights.  I don't know of anything else that is using them but @adamkorynta  or @rma-psmorris might have more.

rma-rripken: As for ordering the acquires and releases to avoid deadlock.  CDA could implement bulk-acquire and bulk-release methods that internally did the ordering.

rma-rripken: I'm not sure how often REGI actually needs multiple locks - I think they are used when someone wants to do Daily Operations and they want to be sure no one changes the project while they are using it.  I don't know if they do that project by project or if they pick a batch of projects and do the whole batch.  And I think hard deadlocks aren't really a thing b/c people just revoke the locks and get them unless the person with the lock says no.  But I think they have to be right there using the app to say no.  So eventually the person wanting the lock will either get it or will have pestered someone to the point that they have coordinated who is doing what.  At least that is what I understand.   No possibility of endless spinning beach balls.

MikeNeilson: Sounds like a reasonable mitigation for now.

MikeNeilson: But a test should indicate and follow actual usage so I ask that the test be updated to reflect your comment above.

rma-rripken: Done - changed to make two projects as suggested.

adamkorynta: sorry I missed your earlier comment. Only REGI uses the Project locks currently.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
```diff
@@ -171,51 +184,76 @@ public void denyLockRevocation(String lockId) {
...
+     * @param projectMask     the project mask
      * @param applicationMask the application mask
-     * @param officeMask the office mask
-     * @param allow true to add to allow list, false to add to deny list
+     * @param userId          the user to add
+     * @param allow           true to add to allow list, false to add to deny list
      */
-    public void updateLockRevokerRights(String office, String userId, String projectMask,
-                                        String applicationMask, String officeMask, boolean allow) {
+    public void updateLockRevokerRights(String office, String officeMask, String projectMask,
```
rma-psmorris: Rename office to sessionOffice? Also applies to calls with similar naming/usage.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectDao.java
```diff
@@ -301,27 +316,44 @@ private static String buildTableQuery(@Nullable String projectIdMask, @Nullable
         return sql;
     }
 
+    /**
```
rma-psmorris, line 312: for the code right above this,
private static String buildTableQuery(@Nullable String projectIdMask, @Nullable String office,
                                          String cursorOffice, String cursorProjectId) {
In terms of clarity, you could swap the Strings to booleans to insure that the Strings are not used in the construction of the query?

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectDao.java
```diff
@@ -383,6 +450,12 @@ public static BigInteger toBigInteger(int revokeTimeout) {
     }
```
rma-psmorris: Might want to rename the int above, int revokeTimeout

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectDao.java
```diff
@@ -383,6 +450,12 @@ public static BigInteger toBigInteger(int revokeTimeout) {
...
 
 
+    /**
+     * Retrieves the locations associated with a project.
+     *
+     * @param office The office ID associated with the project.
+     * @return A list of Location objects representing the project's locations.
+     */
```
rma-psmorris, line 441: down below on: 418/491 builder.withTimeZoneName(ZoneId.of(timeZoneName)); Not all Oracle time zones map to Java time zones. @adamkorynta and I added in extra time zone coverage via Zone's map API in one of our APIs (nucleus?). That same business should occur with time zones in CDA.

rma-psmorris (changes requested):
Reviewed DAOs and DTO/JSON. Feedback in comments. Will follow-up with more feedback as needed.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -149,31 +161,33 @@
...
+    "/forecast-spec/*",
+    "/forecast-instance/*",
+    "/standard-text-id/*",
+    "/projects/*",
+    "/project-locks/*",
+    "/project-lock-rights/*",
+    "/properties/*",
+    "/lookup-types/*",
+    "/embankments/*",
         "/turbines/*"
```
MikeNeilson: You know, maybe we should just change this to * and have javalin also sort out the static files, this thing is getting a smidge long.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
```diff
@@ -147,12 +141,26 @@ private List<ProjectLock> catLocks(String projMask, String appMask, TimeZone tz,
...
-    public void releaseLock(String lockId) {
+    /**
+     * Releases a lock for a given office and lock ID.
+     *
+     * @param office  the office to be used for the session.
+     * @param lockId  the ID of the lock to release
+     */
+    public void releaseLock(String office, String lockId) {
         connection(dsl, c -> CWMS_PROJECT_PACKAGE.call_RELEASE_LOCK(
-                DSL.using(c, SQLDialect.ORACLE18C).configuration(),
```
MikeNeilson, line 153: consider if it makes sense to suppress the line length warning here. Alternatively nesting the calls in a way that makes it faster to spot which items are an argument to what would be nice.

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockCatalogHandlerIT.java
```diff
@@ -0,0 +1,139 @@
...
+public class ProjectLockCatalogHandlerIT extends DataApiTestIT {
+
+    public static final String OFFICE = "SPK";
+
+
+    @Test
+    void test_cat_locks() throws SQLException {
+
+        CwmsDatabaseContainer<?> databaseLink = CwmsDataApiSetupCallback.getDatabaseLink();
+        databaseLink.connection(c -> {
```
MikeNeilson: Use the CwmsDataApiSetupCallback.getWebUser as the second parameter to the databaseLink.connection method.

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockDenyHandlerIT.java
```diff
@@ -0,0 +1,162 @@
...
+
+                assertFalse(lockId[0].isEmpty());
+                // now user1 has a lock.
+            });
+
+
+            // have user2 try to get the lock on another thread.
+            new Thread(() -> {
+                try {
+                    databaseLink.connection(c -> {
```
MikeNeilson: webUser

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockDenyHandlerIT.java
```diff
@@ -0,0 +1,162 @@
...
+
+                        long after = System.currentTimeMillis();
+                        long elapsed = after - before;
+                        assertFalse(Math.abs((1000 * revokeTimeout) - elapsed) < 1000);
+
+                    });
+                } catch (SQLException e) {
+                    throw new RuntimeException(e);
+                }
+            }).start();
```
MikeNeilson: So it *probably* would never matter, but on the change this hangs, this thread should be set to Daemon before start.

That said if we're going to consider more threading in these tests we should probably create an executor service pool with a factory that does that.

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockDenyHandlerIT.java
```diff
@@ -0,0 +1,162 @@
...
+                    .redirects().max(3)
+                    .post("/project-locks/deny")
+                .then()
+                    .log().ifValidationFails(LogDetail.ALL, true)
+                    .assertThat()
+                    .statusCode(is(HttpServletResponse.SC_OK));
+
+        } finally {
+            // clean up
+            databaseLink.connection(c -> {
```
MikeNeilson: webuser

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockReleaseHandlerIT.java
```diff
@@ -0,0 +1,124 @@
...
+public class ProjectLockReleaseHandlerIT extends DataApiTestIT {
+    private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+
+    public static final String OFFICE = "SPK";
+
+
+    @Test
+    void test_release() throws SQLException {
+        CwmsDatabaseContainer<?> databaseLink = CwmsDataApiSetupCallback.getDatabaseLink();
+        databaseLink.connection(c -> {
```
MikeNeilson: webuser

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockRequestHandlerIT.java
```diff
@@ -0,0 +1,250 @@
...
+    public static final String OFFICE = "SPK";
+
+
+    @Test
+    void test_request() throws SQLException {
+        TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
+            String projId = "lockReq";
+            String appId = "test_req";
+        CwmsDatabaseContainer<?> databaseLink = CwmsDataApiSetupCallback.getDatabaseLink();
+        databaseLink.connection(c -> {
```
MikeNeilson: webuser

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockRequestHandlerIT.java
```diff
@@ -0,0 +1,250 @@
...
+        }
+    }
+
+    @Test
+    void test_request_user2() throws SQLException {
+        TestAccounts.KeyUser user2 = TestAccounts.KeyUser.SPK_NORMAL2;
+        String projId = "lockReq2";
+        String appId = "test_req2";
+        CwmsDatabaseContainer<?> databaseLink = CwmsDataApiSetupCallback.getDatabaseLink();
+        databaseLink.connection(c -> {
```
MikeNeilson: getWebUser

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockRequestHandlerIT.java
```diff
@@ -0,0 +1,250 @@
...
+        String appId = "revoke_test";
+        String officeMask = OFFICE;
+        TestAccounts.KeyUser user1 = TestAccounts.KeyUser.SPK_NORMAL;
+        TestAccounts.KeyUser user2 = TestAccounts.KeyUser.SPK_NORMAL2;
+        int revokeTimeout = 4;
+        boolean dontRevoke = false;
+
+        try {
+            // setup
+            databaseLink.connection(c -> {
```
MikeNeilson: getWebUser

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockRequestHandlerIT.java
```diff
@@ -0,0 +1,250 @@
...
+            lockId2[0] = response.path("id");
+            assertNotNull(lockId2[0]);
+
+            long after = System.currentTimeMillis();
+            double elapsed = (double) (after - before);
+            assertThat(elapsed, closeTo(1000 * revokeTimeout, 2000));
+
+        } finally {
+            // clean up
+            databaseLink.connection(c -> {
```
MikeNeilson: getWebUser

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockStatusHandlerIT.java
```diff
@@ -0,0 +1,134 @@
...
+
+@Tag("integration")
+public class ProjectLockStatusHandlerIT extends DataApiTestIT {
+
+    public static final String OFFICE = "SPK";
+
+    @Test
+    void test_lock_status() throws SQLException {
+        CwmsDatabaseContainer<?> databaseLink = CwmsDataApiSetupCallback.getDatabaseLink();
+        databaseLink.connection(c -> {
```
MikeNeilson: getWebUser

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectPublishStatusUpdateHandlerIT.java
```diff
@@ -0,0 +1,128 @@
...
+        CwmsDatabaseContainer<?> databaseLink = CwmsDataApiSetupCallback.getDatabaseLink();
+
+        String projId = "pub_status";
+        String appId = "pub_test";
+
+        TestAccounts.KeyUser user1 = TestAccounts.KeyUser.SPK_NORMAL;
+
+        try {
+            // setup
+            databaseLink.connection(c -> {
```
MikeNeilson: getWebUser

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectPublishStatusUpdateHandlerIT.java
```diff
@@ -0,0 +1,128 @@
...
+                        .body("value", is(greaterThan(before)))
+                        .body("value", is(lessThan(Instant.now().toEpochMilli())))
+                    .statusCode(is(HttpServletResponse.SC_OK));
+
+            });
+
+
+        } finally {
+            // clean up
+            databaseLink.connection(c -> {
```
MikeNeilson: getWebUser

cwms-data-api/src/test/java/cwms/cda/api/project/RemoveAllLockRevokerRightsHandlerIT.java
```diff
@@ -0,0 +1,143 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag("integration")
+public class RemoveAllLockRevokerRightsHandlerIT extends DataApiTestIT {
+
+    public static final String OFFICE = "SPK";
+
+    @Test
+    void test_removeAll_rights() throws SQLException {
```
MikeNeilson, line 86: getWebUser

cwms-data-api/src/test/resources/cwms/cda/data/sql/users.sql
```diff
@@ -30,6 +54,23 @@ begin
...
     cwms_sec.add_user_to_group('l2hectest','TS ID Creator','SPK');
 
+    begin
+        cwms_sec.create_user('l2hectest_vt','l2hectestvt',NULL,'SPK');
+    exception
+        when dup_val_on_index then null; -- user already exists
+    end;
+    cwms_sec.add_cwms_user('l2hectest_vt',NULL,'SPK');
+    cwms_sec.update_edipi('l2hectest_vt',2345678901);
+    cwms_sec.add_user_to_group('l2hectest_vt','All Users', 'HQ');
```
MikeNeilson, line 68: Probably doesn't need HQ. But I don't see any harm right now either.

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectDao.java
adamkorynta: It's in OracleTypeMap::toZoneId:

https://bitbucket.hecdev.net/projects/CWMS/repos/hec-cwms-data-access/browse/cwms-db-dao/src/main/java/usace/cwms/db/dao/util/OracleTypeMap.java#397

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
rma-rripken: done

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectDao.java
rma-rripken: done

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectDao.java
rma-rripken: inlined the two usages.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectDao.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockDenyHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockCatalogHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockDenyHandlerIT.java
rma-rripken: I set the thread to daemon.

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockDenyHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockStatusHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectPublishStatusUpdateHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectPublishStatusUpdateHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/RemoveAllLockRevokerRightsHandlerIT.java
rma-rripken: done

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
rma-rripken: I pulled the configuartion arg into its own variable and adjusted the indents

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockDenyHandlerIT.java
rma-rripken: And added a timeout parameter

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectDao.java
rma-rripken: done

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectDao.java
rma-rripken: inlined the two usages.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
rma-rripken: It doesn't bother me but if we're going to do that lets make it a standalone pr.

MikeNeilson: Agreed. Was mostly just commenting on just how long that list is getting.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectLockRequest.java
```diff
@@ -0,0 +1,147 @@
...
+                ProjectLockId id = new ProjectLockId(lockId);
+                String acceptHeader = ctx.header(Header.ACCEPT);
+                ContentType responseType = Formats.parseHeader(acceptHeader, ProjectLockId.class);
+                String result = Formats.format(responseType, id);
+                ctx.result(result);
+                ctx.contentType(responseType.toString());
+                requestResultSize.update(result.length());
+                ctx.status(HttpServletResponse.SC_CREATED);
+            } else {
+                // We don't have any idea exactly why the create failed - right?
```
adamkorynta: Create fail doesn't throw an error? Probably needs a Jira ticket to correct that behavior

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectLockStatus.java
```diff
@@ -0,0 +1,119 @@
...
+        String prjId = requiredParam(ctx, PROJECT_ID);
+        String appId = requiredParam(ctx, APPLICATION_ID);
+        String office = requiredParam(ctx, OFFICE);
+
+        try (final Timer.Context ignored = markAndTime(GET_ONE)) {
+            ProjectLockDao lockDao = new ProjectLockDao(JooqDao.getDslContext(ctx));
+            boolean locked = lockDao.isLocked(office, prjId, appId);
+            if (locked) {
+                // should we call catLocks instead and get the full lock
+                // with acquireTime and other fields?
```
adamkorynta: Yes, catLocks is the getAll and this is the getOne so it shouldn't return less information

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectLockStatus.java
```diff
@@ -0,0 +1,119 @@
...
+                @OpenApiResponse(status = STATUS_404, description = "Not matching Lock was found.")
+            },
+            tags = {TAGS},
+            path = PATH,
+            method = HttpMethod.GET
+    )
+    @Override
+    public void handle(@NotNull Context ctx) throws Exception {
+
+        String prjId = requiredParam(ctx, PROJECT_ID);
```
adamkorynta: requiredParam checks for a query parameter right? shouldn't project-id be a path parameter?

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectPublishStatusUpdate.java
```diff
@@ -0,0 +1,130 @@
...
+            String office = requiredParam(ctx, OFFICE);
+            String appId = requiredParam(ctx, APPLICATION_ID);
+            String projectId = requiredParam(ctx, PROJECT_ID);
+
+            String sourceId = ctx.queryParam(SOURCE_ID);
+            String tsId = ctx.queryParam(TIMESERIES_ID);
+            Instant begin = queryParamAsInstant(ctx, BEGIN);
+            Instant end = queryParamAsInstant(ctx, END);
+
+            Instant timeOfMessage = prjDao.publishStatusUpdate(
```
adamkorynta: hec-cwms-data-access returns a Date, but the actual usage of the method doesn't use it: https://bitbucket.hecdev.net/projects/CL/repos/cwms-regi-tools/browse/regi-cache/regi-dao/src/main/java/usace/rowcps/regi/status/AtProjectStatusManager.java#72,89,94

I think we can throw that return away.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectPublishStatusUpdate.java
```diff
@@ -0,0 +1,130 @@
...
+                    )},
+            tags = {TAG},
+            path = PATH
+    )
+    @Override
+    public void handle(@NotNull Context ctx) throws Exception {
+
+        try (final Timer.Context ignored = markAndTime("publish")) {
+            ProjectDao prjDao = new ProjectDao(JooqDao.getDslContext(ctx));
+            String office = requiredParam(ctx, OFFICE);
```
adamkorynta, line 105: Query parameters instead of a content body because the status message isn't a representation of a data object in CWMS?

cwms-data-api/src/main/java/cwms/cda/api/project/UpdateLockRevokerRights.java
```diff
@@ -0,0 +1,99 @@
...
+        this.metrics = metrics;
+
+    }
+
+    @OpenApi(
+            description = "Update Lock Revoker Rights.",
+            queryParams = {
+                @OpenApiParam(name = OFFICE, required = true, description =
+                        "Specifies the session office."),
+                @OpenApiParam(name = OFFICE_MASK, description =
```
adamkorynta: I can't think of a reason we'd want to provide an office mask, I think just having a required office makes the api clearer.

adamkorynta: Looking at REGI's usage, I'm thinking we don't need any masks: https://bitbucket.hecdev.net/projects/CL/repos/cwms-regi-tools/browse/regi-cache/regi-dao/src/main/java/usace/rowcps/regi/status/AtProjectLockManager.java#295,463,466,477,479

It's also a little strange that the only call we make to this method is to set it to true right before we call the method to revoke. Almost makes me feel like we don't need this endpoint or revoker rights at all?

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
```diff
@@ -57,73 +56,68 @@ public ProjectLockDao(DSLContext dsl) {
...
+     * @param projMask the project mask
+     * @param appMask the application mask
+     * @return the list of project locks
+     */
+    public List<ProjectLock> catLocks(String officeMask, String projMask, String appMask) {
+        return catLocks(officeMask, projMask, appMask, TimeZone.getTimeZone("UTC"));
     }
 
-    private List<ProjectLock> catLocks(String projMask, String appMask, TimeZone tz, String officeMask) {
+    private List<ProjectLock> catLocks(String officeMask, String projMask, String appMask, TimeZone tz) {
```
adamkorynta: will this timezone ever be anything but "UTC"?

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
```diff
@@ -57,73 +56,68 @@ public ProjectLockDao(DSLContext dsl) {
...
+    /**
+     * Returns the list of project locks based on the given office mask, project mask, and application mask.
+     *
+     * @param officeMask the office mask
+     * @param projMask the project mask
+     * @param appMask the application mask
+     * @return the list of project locks
+     */
+    public List<ProjectLock> catLocks(String officeMask, String projMask, String appMask) {
+        return catLocks(officeMask, projMask, appMask, TimeZone.getTimeZone("UTC"));
```
adamkorynta: creating a TimeZone object from a String, just to pull out the String immediately after

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
```diff
@@ -147,80 +141,129 @@ private List<ProjectLock> catLocks(String projMask, String appMask, TimeZone tz,
...
+     * Denies the revocation of a lock.
+     *
+     * @param lockId the ID of the lock to deny revocation for
+     */
     public void denyLockRevocation(String lockId) {
-        connection(dsl, c ->
-                CWMS_PROJECT_PACKAGE.call_DENY_LOCK_REVOCATION(
-                        DSL.using(c, SQLDialect.ORACLE18C).configuration(),
-                        lockId));
+        connection(dsl, c -> {
```
adamkorynta, line 201: Is the office session id not needed here?

cwms-data-api/src/test/java/cwms/cda/api/project/LockRevokerRightsCatalogHandlerIT.java
```diff
@@ -0,0 +1,144 @@
...
+            DSLContext dsl = getDslContext(c, OFFICE);
+            ProjectLockDao lockDao = new ProjectLockDao(dsl);
+            ProjectDao prjDao = new ProjectDao(dsl);
+
+            String projId = "catRightsIT";
+            String appId = "test_catRights";
+            String officeMask = OFFICE;
+
+            Project testProject = buildTestProject(OFFICE, projId);
+            prjDao.create(testProject);
```
adamkorynta: to make the test easier to maintain, recommend move this to a @BeforeEach method and the cleanup in the finally to an @AfterEach

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockCatalogHandlerIT.java
```diff
@@ -0,0 +1,139 @@
...
+            String projId1 = "catLocks1";
+            String projId2 = "catLocks2";
+            String appId = "catLocks_test";
+            TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
+            String userName = user.getName();
+            String officeMask = OFFICE;
+            Project testProject1 = buildTestProject(OFFICE, projId1);
+            prjDao.create(testProject1);
+            Project testProject2 = buildTestProject(OFFICE, projId2);
+            prjDao.create(testProject2);
```
adamkorynta: If moving setup and cleanup to @Before/AfterEach, I think you could extract a test base class and not need the duplicate code.

cwms-data-api/src/test/java/fixtures/TestAccounts.java
```diff
@@ -9,6 +33,7 @@ public enum KeyUser {
         NONE("none",null,null,null,null, null), // Used for annotations
         GUEST("guest",null,null,null, null, null), // USED as marker label for tests
         SPK_NORMAL("l2hectest","l2hectest","1234567890","l2userkey","ATotallyRandomStringL2hectest","SPK", "CWMS Users", "cac_user"),
+        SPK_NORMAL2("l2hectest_vt","l2hectestvt","2345678901","l2userkey2","DiffrntStringL2hectest_vt","SPK", "CWMS Users", "cac_user"),
```
adamkorynta, line 36: why the VT user? or does that stand for something different in this context than the old validate transform pieces that have been removed from CWMS?

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectLockStatus.java
rma-rripken: This was isLocked and it returned just true/false.  I think I talked to you about how that could just be in the http codes and we didn't need a response body.    Having this be getOne with path parameter and returning a singular lock seems valid too.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectPublishStatusUpdate.java
rma-rripken: I don't see any reason to throw it away. We have it, its easy to return and it gives me something to observe in (minimal) test case.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectPublishStatusUpdate.java
rma-rripken: I'll move projectId into the path.  When I wrote it I was thinking of it as a utility that didn't belong to anything but keying off the project makes sense.

cwms-data-api/src/main/java/cwms/cda/api/project/UpdateLockRevokerRights.java
rma-rripken: Does REGI ever pass false into the update lock revoker rights?

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockReleaseHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockRequestHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockRequestHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockRequestHandlerIT.java
rma-rripken: done

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockRequestHandlerIT.java
rma-rripken: done

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
rma-rripken: The test worked without it, do you think it's needed?

MikeNeilson: I would probably fail during the the integration test that uses the HQ database. That's where a lot of the need for the call to getWebUser came from as that's a user that gets grant extra permissions to make sure test data can get loaded.

So I likely isn't required in *your* test database, but it's most likely required in someone else's test database.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectLockStatus.java
adamkorynta: yeah I think if it was just http codes that would be ok, but since we're doing a response body it should probably be the full state

cwms-data-api/src/main/java/cwms/cda/api/project/UpdateLockRevokerRights.java
adamkorynta: nope, only true and only that single usage of the method. Our integration tests do test to make sure the pieces make sense.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectPublishStatusUpdate.java
adamkorynta: I wasn't so much commenting on the path params vs query params, but was seeing a POST with no body jumped out to me.

cwms-data-api/src/main/java/cwms/cda/api/project/UpdateLockRevokerRights.java
MikeNeilson: Could another usage than REGI benefit from the option to pass false?

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectPublishStatusUpdate.java
adamkorynta: Yeah the only way to observe would be to query the message queues or subscribe to them. Seems like managing the DTO in the java server, java client, javascript client is extra maintenance for something unused. 

If we want to keep it, the DTO should be renamed to something more generic as there are other message calls that we'll end up wanting in the future (unless we start allowing clients to publish to the message broker topics directly). For example: https://bitbucket.hecdev.net/projects/CWMS/repos/hec-cwms-data-access/browse/cwms-db-dao/src/main/java/usace/cwms/db/dao/ifc/msg/CwmsDbMsg.java#96

cwms-data-api/src/test/java/fixtures/TestAccounts.java
rma-rripken: It needs to be a user that is already setup.  In consultation with Peter we went through the users that could possible work and decided vt would be fine.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectPublishStatusUpdate.java
rma-rripken: oh, it might not be typical but I think its fine.  The method is just a trigger.  post makes it clear that it has side effects and should prevent proxies and caches from messing with it.

cwms-data-api/src/test/java/fixtures/TestAccounts.java
MikeNeilson: Yeah, it's odd looking, but it does exist at schema install. I don't really see us removing it from the schema installer either as it's work for no real gain... especially now that someone is using it.

adamkorynta: makes sense, I didn't realize it was still in the schema install

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectPublishStatusUpdate.java
adamkorynta: makes sense to me!

cwms-data-api/src/test/java/fixtures/TestAccounts.java
MikeNeilson: Though l2hectest_vt is only in the schema install if you use SPK as the office, so regardless the user should be added to the system setup for *other* database setups.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectLockStatus.java
rma-rripken: I changed this to GetOne and made project-id a path parameter

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
rma-rripken: Removed the TimeZone arg and the other method.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectLockDao.java
rma-rripken: No

cwms-data-api/src/test/java/cwms/cda/api/project/LockRevokerRightsCatalogHandlerIT.java
rma-rripken: I added  2beforeeach and @aftereach

cwms-data-api/src/test/java/cwms/cda/api/project/ProjectLockCatalogHandlerIT.java
rma-rripken: I added a @beforeeach and @aftereach.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -513,6 +532,33 @@ protected void configureRoutes() {
                 new PropertyController(metrics), requiredRoles,1, TimeUnit.DAYS);
         cdaCrudCache(format("/lookup-types/{%s}", Controllers.NAME),
                 new LookupTypeController(metrics), requiredRoles,1, TimeUnit.DAYS);
+        cdaCrudCache(format("/embankments/{%s}", Controllers.NAME),
```
adamkorynta: embankments and turbine are added above, probably a merge error here

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -546,7 +535,9 @@ private void addProjectLocksHandlers(String path, RouteRole[] requiredRoles) {
         String pathWithoutResource = path.replace(getResourceId(path), "");
 
         get(path, new ProjectLockGetOne(metrics));
+        addCacheControl(path, 5, TimeUnit.MINUTES);
```
adamkorynta: I'm not sold on caching project locks. Seems like you always want the current state for this workflow

rma-psmorris (approved):
I have no feedback on this PR, approved.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectLockRequest.java
rma-rripken: It looks like the pl/sql does throw exceptions.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectPublishStatusUpdate.java
rma-rripken: What do you think about PublishStatusUpdateResult being renamed to MessageTimestamp?  You asked for it to be more general so it could be used for the other Msg calls.  Or EventTimestamp?  Currently it just holds a single Instant so its not really related to Message at all.  I'd call it Instant or Timestamp but I know that would cause all sorts of confusion.  Its meant to just hold the value and as a place we can hang our annotations.  I suppose from that point of view it is related to Message b/c we're using it from Message and configuring it for Message.   So, MessageTimestamp?  MessageInstant?

adamkorynta: I've seen examples like: https://github.com/USACE/cwms-data-api/blob/develop/cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java#L237 where we return a 204 with a simple status message. Can the timestamp be that message and not need a full DTO?

adamkorynta: or it could be a header sent back to the client `X-Message-Timestamp` or `X-Message-Instant`

rma-rripken: At the meeting it was decided to remove the timestamp.

cwms-data-api/src/main/java/cwms/cda/api/project/UpdateLockRevokerRights.java
adamkorynta: I don't want Mike to revoke any of my locks, set them to false.... but then I'd also need to make sure Mike doesn't have permission to reset that to true.....

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectLockRequest.java
adamkorynta: so the `lockId` won't ever come back as null?

adamkorynta (changes requested):
still some comments to address

rripken:
> still some comments to address

Can you be more specific about what you want addressed?  github won't even show me all the comments on one page anymore.

cwms-data-api/src/main/java/cwms/cda/api/project/UpdateLockRevokerRights.java
rripken: I don't understand what you are requesting here

adamkorynta: I don't think this endpoint should include masks for the office given there is also an office parameter. I don't know of a use case where project or application masks are needed, but at least those don't clash.

adamkorynta:
> > still some comments to address
> 
> Can you be more specific about what you want addressed? github won't even show me all the comments on one page anymore.

too bad GitHub doesn't list out comments like BitBucket does....

The remaining two comments were:

- the strange null lockId returned by the dao during ProjectLockRequest. if it throws an error if the user can't obtain the lock, why is there null handling?
- UpdateLocKRevokerRights takes in an office and an office mask. I would remove the office mask as I see no use-case.

rma-rripken:
There is null handling because the pl/sql documentation states that it can return null if the lock request failed.  I don't understand what the issue is.   From a web user perspective they either got the lock or they didn't, they don't have to worry about the null handling just the response code.  You want the Dao to do the null handling itself and turn the null into an exception so that the controller can catch it?

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectLockRequest.java
rma-rripken: No, I didn't say that. pl/sql looks like it does return null if it thinks it got into a race condition.

adamkorynta: The database should throw an exception if that happens so we should create a db ticket.

adamkorynta (dismissed):

MikeNeilson (dismissed):
I don't see anything glaring, though it's rather long so I just skimmed it and will trust Adam's judgement.

adamkorynta (approved):

PR #713, opened 2024-06-19
update embankments to use locationidentifier

adamkorynta:
The changes enhance EmbankmentDao by updating project location and office reference handling. The use of a LocationIdentifier improves code clarity and maintainability. These improvements are also propagated to associated integration and unit tests

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/Embankment.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/Embankment.java
@@ -25,12 +25,14 @@
 package cwms.cda.data.dto.location.kind;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import cwms.cda.api.errors.FieldException;
 import cwms.cda.data.dto.CwmsDTOBase;
 import cwms.cda.data.dto.Location;
+import cwms.cda.data.dto.LocationIdentifier;
 import cwms.cda.data.dto.LookupType;
 import cwms.cda.formatters.Formats;
 import cwms.cda.formatters.annotations.FormattableWith;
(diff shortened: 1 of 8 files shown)


rma-rripken (approved):
I'm sensing AI in the commit message...

adamkorynta:
lol, tried it out. I had to edit though because it was really exaggerating my change "vastly improves maintainability" and "refactors for robust testing"

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/Embankment.java
```diff
@@ -42,7 +44,10 @@
 @JsonDeserialize(builder = Embankment.Builder.class)
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@JsonPropertyOrder({"projectIdentifier", "location"})
```
rma-psmorris, line 47: Is it possible to set the order for "this"?

rma-rripken: Meaning an annotation that adjusts the property order for Embankment DTO? I think your choices are code order, explicitly specified order or alphabetic
Or you want a way to adjust the order on an individual property? I think you can set an index # inside @JsonProperty annotations.  Not sure how it deals with ties.

Or what do you mean by 'this'?

rma-psmorris: The attribute data that the embankment itself supplies. For example:
upstreamSideSlope
downstreamSideSlope
structureLength
Although from the javadoc, it implies that projectIdentifier is first, then location, then the "this" or remainder of embankment attributes.
https://fasterxml.github.io/jackson-annotations/javadoc/2.8/com/fasterxml/jackson/annotation/JsonPropertyOrder.html
I asked this question because in https://github.com/USACE/cwms-data-api/blob/14c96c84812ab340f3dda8897fa15282199fcb7a/cwms-data-api/src/test/resources/cwms/cda/data/dto/location/kind/embankment.json the order is 
project identifier,
"this",
location.

adamkorynta: The rest of the attributes are always last and deserialization doesn't matter. I should have moved the location block up since it serves as an example.

Now that I think bout it though, I'm not as worried about the JSON ordering as the OpenAPI Schema ordering and I don't know how that gets generated. I'll have to take a look into that.

PR #715, opened 2024-06-19
Handle unsupported format exception instead of just formattingexception

adamkorynta:
FormattingException reports an error code 501 to the user, but 406 is more appropriate given it is the client's request that is in error. Updated embankment test to verify various format responses. Introduced new UnsupportedFormatException class to differentiate from a general FormattingException. Embankment serialization and API endpoints now support JSONv1 and default formats.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
@@ -79,7 +79,8 @@ public final class EmbankmentController  implements CrudHandler {
             },
             responses = {
                     @OpenApiResponse(status = STATUS_200, content = {
-                            @OpenApiContent(isArray = true, type = Formats.JSONV1, from = Embankment.class)
+                            @OpenApiContent(isArray = true, type = Formats.JSONV1, from = Embankment.class),
+                            @OpenApiContent(isArray = true, type = Formats.JSON, from = Embankment.class)
                     })
             },
             description = "Returns matching CWMS Embankment Data for a Reservoir Project.",
@@ -116,7 +117,8 @@ public final class EmbankmentController  implements CrudHandler {
             responses = {
                     @OpenApiResponse(status = STATUS_200,
                             content = {
-                                    @OpenApiContent(type = Formats.JSONV1, from = Embankment.class)
+                                    @OpenApiContent(isArray = true, type = Formats.JSONV1, from = Embankment.class),
+                                    @OpenApiContent(isArray = true, type = Formats.JSON, from = Embankment.class)
                             })
             },
             description = "Returns CWMS Embankment Data",
--- a/cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
@@ -79,7 +79,7 @@ public final class EmbankmentController  implements CrudHandler {
             },
             responses = {
                     @OpenApiResponse(status = STATUS_200, content = {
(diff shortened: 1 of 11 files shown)


rma-rripken:
It makes it cleaner to have the DTO do their own validation.  My hesitation is they don't know all the details about how they are going to be used.    There could be different rules for Create vs Update.  The main reason we did validation checks in some of the controllers was to try to provide better error messages - without those validation checks the lower layers still error out but the user gets back 500.   I never got the purpose of the validate method explained to me so its possible this is exactly what Mike envisioned.  @MikeNeilson can you weigh in b/c I think a whole bunch of DTO are gonna get validation rules added to them

rma-rripken:
The end-result is definitely cleaner.  I have to try pretty hard to think of a scenario where this isn't what we'd want to do.

cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
```diff
@@ -79,7 +79,7 @@ private Timer.Context markAndTime(String subject) {
             },
             responses = {
                     @OpenApiResponse(status = STATUS_200, content = {
-                            @OpenApiContent(isArray = true, type = Formats.JSON, from = Embankment.class)
```
MikeNeilson, line 82: I think you want to leave the Formats.JSON here. That will allow it to show up in the drop down properly.

cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
```diff
@@ -143,7 +143,7 @@ public void getOne(Context ctx, String name) {
     @OpenApi(
             requestBody = @OpenApiRequestBody(
                     content = {
-                            @OpenApiContent(from = Embankment.class, type = Formats.JSON)
```
MikeNeilson, line 146: Same comment as above.

cwms-data-api/src/main/java/cwms/cda/data/dto/Blob.java
```diff
@@ -59,5 +60,15 @@ public String toString(){
 
 	@Override
 	public void validate() throws FieldException {
+		if (getOfficeId() == null) {
+			throw new FieldException("An officeId is required when creating a blob");
```
MikeNeilson, line 64: Is it me, or should this constructor take a field name.

... or maybe a builder that allows all possible details to be entered before throwing.

Nothing for you to fix, just pontificating.

cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
rma-rripken: Good point.  When there are aliases we should probably put the more general in the combo box.  We could have both in there.

cwms-data-api/src/main/java/cwms/cda/data/dto/Blob.java
rma-rripken: we may eventually get annoyed enough to come up with an annotation or cleaner way of doing this

cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
adamkorynta: added

cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
adamkorynta: added

rma-rripken (approved):

PR #716, opened 2024-06-19
Bugfix/content type parameters

### LINKED ISSUE #497: Update endpoint: Parameters
Add deprecated header

RyanM-RMA:
Adding aliases, DTO, DTO parsing logic, integration and unit tests for Parameters
Fixes #497


rma-rripken:
I don't love Parameters.   Is Parameters.officeId meaningful?   I'd say just serialize back a List<Parameters> until we write a paged class.

RyanM-RMA:
do we have a way to deserialize a String to a List<Parameter>?  That's the entire reason Parameters exists, but I didn't see any way to do that in Formats.  I might just be blind...lol...

rma-rripken:
Formats has a way to write out a List but I don't think there is a way to read one in.   We didn't previously have everything going thru Formats so it doesn't surprise me that the method you want isn't there. 

You can make jackson do that sort of thing via code like: 
```
ObjectMapper om = JsonV2.buildObjectMapper();
List<ProjectLock> lock = om.readValue(input, new TypeReference <List<ProjectLock>>(){});
```

If Parameters.officeId doesn't make sense - maybe just make Parameters implement the interface instead of extending Base and then the officeId field would go away.   If the officeId field is important than your pr is perfect and we should just merge it.

RyanM-RMA:
I'll just remove the Parameters, and just return the serialized list like you're saying.  Parameters.officeId is pretty useless, since it's the requested office id, I think I just wanted to keep track of it.  We can handle paging later.

cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
rma-rripken, line 1: This probably should be in its own pr

RyanM-RMA: I can do that.  I figured "why not since I'm here", lol.  give me a bit

cwms-data-api/src/test/resources/cwms/cda/data/dto/Parameters.json
rma-rripken, line 1: This probably goes away -right?

cwms-data-api/src/main/java/cwms/cda/data/dao/ParameterDao.java
rma-rripken, line 1: I like the map with method reference - makes the jooq clean to read.

cwms-data-api/src/test/resources/cwms/cda/data/dto/Parameters.json
RyanM-RMA: What goes away?  there's no file here...no...

cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
RyanM-RMA: Removed, I'll get a PR going soon

cwms-data-api/src/test/resources/cwms/cda/data/dto/Parameters.json
rma-rripken, line 1: Just noticed the abbreviations here.  "param" "desc".  Did the legacy format have similar abbreviations?  Not saying its wrong, just thinking about it.

RyanM-RMA: Those were not abbreviated - good catch.  It was name and description.  I'll fix that

cwms-data-api/src/main/java/cwms/cda/data/dao/ParameterDao.java
RyanM-RMA: Thanks!  I thought so too.  I like nice short methods, and this is a nice clean way to do it.

cwms-data-api/src/test/resources/cwms/cda/data/dto/Parameters.json
RyanM-RMA: Updated

rma-rripken (approved):

PR #718, opened 2024-06-19
Adding aliases, and Unit DTO, and tests

### LINKED ISSUE #498: Update endpoint: Units
Add deprecated header

RyanM-RMA:
Starting work on aliases and integration tests for Unit endpoint

Fixes #498

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/UnitsDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/UnitsDao.java
@@ -2,10 +2,16 @@ package cwms.cda.data.dao;
 
 import cwms.cda.data.dto.Unit;
 import org.jooq.DSLContext;
+import org.jooq.Record;
 import usace.cwms.db.jooq.codegen.packages.CWMS_CAT_PACKAGE;
+import usace.cwms.db.jooq.codegen.tables.AV_UNIT;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.jooq.impl.DSL.table;
 
 public class UnitsDao extends JooqDao<String> {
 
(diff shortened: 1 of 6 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dto/Unit.java
rma-rripken, line 1: my personal pref is that the DTO should make it possible for alternateNames list to be null.    I've also been trying to build immutable style DTO with final fields and Builder.  Would be nice for Unit as that seems like a class we might want to cache.

rma-rripken (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/UnitsDao.java
```diff
@@ -12,4 +22,31 @@ public UnitsDao(DSLContext dsl) {
     public String getUnits(String format) {
         return CWMS_CAT_PACKAGE.call_RETRIEVE_UNITS_F(dsl.configuration(), format);
     }
+
+    public List<Unit> getUnits()
+    {
+        Map<Long, List<String>> unitIdToAliasMap = dsl.select()
```
rma-psmorris, line 28: select cu.unit_id, cu.long_name, cu.unit_system, cu.description, cap.abstract_param_id, atu.alias_id
from cwms_20.cwms_unit cu
inner join cwms_20.at_unit_alias atu on (cu.unit_code = atu.unit_code and (db_office_code = 53 /*or db_office_code = SWT*/))
inner join cwms_20.cwms_abstract_parameter cap on (cu.abstract_param_code = cap.abstract_param_code)

rma-psmorris: In my opinion, given AT_UNIT_ALIAS supports office specific aliases, that this query and the unit object should include an office_id field (for aliases). Correspondingly, you could limit the query to only include data from CWMS (53), or CWMS and an optionally parameterized office_id to limit aliases.

rma-psmorris: @RyanM-RMA

PR #719, opened 2024-06-19
remove redundant calls to CwmsDTOBase::validate

adamkorynta:
the Formats deserialization process handles the validate call of content received from the client

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/EmbankmentController.java
@@ -164,7 +164,6 @@ public final class EmbankmentController  implements CrudHandler {
             String formatHeader = acceptHeader != null ? acceptHeader : Formats.JSON;
             ContentType contentType = Formats.parseHeader(formatHeader);
             Embankment embankment = Formats.parseContent(contentType, ctx.body(), Embankment.class);
-            embankment.validate();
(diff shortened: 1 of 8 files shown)


rma-bryson (approved):

rma-rripken:
I missed that your previous Format changes slipped in a validate call.   I'm hesitant.

adamkorynta:
I struggle a little with the purpose of the validate() method. It seems to be designed to ensure that the client's payload deserializes to a working DTO. In which case I do think we need to call it on all deserialization in the Formats class. I didn't see any references to the validate() getting called prior to serialization going back to the client (might be a good idea?). It seems like we should just build it into the serialization through annotations somehow....

MikeNeilson:
An annotation to mark fields as "required for operation" seems like a good idea, just a lot of effort.

The basic idea was enough validation for user provided data to meet minimum requirements if easily known and would save a lot of time going to the database. If the database fails quickly it's not a huge deal.

MikeNeilson:
But since we're using the ObjectMapper more and we can get an object schema, I think Adam is right that validation function isn't as useful as it was in the past.

MikeNeilson:
Or it might have been to handle those field exceptions better.

rma-rripken (dismissed):

rma-rripken (approved):

PR #720, opened 2024-06-19
Add ability to override default context in Gradle run task

adamkorynta:
allow the developer to override the default context in the Gradle run task. an example property for the override is added to `gradle.properties.example`.


rma-rripken:
I accomplished something sort of similar via:
cwms.dataapi.default.office=SPK
but your suggestion seems valid.

rma-rripken (approved):

MikeNeilson (approved):

PR #722, opened 2024-06-20
Deprecating old parseHeader API

RyanM-RMA:
Deprecating the old parseHeader and parseHeaderAndQueryParm methods that don't accept a DTO class.  The new method with the class parameter allows us to use aliasing for format types.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
+++ b/cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
@@ -294,11 +294,27 @@ public class Formats {
         return retVal;
     }
 
+    /**
+     * Parses the supplied header param to determine the content type.
+     *
+     * @param header Accept header value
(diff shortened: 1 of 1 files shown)


cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
```diff
@@ -292,6 +294,7 @@ public static ContentType parseQueryParam(String queryParam, Class<? extends Cwm
         return retVal;
     }
 
+    @Deprecated
```
rma-rripken, line 305: Can you add javadoc comment here with @deprecated suggestion or explanation

RyanM-RMA: Added a javadoc comment and the deprecated bit

rma-rripken (approved):

PR #732, opened 2024-06-24
Removed Equals/Hashcode from LookupType DTO

rma-bryson:


cwms-data-api/src/test/java/cwms/cda/data/dto/LookupTypeTest.java
```diff
@@ -102,6 +102,13 @@ void createLookupType_deserialize() throws Exception {
         String json = IOUtils.toString(resource, StandardCharsets.UTF_8);
         ContentType contentType = new ContentType(Formats.JSON);
         LookupType deserialized = Formats.parseContent(contentType, json, LookupType.class);
-        assertEquals(lookupType, deserialized, "LookupType deserialized from JSON doesn't equal original");
+        assertSame(lookupType, deserialized);
+    }
+
+    public static void assertSame(LookupType lookupType, LookupType deserialized) {
```
MikeNeilson, line 108: It's arguable this should be part of a class of CDA/CWMS assertions. But let's see if we end up with a few more of our own asserts first.

Though it can be useful for clarity and complex operations: https://github.com/opendcs/opendcs/blob/master/src/test-integration/java/org/opendcs/fixtures/assertions/TimeSeries.java (as an example)

MikeNeilson (commented):
This looks fine to me but I'll let the others review handle approval or change requests.

adamkorynta (approved):

rma-psmorris (commented):
No feedback, looks good.

PR #727, opened 2024-06-21
Feature/add outlet endpoint

RyanM-RMA:
Adding Outlet related DTO's

PR #738, opened 2024-06-25
Peer review changes from #727

RyanM-RMA:
This PR is for peer review changes from Peter from #727 related to the outlet DTO's


cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/Outlet.java
```diff
@@ -11,58 +24,64 @@
...
 
 @FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class, aliases = {Formats.DEFAULT, Formats.JSON})
 @JsonDeserialize(builder = Outlet.Builder.class)
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
-public class Outlet extends ProjectStructure
-{
-	private final CharacteristicRef characteristicRef;
+public class Outlet extends ProjectStructure {
+    private final CwmsId characteristicRef;
```
adamkorynta: with the name change, what do you think about `characteristicId` instead?

RyanM-RMA: It's a key to the properties table right?  Is that an id or a ref?  I'm not opposed either way tbh.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/ProjectStructure.java
```diff
@@ -1,53 +1,76 @@
...
+
+    public final CwmsId getProjectId() {
+        return projectId;
+    }
+
+    public final Location getLocation() {
+        return location;
+    }
+
+    protected final List<String> getMissingFields() {
```
adamkorynta: if you make this non-final subclasses can extend and add to the list. Then the validate method could move here and be marked final.

RyanM-RMA: Would we want to force subclasses to implement getMissingFields()? (i.e. abstract the method and do the loc/parentId checks in validate?)  Or just de-finalize it, and let it be overridden as needed?

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/Outlet.java
adamkorynta: If we're doing that distinction, I'd say it's a "properties ref" given the underlying structure is using properties, but a "characteristic id" since it represents a single "Characteristic". Unless that should be plural characteristics....

There aren't any other "ref" fields that I could find, so I would lean towards "id" for consistency.

RyanM-RMA: Alright, characteristicsId it is, cause you can have multiple characteristics in the properties if I recall.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/ProjectStructure.java
adamkorynta: I think in a refactor (out of scope for this PR) we'd make `getMissingFields()` abstract so that all DTO's override. Then the validate() can be consistent across.

adamkorynta: for the two concrete implementations of ProjectStructure so far I think it's ok not to abstract the getMissingFields()

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/Outlet.java
MikeNeilson: Would it instead be better to just include "The Characteristic"?

RyanM-RMA: We could certainly go out and retrieve the properties from the db.  I don't know that Outlets will have a significant amount of data in those.  I wasn't sure if that's something we wanted to include, since it's not always necessary information, and the hec-cwms-data-access JDBC counterpart is currently not retrieving that information, just the ref.

RyanM-RMA: @MikeNeilson is that something you'd like us to do?

MikeNeilson: Not a hard requirement, but as previously discussed it can be useful to just retrieve and thus de-normalize all data, especially if rather small. Which if these are just simple "properties" I imagine they would be.

RyanM-RMA: @MikeNeilson I spoke with @adamkorynta and for this PR I'm going to leave it with just the id.  We realized if we return the properties, we'd also have to decide on how the post/put/patch logic will apply to it.
Adam also recommended that I speak with @krowvin about it, since this may be data he requires, and if it is, we can identify the requirements and discuss behaviors further.  I'll post that in our discussion about gate changes. #317

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/ProjectStructure.java
```diff
@@ -11,43 +24,62 @@
...
+
+    protected List<String> getMissingFields() {
+        List<String> output = new ArrayList<>();
+
+        if (getLocation() == null) {
+            output.add("Location");
+        } else {
+            try {
+                getLocation().validate();
+            } catch (RequiredFieldException ex) {
```
rma-psmorris: Why are you only collecting RequiredFieldException? If validate() kicks out a FieldException, it is going to blow out of this try/catch.

adamkorynta (dismissed):

cwms-data-api/src/test/java/cwms/cda/data/dto/location/kind/OutletTest.java
```diff
@@ -1,78 +1,93 @@
...
-
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-class OutletTest
-{
-	private static final String SPK = "SPK";
-	private static final String PROJECT_LOC = "location";
-	private static final String OUTLET_LOC = PROJECT_LOC + "-outlet";
+class OutletTest {
```
rma-psmorris, line 37: As noted in PR https://github.com/USACE/cwms-data-api/pull/727, this test should have coverage for base location outlets. Does it have this?

cwms-data-api/src/test/resources/cwms/cda/data/dto/location/kind/outlet.json
```diff
@@ -15,8 +15,8 @@
     "horizontal-datum": "NAD84",
     "vertical-datum": "NAVD88"
   },
-  "characteristic-ref": {
+  "characteristics-id": {
     "office-id": "SPK",
-    "characteristic-id": "Ogee Weir Depth"
+    "name": "Ogee Weir Depth"
   }
 }
```
rma-psmorris, line 22: add a base location outlet

rma-psmorris (changes requested):
More feedback

cwms-data-api/src/test/java/cwms/cda/data/dto/location/kind/OutletTest.java
rma-psmorris: this needs to be added as you are making assumptions that the internals of this DTO will never differentiate between base and sub location outlets and both are valid use cases.

RyanM-RMA: I added tests for serializing and deserializing with just a base location id of "outlet".  Also added a serialize from file test for base location only.

cwms-data-api/src/test/resources/cwms/cda/data/dto/location/kind/outlet.json
RyanM-RMA: Added a separate file for this.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/ProjectStructure.java
RyanM-RMA: I hadn't thought that all out apparently.  I added some logic to check the details map, and if it's empty, it adds the message.

RyanM-RMA: I also noticed that the Outlet included the characteristics stuff as required, but it's not required, so I removed that.  And the Embankment wasn't validating the structure type.

rma-psmorris (approved):
No remaining feedback, looks good

adamkorynta (approved):

PR #743, opened 2024-06-27
Implementing StreamDao with tests

rma-bryson:
Also moved basin-connectivity daos into basin-connectivity package


cwms-data-api/src/main/java/cwms/cda/data/dao/StreamDao.java
```diff
@@ -1,130 +1,257 @@
...
+    /**
+     * Retrieve a list of streams
+     * @param officeIdMask - the office id mask
+     * @param streamIdMask - the stream id mask
+     * @return a list of streams
+     */
+    public List<Stream> retrieveStreams(String officeIdMask, String streamIdMask) {
+        return connectionResult(dsl, conn -> {
+            setOffice(conn, officeIdMask);
+            ResultSet streams = CWMS_STREAM_PACKAGE.call_CAT_STREAMS(DSL.using(conn).configuration(), streamIdMask,
```
adamkorynta: @MikeNeilson - do you want these retrievals to be changed to tables queries within CDA? (would start making paging doable) or are we waiting until we get the jOOQ codegen available with table read access for everyone?

MikeNeilson: No specific preference, I'm generally in favor of whatever the easiest way to get the data we need is.

Though if I was listing a preference is would be (in order):
1. JOOQ
2. JDBC
3. hec-cwms-data-access DAOs (but really only becuase it keeps looking like an odd circular dependency to me.)

adamkorynta: Yeah I think we can avoid 2/3 is almost all cases.

For 1. it's just the decision of coding against the PL/SQL packages, jOOQ DSL with codegenned tables/views, or jOOQ DSL with hardcoded table/view names. For what Bryson is doing here, PL/SQL packages hit the easiest way to get data we need.

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamDao.java
```diff
@@ -1,130 +1,257 @@
...
+            String startsDownstream = OracleTypeMap.formatBool(stream.getStartsDownstream());
+            String stationUnits = getStationUnits(stream);
+            String flowIntoStream = null;
+            Double flowIntoStation = null;
+            String flowIntoBank = null;
+            String divertFromStream = null;
+            Double divertFromStation = null;
+            String divertFromBank = null;
+            if(stream.getFlowsIntoStreamNode() != null) {
+                flowIntoStream = stream.getFlowsIntoStreamNode().getStreamId().getName();
```
adamkorynta, line 120: readability: I'd extract `stream.getFlowsIntoStreamNode()` to a variable

adamkorynta: similarly to the other condition blocks below

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamDao.java
```diff
@@ -1,130 +1,257 @@
...
+     */
+    public List<Stream> retrieveStreams(String officeIdMask, String streamIdMask) {
+        return connectionResult(dsl, conn -> {
+            setOffice(conn, officeIdMask);
+            ResultSet streams = CWMS_STREAM_PACKAGE.call_CAT_STREAMS(DSL.using(conn).configuration(), streamIdMask,
+                    DB_STATION_UNITS, null, null, null,
+                    null, null, null, null,
+                    null, null, null, null, null,
+                    null, null, officeIdMask)
+                    .intoResultSet();
```
adamkorynta: Why are you turning into a ResultSet instead of using the jOOQ Result class?

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamDao.java
```diff
@@ -1,130 +1,257 @@
...
+                            .withName(streamId)
+                            .withOfficeId(officeId)
+                            .build())
+                    .withStation(station)
+                    .withBank(bank)
+                    .withStationUnits(stationUnits)
+                    .build();
+            retVal.validate();
+        } catch (FieldException e) {
+            retVal = null; //if missing required fields, don't build a stream node, just return null
```
adamkorynta: This mean the database is returning an invalid StreamNode? When does that happen? that should get logged at the very least. It should probably even just be a error code 500 returned to the client.

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/Bank.java
```diff
@@ -35,9 +35,12 @@ public static Bank fromName(String name){
 
     @JsonCreator
     public static Bank fromCode(String code){
-        for(Bank bank : Bank.values()){
-            if(bank.getCode().equals(code)){
-                return bank;
+        if(code != null)
+        {
```
adamkorynta: formatting

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamDao.java
MikeNeilson: Yeah, if the PL/SQL package work and the flow of data/operations is sufficiently clear definitely use them, that's what their for.

But like the catalog and time series, etc endpoints, just don't be afraid to expand outward to the view and tables if it's make things easier. (And bypassing the wait for a schema release is certainly something that i consider easier, especially if we're trying to fix performance.)

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamDao.java
rma-bryson: yeah, i double checked the db store call and it only stores the station and bank if the divertsFrom/flowsInto stream id is supplied. So there's no cases in practice where you'd get back an invalid streamNode. Going to update this to just build the streamnode conditionally on if the stream id exists for the diversion/flows into stream

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamDao.java
rma-bryson: extracted to variable for both flowsInto and divertsFrom

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamDao.java
rma-bryson: I saw hec-cwms-data-access did something similar so i followed suit. I replaced with Result, similar to the PoolsDao.

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamDao.java
rma-bryson: updated this to just null check the streamId

cwms-data-api/src/main/java/cwms/cda/data/dto/stream/Bank.java
rma-bryson: fixed formatting

adamkorynta (approved):

PR #744, opened 2024-06-27
Release fixes 20240625

MikeNeilson:
Okay, so I gave some bad advice earlier, when using CwmsDatabaseContainer::connection use the `CwmsDataApiSetupCallback.getWebUser()` as the connection user.

Also when using an arbitrary office, make sure within users.sql the &webuser has full permissions in that office to handle all that.

Otherwise most of this was just super nitpicky details and timings about when data was or wasn't deleted.

The most annoying thing was the Office XML, FYI FasterXML's jackson-databind-xml does not and cannot correctly serialize a Collection directly using anything other than the Collection name as the root and "item" as the sub elements.
Hence the XMLv2Office class. Spent way to long on that and probably should've just had it fail or something.

NOTE: tested against schema 24.05.24-RC03, the locations can't be stored correctly from the newer data sets without the additional nation id fix Adam made.


MikeNeilson:
For reference: @rma-bryson @rma-psmorris @zack-rma

MikeNeilson:
The eclipse and idea plugs were added in an attempt to make the IDE integration work better. Unfortunately I think using "cwms-data-api" for one of the sub projects is at least breaking eclipse/vscode. @krowvin got the wonkiest result and I started having issues as well after that change. 

I need to do some manual testing first tomorrow, but I'm hoping to experiment with that in a fresh environment tomorrow and made rename the subproject folder so just api if required.

cwms-data-api/src/test/java/cwms/cda/api/OfficeControllerIT.java
```diff
@@ -160,5 +163,12 @@ enum AliasTest
             _expectedFormat = expectedFormat;
             _expectedGetAllStatusCode = expectedGetAllStatusCode;
         }
+/*
+        @Override
+        public String toString()
```
RyanM-RMA: Dead function

cwms-data-api/src/test/resources/cwms/cda/data/sql/users.sql
```diff
@@ -13,6 +13,10 @@ begin
   cwms_sec.add_user_to_group('&webuser','CWMS Users', 'SPK');
   cwms_sec.add_user_to_group('&webuser','CWMS PD Users', 'SPK');
   cwms_sec.add_user_to_group('&webuser','CWMS DBA Users', 'SPK');
+  cwms_sec.add_user_to_group('&webuser','All Users', 'SWT');
```
adamkorynta, line 16: Should this be for all groups? I presumed since there was the SWT user added below that the webuser had permissions.

cwms-data-api/src/test/java/cwms/cda/api/BasinControllerTestIT.java
```diff
@@ -14,6 +16,7 @@
 import static org.hamcrest.Matchers.is;
 
 @Tag("integration")
+@Disabled("Not data available for test or the test isn't currently loading any data.")
```
RyanM-RMA: Fix the note here, I think you mean "No data available"

adamkorynta: Added an issue: https://github.com/USACE/cwms-data-api/issues/745

cwms-data-api/src/test/java/cwms/cda/api/EmbankmentControllerIT.java
```diff
@@ -57,6 +58,7 @@
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.*;
 
+@Disabled("Data issues in test")
```
adamkorynta, line 61: Added issue to track: https://github.com/USACE/cwms-data-api/issues/746

adamkorynta (dismissed):

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/EmbankmentDaoIT.java
```diff
@@ -55,19 +58,21 @@
...
+    private static final String PROJECT_OFFICE_ID = "SPK";
+    private static final Location PROJECT_LOC = buildProjectLocation("PROJECT1emdao");
+    private static final Location PROJECT_LOC2 = buildProjectLocation("PROJECT2emdao");
+    private static final Location EMBANK_LOC1 = buildEmbankmentLocation("PROJECT-EMBANK_LOC1dao");
+    private static final Location EMBANK_LOC2 = buildEmbankmentLocation("EMBANK_LOC2dao");
+    private static final Location EMBANK_LOC3 = buildEmbankmentLocation("EMBANK_LOC3dao");
 
     @BeforeAll
-    public static void setup() throws Exception {
+    public void setup() throws Exception {
```
RyanM-RMA, line 72: Pretty sure BeforeAll requires static

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/EmbankmentDaoIT.java
```diff
@@ -78,33 +83,39 @@ public static void setup() throws Exception {
...
                 throw new RuntimeException(e);
             }
-        });
+        },
+        CwmsDataApiSetupCallback.getWebUser());
     }
 
     @AfterAll
-    public static void tearDown() throws Exception {
+    public void tearDown() throws Exception {
```
RyanM-RMA, line 91: Pretty sure AfterAll requires static

RyanM-RMA (changes requested):
I only found some minor stuff.  I'm not knowledgeable enough for some of the modified pieces here.

cwms-data-api/src/test/resources/cwms/cda/data/sql/users.sql
MikeNeilson: Honestly not sure, but it's generally safe to just call add_user_to_group so it doesn't hurt to just make sure the permissions exist.

cwms-data-api/src/test/java/cwms/cda/api/OfficeControllerIT.java
MikeNeilson: OH thanks. not sure why the enum toString never got called for the parameterized test display name but I forgot to delete it.

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/EmbankmentDaoIT.java
MikeNeilson: Look at the annotation added to the class. That's only true if the test lifecycle isn't set to "one instance of the class".
Might not have been necessary here as the trackers in DataApiTest are static, but I forgot that while I was experimenting.

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/EmbankmentDaoIT.java
MikeNeilson: See previous comment. Arguable I should change it back, but frankly things are working now so I'd rather not risk breaking them again.

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/EmbankmentDaoIT.java
RyanM-RMA: Oh interesting!  I didn't know that, thank you for pointing that out.  I haven't used the TestInstance annotation before, so I'll have to look into it, and see how I can incorporate it in my work.

rma-rripken (dismissed):

RyanM-RMA (dismissed):

MikeNeilson:
Would you believe in fixing the catalog I broke the catalog? ... column names are a pain.

rma-rripken (dismissed):

rma-rripken (approved):

PR #747, opened 2024-06-28
Query Project children of a kind.

rma-rripken:


cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -474,6 +475,7 @@ protected void configureRoutes() {
         get(forecastFilePath, new ForecastFileController(metrics));
         addCacheControl(forecastFilePath, 1, TimeUnit.DAYS);
 
+        get("/projects/children/", new ProjectChildrenHandler(metrics));
```
rma-psmorris: child-locations?

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectChildrenHandler.java
```diff
@@ -0,0 +1,115 @@
...
+            description = "Get a list of project child locations",
+            queryParams = {
+                @OpenApiParam(name = OFFICE, required = true),
+                @OpenApiParam(name = PROJECT_LIKE,  description = "Posix <a href=\"regexp.html\">regular expression</a> matching "
+                        + "against the project_id."),
+                @OpenApiParam(name = LOCATION_KIND_LIKE, description = "Posix <a "
+                        + "href=\"regexp.html\">regular expression</a> matching against the "
+                        + "location kind.  The pattern will be matched against "
+                        + "the valid location-kinds for Project Children:"
+                        + "{\"EMBANKMENT\", \"TURBINE\", \"OUTLET\", \"LOCK\", \"GATE\"}. "
```
rma-psmorris: I think this will also match PUMP

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
```diff
@@ -0,0 +1,262 @@
...
+                return getGateChildren(office, projLike);
+            default:
+                return null;
+        }
+
+    }
+
+    private Map<String, List<CwmsId>> getGateChildren(String office, @Nullable String projLike) {
+        Map<String, List<CwmsId>> retval = new LinkedHashMap<>();
+        AV_GATE view = AV_GATE.AV_GATE;
```
rma-psmorris: I think this is an un-used view, it should be av_outlet where the rating has a controlling parameter

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
```diff
@@ -0,0 +1,262 @@
...
+        dsl.selectDistinct(view.OFFICE_ID, view.GATE_ID, view.PROJECT_ID)
+                .from(view)
+                .where(view.OFFICE_ID.eq(office))
+                .and(caseInsensitiveLikeRegexNullTrue(view.PROJECT_ID, projLike))
+                .orderBy(view.OFFICE_ID, view.PROJECT_ID, view.GATE_ID)
+                .forEach(row -> {
+                    String projId = row.get(view.PROJECT_ID);
+                    CwmsId embankment = new CwmsId.Builder()
```
rma-psmorris: Is it possible for you to ask your co-pilot to suggest variable name corrections for your source so it can correct embankment to gate. I've seen mis-named variables elsewhere too. Just wondering if there is an easy way to correct these.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
```diff
@@ -0,0 +1,262 @@
...
+                return getLockChildren(office, projLike);
+            case "GATE":
+                return getGateChildren(office, projLike);
+            default:
+                return null;
+        }
+
+    }
+
+    private Map<String, List<CwmsId>> getGateChildren(String office, @Nullable String projLike) {
```
rma-psmorris: You could code this Map as <CwmsId, List<CwmsId>> to provide an office ID to the project.

cwms-data-api/src/test/resources/cwms/cda/data/dto/project_children.json
```diff
@@ -0,0 +1,48 @@
+{
+  "project": {
+    "office-id": "SPK",
+    "name": "TestProject"
+  },
+  "embankments": [
```
rma-psmorris: what do you think about just using the location kind here for the variable name instead of the plural? 
"embankment": [
...

rma-psmorris (changes requested):
Feedback and questions

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
rma-rripken: I can see child is not exactly right.  Do you have a better suggestion? They aren't sub-locations. Are they just locations and b/c the url is /projects/locations it will be clear that they are locations at the projects?

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
rma-rripken: Wouldn't that require CwmsId to have working implementations of hashCode and equals?

cwms-data-api/src/test/resources/cwms/cda/data/dto/project_children.json
rma-rripken: It holds a list so I went with plural but it could be singular or the actual kind "EMBANKMENT".  Is there a reason you prefer the singular?

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
rma-rripken: Or are you suggesting the url be child-locations instead of children?

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
rma-rripken: I tried a couple prompts but nothing was satisfying.  Easier to just fix them by hand.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
rma-rripken: I renamed to /projects/child-locations.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectChildrenHandler.java
rma-rripken: I don't think so.  Did you want to to match PUMP?

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
rma-rripken: ok.  I switched to AV_OUTLET but I don't have data to test against for gates.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
rma-rripken: fixed them.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
MikeNeilson: It would. That's a requirement for Map.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
MikeNeilson: just `locations` would probably be better but then it would make sense to be `projects/<project>/locations` which I don't think we're ready for yet.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
rma-rripken: Right.  I guess I don't understand @rma-psmorris comment.   In the discussion of removing hashCode and equals from all the DTO type I explained how they could be used in Dao implementations.  My example was LocationGroup and LocationCategory but the same can be said for CwmsId.  Why would CwmsId as a key be better than projId?  Do you want that so that we can make the office parameter optional?  Do any of the project-child allow office mismatch between the project and the project-child locations?

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
rma-psmorris: Correct Ryan, I was asking that location be added to the endpoint. I also like Mikes suggestion.

cwms-data-api/src/main/java/cwms/cda/api/project/ProjectChildrenHandler.java
rma-psmorris: I take this back, the relationship between a pump and a project exists only at the water supply contract level. There is no FK from a pump to a Project, it is to a Stream Location.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
rma-psmorris: Did you test for a controlling parameter on the rating?

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
rma-psmorris: Yeah, you are both correct. It would have to be a concatenated String, not a CwmsId. 

The intent of my suggestion was to bubble up the office/project in the return rather than just the project. Up above in the controller you are connecting the queried project id to the argument office so the returned JSON is complete.

cwms-data-api/src/test/resources/cwms/cda/data/dto/project_children.json
rma-psmorris: If someone queries locations and uses the kind then tries to map the kind to here, they can't without some custom handling to understand that the sets are plural here.

Correspondingly, take the results of this query then match the project embankments against the location catalog filtered to locations of kind embankment. The set type "embankments" cant just be used w/o clipping the "s".

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildLocationDao.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    private Map<String, List<CwmsId>> getGateChildLocations(String office, @Nullable String projRegex) {
+        Map<String, List<CwmsId>> retval = new LinkedHashMap<>();
+        // AV_GATE is apparently not used.
+        AV_OUTLET view = AV_OUTLET.AV_OUTLET;
+        dsl.selectDistinct(view.OFFICE_ID, view.PROJECT_ID, view.OUTLET_ID)
+                .from(view)
+                .where(view.OFFICE_ID.eq(office)
+                        .and(view.OPENING_UNIT_EN.isNotNull().or(view.OPENING_UNIT_SI.isNotNull()))
```
rma-rripken, line 107: This is the extra bit added for Gates.  I talked to RyanM and looked at existing code and this seemed easier than trying to pick apart the rating descriptor but I'm not a gate expert - is this a valid check to separate gates from outlets?

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
rma-rripken: I don't think I'm ready for Mike's suggestion.  I'll have to look but I don't think we've done that anywhere else yet. If we haven't done that anywhere else I guess I'm going to use /projects/locations

rma-rripken: @rma-psmorris Are you ok with the dto being named ProjectChildLocations?  Or the Dao/Controller being ProjectChildLocationDao/ProjectChildLocationHandler?   If the class was ProjectLocations would that make sense or be confusing?

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildrenDao.java
rma-rripken: You are talking about separating the gates from the outlets - not about unit or integration tests - right?   I look for outlets that have an opening unit.   I'm not sure that is a sufficient condition but I talked to RyanM about it and there was at least one other place that also did that.

cwms-data-api/src/main/java/cwms/cda/data/dao/project/ProjectChildLocationDao.java
rma-psmorris: This works for all observed cases that we have seen.

cwms-data-api/src/test/resources/cwms/cda/data/dto/project_children.json
rma-rripken: Would a format like this work?   
```
{
  "project": {
    "office-id": "SPK",
    "name": "TestProject"
  },
  "locations-by-kind": [
    {
      "kind": "EMBANKMENT",
      "locations": [
        {
          "office-id": "SPK",
          "name": "TestEmbankment1"
        },
        {
          "office-id": "SPK",
          "name": "TestEmbankment2"
        }
      ]
    },
    {
      "kind": "TURBINE",
      "locations": [
        {
          "office-id": "SPK",
          "name": "TestTurbine1"
        }
      ]
    },
    {
      "kind": "OUTLET",
      "locations": [
        {
          "office-id": "SPK",
          "name": "TestOutlet1"
        }
      ]
    },
    {
      "kind": "LOCK",
      "locations": [
        {
          "office-id": "SPK",
          "name": "TestLock1"
        },
        {
          "office-id": "SPK",
          "name": "TestLock2"
        },
        {
          "office-id": "SPK",
          "name": "TestLock3"
        }
      ]
    },
    {
      "kind": "GATE",
      "locations": [
        {
          "office-id": "SPK",
          "name": "TestGate1"
        }
      ]
    }
  ]
}
```

MikeNeilson: Well, it's verbose, but it's abundantly clear.

adamkorynta (approved):

PR #755, opened 2024-07-01
Devops/build tweaks

MikeNeilson:
@krowvin and @DanielTOsborne have been having issues with gradle integration in their IDEs (Charles I know is VSCode, I think Daniel too.)

After working with Charles a bit I release VScode's gradle extension was having issues with the build setup using the same name for a subproject as the root project, additionally a known dependency wasn't getting picked up by the build as it was in "tomcatLibs" instead of "testImplementation" (there was probably a reason for that, but seems to not exist anymore.)

This reorganizes the project a bit (moving cwms-data-api to data-api) so that VSCode picks up the structure correct and thus intellisense works.

That said, I've had it working a while and Charles' environment was just going all sort of wonky so who knows it it'll fix his, but it should help.

gradle 8.8 also include support for making sure the gradle daemon JVM matches the release target spec, I may do that tomorrow, but it didn't appear to be required, all the testing was without changing my default JVM which for me is 17.


adamkorynta:
rather than moving all the files, I think changing the parent to `cwms-data-api-parent` and leaving `cwms-data-api` along makes sense. We did that rename in the client: https://github.com/HydrologicEngineeringCenter/cwms-data-api-client/blob/main/settings.gradle#L1 (though we just gave it a friendly name since it produces no artifacts).

MikeNeilson:
I'll check that out. I was having issues in VSCode because of it, but maybe that was something in the specific project.

adamkorynta (approved):

MikeNeilson:
Daniel and Charles were the ones having the most problems so I'm going to wait until they have a change to try it out before merging.

PR #758, opened 2024-07-02
cleanup validate() methods 

adamkorynta:
moved validation logic to a single class to help write better validation methods

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
@@ -10,8 +10,6 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import cwms.cda.api.enums.Nation;
-import cwms.cda.api.errors.FieldException;
-import cwms.cda.api.errors.RequiredFieldException;
 import cwms.cda.formatters.Formats;
 import cwms.cda.formatters.annotations.FormattableWith;
 import cwms.cda.formatters.json.JsonV1;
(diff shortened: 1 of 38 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/ProjectStructure.java
```diff
@@ -20,34 +20,34 @@
...
 import cwms.cda.data.dto.CwmsId;
 import cwms.cda.data.dto.Location;
-import java.util.ArrayList;
-import java.util.List;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
-public abstract class ProjectStructure implements CwmsDTOBase {
+public abstract class ProjectStructure extends CwmsDTO {
+    @JsonProperty(required = true)
```
rma-rripken, line 36: Can the validator find the fields annotated as "required" without them also having to be specified in validateInternal?

MikeNeilson: Maybe a default validate function on CwmsDTOBase that scans for the annotations?

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
```diff
@@ -519,31 +516,14 @@ private Nation convertStringToNation(String nation) {
...
-        }
-        if (this.getLatitude() == null) {
-            missingFields.add("Latitude");
-        }
-        if (!missingFields.isEmpty()) {
-            throw new RequiredFieldException(missingFields);
-        }
+    protected void validateInternal(CwmsDTOValidator validator) {
+        super.validateInternal(validator);
+        validator.required(getName(), "name");
```
rma-rripken, line 521: It would be nice if the validation could find the getter from the property name.  Or if it could guess the property name from the name of the getter.   Or even find the required properties from just the annotations...

rma-rripken (approved):
Looks fine.  Cleaner than it was.   I'm not sure this is the final form of validation I think we could use some design inspiration for how to do this even better.

adamkorynta:
I'll merge this in and then take a look at annotations. I was trying (and kinda failing) not to go too wild

PR #759, opened 2024-07-02
refactor equality assertions to helper class

adamkorynta:
taking @MikeNeilson 's suggestion from earlier PR

PR #760, opened 2024-07-02
add turbine changes DTO's

adamkorynta:
adds PhysicalStructureChange DTO which references a list of Settings. There is a TurbineSetting subclass that is discovered through deserialization by the "type" JSON field (added at runtime).

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
@@ -24,10 +24,14 @@
 
 package cwms.cda.data.dto;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.flogger.FluentLogger;
 import cwms.cda.api.errors.ExclusiveFieldsException;
 import cwms.cda.api.errors.FieldException;
 import cwms.cda.api.errors.RequiredFieldException;
+import java.lang.reflect.Field;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Set;
@@ -38,6 +42,8 @@ import java.util.concurrent.Callable;
  * This class is used to validate the fields of a CwmsDTO object.
  */
 public final class CwmsDTOValidator {
+
+    private static final FluentLogger LOGGER = FluentLogger.forEnclosingClass();
 
     private final Set<String> missingFields = new HashSet<>();
     private final Set<String> mutuallyExclusiveFields = new HashSet<>();
@@ -56,6 +62,42 @@ public final class CwmsDTOValidator {
             missingFields.add(fieldName);
         } else if (value instanceof CwmsDTO) {
             ((CwmsDTO) value).validateInternal(this);
+        }
+    }
+
+    public void validateRequiredFields(CwmsDTO cwmsDTO) {
+        Class<? extends CwmsDTO> type = cwmsDTO.getClass();
+        Field[] fields = type.getDeclaredFields();
+        try {
+            for (Field field : fields) {
+                JsonProperty annotation = field.getAnnotation(JsonProperty.class);
+                if (annotation != null && annotation.required()) {
+                    boolean accessible = field.isAccessible();
+                    synchronized (type) {
+                        if (!accessible) {
+                            field.setAccessible(true);
+                        }
+                        Object value = field.get(cwmsDTO);
+                        if (value == null) {
+                            missingFields.add(field.getName());
+                        } else if (value instanceof CwmsDTO) {
+                            validateRequiredFields((CwmsDTO) value);
+                        }
+                        if (!accessible) {
+                            field.setAccessible(false);
+                        }
+                    }
+                }
+            }
+        } catch (IllegalAccessException e) {
+            LOGGER.atWarning().withCause(e).log("Unable to validate required fields are non-null in DTO: " +
+                type);
+        }
+    }
+
+    public void validateCollection(Collection<? extends CwmsDTO> collection) {
+        if (collection != null) {
+            collection.forEach(c -> c.validateInternal(this));
         }
     }
 
@@ -91,7 +133,7 @@ public final class CwmsDTOValidator {
      *
      * @param callable the Callable to be validated
      */
-    public void validate(Callable callable) {
+    public void validate(Callable<?> callable) {
         try {
             callable.call();
         } catch (Exception e) {
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
@@ -74,17 +74,20 @@ public final class CwmsDTOValidator {
                 if (annotation != null && annotation.required()) {
                     boolean accessible = field.isAccessible();
                     synchronized (type) {
-                        if (!accessible) {
-                            field.setAccessible(true);
-                        }
-                        Object value = field.get(cwmsDTO);
-                        if (value == null) {
-                            missingFields.add(field.getName());
-                        } else if (value instanceof CwmsDTO) {
-                            validateRequiredFields((CwmsDTO) value);
-                        }
-                        if (!accessible) {
-                            field.setAccessible(false);
+                        try {
+                            if (!accessible) {
+                                field.setAccessible(true);
+                            }
+                            Object value = field.get(cwmsDTO);
+                            if (value == null) {
+                                missingFields.add(field.getName());
+                            } else if (value instanceof CwmsDTO) {
+                                validateRequiredFields((CwmsDTO) value);
+                            }
+                        } finally {
+                            if (!accessible) {
+                                field.setAccessible(false);
+                            }
                         }
                     }
                 }
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
@@ -75,6 +75,7 @@ public final class PhysicalStructureChange<T extends Setting> extends CwmsDTO {
     private final Set<T> settings;
 
     private PhysicalStructureChange(Builder<T> builder) {
+        //uses CwmsId instead of a separate office-id field
         super(null);
         this.projectId = builder.projectId;
         this.dischargeComputationType = builder.dischargeComputationType;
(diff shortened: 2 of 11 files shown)


adamkorynta:
Based on change: https://github.com/USACE/cwms-data-api/pull/759 that should get merged first.

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
```diff
@@ -59,6 +65,42 @@ public void required(Object value, String fieldName) {
...
+                            field.setAccessible(true);
+                        }
+                        Object value = field.get(cwmsDTO);
+                        if (value == null) {
+                            missingFields.add(field.getName());
+                        } else if (value instanceof CwmsDTO) {
+                            validateRequiredFields((CwmsDTO) value);
+                        }
+                        if (!accessible) {
+                            field.setAccessible(false);
```
RyanM-RMA: Should this be added to a finally block?

cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
```diff
@@ -216,6 +235,11 @@ public static <T extends CwmsDTOBase> T parseContent(ContentType type, InputStre
         return formats.parseContentFromType(type, inputStream, rootType);
     }
 
+    public static <T extends CwmsDTOBase> List<T> parseContentList(ContentType type, String inputStream, Class<T> rootType)
```
RyanM-RMA: Is there a better name than inputStream for the String?  I think the other ones use content - I like that better.

cwms-data-api/src/test/java/cwms/cda/helpers/DTOMatch.java
```diff
@@ -45,8 +53,10 @@ private DTOMatch() {
 
     public static void assertMatch(CwmsId first, CwmsId second, String variableName) {
         assertAll(
-            () -> Assertions.assertEquals(first.getOfficeId(), second.getOfficeId(), variableName + " is not the same. Office ID differs"),
```
MikeNeilson, line 48: Honestly, I would just add an @SuppressWarning for the line length to these methods or the class, even though they're long they read better on one line.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
```diff
@@ -0,0 +1,250 @@
...
+    private final String dischargeUnits;
+    private final Double poolElevation;
+    private final Double tailwaterElevation;
+    private final String elevationUnits;
+    private final String notes;
+    private final Set<T> settings;
+
+    private PhysicalStructureChange(Builder<T> builder) {
+        //uses CwmsId instead of a separate office-id field
+        super(null);
```
MikeNeilson, line 79: A thought, would it make sense to put a CwmsId at the CwmsDto level, and use a method over override the output name? Though we'd have to put an @JsonIgnore at the CwmsDto level.

Just doesn't seem nice to have a special case that ignores that root element.

cwms-data-api/src/main/java/cwms/cda/formatters/OutputFormatter.java
```diff
@@ -13,6 +13,10 @@ default <T extends CwmsDTOBase> T parseContent(String content, Class<T> type) {
         throw new UnsupportedOperationException("Unable to process your request. Deserialization of "
                 + getContentType() + " not yet supported.");
     }
+    default <T extends CwmsDTOBase> List<T> parseContentList(String content, Class<T> type) {
```
MikeNeilson, line 16: It would seem "OutputFormatter" is no longer the correct name. (Not that we need to fix that right now.)

cwms-data-api/src/test/java/cwms/cda/helpers/DTOMatch.java
adamkorynta: ok, revised

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
adamkorynta: yeah this definitely seems weird to me, but when I tried putting CwmsId in CwmsDto that also got weird. We don't have any need for the cwmsId except for getting the office id out of it and it doesn't solve the need for @JsonIgnore on the "office-id" fields. We do already have precedence for handing null to the super in several classes so that's why I went with this solution.

A slightly cleaner idea might be to create a new abstract base class and move the getOfficeId() method up to the CwmsDTOBase interface and then mark the abstract class's implementation of getOfficeId() and the cwmsId field as @JsonIgnore'd.

The other issue is that in classes like PhysicalStructure, we have a CwmsId for the project-id, but really the identifying information for the DTO is the Location which isn't and doesn't have a CwmsId (might be worth having in a v3 of the DTO?).

cwms-data-api/src/main/java/cwms/cda/formatters/OutputFormatter.java
adamkorynta: Well it technically does take in an input and output a "formatted" DTO....

cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
adamkorynta: found my copy/paste from the method above. Fixed

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
adamkorynta: good call, updated

RyanM-RMA (approved):
This looks good to me.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
MikeNeilson: Yeah. I don't see a good solution right now. as long as a getOffice method returns something reasonable though that's the important part. So maybe you're right, CwmsDto should just be an interface with getOffice (or maybe specifically getOfficeForWritePermissions) to be able to set the connection and the objects do what they need. Something to consider later though.

MikeNeilson (approved):

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
```diff
@@ -59,6 +65,45 @@ public void required(Object value, String fieldName) {
         }
     }
 
+    public void validateRequiredFields(CwmsDTO cwmsDTO) {
```
rma-psmorris, line 68: Recommend supporting CwmsDTOBase in this. Providing access to data via bean pattern getter (avoids accessible if the field cant be accessed). Cache either the getter method or the field to the class so the iteration can be skipped.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
```diff
@@ -0,0 +1,250 @@
...
+import java.time.Instant;
+import java.util.Set;
+import java.util.TreeSet;
+
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@JsonDeserialize(builder = PhysicalStructureChange.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@JsonPropertyOrder({"project-id", "change-date", "protected", "discharge-computation-type", "reason-type", "notes"})
+public final class PhysicalStructureChange<T extends Setting> extends CwmsDTO {
```
rma-psmorris, line 57: Recommend decoupling from CwmsDTO given office id is composed in through the location. If necessary look at changing CwmsDTOBase to be an abstract class instead of an interface. Mike's suggestion below about changing CwmsDTO to be an interface providing an office getter also works.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
rma-psmorris: I also like the idea of CwmsDto being an interface providing access to the office, but not having it.

cwms-data-api/src/main/java/cwms/cda/formatters/json/JsonV2.java
```diff
@@ -107,4 +107,13 @@ public <T extends CwmsDTOBase> T parseContent(InputStream content, Class<T> type
             throw new FormattingException("Could not deserialize:" + content, e);
         }
     }
+
+    @Override
+    public <T extends CwmsDTOBase> List<T> parseContentList(String content, Class<T> type) {
+        try {
+            return om.readValue(content, om.getTypeFactory().constructCollectionType(List.class, type));
+        } catch (IOException e) {
+            throw new FormattingException("Could not deserialize:" + content, e);
```
rma-psmorris, line 116: "Could not deserialize `content` as `type`"?

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
rma-psmorris: There is no TurbineChange object? Gate changes are going to sub-class?

rma-psmorris (commented):
Post-merge review

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
adamkorynta: Addressing in: https://github.com/USACE/cwms-data-api/pull/769

Added some tests for performance using the cache.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
adamkorynta: I fixed the inheritance in PR: https://github.com/USACE/cwms-data-api/pull/769

I added a TurbineChange class (essentially just a generic'd placeholder) in order to avoid the lack of generics on deserialization: https://github.com/USACE/cwms-data-api/pull/762/files#diff-dabd158857511fc8e9ec5706cd2d113b5f9e56eadfa1d5c3cb89644d9f1ea99e

Gate changes will also subclass and add more fields like release override etc.

cwms-data-api/src/main/java/cwms/cda/formatters/json/JsonV2.java
adamkorynta: added type in PR: https://github.com/USACE/cwms-data-api/pull/773

PR #762, opened 2024-07-03
add turbine change dao and controller

adamkorynta:
this moves the turbines physical location endpoint to under projects along with the turbine changes

/projects/{name}/turbine-changes
/projects/turbines/{name}

This sets up for outlets to follow suit:

/projects/{name}/outlet-changes (or gate-changes?)
/projects/outlets/{name}

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
@@ -37,7 +37,9 @@ import cwms.cda.data.dto.CwmsDTO;
 import cwms.cda.data.dto.CwmsDTOValidator;
 import cwms.cda.data.dto.CwmsId;
 import cwms.cda.data.dto.LookupType;
+import cwms.cda.formatters.FormattingException;
 import java.time.Instant;
+import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
 
@@ -223,10 +225,14 @@ public abstract class PhysicalStructureChange<T extends Setting> extends CwmsDTO
             return self();
         }
 
-        public B withSettings(Set<T> settings) {
+        public B withSettings(List<T> settings) {
             this.settings = new TreeSet<>(comparing(Setting::getLocationId,
                 comparing(CwmsId::getOfficeId).thenComparing(CwmsId::getName)));
-            this.settings.addAll(settings);
+            boolean receivedDuplicates = this.settings.addAll(settings);
+            if(receivedDuplicates) {
+                throw new FormattingException(
+                    "Received duplicate settings. Only a single setting per location per timestep is supported");
+            }
             return self();
         }
         
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
@@ -160,8 +160,9 @@ public abstract class PhysicalStructureChange<T extends Setting> extends CwmsDTO
         private Double tailwaterElevation;
         private String elevationUnits;
         private String notes;
-        private Set<T> settings;
         private Instant changeDate;
+        private final Set<T> settings = new TreeSet<>(comparing(Setting::getLocationId,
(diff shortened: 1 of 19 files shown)


adamkorynta:
based on: https://github.com/USACE/cwms-data-api/pull/760

rma-rripken (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
```diff
@@ -157,8 +146,8 @@ protected void validateInternal(CwmsDTOValidator validator) {
         validator.validateRequiredFields(this);
         validator.validateCollection(getSettings());
     }
-
-    public static final class Builder<T extends Setting> {
+    
+    public abstract static class Builder<B extends Builder<B, T>, T extends Setting> {
```
MikeNeilson, line 150: I can be a little annoying to type been can be really helpful when reading code to use a generic name that's distruptive. Like I have no idea what T is here until I go look at the TurbineChange class

MikeNeilson: I usually use an otherwise normal variable name but then end it with "Type". Imagine reading this with https://github.com/opendcs/gherkin-to-kiwi/blob/main/lib/src/main/java/org/opendcs/testing/util/FailableResult.java with just S and F... or the way some people do it L and R.

MikeNeilson (approved):

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
adamkorynta: I'm not totally on board there. The Java convention explicitly call out the conventions for generic naming and refute the normal word name in order to not confuse with other classes or interfaces. https://docs.oracle.com/javase/tutorial/java/generics/types.html

I'd be even more confused looking for SettingType and not finding it to then realize it's a generic.

rma-rripken: I agree that the single character naming for java generics is the standard.  I'd be willing to give meaningful short names a try where longer names might make things clearer.   An alternative is to add documentation to the class declaration which explains what the generic holds.  Sometimes the generics don't make a lot of sense and they just have to be added in order to make things work.  In those cases I'm not sure any amount of naming or documentation would help.  In my experience anything above a class that takes a pair of generics wasn't designed from scratch but written to accommodate a specific use case.   When that is true a comment providing specific concrete examples can be as helpful as a long variable name.  Something like  ```// S - The type of the Settings class which corresponds to the Change. For example, For the TurbineChange class S would be TurbineSettings.```   Or something like that...

MikeNeilson: 1. I'm going to call oracle just totally full of shit here, we write code for others to read.
2. I'm asserting this for code we create, not filling in needs for other things.

But anyways. Javadoc *does* handle those parameters at the class level, you can see that in my example above. So we should at least put the meaningful description there.

Even C++ has gotten away from single character names, at least for anything that isn't blindingly obvious (granted with a c++ template it goes way beyond what a java generic, but the point stands, it's a *name*.

RyanM-RMA: Personally, I like the generic naming convention from C#.  It's basically a single character if it's self explanatory, otherwise you prefix your generic name with T like TSettingsType.
https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#type-parameter-naming-guidelines
That probably works better for something like C++ or C# because they also prefer prefixes on class names like interfaces and abstractions.  So you're already in that environment of looking at the class name to identify what kind of thing it is.

MikeNeilson: You don't actually see prefixes in C++ much anymore, gets pretty ugly looking, but we can hash this out more at our call today... or a vigorous discussion... or probably both.

But I do like that compromise. I agree that 80-90% of the time it's usually pretty obvious like if there's an " extends ...", but sometimes it just ain't and we should prefer clarity over convention.

PR #764, opened 2024-07-05
Feature/add outlet controller

RyanM-RMA:


cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -407,6 +412,20 @@ protected void configureRoutes() {
         get("/", ctx -> ctx.result("Welcome to the CWMS REST API")
                 .contentType(Formats.PLAIN));
         // Even view on this one requires authorization
+        //Temporarily at top to speed up testing.
```
adamkorynta: Not sure how this speeds up testing.... Are you going to move it down?

cwms-data-api/src/main/java/cwms/cda/api/location/kind/BaseOutletHandler.java
```diff
@@ -0,0 +1,51 @@
...
+import com.codahale.metrics.Histogram;
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
+import cwms.cda.api.Controllers;
+import io.javalin.http.Handler;
+import static com.codahale.metrics.MetricRegistry.name;
+import static cwms.cda.api.Controllers.RESULTS;
+import static cwms.cda.api.Controllers.SIZE;
+
+abstract class BaseOutletHandler implements Handler {
```
adamkorynta, line 32: I don't see anything specific to outlets in this class

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -407,6 +412,20 @@ protected void configureRoutes() {
...
+        String outletPath = format("/projects/{%s}/outlets/{%s}", Controllers.OFFICE, NAME);
```
adamkorynta: Office should come before the name, not before outlets. This makes it seem like the project is named office

adamkorynta: I think the endpoint would be let confusing if it matched the turbine/embankments where it is just `/projects/outlets/{name}` since the project doesn't own the physical locations, the outlets are just required to be tied to a project. The project owns the virtual outlets so those paths look correct to me.

adamkorynta (changes requested):

adamkorynta:
Note from Ryan: integration tests will follow in a separate PR

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
RyanM-RMA: it speeds up my testing, and yes I will.  Thank you for pointing that out.

adamkorynta: How does it speed up your testing? I'm curious because I don't see it and would like to benefit from the speed-up

RyanM-RMA: Any time I refresh the CDA website, it puts the outlets at the top of the list of endpoints.  so I don't have to scroll down and find it in the mass of endpoints.

RyanM-RMA: It's not a code-side speed up, it's a me-side speed up, lol.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
RyanM-RMA: > Office should come before the name, not before outlets. This makes it seem like the project is named office

It stops working when you put it before name.  I can switch to the new one you pointed out - I like that.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
adamkorynta: ohhhh, gotcha

cwms-data-api/src/main/java/cwms/cda/api/location/kind/BaseOutletHandler.java
RyanM-RMA: True, it's more of a BaseHandler class.  I wasn't sure if I should introduce it as an abstract class everyone can use or not, because it seemed like the norm was to write it all out.  I can move it to the root package and make it less specific if you want.

MikeNeilson: Definitely a good way to start, but this definitely looks like something they all use. Though probably reasonable to keep the specific name until someone wants to go through refactor all of them.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
RyanM-RMA: I've updated this to use /projects/outlets/{name}

cwms-data-api/src/main/java/cwms/cda/api/location/kind/BaseOutletHandler.java
RyanM-RMA: @adamkorynta @MikeNeilson is this something that needs to be resolved?

MikeNeilson: Only if you really want to slog through all usages of `Handler.`

adamkorynta: Handler is pretty small, but CrudHandler is the big lift. I think the BaseHandler class shouldn't implement an interface that way both Handler/CrudHandler impls can use the same logic. I'd do all that in a separate PR though.

RyanM-RMA: If we do a base class agnostic of the Handler and CrudHandler interface, that's 51 classes that'll need an update.  I'm down for doing that in a separate PR, but where does that time go?  We can always create a ticket, and put that on hold until we can decide on that?

MikeNeilson: I would just create the ticket.

MikeNeilson: It'll either get approved, or one of us will just get that annoyed when we see it again.

RyanM-RMA: #811 created for this.

adamkorynta (approved):

PR #769, opened 2024-07-09
cleanup inheritance structure of CwmsDTO vs CwmsDTOBase

adamkorynta:
classes that were passing in null office ids to CwmsDTO are now using CwmsDTOBase and CwmsDTOBase is made abstract with the default validation logic

Also this updates the validator to use the Java Bean getter instead of using field access and having to modify access. Added caching to reduce performance hit in subsequent calls.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/Blob.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/Blob.java
@@ -15,6 +15,7 @@ public class Blob extends CwmsDTO
 	private String id;
 	private String description;
 	private String mediaTypeId;
+	@JsonProperty(required=true)
 	private byte[] value;
 
 	private Blob() {
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/Blob.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/Blob.java
@@ -3,12 +3,9 @@ package cwms.cda.data.dto;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
-import cwms.cda.api.errors.FieldException;
 import cwms.cda.formatters.Formats;
-import cwms.cda.formatters.FormattingException;
 import cwms.cda.formatters.annotations.FormattableWith;
 import cwms.cda.formatters.json.JsonV2;
-import cwms.cda.formatters.xml.XMLv2;
 
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
 @FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
@@ -58,11 +55,4 @@ public class Blob extends CwmsDTO
         return getOfficeId() + "/" + id + ";description=" + description;
 	}
 
-	@Override
-	protected void validateInternal(CwmsDTOValidator validator) {
-		super.validateInternal(validator);
-		validator.required(getOfficeId(), "office-id");
-		validator.required(getId(), "id");
-		validator.required(getValue(), "value");
-	}
 }
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOBase.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOBase.java
@@ -9,7 +9,6 @@ public abstract class CwmsDTOBase {
 
     public final void validate() throws FieldException {
         CwmsDTOValidator validator = new CwmsDTOValidator();
-        validator.validateRequiredFields(this);
         validateInternal(validator);
         validator.validate();
     }
@@ -19,6 +18,6 @@ public abstract class CwmsDTOBase {
      * @param validator validator that will aggregate all validation errors
      */
     protected void validateInternal(CwmsDTOValidator validator) {
-
+        validator.validateRequiredFields(this);
     }
 }
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOBase.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOBase.java
@@ -1,12 +1,24 @@
 package cwms.cda.data.dto;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-
 import cwms.cda.api.errors.FieldException;
 /*
  * Suitable base for objects not owned by an office, like a list of locations.
  * Or parameters, or the list of offices themselves.
  */
-public interface CwmsDTOBase {
-    public  void validate() throws FieldException; // additional validation logic.
+public abstract class CwmsDTOBase {
+
+    public final void validate() throws FieldException {
+        CwmsDTOValidator validator = new CwmsDTOValidator();
+        validator.validateRequiredFields(this);
+        validateInternal(validator);
+        validator.validate();
+    }
+
+    /**
+     * This method can be overridden to provide additional validation logic
+     * @param validator validator that will aggregate all validation errors
+     */
+    protected void validateInternal(CwmsDTOValidator validator) {
+
+    }
 }
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
@@ -39,7 +39,7 @@ import java.util.concurrent.Callable;
 
 
 /**
- * This class is used to validate the fields of a CwmsDTO object.
+ * This class is used to validate the fields of a CwmsDTOBase object.
  */
 public final class CwmsDTOValidator {
 
@@ -52,7 +52,7 @@ public final class CwmsDTOValidator {
     /**
      * Validates the presence of a required field in a given object or DTO.
      * Adds the field name to a set of missing fields if the value is null.
-     * If the value is an instance of CwmsDTO, it validates the CwmsDTO as well.
+     * If the value is an instance of CwmsDTOBase, it validates the CwmsDTOBase as well.
      *
      * @param value     the value of the field to be checked
      * @param fieldName the name of the field to be checked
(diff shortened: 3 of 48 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dto/Blob.java
```diff
@@ -58,11 +55,4 @@ public String toString(){
         return getOfficeId() + "/" + id + ";description=" + description;
 	}
 
-	@Override
-	protected void validateInternal(CwmsDTOValidator validator) {
-		super.validateInternal(validator);
-		validator.required(getOfficeId(), "office-id");
-		validator.required(getId(), "id");
-		validator.required(getValue(), "value");
```
RyanM-RMA, line 66: Annotate value as required

cwms-data-api/src/main/java/cwms/cda/data/dto/Pool.java
```diff
@@ -3,27 +3,84 @@
...
 import cwms.cda.formatters.annotations.FormattableWith;
 import cwms.cda.formatters.json.JsonV2;
 import cwms.cda.formatters.Formats;
 
 import usace.cwms.db.dao.ifc.pool.PoolNameType;
-import usace.cwms.db.dao.ifc.pool.PoolType;
 
 @FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class)
-public class Pool extends PoolType implements CwmsDTOBase {
+public class Pool extends CwmsDTOBase {
```
RyanM-RMA, line 11: Uses tabs, should we format?

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOBase.java
```diff
@@ -1,12 +1,24 @@
...
-import com.fasterxml.jackson.annotation.JsonProperty;
-
 import cwms.cda.api.errors.FieldException;
 /*
  * Suitable base for objects not owned by an office, like a list of locations.
  * Or parameters, or the list of offices themselves.
  */
-public interface CwmsDTOBase {
-    public  void validate() throws FieldException; // additional validation logic.
+public abstract class CwmsDTOBase {
```
MikeNeilson, line 8: FYI, interfaces can have default methods now. even in Java 8.

MikeNeilson (commented):
Why abstract vs interface with default? No instance variables are required, just the reference to this with the method would have.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
```diff
@@ -43,7 +42,7 @@
 import java.util.Set;
 import java.util.TreeSet;
 
-public abstract class PhysicalStructureChange<T extends Setting> extends CwmsDTO {
+public abstract class PhysicalStructureChange<T extends Setting> extends CwmsDTOBase {
```
RyanM-RMA, line 45: Does the `validateInternal` method here need to be updated?

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOBase.java
adamkorynta: yeah.... the intention of that feature is to provide ways to modify interfaces without breaking backwards compatibility. Given our codebase is self-contained, the compatibility thing isn't an issue. Additionally, you can't mark default methods as final and we don't want the entire validate() method logic to be overwritten (until we come up with a good exception of course), but we can still provide an extension point through the validateInternal() method. Might be sticking too close to the open-closed principle there.

RyanM-RMA (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dto/Pool.java
MikeNeilson: Yeah, basically if it it isn't a huge imposition fix the formatting.

Basically use judgement. The Github view is good about hiding whitespace changes so it's generally safe to convert tabs to spaces and remove extra spaces at EOL without making review difficult.

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOBase.java
MikeNeilson: Ah, that makes sense.

cwms-data-api/src/main/java/cwms/cda/data/dto/Pool.java
adamkorynta: yeah really easy to do in DTO's like this given there isn't logic to check. Reformatted

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
adamkorynta: yes, fixed

RyanM-RMA (approved):

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
```diff
@@ -29,85 +29,118 @@
...
 
 /**
- * This class is used to validate the fields of a CwmsDTO object.
+ * This class is used to validate the fields of a CwmsDTOBase object.
  */
 public final class CwmsDTOValidator {
 
     private static final FluentLogger LOGGER = FluentLogger.forEnclosingClass();
-
+    private static final Map<Class<?>, List<Method>> REQUIRED_FIELD_GETTERS = new ConcurrentHashMap<>();
```
rma-psmorris, line 57: Did you consider using Java References here to allow for garbage collection?

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
```diff
@@ -29,85 +29,118 @@
...
+                            retval.add(readMethod);
                         }
                     }
                 }
             }
-        } catch (IllegalAccessException e) {
-            LOGGER.atWarning().withCause(e).log("Unable to validate required fields are non-null in DTO: " + type);
+        } catch (IntrospectionException e) {
+            LOGGER.atWarning().withCause(e)
+                .log("Unable to validate required field are non-null in DTO: %s", type);
```
rma-psmorris, line 123: "Unable to validate required field are non-null in DTO" --  "required fields are" or "required field is"?

cwms-data-api/src/main/java/cwms/cda/data/dto/LookupType.java
```diff
@@ -25,23 +25,23 @@
...
 import cwms.cda.formatters.json.JsonV1;
 
-import java.util.Objects;
-
 @FormattableWith(contentType = Formats.JSON, formatter = JsonV1.class)
 @JsonDeserialize(builder = LookupType.Builder.class)
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
-public final class LookupType implements CwmsDTOBase {
+public final class LookupType extends CwmsDTOBase {
```
rma-psmorris, line 40: This could be a CwmsDTO?

cwms-data-api/src/main/java/cwms/cda/data/dto/Pool.java
```diff
@@ -1,246 +1,249 @@
...
-	@Override
-	public void validate() throws FieldException {
-		// TODO Auto-generated method stub
-
-	}
-
-
+public class Pool extends CwmsDTOBase {
+
+    private final PoolNameType poolName;
```
rma-psmorris, line 13: Recommend this dependency be changed. No longer use PoolNameType and instead add its attributes to Pool. If not possible, then introduce a new PoolId class that has the office and pool name variables.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeZone.java
```diff
@@ -20,30 +21,26 @@
 @FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
 @FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
-public final class TimeZone extends CwmsDTO
-{
-	private String timeZone;
+public final class TimeZone extends CwmsDTOBase {
```
rma-psmorris, line 24: Recommend changing this class name to be TimeZoneName or TimeZoneId to clarify it as not having TimeZone business rules and separate it from the java class sharing the same name.

cwms-data-api/src/main/java/cwms/cda/data/dto/Unit.java
```diff
@@ -23,7 +23,7 @@
 @FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
 @FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
-public final class Unit extends CwmsDTO
+public final class Unit extends CwmsDTOBase
```
rma-psmorris, line 26: I'm not certain if this class was implemented with knowledge of the database notion of DisplayUnits? See source for AV_DISPLAY_UNIT which brings in office specific parameter units

rma-psmorris (commented):
Post-merge feedback.

cwms-data-api/src/main/java/cwms/cda/data/dto/Pool.java
adamkorynta: Added as an item to accomplish in: https://github.com/USACE/cwms-data-api/issues/786

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeZone.java
adamkorynta: Renamed to `TimeZoneId` in: https://github.com/USACE/cwms-data-api/pull/787

cwms-data-api/src/main/java/cwms/cda/data/dto/Unit.java
adamkorynta: @RyanM-RMA - I'll let you chime in here.

cwms-data-api/src/main/java/cwms/cda/data/dto/LookupType.java
adamkorynta: Updated LookupType and Property: https://github.com/USACE/cwms-data-api/pull/788
VerticalDatumInfo is in a similar boat, but unfortunately uses "office" instead of "office-id" so I'm leaving it for backwards compatibility.

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
adamkorynta: Addressed in: https://github.com/USACE/cwms-data-api/pull/789

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsDTOValidator.java
adamkorynta: yeah, I did but there would be no added benefit. the classloader already caches the Class objects so there is no needed Soft/Weak reference here. Additionally, the Methods are strongly referenced by the Class so reference doesn't help here either.

cwms-data-api/src/main/java/cwms/cda/data/dto/Unit.java
RyanM-RMA: It was not implemented with that knowledge.  The table and view it's using, AT_UNIT_ALIAS, and AV_UNIT, both contain office IDs, so we could make a ticket to update them to include those office ids.  Seems a bit out of scope for this work though.

RyanM-RMA: Actually, looking at the original output from this (it was using CWMS_CAT.RETRIEVE_UNITS_F before), that output didn't include the office id, which is probably why it was ignored in the changes I made.

rma-psmorris: The importance of office ID as used in display vs the non-office ID CWMS defaults is that the unit will be the one that the district has determined to be their default. @RyanM-RMA - How is that out of scope?

RyanM-RMA: I say out of scope because it's related to recent maintenance ticket work, not our fixed cost task.

RyanM-RMA: This is the ticket for that work.
https://github.com/USACE/cwms-data-api/issues/498

PR #773, opened 2024-07-11
add content "type" to error message on deserialization

adamkorynta:


MikeNeilson (commented):
Have we confirmed that the content type passes through sanitation first?

It's arbitrary user input sent back to the user so it needs to be sanitized.

adamkorynta:
The newly added message pieces are the class types and not user input. The templating to the unsupported ones in the interface use the getContentType() method from OutputFormatter and all implementations use constants rather than user input. 

The actual content itself is not sanitized and is passed in the message, seems useful for logs, but I don't see much need to pass that back to the client. I didn't want to change that behavior in this PR though, just addressing Peter's comment from: https://github.com/USACE/cwms-data-api/pull/760#discussion_r1672741196. I also think that if we wanted to figure out which invalid content isn't getting deserialized, we could turn up some logging rather than always putting into the exceptions so this ends up being redundant.

MikeNeilson (approved):

rma-psmorris (approved):
Looks good, not sure if you have more conversation with Mike on his feedback

PR #780, opened 2024-07-12
Update sections of build for Integration test automation

MikeNeilson:
Okay, so now this contains actual changes to support the build running correctly.


MikeNeilson:
The next build, it's in the queue, should provide "correct" failures. Though I think some are still just because the test isn't done yet and hasn't been marked disabled.

adamkorynta (approved):

PR #787, opened 2024-07-15
rename TimeZone class names to TimeZoneId

adamkorynta:
this prevents confusion with `java.util.TimeZone` and clarifies that only the id is collected


rma-rripken (approved):

rma-psmorris:
Post merge review: approved, looks good.

PR #788, opened 2024-07-15
move lookuptype and property to inherit from CwmsDTO

adamkorynta:
these two classes have an office id so they should inherit from CwmsDTO instead of CwmsDTOBase


rma-rripken (approved):

rma-psmorris (commented):
Post merge review. Approved, looks good.

PR #793, opened 2024-07-17
Added fetchSize to Location getAll query.

rma-rripken:
Suggested by @krowvin in post-CURG feedback.
Possible 40% speed improvement on a synthetic test.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
@@ -119,9 +119,10 @@ public class LocationsDaoImpl extends JooqDao<Location> implements LocationsDao 
         }
 
         return dsl.select(AV_LOC.asterisk())
-                .from(AV_LOC)
-                .where(whereCondition)
-                .fetch(this::buildLocation);
+                    .from(AV_LOC)
+                    .where(whereCondition)
+                    .fetchSize(500)
+                    .fetch(this::buildLocation);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


rma-rripken:
With this change I'm seeing getAll with 1000 locations taking 600 ms.    About 60% of that time is in db fetch and 40% in JSON serialization.  Its not obvious to me how to make the JSON serialization faster.
image

adamkorynta (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
adamkorynta, line 1: Does this also speed up the catalog endpoint?

rma-rripken: No.  The final query in the loc catalog includes fetchSize so this is already being done there.

krowvin: We also talked about how if users have a map they may just want to use a preconfigured `location-group` of the gages they would like to fetch. 

I know we want to be designing things in a database agnostic method. Meaning we do not want to lock ourselves into the constructs of Oracle like using the JSON methods available in Oracle. 

I also know we want to use the procedure calls already defined. But if those are what are building the string does it not make sense to at least test and/or start work on a new set of SQL/Procedures for when we *do* go database agnostic? Because that would be the time we'd have to do rewrites anyways from PLSQL. 

What a big conversation i'm attempting to have in a thread of a commit.

Found this though, you guys privvy to it? 
https://www.jooq.org/doc/latest/manual/sql-execution/exporting/exporting-json/

PR #794, opened 2024-07-17
Added static version cache to Dao.

rma-rripken:
Also moved timeseries getDbVersion call out of the catalog forEach loop.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
@@ -29,6 +29,8 @@ import cwms.cda.data.dto.TsvId;
 import cwms.cda.data.dto.VerticalDatumInfo;
 import cwms.cda.data.dto.catalog.CatalogEntry;
 import cwms.cda.data.dto.catalog.TimeseriesCatalogEntry;
+import cwms.cda.formatters.FormattingException;
+import cwms.cda.formatters.xml.XMLv1;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.sql.Connection;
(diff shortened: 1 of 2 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -566,7 +564,7 @@ public Catalog getTimeSeriesCatalog(String page, int pageSize, CatalogRequestPar
                         .units(row.get(AV_CWMS_TS_ID.AV_CWMS_TS_ID.UNIT_ID))
                         .interval(row.get(AV_CWMS_TS_ID.AV_CWMS_TS_ID.INTERVAL_ID))
                         .intervalOffset(row.get(AV_CWMS_TS_ID.AV_CWMS_TS_ID.INTERVAL_UTC_OFFSET));
-                if (this.getDbVersion() > Dao.CWMS_21_1_1) {
+                if (dbVersion > Dao.CWMS_21_1_1) {
```
adamkorynta, line 567: This clause can get removed, all dbs are now newer versions according to confluence doc.

adamkorynta (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
MikeNeilson: True, but we should actually formalize when that happens a bit more. Totally okay in this case but kinda of have the same issue in OpenDCS, can't keep code around to support everything forever without a rather high maintenance burden.

MikeNeilson (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-rripken: I'll merge this and then remove the check in a new pr.

PR #797, opened 2024-07-18
fix tomcat download link to go to archive instead of the latest release

adamkorynta:
docker build was giving a 404 due to new 9.0.91 release


MikeNeilson (dismissed):
Probably a good idea to also switch it to 9.0.91.. plus in the libs toml file for tests.

adamkorynta:
I can do that, wasn't sure if we waited for districts to update before targeting new versions.

rma-rripken (dismissed):

rma-rripken (approved):

rma-rripken (approved):

PR #808, opened 2024-07-19
Adjusting outlet location deletion process

RyanM-RMA:
Making sure we delete locations, and don't fail if it doesn't exist.

### DIFF
--- a/cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/OutletDaoIT.java
+++ b/cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/OutletDaoIT.java
@@ -95,11 +95,7 @@ class OutletDaoIT extends ProjectStructureDaoIT {
 
     static void storeLocation(DSLContext context, Location loc) throws IOException {
         LocationsDaoImpl locationsDao = new LocationsDaoImpl(context);
-        try {
-            locationsDao.deleteLocation(loc.getName(), loc.getOfficeId(), true);
-        } catch (NotFoundException e) {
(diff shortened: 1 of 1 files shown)


adamkorynta (dismissed):

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/OutletDaoIT.java
```diff
@@ -82,22 +84,40 @@ static void setup() throws Exception {
...
+            locationsDao.deleteLocation(locId, officeId, true);
+        } catch (NotFoundException ex) {
+            LOGGER.atFinest().withCause(ex).log("No data found for " + officeId + "." + locId);
+        }
+    }
+
+    static void storeLocation(DSLContext context, Location loc) throws IOException {
+        LocationsDaoImpl locationsDao = new LocationsDaoImpl(context);
+        try {
+            locationsDao.deleteLocation(loc.getName(), loc.getOfficeId(), true);
```
MikeNeilson: why not just call the above function?

MikeNeilson (changes requested):

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/OutletDaoIT.java
RyanM-RMA: Oh good point.  I think I made that method after this and just didn't think about it.  Updated, thank you

MikeNeilson (approved):

Issue #570, opened 2024-03-07
Deprecate the "timezone" query parameter for v2 and newer end-points


rma-rripken:
As described in https://github.com/USACE/cwms-data-api/issues/556#issuecomment-1984710644  the "timezone" parameter is a little bit weird when used in JSONv2 and XMLv2.   It basically only used as a way for the user to specify a default timezone to be used if the user didn't include timezone info in their "begin" and  "end" strings.

It makes it so the user could specify 
begin = 2022-01-06T00:00:00
end = 2022-01-06T12:00:00
timezone = UTC
or 
begin = 2022-01-06T00:00:00
end = 2022-01-06T12:00:00
timezone = America/Los_Angeles

I think the V2 end-points sort of inherited this behavior.   But its too helpful.

The V1 end-points also passed timezone into the pl/sql and the pl/sql promised to return values in the specified timezone.

The V2 end-points return data in UTC.  Maybe V2 isn't always UTC but I can't think of when it isn't and if it isn't we need to document that better.

Instead of continuing to "inherit" the old behavior (in my opinion) we should force the clients to give us better inputs (timestamps with tz) and throw errors if the inputs seem weird.

DanielTOsborne:
Downside of this is that on tomcat by default, the square brackets [] are not allowed in the URL query string.
Also, as I learned after I wrote that, it's non-trivial to format/parse that time format in Javascript.

My recommendation would be to only allow naïve datetimes if the timezone parameter is used, and throw an error if the timezone is specified in both places. Additionally, enforce the start and end times to be the same format, throw an error if not.
In practice, that's how I ended up using it anyway.

I made it lax in what was allowed, but that wasn't the best idea. It should be strict, but still flexible for the user _where it makes sense_ (and allowing tz ambiguity does not make sense).

krowvin:
Would they just need to be uri encoded? 

I. E.
`%5B...%5D`

Unless you're saying tomcat won't handle that either? 

Edit: agreed, I too expected time zone to be sufficient to prevent end users having to deal with it. 

[The browser will return the time zone string no problem. ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTimeZones)

DanielTOsborne:
> Would they just need to be uri encoded?
> 
> I. E. `%5B...%5D`
> 
> Unless you're saying tomcat won't handle that either?
> 
[...]

PR #813, opened 2024-07-23
Adding a test to verify parsing of the timestamp example used in Time…

rma-rripken:
…seriesController and basically every Controller.

### DIFF
--- a/cwms-data-api/src/test/java/cwms/cda/helpers/DateUtilsTest.java
+++ b/cwms-data-api/src/test/java/cwms/cda/helpers/DateUtilsTest.java
@@ -2,8 +2,10 @@ package cwms.cda.helpers;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import cwms.cda.api.Controllers;
 import cwms.cda.data.dto.TimeSeries;
 import java.time.DateTimeException;
 import java.time.Instant;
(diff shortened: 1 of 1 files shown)


rma-rripken:
@krowvin What examples are failing to parse?  

I did some grepping of the code and as far as I can tell:
Every Instant parsed in a Controller goes thru one of:
Controllers.requiredInstant which goes thru queryParamAsZdt
Or queryParamAsInstant which goes thru queryParamAsZdt
Or DateUtil.parseUserDate(...).toIInstant()

Every ZonedDateTime in a Controller goes thru:
Controllers.requiredZdt which calls queryParamAsZdt
or queryParamAsZdt which calls DateUtils.parseUserDate
or DateUtils.parseUserDate

MikeNeilson:
The example provided in the swagger ui was something that didn't validate.

krowvin:
> The example provided in the swagger ui was something that didn't validate.  

Yes, this

But also in the swagger UI there are date fields that do not provide an expected format. In the example Value it just lists the primitive `string`. 

EDIT: Example is the Levels endpoint here (older version) https://cwms-data.usace.army.mil/cwms-data/swagger-ui.html
image

For example the levels endpoint takes an effective-date. Without telling users explicitly what date to type in; it leaves them to enter a variety of dates until they get it. 

Nevermind the other level issues that exist existed prior to recent changes/database fixes/data. 

And maybe it's just that one date field! I tried to think of a clever way to search for all the various annotations, but gave up.

rma-rripken:
Ok.  Mike's quote is that it "was" something - not that it still is?  Right?  As far as I can see the date parsing is currently fine and the documentation example is fine.   This was brought up under the "issues" agenda item so I want to be clear whether its currently an issue or maybe its just an issue for older CDA sites?

With regard to a good example not being provided in Swagger-ui and just seeing "string" - I don't think we are able to change that with the javalin version we currently have to use.  Which is one reason we started putting examples in the field descriptions.

In this issue I suggested we create a timestamp documentation page and link to it from every timestamp type end-point. Similar to how we've handled regex fields.  My questions and suggestion got no response and the issue is still tagged needs-approval
https://github.com/USACE/cwms-data-api/issues/570



> Nevermind the other level issues that exist existed prior to recent changes/database fixes/data. 

I'm not sure what is meant by this - if you know of level issues I hope you report, create issues and track them.

krowvin:
>  I hope you report  

The issues are known and the latest release will fix it

As for the tests I will close this issue as it appears you were testing CDA's internal handling of the datetimes and not just presentation on swagger pages. 

An additional docs page on datetimes and the formats that can be pointed to would certainly solve this. And perhaps anywhere a date is set for input in the swagger page we can reference the new doc. [docs here]etc

krowvin (approved):
Looks good. just doing a spot check of the dates to confirm the generated date matches the expected.

PR #816, opened 2024-07-26
Pump accounting controller

zack-rma:
Implements water supply pump accounting controller and tests.


cwms-data-api/src/main/java/cwms/cda/api/watersupply/AccountingCatalogController.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    @NotNull
+    protected WaterSupplyAccountingDao getWaterSupplyAccountingDao(DSLContext dsl) {
+        return new WaterSupplyAccountingDao(dsl);
+    }
+
+    @OpenApi(
+        queryParams = {
+            @OpenApiParam(name = START, description = "The start time of the time window for "
```
adamkorynta, line 98: Should use the same language as other time parameters. Also, these defaults would retrieve a huge amount of data. I'd just make the two fields required. This is also missing the time zone field.

cwms-data-api/src/main/java/cwms/cda/api/watersupply/AccountingCatalogController.java
```diff
@@ -0,0 +1,202 @@
...
+                + "inclusive or not. Defaults to TRUE.", type = Boolean.class),
+            @OpenApiParam(name = END_TIME_INCLUSIVE, description = "Whether or not the end time is inclusive "
+                + "or not. Defaults to TRUE.", type = Boolean.class),
+            @OpenApiParam(name = ASCENDING, description = "Whether or not the entries should be returned "
+                + "in ascending order. Defaults to TRUE.", type = Boolean.class),
+            @OpenApiParam(name = ROW_LIMIT, description = "The maximum number of rows to return. "
+                + "Defaults to 0, which means no limit.", type = Integer.class)
+        },
+        pathParams = {
+            @OpenApiParam(name = OFFICE, description = "The office ID the pump accounting is associated with.",
```
adamkorynta: This is the project's office right?

cwms-data-api/src/main/java/cwms/cda/api/watersupply/AccountingCatalogController.java
```diff
@@ -0,0 +1,202 @@
...
+                ctx.status(HttpServletResponse.SC_NOT_FOUND).json(error);
+                return;
+            }
+
+            Instant startInstant = DateUtils.parseUserDate(startTime, "UTC").toInstant();
+            Instant endInstant = DateUtils.parseUserDate(endTime, "UTC").toInstant();
+
+            WaterSupplyAccountingDao waterSupplyAccountingDao = getWaterSupplyAccountingDao(dsl);
+            List<WaterSupplyAccounting> accounting = waterSupplyAccountingDao.retrieveAccounting(contractId, waterUser,
+                    projectLocation, null, startInstant, endInstant, startInclusive, endInclusive,
```
adamkorynta: What is the null parameter here? Since this is the only usage of the method, there shouldn't be nulls

cwms-data-api/src/main/java/cwms/cda/api/watersupply/AccountingCreateController.java
```diff
@@ -0,0 +1,164 @@
...
+            final String office = ctx.pathParam(OFFICE);
+            DSLContext dsl = getDslContext(ctx);
+            String formatHeader = ctx.header(Header.ACCEPT) != null ? ctx.header(Header.ACCEPT) : Formats.JSONV1;
+            ContentType contentType = Formats.parseHeader(formatHeader, WaterSupplyAccounting.class);
+            ctx.contentType(contentType.toString());
+            WaterSupplyAccounting accounting = Formats.parseContent(contentType, ctx.body(),
+                    WaterSupplyAccounting.class);
+            WaterSupplyAccountingDao waterSupplyAccountingDao = getWaterSupplyAccountingDao(dsl);
+            LookupTypeDao lookupTypeDao = new LookupTypeDao(dsl);
+            List<LookupType> lookupList = lookupTypeDao
```
adamkorynta, line 122: The database doesn't already perform these checks in the store call?

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/watersupply/AccountingCatalogController.java
zack-rma: Updated language to better match other controllers. Made fields required, added time zone.

cwms-data-api/src/main/java/cwms/cda/api/watersupply/AccountingCreateController.java
zack-rma: I double checked, it seems the database will drop any data points that have a transfer type that does not exist when storing accounting data. No error is thrown, those values are just not stored and do not come back upon retrieval.

cwms-data-api/src/main/java/cwms/cda/api/watersupply/AccountingCatalogController.java
zack-rma: Yes, updated the language to reflect that

cwms-data-api/src/main/java/cwms/cda/api/watersupply/AccountingCatalogController.java
zack-rma: This is units, I added this in as a controller parameter.

cwms-data-api/src/main/java/cwms/cda/api/watersupply/AccountingCreateController.java
adamkorynta: This should get logged as a ticket in Jira for the database schema. Can you add this as a comment in the CDA code?

zack-rma: Added a comment and created an issue ticket: https://jira.hecdev.net/browse/CWDB-316

adamkorynta (approved):

PR #825, opened 2024-08-02
bugfixes for basin controller

adamkorynta:
standardize on delete "method" query parameter name fix getOne serialization that was using an array
fix rename to not require a body


cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
```diff
@@ -345,6 +344,18 @@ public static String requiredParam(io.javalin.http.Context ctx, String name) {
         return param;
     }
 
+    /**
+     * Returns the first matching query param or throws RequiredQueryParameterException.
+     * @param ctx Request Context
+     * @param name Query parameter name
+     * @return value of the parameter
+     * @throws RequiredQueryParameterException if the parameter is not found
+     */
```
rma-bryson, line 353: Should we add a unit test for this?

cwms-data-api/src/main/java/cwms/cda/api/BasinController.java
```diff
@@ -312,19 +309,13 @@ public void create(@NotNull Context ctx) {
...
                 .withOfficeId(ctx.queryParam(OFFICE))
                 .build();
-        cwms.cda.data.dto.basin.Basin retBasin = basinDao.getBasin(basinId, "EN");
-        if (retBasin == null) {
-            CdaError error = new CdaError("No matching basin " + name);
-            ctx.status(HttpServletResponse.SC_NOT_FOUND).json(error);
-            return;
-        }
-        basinDao.deleteBasin(basinId, deleteMethod);
+        basinDao.deleteBasin(basinId, deleteMethod.getRule().getRule());
```
rma-bryson: getRule().getRule() -> Worth renaming one of those for clarity?

adamkorynta: updated to use DeleteRule in the DAO for consistency

cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
adamkorynta: good call, added

rma-bryson (approved):

PR #826, opened 2024-08-02
Bugfix/client integration test fixes

adamkorynta:
fixes client integration test failures


MikeNeilson (commented):
Looks reasonable thus far.

rma-rripken (approved):

Issue #786, opened 2024-07-15
Remove dependencies on hec-cwms-data-access


adamkorynta:
The hec-cwms-data-access library depends on CDA through the CDA Client Java API which creates a cyclic dependency between code repositories. There are only a few references left that need to be addressed. This effort should likely get split into separate Issues when ready to work on (or at least separate PRs).

OracleTypeMap - this is a utility class transforming data types like Boolean to "T"/"F" strings and Date to Timestamp
PoolTypeName - this is essentially a CwmsId, but would break backwards compatibility if the naming changed
CwmsDbLevel- the jOOQ query is easily replaced by calling against the codegen directly
CwmsDbLoc- the jOOQ query is easily replaced by calling against the codegen directly
CwmsDbTs- the jOOQ query is easily replaced by  calling against the codegen directly
CwmsDbStreamJooq - the jOOQ query is easily replaced by calling against the codegen directly

adamkorynta:
awesome that this got done!

PR #828, opened 2024-08-07
preliminary code for adding reservoir lock support

adamkorynta:
currently missing fields for the Lock DTO that wil be resolved once https://jira.hecdev.net/browse/CTO-147 is completed

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
@@ -110,4 +110,19 @@ public abstract class Dao<T> {
 
     public abstract Optional<T> getByUniqueName(String uniqueName, String office);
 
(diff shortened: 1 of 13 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
zack-rma, line 1: OracleTypeMap formatBool has the same functionality

adamkorynta: we want to stop using OracleTypeMap so that we don't need to pull in the cwms-db-dao dependencies.

adamkorynta: https://github.com/USACE/cwms-data-api/issues/786

PR #834, opened 2024-08-08
Added shared loc alias ID handling to Outlet DTO and DAO

zack-rma:
Tested with recent DB changes, tested using values entered with CWMSVue


cwms-data-api/src/test/resources/cwms/cda/data/dto/location/kind/base_outlet.json
```diff
@@ -22,5 +22,6 @@
   "rating-group-id": {
     "office-id": "SPK",
     "name": "Rating-BIGH-TG1"
-  }
+  },
+  "shared-loc-alias-id": "rating spec id"
```
adamkorynta: This should be an example of a rating spec if

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -105,7 +103,6 @@ public static void tearDown() throws Exception {
         tearDownProject();
     }
 
-    @Disabled("Disabled due to a DB issue.  See https://jira.hecdev.net/browse/CWDB-296")
```
adamkorynta: This should stay disabled until the integration tests point to a production database version with the bug fix. Currently the fix is only in our cda_staging branch of the schema.

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
```diff
@@ -236,4 +240,22 @@ public void renameOutlet(String officeId, String oldOutletId, String newOutletId
             CWMS_OUTLET_PACKAGE.call_RENAME_OUTLET(DSL.using(conn).configuration(), oldOutletId, newOutletId, officeId);
         });
     }
+
+    private static class LocGroupReturn {
```
adamkorynta: Unneeded maintenance burden adding an inner class like that, your method can just return the LocationGroup itself and let the mapToOutlet function pull out the necessary parts

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -216,6 +216,95 @@ public List<LocationGroup> getLocationGroups(@Nullable String officeId,
...
+                        .and(alcg.LOC_GROUP_ID.eq(alga.GROUP_ID)));
+
+
+        Condition condition = noCondition();
+        if (locCategoryLike != null && !locCategoryLike.isEmpty()) {
+            condition = caseInsensitiveLikeRegex(alcg.LOC_CATEGORY_ID, locCategoryLike);
+        }
+
+        if (sharedLocAliasLike != null && !sharedLocAliasLike.isEmpty()) {
+            condition = condition.and(caseInsensitiveLikeRegex(alcg.SHARED_LOC_ALIAS_ID, sharedLocAliasLike));
```
RyanM-RMA: We need to filter on the SHARED_REF_LOCATION_ID not the SHARED_LOC_ALIAS_ID.  the SHARED_LOC_ALIAS_ID is the rating spec, the SHARED_REF_LOCATION_ID is the parent location.

RyanM-RMA: Also rename the sharedLocAliasLike parameter

cwms-data-api/src/test/resources/cwms/cda/data/dto/location/kind/base_outlet.json
zack-rma: A leftover from testing, updated to example rating spec id

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Disabled test again

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
zack-rma: Reworked to return LocationGroup object, mapping occurs in mapToOutlet

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/Outlet.java
```diff
@@ -36,18 +36,20 @@
 @JsonDeserialize(builder = Outlet.Builder.class)
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
-@JsonPropertyOrder({"projectId", "location", "ratingCatId", "ratingGroupId"})
+@JsonPropertyOrder({"projectId", "location", "ratingCatId", "ratingGroupId", "sharedLocAliasId"})
 public final class Outlet extends ProjectStructure {
     public static final String RATING_LOC_GROUP_CATEGORY = "Rating";
     private final CwmsId ratingCategoryId;
     private final CwmsId ratingGroupId;
+    private final String sharedLocAliasId;
```
RyanM-RMA: I'd call this ratingSpecId

adamkorynta (changes requested):
looks good, make sure to add tests

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
```diff
@@ -184,15 +182,21 @@ public void deleteVirtualOutlet(String officeId, String projectId, String virtua
...
     }
 
-    private Outlet mapToOutlet(PROJECT_STRUCTURE_OBJ_T projectStructure,
+    private Outlet mapToOutlet(PROJECT_STRUCTURE_OBJ_T outlet,
                                List<LocationGroup> groups) {
-        Location location = LocationUtil.getLocation(projectStructure.getSTRUCTURE_LOCATION());
-        CwmsId ratingGroupId = getRatingGroupId(location.getName(), groups);
-        CwmsId projectId = LocationUtil.getLocationIdentifier(projectStructure.getPROJECT_LOCATION_REF());
+        Location location = LocationUtil.getLocation(outlet.getSTRUCTURE_LOCATION());
+        LocationGroup locGroupReturn = getRatingGroupId(location.getName(), groups);
```
adamkorynta: getRatingId can return null, that needs to get handled

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
```diff
@@ -78,24 +79,21 @@ public Outlet retrieveOutlet(String officeId, String locationId) {
         });
     }
 
-    private CwmsId getRatingGroupId(String locationId, List<LocationGroup> groups) {
+    private LocationGroup getRatingGroupId(String locationId, List<LocationGroup> groups) {
```
RyanM-RMA: Rename to getRatingGroup

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -216,6 +216,95 @@ public List<LocationGroup> getLocationGroups(@Nullable String officeId,
...
+    /**
+     * Get all location groups for a given office and category,
+     * as well as a where clause to filter the shared_ref_location_id.
+     * @Param officeId The office id to use for the query.
+     * @Param locCategoryLike A regex to use to filter the location categories.  May be null.
+     * @Param sharedLocAliasLike A where clause to filter the shared_loc_alias_id.  May be null.
+     * @return A list of all location groups for the given parameters.
+     */
+
+    public List<LocationGroup> getLocationGroups(String officeId, String locCategoryLike, String sharedLocAliasLike) {
```
adamkorynta: where is this new method getting called?

zack-rma: retrieveOutletsForProject in OutletDao

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
```diff
@@ -184,15 +182,21 @@ public void deleteVirtualOutlet(String officeId, String projectId, String virtua
...
 
-    private Outlet mapToOutlet(PROJECT_STRUCTURE_OBJ_T projectStructure,
+    private Outlet mapToOutlet(PROJECT_STRUCTURE_OBJ_T outlet,
                                List<LocationGroup> groups) {
-        Location location = LocationUtil.getLocation(projectStructure.getSTRUCTURE_LOCATION());
-        CwmsId ratingGroupId = getRatingGroupId(location.getName(), groups);
-        CwmsId projectId = LocationUtil.getLocationIdentifier(projectStructure.getPROJECT_LOCATION_REF());
+        Location location = LocationUtil.getLocation(outlet.getSTRUCTURE_LOCATION());
+        LocationGroup locGroupReturn = getRatingGroupId(location.getName(), groups);
+        CwmsId ratingGroupId = locGroupReturn == null ? null
```
RyanM-RMA: Rework this so it's not using ternaries.

RyanM-RMA (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
adamkorynta: oh the projecId is the shared alias loc

RyanM-RMA: I made a comment on that, it's actually the shared reference location id

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
adamkorynta: or is it allowed to be null?

RyanM-RMA: it shouldn't be null, but there's no guarantee, and if it's null then the group id and spec should be null.  he handles it correctly, but I think we can use an optional here instead.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
zack-rma: Renamed parameter

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/Outlet.java
zack-rma: Renamed as suggested

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
zack-rma: Renamed as suggested

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
zack-rma: Switched to Optional

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -187,6 +206,121 @@ void test_outlet_rename() {
         ;
     }
 
+    @Test
+    void test_rating_spec_id() {
+        // Structure of test:
+        // 1) Create the Rating Location group
```
RyanM-RMA: I'd expect this organized like this:

1. Create the outlet
2. Retrieve the outlet, and verify that the rating-spec-id is null
3. Retrieve the location group using the outlet's category and group id, deserialize the body to an object
4. Modify the location group shared-loc-alias-id, then store it back to the database
5. Retrieve your outlet, and verify the rating-spec-id comes back as expected
6. Delete the outlet
7. Delete the location group

This verifies that:
- The automatic DB location group creation for outlets is working as expected, and starts off with null
- When you change the location group's shared-loc-alias-id it's reflected in the outlet

We don't need to verify that the outlet was deleted, that's already part of the crud testing.
Don't validate the behavior of the location groups API, it's already tested in other integration tests.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -96,6 +114,7 @@ public static void tearDown() throws Exception {
 
             outletDao.deleteOutlet(EXISTING_CONDUIT_GATE.getOfficeId(), EXISTING_CONDUIT_GATE.getName(),
                                    DeleteRule.DELETE_ALL);
+            deleteLocation(context, RATED_OUTLET_LOCATION.getOfficeId(), RATED_OUTLET_LOCATION.getName());
```
RyanM-RMA: Add location group deletion just in case.  Use the ID in the NEW_RATED_OUTLET object.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -80,6 +97,7 @@ public static void setup() throws Exception {
                 deleteLocation(context, RENAMED_CONDUIT_GATE.getOfficeId(), RENAMED_CONDUIT_GATE.getName());
                 storeLocation(context, NEW_CONDUIT_GATE_2);
                 storeLocation(context, EXISTING_CONDUIT_GATE);
+                storeLocation(context, RATED_OUTLET_LOCATION);
```
RyanM-RMA: Add location group deletion just in case.  Use the ID in the NEW_RATED_OUTLET object.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -54,20 +57,34 @@ class OutletControllerTestIT extends ProjectStructureIT {
...
+    private static final String RATING_PROJECT_ID = "BIGH-TG1";
+    private static final String RATING_SPEC_ID = RATING_PROJECT_ID
+            + ".Opening-Low_Flow_Gates,Elev;Flow-Low_Flow_Gates.Standard.Production";
+    private static final CwmsId RATING_GROUP = new CwmsId.Builder().withName("Rating-" + RATING_PROJECT_ID
+            + "-RatedGate").withOfficeId(OFFICE_ID).build();
+    private static final Location RATED_OUTLET_LOCATION
+            = buildProjectStructureLocation(RATING_PROJECT_ID, OUTLET_KIND);
+    private static final Outlet NEW_RATED_OUTLET
+            = buildTestOutlet(RATED_OUTLET_LOCATION, PROJECT_LOC, RATING_GROUP, RATING_SPEC_ID);
+    private static final LocationGroup RATING_GROUP_LOCATION_GROUP
```
RyanM-RMA: This is not necessary for the tests, let the DB create the location group, and use the NEW_RATED_OUTLET's variables for retrieving the location group.  This will help solidify that the Outlet object is working as expected.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -216,6 +216,95 @@ public List<LocationGroup> getLocationGroups(@Nullable String officeId,
...
+    /**
+     * Get all location groups for a given office and category,
+     * as well as a where clause to filter the shared_ref_location_id.
+     * @Param officeId The office id to use for the query.
+     * @Param locCategoryLike A regex to use to filter the location categories.  May be null.
+     * @Param sharedRefLocLike A where clause to filter the shared_loc_alias_id.  May be null.
+     * @return A list of all location groups for the given parameters.
+     */
+
+    public List<LocationGroup> getLocationGroups(String officeId, String locCategoryLike, String sharedRefLocLike) {
```
RyanM-RMA, line 227: Unit test this since it's part of the public LocationGroupDao API

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -188,6 +206,121 @@ void test_outlet_rename() {
         ;
     }
 
+    @Test
+    void test_rating_spec_id() {
```
RyanM-RMA: Let's go over the design of this.  We need to have two unit tests for controlled and uncontrolled outlets.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -325,12 +458,13 @@ void test_outlet_crud() {
             .statusCode(is(HttpServletResponse.SC_NOT_FOUND));
     }
 
-    private static Outlet buildTestOutlet(Location outletLoc, Location projectLoc, CwmsId ratingId) {
+    private static Outlet buildTestOutlet(Location outletLoc, Location projectLoc, CwmsId ratingId, String ratingSpecId) {
         return new Outlet.Builder().withProjectId(new CwmsId.Builder().withName(projectLoc.getName())
```
RyanM-RMA, line 695: Overload the ratingSpecId into this, rather than update stuff to use null.  Makes the code a bit more readable in the mess that's already there.

cwms-data-api/src/test/resources/cwms/cda/data/dto/location/kind/base_outlet.json
```diff
@@ -22,5 +22,6 @@
   "rating-group-id": {
     "office-id": "SPK",
     "name": "Rating-BIGH-TG1"
-  }
+  },
+  "rating-spec-id": "Rating-BIGH-TG1.Elev;Flow-Uncontrolled_Spillway.Linear.Production"
```
RyanM-RMA: This should use Elev,Opening-Tainter_Gate;Flow-Uncontrolled_Spillway.  I know it's a bit petty, but a tainter gate is not an uncontrolled spillway.  I don't know if this needs updating - @adamkorynta thoughts?

RyanM-RMA (changes requested):

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Updated test case to match this pattern

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Removed unnecessary LocationGroup creation, grabbed the object using the Outlet

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
zack-rma: Added unit test for new functionality

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Added overload, removed null fields in method calls

cwms-data-api/src/test/resources/cwms/cda/data/dto/location/kind/base_outlet.json
zack-rma: Updated naming, open to suggestions

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -313,12 +503,20 @@ void test_rating_spec_id() {
...
             .redirects().max(3)
-            .delete("/location/group/" + RATING_GROUP.getName())
+            .delete("/location/group/" + RATING_GROUP_CONTROLLED.getName())
         .then()
             .log().ifValidationFails(LogDetail.ALL,true)
         .assertThat()
             .statusCode(is(HttpServletResponse.SC_NO_CONTENT));
+    }
 
+    private static @NotNull LocationGroup getLocationGroup(LocationGroup locGroup, String ratingSpecId) {
```
RyanM-RMA: Rename this to something more meaningful, getLocationGroup sounds like it's going out to retrieve a group, or get it from a file or cache.  Come to me with some ideas before modifying it.

RyanM-RMA: is the NotNull annotation required?  If not, remove it and the import

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
RyanM-RMA: This still needs to be resolved

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
RyanM-RMA: This still needs to be resolved

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
RyanM-RMA, line 1: Did you run the OutletDaoIT tests?  I'm getting null pointers when I run them on this branch and they pass on my other branches.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -188,6 +212,313 @@ void test_outlet_rename() {
...
+            .queryParam(Controllers.OFFICE, OFFICE_ID)
+        .when()
+            .redirects().follow(true)
+            .redirects().max(3)
+            .get("projects/outlets/" + RATED_OUTLET_LOCATION_UNCONTROLLED.getName())
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_OK))
+            .body("rating-spec-id", equalTo(null))
```
RyanM-RMA: Use nullValue() instead of equalTo(null).

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -188,6 +212,313 @@ void test_outlet_rename() {
...
+            .queryParam(CATEGORY_ID, NEW_RATED_OUTLET_UNCONTROLLED.getRatingCategoryId().getName())
+        .when()
+            .redirects().follow(true)
+            .redirects().max(3)
+            .get("/location/group/" + RATING_GROUP_UNCONTROLLED.getName())
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_OK))
+            .body("shared-loc-alias-id", equalTo(null))
```
RyanM-RMA: nullValue()

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -188,6 +212,313 @@ void test_outlet_rename() {
...
+            .extract().as(LocationGroup.class);
+
+        LocationGroup modifiedLocGroup = getLocationGroup(locGroup, RATING_SPEC_ID_UNCONTROLLED);
+
+        json = Formats.format(Formats.parseHeader(Formats.JSONV1, LocationGroup.class), modifiedLocGroup);
+
+        // Update the location group - delete with cascade and create a new one
+        given()
```
RyanM-RMA, line 295: Why are we deleting here?

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -188,6 +212,313 @@ void test_outlet_rename() {
...
+            .queryParam(Controllers.OFFICE, OFFICE_ID)
+        .when()
+            .redirects().follow(true)
+            .redirects().max(3)
+            .get("projects/outlets/" + RATED_OUTLET_LOCATION_CONTROLLED.getName())
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_OK))
+            .body("rating-spec-id", equalTo(null))
```
RyanM-RMA: nullValue()

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -188,6 +212,313 @@ void test_outlet_rename() {
...
+            .queryParam(CATEGORY_ID, NEW_RATED_OUTLET_CONTROLLED.getRatingCategoryId().getName())
+        .when()
+            .redirects().follow(true)
+            .redirects().max(3)
+            .get("/location/group/" + RATING_GROUP_CONTROLLED.getName())
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_OK))
+            .body("shared-loc-alias-id", equalTo(null))
```
RyanM-RMA: nullValue()

RyanM-RMA (changes requested):

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
RyanM-RMA: Confirmed that this is required

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
RyanM-RMA: modifyRatingSpecId

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
zack-rma: No null pointers on my end, the tests pass for me

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Swapped

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Swapped

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Swapped

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Swapped

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Removed annotations and relocated the function

zack-rma: Renamed as discussed

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Added location group removal

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: Added location group removal

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
```diff
@@ -95,7 +96,16 @@ public static void setup() throws Exception {
         databaseLink.connection(c -> {
             DSLContext context = getDslContext(c, OFFICE_ID);
             OutletDao outletDao = new OutletDao(context);
+            LocationGroupDao locationGroupDao = new LocationGroupDao(context);
             try {
+                try {
+                    locationGroupDao.delete(NEW_RATED_OUTLET_CONTROLLED.getRatingCategoryId().getName(),
```
RyanM-RMA: This isn't doing what you think it's doing.  If the first delete fails, the second delete will be ignored.  You want to do a try/catch around each one.  Make it into a function like deleteLocation, so you don't muddy up the code.

RyanM-RMA (changes requested):

cwms-data-api/src/test/java/cwms/cda/api/location/kind/OutletControllerTestIT.java
zack-rma: That makes sense. I've replaced it with a location group cleanup function.

RyanM-RMA (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -216,6 +216,95 @@ public List<LocationGroup> getLocationGroups(@Nullable String officeId,
...
+        if (officeId != null) {
+            if (CWMS.equalsIgnoreCase(officeId)) {
+                connectBy = onStep.where(alcg.CAT_DB_OFFICE_ID.eq(CWMS)
+                        .and(alcg.GRP_DB_OFFICE_ID.eq(CWMS))
+                        .and(condition)
+                );
+            } else {
+                connectBy = onStep.where(alcg.CAT_DB_OFFICE_ID.in(CWMS, officeId)
+                        .and(alcg.GRP_DB_OFFICE_ID.in(CWMS, officeId))
+                        .and(alga.DB_OFFICE_ID.isNull().or(alga.DB_OFFICE_ID.eq(officeId)))
```
adamkorynta: you're looking specifically for location groups so you don't want null ones

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -216,6 +216,95 @@ public List<LocationGroup> getLocationGroups(@Nullable String officeId,
...
+                        .and(condition)
+                );
+            }
+        } else {
+            connectBy = onStep.where(alcg.LOC_GROUP_ID.isNotNull());
```
adamkorynta: you want this condition always since you're returning location groups regardless of `officeId` input

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -216,6 +216,95 @@ public List<LocationGroup> getLocationGroups(@Nullable String officeId,
...
+        if (locCategoryLike != null && !locCategoryLike.isEmpty()) {
+            condition = caseInsensitiveLikeRegex(alcg.LOC_CATEGORY_ID, locCategoryLike);
+        }
+
+        if (sharedRefLocLike != null && !sharedRefLocLike.isEmpty()) {
+            condition = condition.and(caseInsensitiveLikeRegex(alcg.SHARED_REF_LOCATION_ID, sharedRefLocLike));
+        }
+
+        if (officeId != null) {
+            if (CWMS.equalsIgnoreCase(officeId)) {
```
adamkorynta: why split the CWMS and specific office branches when you'll wind up with the same outcome?

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -216,6 +216,95 @@ public List<LocationGroup> getLocationGroups(@Nullable String officeId,
...
+                connectBy = onStep.where(alcg.CAT_DB_OFFICE_ID.eq(CWMS)
+                        .and(alcg.GRP_DB_OFFICE_ID.eq(CWMS))
+                        .and(condition)
+                );
+            } else {
+                connectBy = onStep.where(alcg.CAT_DB_OFFICE_ID.in(CWMS, officeId)
+                        .and(alcg.GRP_DB_OFFICE_ID.in(CWMS, officeId))
```
adamkorynta: Seems misleading to take in input `SPK` but then return `CWMS` owned groups. For your specific outlet use-case aren't they always office specific groups and not `CWMS` groups?

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
adamkorynta: also, all of these checks should ignore case (by calling upper on both sides of the `eq`)

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -216,6 +216,95 @@ public List<LocationGroup> getLocationGroups(@Nullable String officeId,
...
+        final RecordMapper<Record, Pair<LocationGroup, AssignedLocation>> mapper = grpRecord -> {
+            LocationCategory category = buildLocationCategory(grpRecord);
+
+            LocationGroup group = buildLocationGroup(grpRecord, category);
+            AssignedLocation loc = buildAssignedLocation(grpRecord);
+
+            return new Pair<>(group, loc);
+        };
+
+        Map<LocationGroup, List<AssignedLocation>> map = new LinkedHashMap<>();
```
adamkorynta: This variable is declared and instantiated way before it is actually used

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
```diff
@@ -78,24 +80,20 @@ public Outlet retrieveOutlet(String officeId, String locationId) {
         });
     }
 
-    private CwmsId getRatingGroupId(String locationId, List<LocationGroup> groups) {
+    private Optional<LocationGroup> getRatingGroup(String locationId, List<LocationGroup> groups) {
         return groups.stream()
                      .filter(group -> group.getAssignedLocations()
                                            .stream()
                                            .anyMatch(loc -> loc.getLocationId().equalsIgnoreCase(locationId)))
```
adamkorynta: nitpick:
for readability, try to keep the stream as tight as possible
```
group.getAssignedLocations()
              .stream()
              .map(AssignedLocation::getLocationId)
              .anyMatch(id -> id.equalsIgnoreCase(locationId))
```

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
adamkorynta: since you're only returning groups it seems like you'd want to remove the where clause of the `CAT_DB_OFFICE_ID`

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
RyanM-RMA: This is my code, and he's right about this, I should have written it this way.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
zack-rma: Removed null check and ignored case

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
zack-rma: Modified to always be part of the condition

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
zack-rma: Simplified to remove CWMS comparison

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
zack-rma: Removed unnecessary CWMS comparison, replaced with officeID comparison

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
zack-rma: Moved closer to use

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
zack-rma: Cleaned up stream

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
adamkorynta: Don't duplicate the condition

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -271,22 +269,15 @@ public List<LocationGroup> getLocationGroups(String officeId, String locCategory
...
-                        .and(condition)
-                );
-            } else {
-                connectBy = onStep.where(alcg.CAT_DB_OFFICE_ID.in(CWMS, officeId)
-                        .and(alcg.GRP_DB_OFFICE_ID.in(CWMS, officeId))
-                        .and(alga.DB_OFFICE_ID.isNull().or(alga.DB_OFFICE_ID.eq(officeId)))
-                        .and(condition)
-                );
-            }
+            connectBy = onStep.where(DSL.upper(alcg.CAT_DB_OFFICE_ID).in(officeId.toUpperCase())
```
adamkorynta: You're searching for location groups with the given office id, not categories with that office id

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -271,22 +269,15 @@ public List<LocationGroup> getLocationGroups(String officeId, String locCategory
...
+                    .and(DSL.upper(alcg.GRP_DB_OFFICE_ID).in(officeId.toUpperCase()))
```
adamkorynta: should be an `eq` not an `in`

adamkorynta (dismissed):

RyanM-RMA (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -215,6 +215,85 @@ public List<LocationGroup> getLocationGroups(@Nullable String officeId,
...
+        if (officeId != null) {
+            connectBy = onStep.where(DSL.upper(alcg.GRP_DB_OFFICE_ID).eq(officeId.toUpperCase()).and(condition));
+        } else {
+            connectBy = onStep.where(condition);
+        }
+
+
+        Map<LocationGroup, List<AssignedLocation>> map = new LinkedHashMap<>();
+        connectBy.orderBy(alcg.LOC_CATEGORY_ID, alcg.LOC_GROUP_ID, alga.ATTRIBUTE)
+                .fetchSize(1000)  // This made the query go from 2 minutes to 10 seconds?
```
rma-psmorris, line 280: does this 1000 value differ from query to query or is there a constant or global function that can be used for it?

rma-psmorris (commented):
Post-merge review

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
MikeNeilson: That's likely to be rather query specific as it let's Oracle decide how much to gather and prepare at once as the downstream system calls `ResultSet::next`

zack-rma: I believe it remains constant between queries. I can turn it into a constant since it's also used in the function that I based this one off of. It seems Ryan R originally added this optimization about a year ago, and I didn't want to break it if it was still necessary.

rma-psmorris: https://github.com/search?q=repo%3AUSACE%2Fcwms-data-api%20fetchSize&type=code
Shows constants being 1000 vs 500, with default use for most cases I suspect. Default for Oracle appears to be 10:
https://docs.oracle.com/en/database/oracle/oracle-database/23/jjdbc/resultset.html#GUID-BD27B43E-525F-44C4-973F-18D2BB8BA007
10 is too small and that should be set in the connection pool to a better number. The 500 vs 1000 could be captured in a global function to allow variation or as global constants to allow easy change. Having them as int local values makes maintenance difficult.

zack-rma: I'll look into making a global variable for this somewhere that is available to all the DAOs that use it. I'm open to suggestions for where to put it, but I'll try to find a helper file or base class that makes sense for this.

zack-rma: I added a couple global variables in the JooqDao class, since that is shared between the DAOs that use fetchSize. Take a look at the PR here: https://github.com/USACE/cwms-data-api/pull/877

PR #841, opened 2024-08-12
switch project update to project rename

adamkorynta:
functionality needed by the client API

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/ProjectController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/ProjectController.java
@@ -230,32 +230,33 @@ public class ProjectController implements CrudHandler {
     }
 
     @OpenApi(
-            description = "Updates a project",
-            pathParams = {
-                @OpenApiParam(name = NAME, description = "The id of the project to be updated"),
+        description = "Rename a project",
+        pathParams = {
+            @OpenApiParam(name = NAME, description = "The name of the project to be renamed"),
+        },
+        queryParams = {
+            @OpenApiParam(name = NAME, description = "The new name of the project"),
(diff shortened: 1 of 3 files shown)


cwms-data-api/src/main/java/cwms/cda/api/ProjectController.java
```diff
@@ -230,32 +230,33 @@ public void create(@NotNull Context ctx) {
...
-            method = HttpMethod.PATCH,
-            tags = {TAG}
+            required = true
+        ),
+        method = HttpMethod.PATCH,
+        tags = {TAG}
     )
     @Override
-    public void update(@NotNull Context ctx, @NotNull String name) {
+    public void update(@NotNull Context ctx, @NotNull String oldName) {
```
MikeNeilson, line 252: Just to be pedantic, this is the only possible update for a Project at this level, just the name?

adamkorynta: It's not, but it gets back to the issues we've had with all the renames. 

Currently none of the POST operations block updating an existing record (usually there's a fail-if-exists flag). Therefore all of the functionality of an "update" will work via POST. 

Having both a rename operation and an update on the same endpoint is confusing given a rename only needs and old/new name, while an update would need to selectively update non-null fields (something not supported by most of the procedures) and skip DTO validation.

Given update can be performed via the POST operation, patch has been used for the rename function. I can see this changing in the future if/when we want true PUT/PATCH support and then moving the rename to a `/<type>/{name}/rename` endpoint.

MikeNeilson: Confusing as hell, but makes sense.

MikeNeilson (approved):

PR #844, opened 2024-08-13
Basin Bugfix: Updated parameter naming

zack-rma:
Renamed parameters for basin id


adamkorynta:
I think the opposite change should happen. Update ApiServlet to use `name` instead of `basin-id` for consistency. It should be backwards compatible with other clients since it is only a path parameter.

adamkorynta (approved):

PR #855, opened 2024-08-20
﻿First draft of gate change and settings

RyanM-RMA:
Includes DTO and DAO implementations, and unit and integration tests respectively

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
@@ -27,10 +27,16 @@ import cwms.cda.data.dto.AssignedLocation;
 import cwms.cda.data.dto.CwmsId;
 import cwms.cda.data.dto.Location;
 import cwms.cda.data.dto.LocationGroup;
+import cwms.cda.data.dto.LookupType;
+import cwms.cda.data.dto.location.kind.GateChange;
+import cwms.cda.data.dto.location.kind.GateSetting;
 import cwms.cda.data.dto.location.kind.Outlet;
 import cwms.cda.data.dto.location.kind.ProjectStructure;
 import cwms.cda.data.dto.location.kind.VirtualOutlet;
 import cwms.cda.data.dto.location.kind.VirtualOutletRecord;
+import java.math.BigInteger;
+import java.sql.Timestamp;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
@@ -39,10 +45,15 @@ import org.jooq.Configuration;
 import org.jooq.DSLContext;
 import org.jooq.impl.DSL;
 import usace.cwms.db.jooq.codegen.packages.CWMS_OUTLET_PACKAGE;
+import usace.cwms.db.jooq.codegen.udt.records.GATE_CHANGE_OBJ_T;
+import usace.cwms.db.jooq.codegen.udt.records.GATE_CHANGE_TAB_T;
+import usace.cwms.db.jooq.codegen.udt.records.GATE_SETTING_OBJ_T;
+import usace.cwms.db.jooq.codegen.udt.records.GATE_SETTING_TAB_T;
 import usace.cwms.db.jooq.codegen.udt.records.LOCATION_REF_T;
 import usace.cwms.db.jooq.codegen.udt.records.PROJECT_STRUCTURE_OBJ_T;
 import usace.cwms.db.jooq.codegen.udt.records.STR_TAB_T;
 import usace.cwms.db.jooq.codegen.udt.records.STR_TAB_TAB_T;
+import static cwms.cda.data.dao.location.kind.LocationUtil.getLocationRef;
 
 public class OutletDao extends JooqDao<Outlet> {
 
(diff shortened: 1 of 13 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
```diff
@@ -242,4 +253,116 @@ public void renameOutlet(String officeId, String oldOutletId, String newOutletId
...
+            LOCATION_REF_T locationRef = getLocationRef(projectId);
+            Timestamp startTimestamp = Timestamp.from(startTime);
+            Timestamp endTimestamp = Timestamp.from(endTime);
+            BigInteger rowLimitBig = BigInteger.valueOf(rowLimit);
+            GATE_CHANGE_TAB_T changeTab = CWMS_OUTLET_PACKAGE.call_RETRIEVE_GATE_CHANGES(
+                    DSL.using(conn).configuration(), locationRef, startTimestamp, endTimestamp, "UTC", unitSystem,
+                    formatBool(startInclusive), formatBool(endInclusive), rowLimitBig);
+
+            List<GateChange> output = new ArrayList<>();
+            if (changeTab != null) {
```
rma-bryson: If changeTab can come back null, does that mean no gate changes were found? If so, I would throw a NotFoundException

cwms-data-api/src/test/java/cwms/cda/data/dto/location/kind/GateChangeTest.java
rma-bryson, line 1: Add test of missing required fields

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
```diff
@@ -56,7 +56,6 @@ public abstract class PhysicalStructureChange<T extends Setting> extends CwmsDTO
     private final Double newTotalDischargeOverride;
     private final Double oldTotalDischargeOverride;
     private final String dischargeUnits;
-    private final Double poolElevation;
```
rma-bryson, line 59: I believe you could just keep the poolElevation in this base class. Then in the class that requires it (GateChange), you can override the validator

`@Override
    protected void validateInternal(CwmsDTOValidator validator)
    {
        validator.required(getPoolElevation(), "poolElevation");
        super.validateInternal(validator);
    }`

cwms-data-api/src/test/java/cwms/cda/data/dto/location/kind/GateSettingTest.java
rma-bryson, line 1: Add missing required fields test

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/GateChange.java
```diff
@@ -0,0 +1,87 @@
...
+@JsonPropertyOrder({"project-id", "change-date", "reference-elevation", "pool-elevation", "protected", "discharge-computation-type", "reason-type", "notes"})
+public class GateChange  extends PhysicalStructureChange<GateSetting> {
+    private final Double referenceElevation;
+
+    @JsonProperty(required = true)
+    private final double poolElevation;
+
+    GateChange(Builder builder) {
+        super(builder);
+        this.poolElevation = Objects.requireNonNull(builder.poolElevation, "Pool elevation is required when creating a GateChange");
```
rma-bryson: Remove this Objects.requireNonNull. Null checks should get handled by the validation process on required fields.

rma-bryson (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
RyanM-RMA: Threw a NotFoundException, that's a good idea, I was going off what TurbineDao did, but I think this is a better approach.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
RyanM-RMA: Validation isn't really the reason I did any of this.  The public API for the two change classes is what I was considering when making this change.  poolElevation is required for gate changes, but not for turbines, so for turbines the method `public Double getPoolElevation()` makes sense.  Whereas this doesn't make sense for the gate changes.  We can't override the `getPoolElevation()` to return a double instead, which is why the API is split apart, and since poolElevation is a private field it's also separated with the public API.
If you can think of a better approach to keeping that API that way, let me know and I'd be glad to consider that.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/GateChange.java
RyanM-RMA: Removed

cwms-data-api/src/test/java/cwms/cda/data/dto/location/kind/GateSettingTest.java
RyanM-RMA: Added test for this

cwms-data-api/src/test/java/cwms/cda/data/dto/location/kind/GateChangeTest.java
RyanM-RMA: Added test for this

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/PhysicalStructureChange.java
RyanM-RMA: I updated it to use Double's, as we had discussed, and had to adjust the GateSetting to do the same.  I like the idea of the validation handling this.

rma-bryson (approved):

cwms-data-api/src/test/resources/cwms/cda/data/dto/location/kind/gate-settings.json
```diff
@@ -0,0 +1,24 @@
+[
+  {
+    "type": "gate-setting",
```
rma-psmorris, line 3: this data structure needs to scale similar to time series

RyanM-RMA: Do you have some time today or this week where we can talk about this?

PR #859, opened 2024-09-03
Feature/add gate change controller

RyanM-RMA:


cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeCreateController.java
```diff
@@ -0,0 +1,83 @@
...
+            tags = {OutletController.TAG},
+            responses = {
+                    @OpenApiResponse(status = STATUS_201, description = "Gate Changes successfully stored to CWMS.")
+            }
+    )
+    @Override
+    public void handle(@NotNull Context context) throws Exception {
+        boolean failIfExists = context.queryParamAsClass(FAIL_IF_EXISTS, Boolean.class).getOrDefault(true);
+        String formatHeader = context.header(Header.ACCEPT) != null ? context.header(Header.ACCEPT) : Formats.JSONV1;
+        ContentType contentType = Formats.parseHeader(formatHeader, GateChange.class);
```
MikeNeilson, line 73: Should we just have parseHeader return the configured default if null is passed in, or an error if there is not default?

MikeNeilson: NOTE: we don't need to do that in this PR, just realize how common this could be and we are trying to go for convention over configuration here... well or at least put the onerous on format choices on the DTOs not the controllers anyways.

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeGetAllController.java
```diff
@@ -0,0 +1,117 @@
...
+        int pageSize = context.queryParamAsClass(PAGE_SIZE, Integer.class).getOrDefault(DEFAULT_PAGE_SIZE);
+
+        try (Timer.Context ignored = markAndTime(GET_ALL)) {
+            DSLContext dsl = JooqDao.getDslContext(context);
+            OutletDao dao = new OutletDao(dsl);
+            List<GateChange> changes = dao.retrieveOperationalChanges(projectId, startTime.toInstant(),
+                                                                      endTime.toInstant(), startInclusive, endInclusive,
+                                                                      unitSystem, pageSize);
+            String formatHeader = context.header(Header.ACCEPT) != null ? context.header(Header.ACCEPT) : Formats.JSONV1;
+            ContentType contentType = Formats.parseHeader(formatHeader, GateChange.class);
```
MikeNeilson, line 111: My previous comment makes more sense here, than in POST, POST should probably error if it's not told at all.

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/ProjectStructureIT.java
```diff
@@ -146,4 +157,52 @@ public static void deleteLocation(DSLContext context, String officeId, String lo
...
+		LocationGroupDao locationGroupDao = new LocationGroupDao(context);
+		try {
+			locationGroupDao.delete(outlet.getRatingCategoryId().getName(), outlet.getRatingGroupId().getName(), true, OFFICE_ID);
+		} catch (NotFoundException e) {
+			LOGGER.atFinest().withCause(e).log("No data found for category:" + outlet.getRatingCategoryId().getName()
+													   + ", group-id:" + outlet.getRatingGroupId().getName());
+		}
+	}
+
+	public static <T extends CwmsDTOBase> void containsDto(List<T> outlets, T expectedDto, BiPredicate<T, T> identifier, BiConsumer<T, T> dtoMatcher) {
```
MikeNeilson: These should probably just be matchers to maintain the expect test assert naming scheme in the test cases.

They're specific enough to outlets that keeping them in an "outlets" package makes sense. But I would prefer to see assert(doesNot)Contain(s)Dto in the test case itself.

MikeNeilson (commented):
Overall looks reasonable. One suggestion that maybe reasonable to do now, others are thoughts for the future.

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeCreateController.java
RyanM-RMA: So, if the formatHeader is null, it'll end up throwing an UnsupportedFormatException, see the code here:
https://github.com/USACE/cwms-data-api/blob/develop/cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java#L377

MikeNeilson: Yeah, I'm not saying we change it now, I just keep seeing this type of two line code repeated and am wondering if we should shift the responsibility in a future PR.

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/ProjectStructureIT.java
RyanM-RMA: Not sure I'm understanding this comment.  What do you mean by matchers, and the assert contains bit?  Are you saying I should rename this and its inverse to assertContainsDto and assertDoesNotContainDto?

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeCreateController.java
adamkorynta: We already have a DEFAULT constant that's handled in the annotations (or at least can be), I agree that we should just have the `parseHeader` substitute null for that DEFAULT and then controllers no longer need to null check. I also agree that a separate PR makes sense for that change. @RyanM-RMA - I can do that change if you're in agreement

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
```diff
@@ -266,17 +267,18 @@ public List<GateChange> retrieveOperationalChanges(CwmsId projectId, Instant sta
             Timestamp endTimestamp = Timestamp.from(endTime);
             BigInteger rowLimitBig = BigInteger.valueOf(rowLimit);
             GATE_CHANGE_TAB_T changeTab = CWMS_OUTLET_PACKAGE.call_RETRIEVE_GATE_CHANGES(
-                    DSL.using(conn).configuration(), locationRef, startTimestamp, endTimestamp, "UTC", unitSystem,
+                    DSL.using(conn).configuration(), locationRef, startTimestamp, endTimestamp, "UTC", unitSystem.getValue(),
                     formatBool(startInclusive), formatBool(endInclusive), rowLimitBig);
 
             List<GateChange> output = new ArrayList<>();
```
adamkorynta: no need to create an array, let the stream processing handle it

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/ProjectStructureIT.java
MikeNeilson: Yeah. There are some other examples of the matchers in the test fixtures package.

They should also include a String message, nothing wrong with the default you have there, but seems worth making a bit more generic now.

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/TurbineChange.java
```diff
@@ -45,8 +45,15 @@
 @JsonPropertyOrder({"project-id", "change-date", "protected", "discharge-computation-type", "reason-type", "notes"})
 public final class TurbineChange extends PhysicalStructureChange<TurbineSetting> {
 
+    private Double poolElevation;
+
     TurbineChange(Builder builder) {
         super(builder);
+        poolElevation = builder.poolElevation;
```
adamkorynta: This is getting set in `PhysicalStructureChange` as well (and you're adding the getter to this class), does it belong in both?

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeCreateController.java
RyanM-RMA: A separate PR makes sense.  Looking at the handling of the DEFAULT constant, I think that might need some adjustment, since it'll return Formats.JSON, and doesn't check the class aliases for the actual format to return.

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/ProjectStructureIT.java
RyanM-RMA: Oh, do you mean the assertMatch methods in DTOMatch class?  I could probably add something there for this so it's more reusable.

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeCreateController.java
MikeNeilson: probably better at that level to return `*/*` which would make sure the system just went through and found the right default for the DTO.

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeGetAllController.java
RyanM-RMA: Mind if we resolve this comment as part of the one above?  Seems like that one has taken over this conversation.

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/ProjectStructureIT.java
```diff
@@ -146,4 +157,52 @@ public static void deleteLocation(DSLContext context, String officeId, String lo
...
+
+	public static <T extends CwmsDTOBase> void doesNotContainDto(List<T> outlets, T expectedDto, BiPredicate<T, T> identifier) {
+		T receivedDto = outlets.stream()
+									   .filter(dto -> identifier.test(dto, expectedDto))
+									   .findFirst()
+									   .orElse(null);
+		assertNull(receivedDto);
+	}
+
+	public static void createRatingSpecForOutlet(DSLContext context, Outlet outlet, String specId) {
```
adamkorynta: Move outlet code to implementation class instead of base class of turbines and outlets. Also rename occurrences of `outlets` in the file to avoid miscommunication.

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/ProjectStructureIT.java
MikeNeilson: Yep, don't get too hung up on generic - if it gets complicated, but we do want to match that style.

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeGetAllController.java
```diff
@@ -0,0 +1,117 @@
...
+            },
+            description = "Returns matching CWMS gate change data for a Reservoir Project.",
+            tags = {OutletController.TAG}
+    )
+    @Override
+    public void handle(@NotNull Context context) {
+        String office = context.pathParam(OFFICE);
+        String id = context.pathParam(PROJECT_ID);
+        CwmsId projectId = new CwmsId.Builder().withName(id).withOfficeId(office).build();
+        Timestamp startTime = Timestamp.from(requiredZdt(context, BEGIN).toInstant());
```
adamkorynta: Instead of using `Timestamp` use `requiredInstant` and map to `Timestamp` at the jOOQ level.

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeDeleteController.java
```diff
@@ -0,0 +1,83 @@
...
+            description = "Deletes matching CWMS gate change data for a Reservoir Project.",
+            tags = {OutletController.TAG},
+            method = HttpMethod.DELETE
+    )
+    @Override
+    public void handle(@NotNull Context context) throws Exception {
+        String office = context.pathParam(OFFICE);
+        String id = context.pathParam(PROJECT_ID);
+        CwmsId projectId = new CwmsId.Builder().withName(id).withOfficeId(office).build();
+        Timestamp startTime = Timestamp.from(requiredZdt(context, BEGIN).toInstant());
```
adamkorynta: Same `Timestamp` comment

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeGetAllController.java
```diff
@@ -0,0 +1,117 @@
...
+                                    "A positive integer is interpreted as the maximum number of changes from the " +
+                                    "beginning of the time window. A negative integer is interpreted as the maximum number " +
+                                    "from the end of the time window. " + "Default " + DEFAULT_PAGE_SIZE + "." +
+                                    "A page cursor will not be returned by this DTO. Instead, the next page can be determined " +
+                                    "by querying the next set of changes using the last returned change date and using " +
+                                    "start-time-inclusive=false")
+            },
+            responses = {
+                    @OpenApiResponse(status = STATUS_200, content = {
+                            @OpenApiContent(from = GateChange.class, isArray = true, type = Formats.JSONV1)
```
adamkorynta: missing the `JSON` format

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeCreateController.java
adamkorynta: Yeah, I can't think of why we would have any nulls that wouldn't be handled the same as `DEFAULT` (which is `*/*`)

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/ProjectStructureIT.java
RyanM-RMA: I moved the methods into DTOMatch, they were already generic enough, just needed a scope change and rename.

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeCreateController.java
adamkorynta: Oh I see what you mean @RyanM-RMA - if there is no matching content type for the DTO and the we pass in `*/*` it gets replaced with "application/json".  I think that should still be ok.

RyanM-RMA: I think we should resolve it to the appropriate version for the class, given the aliasing.

adamkorynta: If the user passes in null, but the DTO has no `DEFAULT` handling, how do we know what is the appropriate version? The first annotation?

MikeNeilson: Error, we should assume if no default was set, that it was intentional.

adamkorynta: Handling under: https://github.com/USACE/cwms-data-api/pull/861

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
RyanM-RMA: Good catch

cwms-data-api/src/main/java/cwms/cda/data/dto/location/kind/TurbineChange.java
RyanM-RMA: Forgot to delete those I think

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/ProjectStructureIT.java
RyanM-RMA: I moved these into a base class, so I can share them across the extended Outlet classes

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeGetAllController.java
RyanM-RMA: Added JSON

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeGetAllController.java
RyanM-RMA: Oh, didn't know that was a thing, thanks.

cwms-data-api/src/main/java/cwms/cda/api/location/kind/GateChangeDeleteController.java
RyanM-RMA: Updated this one too

cwms-data-api/src/test/java/cwms/cda/data/dao/location/kind/ProjectStructureIT.java
adamkorynta: and not worth creating an OutletStructureIT base class, I'll buy it!

adamkorynta: oh, looks like you did make one

adamkorynta (dismissed):

adamkorynta (approved):

PR #861, opened 2024-09-04
handle null content type headers the same as */*

adamkorynta:
Also updated DTO's and controllers for consistency, removing deprecated methods.

### DIFF
--- a/cwms-data-api/src/test/resources/cwms/cda/data/sql/load_testusers.sql
+++ b/cwms-data-api/src/test/resources/cwms/cda/data/sql/load_testusers.sql
@@ -1,9 +1,8 @@
+declare
+    group_list cwms_20.char_32_array_type;
 begin
-    create user "l2hectest" identified by "l2hectest";
-    create user "l2hectest_vt" identified by "l2hectestvt";
-    create user "m5hectest" identified by "swt99db";
-    create user "user2" identified by "user2";
-    create user "&webuser" identified by "&password";
-    create user "&user" identified by "&password";
-    create user "&pduser" identified by "&password";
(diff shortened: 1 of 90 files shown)


cwms-data-api/src/test/resources/cwms/cda/data/sql/load_testusers.sql
```diff
@@ -1,8 +1,9 @@
-declare
-    group_list cwms_20.char_32_array_type;
 begin
-    group_list := cwms_20.CHAR_32_ARRAY_TYPE('CWMS Users');
-    cwms_20.cwms_sec.create_user('user','blah',group_list, 'HQ');    
-    cwms_20.cwms_sec.create_user('user2','blah',NULL,'SPK');    
-    cwms_20.cwms_sec.create_user('s0webtest','&password.',group_list,'HQ');    
+    create user "l2hectest" identified by "l2hectest";
```
MikeNeilson: None of these users should be actual oracle users in these tests.

MikeNeilson: Except maybe s0webtest, the rest operate as "cac only".

MikeNeilson (changes requested):

cwms-data-api/src/test/resources/cwms/cda/data/sql/load_testusers.sql
adamkorynta: thanks for this catch, I didn't mean to change this file.

MikeNeilson (approved):

PR #863, opened 2024-09-04
Updated Water Contract Controller DeleteMethod Default Value

zack-rma:
Updated DeleteMethod to allow null, defaults to DELETE_KEY

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterPumpDisassociateController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterPumpDisassociateController.java
@@ -65,8 +65,8 @@ public final class WaterPumpDisassociateController extends WaterSupplyController
             @OpenApiParam(name = PUMP_TYPE, required = true, type = PumpType.class,
                     description = "The type of pump to be disassociated from the contract."
                             + " Expected values: IN, OUT, OUT BELOW"),
-            @OpenApiParam(name = METHOD, type = boolean.class, required = true,
-                    description = "Whether to delete the associated accounting data.")
+            @OpenApiParam(name = METHOD, type = boolean.class,
+                    description = "Whether to delete the associated accounting data. Defaults to FALSE.")
         },
         pathParams = {
             @OpenApiParam(name = NAME, description = "The name of the pump to be "
@@ -88,7 +88,7 @@ public final class WaterPumpDisassociateController extends WaterSupplyController
     public void handle(@NotNull Context ctx) {
         try (Timer.Context ignored = markAndTime(DELETE)) {
             DSLContext dsl = getDslContext(ctx);
-            boolean deleteAccounting = requiredParamAs(ctx, METHOD, boolean.class);
+            boolean deleteAccounting = ctx.queryParamAsClass(METHOD, boolean.class).getOrDefault(false);
             String officeId = ctx.pathParam(OFFICE);
             String pumpName = ctx.pathParam(NAME);
             String projectName = ctx.pathParam(PROJECT_ID);
--- a/cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterPumpDisassociateController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterPumpDisassociateController.java
@@ -28,7 +28,6 @@ package cwms.cda.api.watersupply;
 
 import static cwms.cda.api.Controllers.CONTRACT_NAME;
 import static cwms.cda.api.Controllers.DELETE;
-import static cwms.cda.api.Controllers.METHOD;
 import static cwms.cda.api.Controllers.NAME;
 import static cwms.cda.api.Controllers.OFFICE;
 import static cwms.cda.api.Controllers.PROJECT_ID;
@@ -55,6 +54,7 @@ import org.jooq.DSLContext;
 
 public final class WaterPumpDisassociateController extends WaterSupplyControllerBase implements Handler {
     private static final String PUMP_TYPE = "pump-type";
+    private static final String DELETE_ACCOUNTING = "delete-accounting";
 
     public WaterPumpDisassociateController(MetricRegistry metrics) {
         waterMetrics(metrics);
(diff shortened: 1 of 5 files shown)


cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterUserDeleteController.java
```diff
@@ -80,8 +80,7 @@ public void handle(@NotNull Context ctx) {
             DSLContext dsl = getDslContext(ctx);
             String office = ctx.pathParam(OFFICE);
             String locationId = ctx.pathParam(PROJECT_ID);
-            DeleteMethod deleteMode = getDeleteMethod(ctx.queryParamAsClass(METHOD, String.class)
-                    .getOrDefault(DeleteMethod.DELETE_ALL.toString()));
+            DeleteMethod deleteMode = getDeleteMethod(ctx.queryParam(METHOD));
```
RyanM-RMA, line 83: Don't forget to update the DAO to allow for null

RyanM-RMA (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterUserDeleteController.java
zack-rma: Updated DAO

RyanM-RMA (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterPumpDisassociateController.java
```diff
@@ -65,8 +65,8 @@ public WaterPumpDisassociateController(MetricRegistry metrics) {
             @OpenApiParam(name = PUMP_TYPE, required = true, type = PumpType.class,
                     description = "The type of pump to be disassociated from the contract."
                             + " Expected values: IN, OUT, OUT BELOW"),
-            @OpenApiParam(name = METHOD, type = boolean.class, required = true,
-                    description = "Whether to delete the associated accounting data.")
+            @OpenApiParam(name = METHOD, type = boolean.class,
```
adamkorynta: METHOD should use DeleteMethod DTO and not a boolean.

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterPumpDisassociateController.java
RyanM-RMA: I think this is a nomenclature thing, METHOD isn't the right word.  It should be something like "Delete Accounting" or something, cause it's just a boolean not a DeleteMethod.

adamkorynta: isn't that the distinction between Delete Key and Delete Data? changing the query parameter is fine too

zack-rma: How about renaming the query parameter to DELETE_ACCOUNTING ("delete-accounting")?

adamkorynta: that's fine, unless you can get swagger to pick only two of the three DeleteMethod enums

RyanM-RMA: I hadn't seen your comment from yesterday on this, can you explain why are you opposed to using a boolean for this vs the DeleteMethod piece?

adamkorynta: I'm not opposed to a boolean, I'm just more in favor of consistency with other delete methods when it makes sense

RyanM-RMA: Just realized we never finished this conversation.  I can see the benefit of that.  Would it look something like DELETE_KEY == false and DELETE_DATA || DELETE_ALL == true?  Or are you thinking just DELETE_KEY and DELETE_DATA are acceptable?  This would be if the naming stuck with METHOD, since that's also consistent with the other naming usages.

adamkorynta: Hmm that's a good point. If you specified `DELETE_DATA`, then I'd expect the accounting to get deleted, but the pump not disassociated? I was envisioning DELETE_KEY and DELETE_ALL would be the only two options, but I don't even know if swagger can do that. 

Sounds to me like DeleteMethod doesn't make sense here and a simple boolean is sufficient?

RyanM-RMA: Yeah...being that you have 2 options instead of 3, I think that's more appropriate to be a boolean.  I can see the benefit of it being consistent, but that would throw in some additional expected behaviors that's not supported.

RyanM-RMA: @zack-rma please move forward with updating this to use a boolean, use the name you provided for DELETE_ACCOUNTING and let me know when you're done so I can re-review.

adamkorynta (dismissed):

cwms-data-api/src/test/java/cwms/cda/api/WaterPumpDisassociateControllerTestIT.java
```diff
@@ -415,7 +416,7 @@ void test_remove_does_not_exist() throws Exception {
         // Remove pump
         given()
             .log().ifValidationFails(LogDetail.ALL, true)
-            .queryParam(METHOD, DeleteMethod.DELETE_ALL)
+            .queryParam(DELETE_ACCOUNTING, DeleteMethod.DELETE_ALL)
```
RyanM-RMA: Needs to be a boolean now, yeah?

RyanM-RMA (changes requested):

cwms-data-api/src/test/java/cwms/cda/api/WaterPumpDisassociateControllerTestIT.java
zack-rma: Updated to boolean

RyanM-RMA (approved):

adamkorynta (approved):

PR #877, opened 2024-09-18
Added global fetchSize variables to DAOs

zack-rma:
Added global fetchSize variables to JooqDao and appropriate DAOs


cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -69,6 +68,8 @@
 public abstract class JooqDao<T> extends Dao<T> {
     protected static final int ORACLE_CURSOR_TYPE = -10;
     private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+    public static final int FETCH_SIZE = 1000;
```
MikeNeilson: I feel like this should be `DEFAULT_FETCH_SIZE` and `DEFAULT_SMALL_FETCH_SIZE` or the like.

Over time we will likely find that we need to fine tune this for different queries though this is a solid baseline, especially over the apparent default of 10.

MikeNeilson (commented):
Excellent idea. Some thoughts on naming but not enough to assert a denial if others dissagree with me.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
rma-psmorris: If wanting to tune this, I recommend that this variable be changed to public static getDefaultFetchSize method that supports a System property override if System property changes are easily done.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -69,6 +68,8 @@
...
+    public static final int HALF_FETCH_SIZE = FETCH_SIZE / 2;
```
rma-psmorris: I wouldnt code this as a half fetch size but instead as Mike called out above. SMALL and set it to 500. I think the relationship between the two values in the code is arbitrary.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
MikeNeilson: I doubt we'd want to do a global override of the fetch size. If it's tweaked, it's going to be per query based on that queries specific need.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
zack-rma: I updated the naming and set the small fetch size to 500

adamkorynta:
The default size can get set via properties on the connection pool. See `connectionProperties` https://tomcat.apache.org/tomcat-7.0-doc/jdbc-pool.html

MikeNeilson:
I'd have to check in more detail, but there are two pool classes in Tomcat by default and I believe we use the other one. May have a similar setting though.

MikeNeilson (approved):
Excellent improvement of the software in general. Thanks for taking the time to address performance fixes with the effort of related work and making an appropriate refactor that improves performance and clarity.

PR #878, opened 2024-09-23
Feature/scaled testing

RyanM-RMA:
This PR contains a scaled integration test that initially stores gate changes for 1 year of 5 minute data for 18 gates.  If this data already exists, it does not attempt to store the data, because it's a very time consuming process.
The actual test retrieves data for the entire range, but requests a row limit equal to 2 weeks of data and then logs the length of data and characters it received.  This will help us identify if the JSON is too bulky.
Additionally, because the data storage is separated, the unit test will provide timing, and metrics are stored for the actual data retrieval.


cwms-data-api/src/test/java/cwms/cda/api/location/kind/GateChangeStressTest.java
```diff
@@ -0,0 +1,250 @@
...
+            .withOfficeId(OFFICE_ID)
+            .build();
+
+    private static final TemporalAmount INTERVAL = Duration.of(5, ChronoUnit.MINUTES);
+    private static final ZonedDateTime START = ZonedDateTime.of(2020, 1, 1, 0, 0, 0, 0, NumericalConstants.UTC_ZONEID);
+    private static final ZonedDateTime END = START.plusYears(1).plus(INTERVAL);
+
+    @BeforeAll
+    public static void setup() throws Exception {
+        //1 year of 5 minute gate change data for 18 gates
```
MikeNeilson: Who changes a gate every 5 minutes? Or is this just an extreme example to cause problems?

cwms-data-api/src/test/java/cwms/cda/api/location/kind/GateChangeStressTest.java
```diff
@@ -0,0 +1,250 @@
...
+            changes.forEach(changeBatch -> {
+                LOGGER.atSevere().log("Storing " + changeBatch.size() + " changes from " + changeBatch.get(0).getChangeDate() + " to " + changeBatch.get(changeBatch.size() - 1));
+                dao.storeOperationalChanges(changeBatch, true);
+            });
+
+        }, CwmsDataApiSetupCallback.getWebUser());
+    }
+
+    @Test
+    void stress_test() {
```
MikeNeilson: Is there a better word than stress here? When I saw that immediately thought "parallel" not bulk data.

RyanM-RMA: should be scaled, sorry

cwms-data-api/src/test/java/cwms/cda/api/location/kind/GateChangeStressTest.java
RyanM-RMA: This was the most extreme example of something like SCADA data, or some other automated process.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/GateChangeStressTest.java
RyanM-RMA: or scaling?

MikeNeilson: Honestly that seems even more like something that should be parallel.

But yeah, that seems reasonable. Or at least as reasonable as the english language allows. Perhaps someone else will come up with something.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/GateChangeStressTest.java
MikeNeilson: Crazy. Thanks for the info.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/GateChangeScaleTest.java
```diff
@@ -0,0 +1,249 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import usace.cwms.db.jooq.codegen.packages.CWMS_PROJECT_PACKAGE;
+import static cwms.cda.api.Controllers.*;
+import static cwms.cda.data.dao.DaoTest.getDslContext;
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.*;
+
+class GateChangeScaleTest extends BaseOutletDaoIT {
```
RyanM-RMA, line 71: Disable the test, so it's not run in automation

MikeNeilson: Don't disable the test. use `@EnableIfProperty` (that might not be exact) so we don't need to change code to run the test when needed.

RyanM-RMA: Oh, that's a good idea, thank you.  Didn't know that was a thing.

MikeNeilson: FYI, I'd use Property over environment variable. I don't think we have the gradle build pass in the environment but we do pass in properties. There's a few other tests that I believe use this mechanism you can look at for an example.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/GateChangeScaleTest.java
```diff
@@ -77,6 +82,18 @@ class GateChangeScaleTest extends BaseOutletDaoIT {
     private static final TemporalAmount INTERVAL = Duration.of(5, ChronoUnit.MINUTES);
     private static final ZonedDateTime START = ZonedDateTime.of(2020, 1, 1, 0, 0, 0, 0, NumericalConstants.UTC_ZONEID);
     private static final ZonedDateTime END = START.plusYears(1).plus(INTERVAL);
+    private static final ZonedDateTime TS_START = ZonedDateTime.of(2024, 5, 31, 0, 0, 0, 0, NumericalConstants.UTC_ZONEID);
+    private static final Duration TS_DURATION = Duration.ofHours(120);
+
+    private static final String CWBI = "https://cwms-data-test.cwbi.us";
```
MikeNeilson, line 88: Tests should not be pointed at a specific URL.
1. This is brittle, we cannot rely on cwbi-test services to be available; by definition even if in practice it usually is
2. It limits who can do the testing.

In the setup section default this to the test environment URL; if it's important to point it to another location at times allow a property override.

cwms-data-api/src/test/java/cwms/cda/api/location/kind/GateChangeScaleTest.java
```diff
@@ -86,87 +103,224 @@ public static void setup() throws Exception {
...
+        LOGGER.atInfo().log("CWBI Output character length: " + body.length() + "\n" +
+                                    Charset.defaultCharset().displayName() + " byte length: " + body.getBytes().length);
+    }
+
+    @RepeatedTest(10)
+    void retrieve_test_ts_data(RepetitionInfo info) {
+        LOGGER.atInfo().log("Retrieving TS data with page size: " + PAGE_SIZE);
+        ZonedDateTime startTime = TS_START;
+
+        for (int i = 1; i < info.getCurrentRepetition(); i++) {
```
MikeNeilson, line 259: Interesting use of repeated test to create a sliding window.

MikeNeilson (changes requested):

PR #880, opened 2024-09-24
CWDB-225 - Implemented MeasurementDao with unit and integration tests

### LINKED ISSUE #394: add support for Measurements
The CWMSVue rating editor uses the streamflow measurement API. It calls on package procedures that return XML. It is not necessary that CDA do the same.

getOne - get all measurements for one location with optional date range and optional number range
getAll - get all measurements for multiple locations using parameterized location filter with optional date range and optional number range
post - store multiple measurements for a single location, with fail-if-exists parameter
delete - needs to parameterize the location, start/end date range OR measurement # min/max range

rma-bryson:


MikeNeilson (commented):
Looks reasonable to me.

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
```diff
@@ -0,0 +1,500 @@
...
+        });
+    }
+
+    /**
+     * Delete a measurement
+     *
+     * @param officeId   - the office id
+     * @param locationId - the location id of the measurement to delete
+     */
+    public void deleteMeasurements(String officeId, String locationId) {
```
RyanM-RMA: The null variables below should be passed in, except the Timestamps, those should be an Instant.  This provides more flexibility for the caller.  They can be added to the Controller later as optional parameters.

cwms-data-api/src/test/java/cwms/cda/data/dao/MeasurementDaoTest.java
```diff
@@ -0,0 +1,238 @@
...
+        assertEquals("Test comment", measurement.getWmComments());
+        assertEquals("ft2", measurement.getAreaUnit());
+        assertEquals("cfs", measurement.getFlowUnit());
+        assertEquals("ft", measurement.getHeightUnit());
+        assertEquals("fps", measurement.getVelocityUnit());
+        assertEquals("F", measurement.getTempUnit());
+        assertTrue(measurement.isUsed());
+
+        //Assertions for Streamflow meas fields
+        assertEquals(100.0, measurement.getStreamflowMeasurement().getFlow());
```
RyanM-RMA: Use assertEquals for doubles where it includes a fudge factor.  The fudge factor should be pretty small, like 0.001.  This applies to all your double assertions.

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
```diff
@@ -0,0 +1,500 @@
...
+        });
+    }
+
+    /**
+     * Store a measurement
+     *
+     * @param measurement  - the measurement to store
+     * @param failIfExists - if true, fail if the measurement already exists
+     */
+    public void storeMeasurement(Measurement measurement, boolean failIfExists) {
```
RyanM-RMA, line 124: Why do we only have a store single measurement option instead of a list of measurements?  This is going to be incredibly slow if we're importing a thousand measurements.

RyanM-RMA (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
rma-bryson: Updated to parameterize the dao with null fields, which means the controller will be responsible for deciding which fields are passed in as null.

cwms-data-api/src/test/java/cwms/cda/data/dao/MeasurementDaoTest.java
rma-bryson: right, missed that in the unit tests. Updated

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
rma-bryson: I don't think CDA supports bulk store calls anywhere else. From the hec-cwms-data-access perspective, I would iterate over a collection of measurements and make individual POST calls. 

Similar to how the measurement jdbc dao currently iterates over the set and makes individual store calls into the jooq layer.

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
```diff
@@ -123,20 +117,13 @@ public void storeMeasurement(Measurement measurement, boolean failIfExists) {
...
-            Number maxFlow = null;
-            String minNum = null;
-            String maxNum = null;
-            String agencies = null;
-            String qualities = null;
-            Timestamp minTimestamp = null;
-            Timestamp maxTimestamp = null;
-            String unitSystem = null;
+            Timestamp minTimestamp = minDateMask == null ? null : Timestamp.from(maxDateMask);
+            Timestamp maxTimestamp = minDateMask == null ? null : Timestamp.from(maxDateMask);
```
RyanM-RMA: Why are you doing it this way instead of at line 88/89 where it's different?
`Timestamp maxTimestamp = OracleTypeMap.buildTimestamp(maxDateMask == null ? null : Date.from(maxDateMask));`

RyanM-RMA (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
RyanM-RMA: As we discussed on Teams, please try doing a multi-store first, and if it works, please use that instead.  The DB does indicate it supports multiple storage.
Additionally, we should make sure we update the JDBC dao during HCDA updates to use the multi-store, because that would be an additional speed improvement and keep the behavior consistent.

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
rma-bryson: because I just forgot we had something that builds that. fixed it.

RyanM-RMA: Oh man, I missed the minDateMask bug too, great catch.

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
rma-bryson: Tested out the bulk store and looks like it works successfully. Added unit tests and integrations tests.

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
```diff
@@ -132,6 +138,28 @@ public void storeMeasurements(List<Measurement> measurements, boolean failIfExis
...
+    public void updateMeasurement(Measurement measurement) {
+        verifyMeasurementExists(measurement);
+        storeMeasurement(measurement, false);
+    }
+
+    /**
+     * Updates a list of existing measurements
+     * @param measurements - the measurements to update
+     */
+    public void updateMeasurements(List<Measurement> measurements)
```
RyanM-RMA, line 170: What's the purpose of these update methods?

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
```diff
@@ -132,6 +138,28 @@ public void storeMeasurements(List<Measurement> measurements, boolean failIfExis
...
+    {
+        for(Measurement measurement : measurements)
+        {
+            verifyMeasurementExists(measurement);
```
RyanM-RMA: This could potentially do a lot of database calls, I think this should be thought out a bit differently.  I also don't understand why we care if the measurement exists or not.

RyanM-RMA (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
rma-bryson: Since we don't have a rename, what we've done in the past for a patch calls is treated update = check for existing + store + don't fail if exists.

RyanM-RMA: ah gotcha, so it's fail if doesn't exist.

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
RyanM-RMA: ok, so I looked over this again and I understand a bit more.  I think you'll want to do a custom query checking against the location and the numbers available for that location.  so iterate over the unique locations, collect all of the numbers for each location, then you'd do something like `select * from AV_STREAMFLOW_MEAS where LOCATION_ID = '{your loc}' and OFFICE_ID = '{loc_office_id}' and NUMBER in [{your collection of numbers as strings}]` and then compare the lengths of the rows found to see if you found everything.  If not, then you collect the missing numbers and report them.  You could even report them per location.  That would give far more detail than a single exception for one measurement, and it would be 1 db call per unique location.

rma-bryson: Updated to use jooq query of specific number match to compare against

RyanM-RMA (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/MeasurementDao.java
adamkorynta: I disagree with the API choice here. We shouldn't be requiring null parameters for unused inputs. I'm asking Bryson to revert in:https://github.com/USACE/cwms-data-api/pull/890/files#diff-2e2e24ebae962a792534301fa3000a230f3241295a9110caa403c5f92e08264bR216.

PR #887, opened 2024-09-28
LocationLevel retrieve constant values in time window

### LINKED ISSUE #599: Allow retrieval of constant value location levels over a date range
For non-timeseries location levels, I'd like to specify a date range and see all the values it was set to within that date range.

For example...
Assuming the levels table contains these:
```
LOCATION_LEVEL_ID				LEVEL_DATE	LEVEL_UNIT	CONSTANT_LEVEL
Lake Isabella.Stor.Inst.0.Top of Gross Pool	01-JAN-00	ac-ft		568100.0
Lake Isabella.Stor.Inst.0.Top of Gross Pool	01-OCT-10	ac-ft		361249.99999999994
Lake Isabella.Stor.Inst.0.Top of Gross Pool	13-JUN-23	ac-ft		568100.0

```
Notice the LEVEL_DATE is different.

I'd like to supply a start and end date, and get a result like this:
```
{
  "office-id": "SPK",
  "seasonal-values": [],
  "constant-value": 568100,
  "constant-values": [
    ["1900-01-01T00:00:00-07:00", 568100],
    ["2010-10-01T00:00:00-07:00", 361249.99999999994],
    ["2024-06-13T00:00:00-07:00", 568100],
  ],
  "level-date-start": "1900-01-01T00:00:00-07:00",
  "level-date-end": "2024-06-13T00:00:00-07:00",
  "level-units-id": "ac-ft",
  "level-date": "2024-06-13T00:00:00-07:00",
  "interval-months": 0,
  "interval-minutes": 0,
  "attribute-units-id": "ac-ft"
}
```

For the regular single point-in-time query, I'd expect "constant-values" array to be empty, if present at all. At least if compatibility with the current version is desired. If not, making the entire thing an array even for a single result is fine with me.
For compatibility, I left the current behavior in there as well, where it returns "constant-value" and "level-date", which is set to the most recent value in the date range requested. That can be changed if desired too.

I'm open to various output formats, I listed the above as an example.
Personally, I'd prefer the unix epoch timestamp, like in timeseries, but this levels call already has its own date format. Though, the response doesn't include the timezone, so it's not round-trippable. That's a separate issue though.

zack-rma:
Fixes #599, LocationLevel retrieve constant values in time window. Includes tests to verify functionality


MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -344,6 +316,12 @@ public void getAll(@NotNull Context ctx) {
                         + "specified), as well as the time zone of any times in the response."
                         + " If this field is not specified, the default time zone of UTC "
                         + "shall be used."),
+                @OpenApiParam(name = START, type = Instant.class, description = "Specifies the start of the time "
```
adamkorynta, line 335: Adding a start and end date for getOne does not make sense to me as it conflicts with the effective date parameter.

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
MikeNeilson: That's fair, but because there are multiple effective dates there does need to be a way to retrieve them all without it necessarily being a "timeseries" as that's a different concept (seasonal, etc) of location level.

adamkorynta: I don't see how that's different than the behavior of the getAll endpoint.

MikeNeilson: We'll see how Daniel responds.

zack-rma: @DanielTOsborne

zack-rma: @DanielTOsborne Following up for your thoughts on this

rma-rripken: @DanielTOsborne  or @MikeNeilson  Can we get a decision on this?  If we can't decide what makes sense for getOne we could always not have a getOne and force users through getAll.  I guess I'd like to see this merged and a new issue/pr opened to figure out next steps.  People could try this out in the meantime

MikeNeilson: Agreed. I believe people are already using the getOne though so I don't think we should remove it at this time. But PR itself should be appropriately updated and the work merged so we can move on.

If we don't get responses during development we're definitely going to get them after release.

zack-rma:
Closing since branch is too out of date to reasonably merge. Superseded by PR: https://github.com/USACE/cwms-data-api/pull/1202

PR #889, opened 2024-09-30
Issue 765 - Addressed LocationLevel Catalog requiring start date

### LINKED ISSUE #765: Getting location level list requires knowing valid begin time
A user recently tried to list location levels matching a mask and received no results. Changing begin to 1900-01-01 provided all appropriate results, but doesn't seem the behavior we desire.

Either the simple GET endpoint of levels should act like the catalog giving the requestor the known results, or we should add a new dataset to the /catalog endpoint.

zack-rma:
Fixes [#765](https://github.com/USACE/cwms-data-api/issues/765) - Adjusted start date default to 1900

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
@@ -67,7 +67,6 @@ import com.fasterxml.jackson.dataformat.xml.JacksonXmlModule;
 import com.fasterxml.jackson.dataformat.xml.XmlMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import cwms.cda.api.enums.UnitSystem;
-import cwms.cda.api.errors.CdaError;
 import cwms.cda.data.dao.LocationLevelsDao;
 import cwms.cda.data.dao.LocationLevelsDaoImpl;
 import cwms.cda.data.dto.LocationLevel;
@@ -220,8 +219,8 @@ public class LevelsController implements CrudHandler {
                         + "specified or default units above the NGVD-29 datum."),
                 @OpenApiParam(name = BEGIN, description = "Specifies the start of the time "
                         + "window for data to be included in the response. If this field is "
-                        + "not specified, any required time window begins 24 hours prior to "
-                        + "the specified or default end time."),
+                        + "not specified, any required time window begins 1/1/1900 at "
+                        + "the specified or default timezone."),
                 @OpenApiParam(name = END, description = "Specifies the end of the time "
                         + "window for data to be included in the response. If this field is "
                         + "not specified, any required time window ends at the current time"),
@@ -295,7 +294,7 @@ public class LevelsController implements CrudHandler {
                 if (begin != null) {
                     beginZdt = DateUtils.parseUserDate(begin, timezone);
                 } else {
-                    beginZdt = endZdt.minusHours(24);
+                    beginZdt = ZonedDateTime.of(1900, 1, 1, 0, 0, 0, 0, tz);
                 }
 
                 LocationLevels levels = levelsDao.getLocationLevels(cursor, pageSize, levelIdMask,
--- a/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
@@ -89,7 +89,6 @@ import io.javalin.plugin.openapi.annotations.OpenApiParam;
 import io.javalin.plugin.openapi.annotations.OpenApiRequestBody;
 import io.javalin.plugin.openapi.annotations.OpenApiResponse;
 import java.math.BigDecimal;
-import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -295,7 +294,7 @@ public void getAll(@NotNull Context ctx) {
                 if (begin != null) {
                     beginZdt = DateUtils.parseUserDate(begin, timezone);
                 } else {
-                    beginZdt = endZdt.minusHours(24);
+                    beginZdt = ZonedDateTime.of(1900, 1, 1, 0, 0, 0, 0, tz);
```
MikeNeilson: While generally everything practical does start at 1900, it would probably make sense to set this as far back as ZonedDateTime allows. Does ZonedDateTime have an "earliest date" function?

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: I looked into this a bit. ZonedDateTime does not appear to have an earliest time function, but I tried generating an earliest time using ZonedDateTime from Instant.MIN. Unfortunately, this is outside of the date values supported, so I went with the earliest supported Instant EpochMilli of -999999999.

MikeNeilson: Works for me. Thanks.

MikeNeilson (dismissed):

zack-rma:
Rebased to fix build errors

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -295,7 +295,8 @@ public void getAll(@NotNull Context ctx) {
                 if (begin != null) {
                     beginZdt = DateUtils.parseUserDate(begin, timezone);
                 } else {
-                    beginZdt = endZdt.minusHours(24);
+                    // uses earliest supported time in CWMS
+                    beginZdt = ZonedDateTime.ofInstant(Instant.ofEpochMilli(-999999999L), tz);
```
adamkorynta: this instant is `1969-12-20T10:13:20.001Z` which is definitely not the earliest supported time in CWMS

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -220,8 +220,8 @@ public void delete(@NotNull Context ctx, @NotNull String levelId) {
                         + "specified or default units above the NGVD-29 datum."),
                 @OpenApiParam(name = BEGIN, description = "Specifies the start of the time "
                         + "window for data to be included in the response. If this field is "
-                        + "not specified, any required time window begins 24 hours prior to "
-                        + "the specified or default end time."),
+                        + "not specified, any required time window begins 1/1/1900 at "
```
adamkorynta: This date doesn't match up with what is actually used.

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
adamkorynta: Tests should be added to test edge case of the earliest value

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
adamkorynta: I would think leaving off the beginning or ending empty would just remove them from the select query bounds rather filling in a default.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: I realized the instant value I used is actually the minimum year allowed by CWMS.

I added a test case for the minimum time allowed by the database, which is much more recent than the CWMS allowed value.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: I updated the description to better match actual behavior. I also changed the default values to null, which causes the DAO to remove them from the query.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -280,23 +278,14 @@ public void getAll(@NotNull Context ctx) {
...
-
-                ZonedDateTime endZdt = end != null ? DateUtils.parseUserDate(end, timezone) :
-                        ZonedDateTime.now(tz);
-                ZonedDateTime beginZdt;
-                if (begin != null) {
-                    beginZdt = DateUtils.parseUserDate(begin, timezone);
-                } else {
-                    beginZdt = endZdt.minusHours(24);
-                }
+                ZonedDateTime endZdt = end != null ? DateUtils.parseUserDate(end, timezone) : null;
```
adamkorynta: use `requiredInstant(ctx, END);` or `requiredZdt(ctx, END);` (not sure why this time needs to be zoned).

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -315,12 +304,9 @@ public void getAll(@NotNull Context ctx) {
                 ctx.status(HttpServletResponse.SC_OK);
                 ctx.result(results);
                 requestResultSize.update(results.length());
-                if (isLegacyVersion)
-                {
+                if (isLegacyVersion) {
```
adamkorynta, line 308: isn't this always false given the if statement above?

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: The beginning and end parameters are not required fields. I replaced the ternaries with `queryParamAsZdt` instead.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Not necessarily. If the format parameter is specified, the statement above will be skipped and the version can still be either legacy or not.

adamkorynta: oh I see, it's one of those that uses the format parameter, not just the header

adamkorynta (approved):

PR #890, opened 2024-10-02
CWDB-225 - Implemented Measurement Controller

### LINKED ISSUE #394: add support for Measurements

rma-bryson:


cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -525,6 +528,10 @@ protected void configureRoutes() {
                 new StreamLocationController(metrics), requiredRoles,5, TimeUnit.MINUTES);
         cdaCrudCache(format("/stream-reaches/{%s}", NAME),
                 new StreamReachController(metrics), requiredRoles,1, TimeUnit.DAYS);
+        String measurements = "/measurements/";
+        patch(measurements ,new MeasurementPatchController(metrics));
```
zack-rma: Nitpick: formatting

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    private Timer.Context markAndTime(String subject) {
+        return Controllers.markAndTime(metrics, getClass().getName(), subject);
+    }
+
+    @OpenApi(
+            queryParams = {
+                    @OpenApiParam(name = OFFICE_MASK, description = "Office id for filtering measurements."),
+                    @OpenApiParam(name = ID_MASK, description = "Location id for filtering measurements."),
```
zack-rma: I would suggest making these descriptions a bit more verbose. Maybe mention that it's a mask and whether it supports * matching.

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+                    @OpenApiParam(name = MIN_NUMBER, description = "Minimum number for filtering measurements."),
+                    @OpenApiParam(name = MAX_NUMBER, description = "Maximum number for filtering measurements."),
+                    @OpenApiParam(name = MIN_DATE, description = "Minimum date for filtering measurements."),
```
zack-rma: I would also recommend adding expected date formatting, since that seems to be a common issue that comes up.

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+        String officeId = ctx.queryParam(OFFICE_MASK);
+        String locationId = ctx.queryParam(ID_MASK);
+        String unitSystem = ctx.queryParamAsClass(UNIT_SYSTEM, String.class).getOrDefault(UnitSystem.EN.value());
+        Instant minDate = parseInstant(ctx.queryParam(MIN_DATE));
+        Instant maxDate = parseInstant(ctx.queryParam(MAX_DATE));
+        String minNum = ctx.queryParam(MIN_NUMBER);
+        String maxNum = ctx.queryParam(MAX_NUMBER);
+        Number minHeight = null;
+        Number maxHeight = null;
+        Number minFlow = null;
```
zack-rma: Are these not fields to be used as filters? Is there a reason they aren't available as input parameters?

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+
+    private Timer.Context markAndTime(String subject) {
+        return Controllers.markAndTime(metrics, getClass().getName(), subject);
+    }
+
+    @OpenApi(
+            queryParams = {
+                    @OpenApiParam(name = OFFICE_MASK, description = "Office id for filtering measurements."),
+                    @OpenApiParam(name = ID_MASK, description = "Location id for filtering measurements."),
+                    @OpenApiParam(name = MIN_NUMBER, description = "Minimum number for filtering measurements."),
```
zack-rma: Would these be the min and max of the measurement value? I feel that the description could be a bit clearer as to what these are filtering on.

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+            throw new UnsupportedOperationException("Not supported with required location Id");
+        }
+    }
+
+    @OpenApi(
+            queryParams = {
+                    @OpenApiParam(name = OFFICE, required = true, description = "Specifies the office of the measurements to delete"),
+                    @OpenApiParam(name = MIN_NUMBER, description = "Specifies the min number of the measurement to delete."),
+                    @OpenApiParam(name = MAX_NUMBER, description = "Specifies the max number of the measurement to delete."),
+                    @OpenApiParam(name = MIN_DATE, description = "Specifies the minimum date of the measurement to delete."),
```
zack-rma: Same comment about expected date format

cwms-data-api/src/main/java/cwms/cda/api/MeasurementPatchController.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @OpenApi(
+            requestBody = @OpenApiRequestBody(
+                    content = {
+                            @OpenApiContent(isArray = true, from = Measurement.class, type = Formats.JSONV1),
+                            @OpenApiContent(from = Measurement.class, type = Formats.JSON)
+                    },
+                    required = true),
+            description = "Update Measurement",
```
zack-rma: Consider specifying what kind of update this results in. Some updates are just renames, this appears be a full data update.

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
```diff
@@ -0,0 +1,630 @@
...
+                .queryParam(Controllers.MIN_NUMBER, number)
+                .queryParam(Controllers.MAX_NUMBER, number)
+                .queryParam(Controllers.UNIT_SYSTEM, UnitSystem.EN.getValue())
+        .when()
+                .redirects().follow(true)
+                .redirects().max(3)
+                .get("/measurements/")
+        .then()
+                .log().ifValidationFails(LogDetail.ALL, true)
+                .assertThat()
```
zack-rma: Nitpick: indentation

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
```diff
@@ -0,0 +1,630 @@
...
+                .body("[0].usgs-measurement.control-condition", equalTo(updatedMeasurement.getUsgsMeasurement().getControlCondition()))
+                .body("[0].usgs-measurement.flow-adjustment", equalTo(updatedMeasurement.getUsgsMeasurement().getFlowAdjustment()))
+                .body("[0].usgs-measurement.shift-used", equalTo(updatedMeasurement.getUsgsMeasurement().getShiftUsed().floatValue()))
+                .body("[0].usgs-measurement.percent-difference", equalTo(updatedMeasurement.getUsgsMeasurement().getPercentDifference().floatValue()))
+                .body("[0].usgs-measurement.delta-height", equalTo(updatedMeasurement.getUsgsMeasurement().getDeltaHeight().floatValue()))
+                .body("[0].usgs-measurement.delta-time", equalTo(updatedMeasurement.getUsgsMeasurement().getDeltaTime().floatValue()))
+                .body("[0].usgs-measurement.air-temp", equalTo(updatedMeasurement.getUsgsMeasurement().getAirTemp().floatValue()))
+                .body("[0].usgs-measurement.water-temp", equalTo(updatedMeasurement.getUsgsMeasurement().getWaterTemp().floatValue()));
+
+        // Delete the Measurement
```
zack-rma, line 221: Consider adding measurement deletion to the afterall cleanup routine, just in case the test fails before cleanup can occur within the test.

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
```diff
@@ -0,0 +1,630 @@
...
+        .then()
+                .log().ifValidationFails(LogDetail.ALL, true)
+        .assertThat()
+                .statusCode(is(HttpServletResponse.SC_NOT_FOUND));
+    }
+
+    @Test
+    void test_delete_does_not_exist() {
+        TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
+        // Delete a Embankment
```
zack-rma: update comment

zack-rma (commented):
Looks good, just a few suggestions and points of clarification

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: We aren't supporting these fields currently for filtering via the GET endpoint input parameters. I separated them out here in case that ever becomes something supported in the future.

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: For measurements, there is a number field which isn't really a number. Its a String value that acts as an identifier, but there is order to it. 

Normally it is a number like "1", "2", etc.. but it could be "1A", "1.B", etc. 

I updated the description to use "number-id" to be more clear.

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
rma-bryson: Good idea. Added measurement deletion call to afterall.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
rma-bryson: fixed

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: added expected ISO-8601 date formatting

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: added expected ISO-8601 date formatting

cwms-data-api/src/main/java/cwms/cda/api/MeasurementPatchController.java
rma-bryson: Added clarifying description

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
rma-bryson: fixed

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
rma-bryson: fixed

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
```diff
@@ -0,0 +1,654 @@
...
+    public static void tearDown() {
+        for (Stream stream : STREAMS_CREATED) {
+            try {
+                CwmsDatabaseContainer<?> db = CwmsDataApiSetupCallback.getDatabaseLink();
+                db.connection(c -> {
+                    StreamDao streamDao = new StreamDao(getDslContext(c, OFFICE_ID));
+                    try {
+                        streamDao.deleteStream(stream.getId().getOfficeId(), stream.getId().getName(), DeleteRule.DELETE_ALL);
+                    } catch (Exception e) {
+                        // ignore
```
zack-rma: You may want to log these exceptions with a logger, just so we can be aware when this fails to delete a stream.

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
```diff
@@ -0,0 +1,654 @@
...
+        for(String measLoc: STREAM_LOC_IDS)
+        {
+            try {
+                CwmsDatabaseContainer<?> db = CwmsDataApiSetupCallback.getDatabaseLink();
+                db.connection(c -> {
+                    MeasurementDao measDao = new MeasurementDao(getDslContext(c, OFFICE_ID));
+                    try {
+                        measDao.deleteMeasurements(OFFICE_ID, measLoc, null, null, null, null, null, null, null, null, null, null, null);
+                    } catch (Exception e) {
+                        // ignore
```
zack-rma: Same comment as above

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
```diff
@@ -0,0 +1,654 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import static org.hamcrest.Matchers.*;
+
+@Tag("integration")
+final class MeasurementControllerTestIT extends DataApiTestIT {
+
+    private static final String OFFICE_ID = TestAccounts.KeyUser.SPK_NORMAL.getOperatingOffice();
+    private static final List<Stream> STREAMS_CREATED = new ArrayList<>();
```
zack-rma: Consider naming these lists something like `STREAMS_TO_CLEANUP` to clarify their purpose.

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
```diff
@@ -0,0 +1,654 @@
...
+                    }
+                }, CwmsDataApiSetupCallback.getWebUser());
+            } catch (SQLException ex) {
+                throw new RuntimeException(ex);
+            }
+        }
+    }
+
+    @Test
+    @Disabled
```
zack-rma: Why are all the tests disabled? Schema support?

zack-rma (commented):
Some suggestions and a question

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
rma-bryson: I added a logging at info level to display a failure

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
rma-bryson: I added a logging at info level to display a failure, though it is expected that this will fail if the deletion works as expected in the places where deletion is being tested.

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
rma-bryson: Renamed to TEST_STREAMS. The streams get created, used, and then cleaned up for test purposes, so its more than just cleanup.

cwms-data-api/src/test/java/cwms/cda/api/MeasurementControllerTestIT.java
rma-bryson: yeah, the tests will run against the prod schema and fail, since it doesn't have the measurement updates.

zack-rma: Maybe consider adding a note to the disabled tag about that until the schema updates are merged in

MikeNeilson: Better than a note, Use `@EnableIfProperty`

We may also want to consider a more advance enable mechanism that checks the database version to make some of these easier to deal with.

@adamkorynta thoughts?

adamkorynta: here's my idea: https://github.com/USACE/cwms-data-api/pull/917

rma-bryson: Updated to use `@MinimumSchema` tag

adamkorynta (changes requested):
some failing integration tests still

rma-bryson:
@MikeNeilson - With Measurements we wanted to support a bulk store, and there is no rename, so update is really just a check-for-existing and then store... The test db doesn't have display units for the office (SPK) needed for bulk retrieving existing measurements efficiently (there is jooq code in the MeasurementDao that retrieves in bulk from the meas view, with the given meas numbers), which causes the tests to fail, without configuring the at_display_units table manually for SPK . If configured, the existing check does work as expected. 

EDIT: after further discussion we decided its best to just not support an update at all for measurements, as it doesn't make much sense. No PATCH, just POST.

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+            throw new UnsupportedOperationException(NOT_SUPPORTED_YET);
+        }
+
+    }
+
+    @OpenApi(
+            requestBody = @OpenApiRequestBody(
+                    content = {
+                            @OpenApiContent(isArray = true, from = Measurement.class, type = Formats.JSONV1),
+                            @OpenApiContent(from = Measurement.class, type = Formats.JSON)
```
adamkorynta: this needs the `isArray` too

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+    public void create(Context ctx) {
+
+        try (Timer.Context ignored = markAndTime(CREATE)) {
+            String formatHeader = ctx.req.getContentType();
+            ContentType contentType = Formats.parseHeader(formatHeader, Measurement.class);
+            List<Measurement> measurements = parseMeasurements(ctx, contentType);
+            boolean failIfExists = ctx.queryParamAsClass(FAIL_IF_EXISTS, Boolean.class).getOrDefault(true);
+            DSLContext dsl = getDslContext(ctx);
+            MeasurementDao dao = new MeasurementDao(dsl);
+            if(measurements.size() == 1) {
```
adamkorynta: why two different calls?

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+    public void delete(@NotNull Context ctx, @NotNull String locationId) {
+        String officeId = requiredParam(ctx, OFFICE);
+        String minNum = ctx.queryParam(MIN_NUMBER);
+        String maxNum = ctx.queryParam(MAX_NUMBER);
+        Instant minDate = parseInstant(ctx.queryParam(MIN_DATE));
+        Instant maxDate = parseInstant(ctx.queryParam(MAX_DATE));
+        try (Timer.Context ignored = markAndTime(DELETE)) {
+            DSLContext dsl = getDslContext(ctx);
+            MeasurementDao dao = new MeasurementDao(dsl);
+            dao.deleteMeasurements(officeId, locationId, minDate, maxDate, UnitSystem.EN.getValue(), null,
```
adamkorynta: why so many nulls? what are the other usages of the dao?

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+                    @OpenApiResponse(status = "204", description = "Measurement successfully deleted."),
+                    @OpenApiResponse(status = "404", description = "Measurement not found.")
+            }
+    )
+    @Override
+    public void delete(@NotNull Context ctx, @NotNull String locationId) {
+        String officeId = requiredParam(ctx, OFFICE);
+        String minNum = ctx.queryParam(MIN_NUMBER);
+        String maxNum = ctx.queryParam(MAX_NUMBER);
+        Instant minDate = parseInstant(ctx.queryParam(MIN_DATE));
```
adamkorynta: instead of writing your own parser, use the utilities in the base class

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+            throw new UnsupportedOperationException(NOT_SUPPORTED_YET);
+        }
+    }
+
+    @OpenApi(
+            queryParams = {
+                    @OpenApiParam(name = OFFICE, required = true, description = "Specifies the office of the measurements to delete"),
+                    @OpenApiParam(name = MIN_NUMBER, description = "Specifies the min number-id of the measurement to delete."),
+                    @OpenApiParam(name = MAX_NUMBER, description = "Specifies the max number-id of the measurement to delete."),
+                    @OpenApiParam(name = MIN_DATE, description = "Specifies the minimum date (in ISO-8601 format) of the measurement to delete."),
```
adamkorynta: maintain consistency with dates in the other API's by providing a time zone

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+    public void update(@NotNull Context ctx, @NotNull String locationId) {
+        try (final Timer.Context ignored = markAndTime(GET_ONE)) {
+            throw new UnsupportedOperationException(NOT_SUPPORTED_YET);
+        }
+    }
+
+    @OpenApi(
+            queryParams = {
+                    @OpenApiParam(name = OFFICE, required = true, description = "Specifies the office of the measurements to delete"),
+                    @OpenApiParam(name = MIN_NUMBER, description = "Specifies the min number-id of the measurement to delete."),
```
adamkorynta, line 241: delete should be operating on a location

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
```diff
@@ -0,0 +1,247 @@
...
+            retVal = Instant.parse(date);
+        }
+        return retVal;
+    }
+
+    static List<Measurement> parseMeasurements(@NotNull Context ctx, ContentType contentType) {
+        List<Measurement> measurements;
+        try {
+            measurements = Formats.parseContentList(contentType, ctx.body(), Measurement.class);
+        } catch (FormattingException e) {
```
adamkorynta: This should throw an error to the client.

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: This is the only usage of the dao. I also updated this to support all of these as optional query params. I originally only supported a limited number of ones that I knew were in use. But I don't see a reason to not just let them all be passed-through.

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: Originally, I was supporting both single and multi-store... meaning also supporting json without the [...] array brackets. But I think its better to just use arrays with a single entry in that case to reduce complexity and maintenance in the api.

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: updated

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: updated to base class utility

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: added timezone param

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: added path param for location id

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: Updated to throw error to client now that only lists are supported

cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
```diff
@@ -375,6 +382,16 @@ public static <T> T requiredParamAs(io.javalin.http.Context ctx, String name, Cl
             .getOrThrow(e -> new RequiredQueryParameterException(name));
     }
 
+    @Nullable
+    public static Double queryParamAsDouble(Context ctx, String param) {
```
adamkorynta, line 383: Should get a unit test

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
adamkorynta: the schema descriptions should be consistent for dates and time zone. it has precise language.

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/MeasurementController.java
rma-bryson: Made descriptions consistent with other uses of date and tz params. I also updated `min_date` and `max_date` to be `begin` and `end` for consistency.

cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
rma-bryson: Added unit test

adamkorynta (approved):

PR #897, opened 2024-10-03
Change group name for clarity

MikeNeilson:
Realized when we get the emails about dependencies every batch is called "alldependencies", changed the group name for the github actions so the reader of the email knows more directly what they'll be looking at.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,7 +19,7 @@ updates:
     schedule:
       interval: "weekly"
     groups:
-      alldependencies:
+      github-action-dependencies:
         patterns:
           - "*"
         update-types:


adamkorynta (approved):

.github/dependabot.yml
```diff
@@ -19,12 +19,12 @@ updates:
...
     groups:
-      alldependencies:
+      github-action-dependencies:
         patterns:
           - "*"
         update-types:
           - "minor"
           - "patch"
     reviewers:
       - "@MikeNeilson"
```
adamkorynta, line 29: I presume making a CDA-developers GitHub Team requires USACE org creation?

MikeNeilson: I think so. I know we have one, but I recall having issues getting contractors into it previously. But that may have been before I created this. I will investigate and perhaps change in a separate PR.

PR #900, opened 2024-10-04
Issue 886 - Updated LocationGroup and TimeSeriesGroup patch

### LINKED ISSUE #886: Patch Timeseries group office permission issues
closed issue https://github.com/USACE/cwms-data-api/issues/879 and opened this one since it is really a different issue.  

When patching a time series group owned by CWMS but with timeseries owned by a district office I get a user not authorized. I had to give myself write permissions to the CWMS office in order to get past that error. I don't think that is correct it should be checking the permission on the assigned-time-series office-id. At least that is they way it works in CWMS-Vue. Users don't need write permission to CWMS but to the time series office-ID.

Once I gave myself permission to the CWMS office I get a

cwms.cda.api.errors.NotFoundException: ORA-20025: LOCATION_ID_NOT_FOUND: The Location: "Westhope" does not exist.

But Westhope does exist for MVP. I wonder if it is checking CWMS office instead of MVP for this. 

curl -X 'PATCH'
'https://cwms-data-test.cwbi.us/cwms-data/timeseries/group/SHEF%20Data%20Acquisition?replace-assigned-ts=false&office=CWMS'
-H 'accept: /'
-H 'Authorization: apikey '
-H 'Content-Type: application/json'
-d '{
"office-id": "CWMS",
"id": "SHEF Data Acquisition",
"time-series-category": {
"office-id": "CWMS",
"id": "Data Acquisition"
},
"assigned-time-series": [
{
"office-id": "MVP",
"timeseries-id": "Westhope.Stage.Inst.15Minutes.0.CEMVP-GOES-Raw",
"alias-id": "DE32E520.HG.RZZ.0015:TZ=UTC;DLTime=false;Units=ft"
}
]
}'

Also the request body in swagger needs to be updated it shows

"assigned-time-series": [
{
"officeId": "string",
"timeseriesId": "string",
"tsCode": 0,
"aliasId": "string",
"refTsId": "string",
"attribute": 0
}

It should be

"assigned-time-series": [
{
"office-id": "string",
"timeseries-id": "string",
"ts-code": 0,
"alias-id": "string",
"refTsId": "string", -> not sure what this should be.
"attribute": 0
}

zack-rma:
Issue #886  - Updated LocationGroup and TimeSeriesGroup patch office handling to allow for assigning non-CWMS TS/locations to a CWMS-owned group


cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesGroupController.java
```diff
@@ -239,31 +239,31 @@ public void create(@NotNull Context ctx) {
...
         tags = {TAG}
     )
     @Override
-    public void update(@NotNull Context ctx, String oldGroupId) {
-
+    public void update(@NotNull Context ctx, @NotNull String oldGroupId) {
         try (Timer.Context ignored = markAndTime(CREATE)) {
             DSLContext dsl = getDslContext(ctx);
-
+            final String CWMS_OFFICE = "CWMS";
```
MikeNeilson: This should probably be a higher level constant as the value is shared between two classes.

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesGroupControllerTestIT.java
```diff
@@ -583,4 +651,313 @@ void test_add_assigned_locs() throws Exception {
             .log().ifValidationFails(LogDetail.ALL,true)
             .statusCode(is(HttpServletResponse.SC_NO_CONTENT));
     }
+
+    @Test
+    void test_patch_permissions_CWMS() throws Exception {
+        ObjectMapper mapper = new ObjectMapper();
+
+        InputStream resource = this.getClass().getResourceAsStream(
+                "/cwms/cda/api/timeseries_permission_create4.json");
```
MikeNeilson: Please provide a more meaningful name for these. I tried to skip directly to them to get a sense of data and got confused.

MikeNeilson (changes requested):
Looks reasonable. would like some names changed to make review easier before digging further.

cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
```diff
@@ -235,31 +235,32 @@ public void create(@NotNull Context ctx) {
                 + "unassign all existing locations before assigning new locations specified in the content body "
                 + "Default: false"),
             @OpenApiParam(name = OFFICE, required = true, description = "Specifies the "
-                + "owning office of the location group to be updated"),
+                + "user office making the request"),
```
adamkorynta: We probably want to make this as clear as we can. It is the office used for the owning category, group and location with the caveat that for CWMS groups it is for the location only.

adamkorynta: The description field should also get updated to describe all the operations supported by this endpoint.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesGroupController.java
```diff
@@ -239,31 +239,31 @@ public void create(@NotNull Context ctx) {
                 + "unassign all existing time series before assigning new time series specified in the content body "
                 + "Default: false"),
             @OpenApiParam(name = OFFICE, required = true, description = "Specifies the "
-                + "owning office of the time series group to be updated"),
+                + "user office making the request"),
```
adamkorynta: same comment as the other controller

cwms-data-api/src/test/java/cwms/cda/api/LocationGroupControllerTestIT.java
```diff
@@ -36,32 +44,69 @@
...
 
 import static cwms.cda.api.Controllers.*;
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.*;
 
 @Tag("integration")
-@Disabled()
 class LocationGroupControllerTestIT extends DataApiTestIT {
+    private static final Logger LOGGER = Logger.getLogger(LocationGroupControllerTestIT.class.getName());
+    private List<LocationGroup> groups = new ArrayList<>();
```
adamkorynta: It looks like these field as specifically used for cleanup operations. Rename the field variable `groupsToCleanup` or something similar for clarity.

adamkorynta (changes requested):

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesGroupControllerTestIT.java
adamkorynta: same

cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
MikeNeilson: It also seems to break the several concept we have used through CDA, though so does having an object like this owned by "CWMS" so I'm not sure there's a better way that reusing the office field in this way. 

Alternatively instead of the user needs to specifiy this here, for at least the case of assigning location, or time series in the other, it may make more sense to process them one at a time as one could provide a list of locations/timeseries each with a different office they have permissions to. Perhaps not likely now, but I could see it becoming a thing.

adamkorynta: The loc_alias_t is missing the office id component that would make this more explicit, the thing we thought of was that CDA could start that move by updating the location-id to a CwmsId and then iterate one by one. That would be a backwards incompatible change though given the JSON format restructure.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesGroupController.java
zack-rma: Moved this constant to the Controllers class.

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesGroupControllerTestIT.java
zack-rma: Renamed to timeseries_create_SPK, hopefully this is a bit clearer

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesGroupController.java
zack-rma: I expanded the description a bit to be more representative of the actual use of the parameter

cwms-data-api/src/test/java/cwms/cda/api/LocationGroupControllerTestIT.java
zack-rma: Renamed as suggested.

cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
zack-rma: I added more detail to both the endpoint description and the parameter description.

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesGroupControllerTestIT.java
adamkorynta: The confusion is the numbering. The time series all look similar, you could just edit the id at runtime to get your multiple ids in the group.

cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
adamkorynta: Can you add a ticket to address the above? Needing the location office to get included that's separate from the category/group office.

adamkorynta (changes requested):
two outstanding comments

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesGroupControllerTestIT.java
zack-rma: I reduced the number of test JSON files from 4 to 1 with TSID replacement to differentiate the Timeseries names

cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
zack-rma: I've created a DB ticket here: [https://jira.hecdev.net/browse/CWDB-308](https://jira.hecdev.net/browse/CWDB-308)

adamkorynta (dismissed):

MikeNeilson (dismissed):

zack-rma:
Rebased to fix build errors

adamkorynta (approved):

PR #907, opened 2024-10-08
Restrict access to the auth end point to users that are directly auth…

### LINKED ISSUE #906: Prevent creation of API Keys, while using an API key
Only directly authenticated users (e.g. CWMSLogin, OIDC) should be able to create keys. Apikey tokens should not allow the creation of a new token.

MikeNeilson:
…enticated.

Fixes #906.

As the required roles are (or at least should be) listed on the openapi specification, there is no leak of information to inform the user exactly which role they are missing.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
@@ -12,6 +12,7 @@ import cwms.cda.datasource.SessionTimeZonePreparer;
 import cwms.cda.helpers.ResourceHelper;
 import cwms.cda.security.CwmsAuthException;
 import cwms.cda.security.DataApiPrincipal;
+import cwms.cda.security.MissingRolesException;
 import cwms.cda.security.Role;
 import io.javalin.core.security.RouteRole;
 import io.javalin.http.Context;
@@ -35,6 +36,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
 import javax.sql.DataSource;
 import org.jetbrains.annotations.NotNull;
 import org.jooq.DSLContext;
@@ -321,8 +324,7 @@ public class AuthDao extends Dao<DataApiPrincipal> {
                 logger.atFinest().log("User has required roles.");
                 return;
             } else {
-                logger.atFine().log();
-                throw new CwmsAuthException(getFailMessage(ctx,routeRoles,p),403);
+                throw new MissingRolesException(getMissingRoles(ctx, routeRoles, p));
             }
         } else {
             throw new CwmsAuthException("No credentials provided.",401);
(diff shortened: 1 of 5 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -351,14 +353,13 @@ public void prepareGuestContext(Context ctx) {
...
-    private static String getFailMessage(@NotNull Context ctx,
+    private static String[] getMissingRoles(@NotNull Context ctx,
                                         @NotNull Set<RouteRole> requiredRoles,
                                         @NotNull DataApiPrincipal p) {
         Set<RouteRole> specifiedRoles = p.getRoles();
         Set<RouteRole> missing = new LinkedHashSet<>(requiredRoles);
         missing.removeAll(specifiedRoles);
-
-        return "Request for: " + ctx.req.getRequestURI() + " denied. Missing roles: " + missing;
+        return missing.stream().map(r -> r.toString()).collect(Collectors.toList()).toArray(new String[0]);
```
adamkorynta: Seems silly to return an array instead of a List when you're already creating a List. `stream().collect(join(","))` will do the same thing when processing the message.

adamkorynta: and in the interest of security, the list can be immutable while the array cannot be

adamkorynta (dismissed):
nitpick: array vs list, but otherwise looks good

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: yeah, it was mostly because of the downstream usage but I agree with you on that one. I think I'll just change it now. Not like it's API user facing.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -353,13 +354,16 @@ public void prepareGuestContext(Context ctx) {
...
 
-    private static String[] getMissingRoles(@NotNull Context ctx,
+    private static List<String> getMissingRoles(@NotNull Context ctx,
                                         @NotNull Set<RouteRole> requiredRoles,
                                         @NotNull DataApiPrincipal p) {
         Set<RouteRole> specifiedRoles = p.getRoles();
         Set<RouteRole> missing = new LinkedHashSet<>(requiredRoles);
         missing.removeAll(specifiedRoles);
-        return missing.stream().map(r -> r.toString()).collect(Collectors.toList()).toArray(new String[0]);
+        return Collections.unmodifiableList(
```
adamkorynta, line 363: FYI this works:
```
return missing.stream()               
.map(Object::toString)               
.collect(Collectors.toUnmodifiableList());
```

adamkorynta (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: ugh. thank you, my autocomplete was apparently lying to me. I was sure that had to be a thing.

MikeNeilson: Actually no, my autocomplete was correct, ::toUnmodifiableList was introduced after Java 8.

adamkorynta: doh!

PR #908, opened 2024-10-10
Issue 650 - Updated LocationLevel getOne to handle custom units

### LINKED ISSUE #650: Location Levels should allow arbitrary units

zack-rma:
Fixes #650 - Updated LocationLevel getOne to handle custom units and unit systems


zack-rma:
I also discovered that the effective date query parameter for getOne is used as the end date in the PL/SQL call. Effectively, any date provided that is after the effective date of the target locationLevel will work to find that locationLevel. This also results in the returned effective date being set to the provided value, not the DB value.

Not sure about the impact of this, maybe it should get its own issue? @MikeNeilson

MikeNeilson (dismissed):

MikeNeilson:
> I also discovered that the effective date query parameter for getOne is used as the end date in the PL/SQL call. Effectively, any date provided that is after the effective date of the target locationLevel will work to find that locationLevel. This also results in the returned effective date being set to the provided value, not the DB value.
> 
> Not sure about the impact of this, maybe it should get its own issue? @MikeNeilson

Sorry, I've re-read this several times and I'm still having difficulty following the issue; I mean it seems like an issue but it's not mapping in my head. Any way to make a simple diagram, flow chart, just plain example data where the issue would be more obvious to myself and others.?

adamkorynta:
If I'm following correctly: lets say I pass in today's date, it'll give me back the loc level value from the last effective date (let's say last year), but the DTO returns today's date as the effective date.

MikeNeilson:
Ah. I see, yeah the correct effective date should be what's in the DTO.

zack-rma:
Rebased to fix build errors.

zack-rma:
Rebased to fix build issue

zack-rma:
Build errors appear to be unrelated to these changes

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -330,6 +331,11 @@ public LocationLevel retrieveLocationLevel(String locationLevelName, String pUni
         return connectionResult(dsl, c -> {
             String units = pUnits;
             Configuration configuration = getDslContext(c, officeId).configuration();
+            if (units != null && (units.equalsIgnoreCase("SI")
+                    || units.equalsIgnoreCase("EN"))) {
+                units = CWMS_UTIL_PACKAGE.call_GET_DEFAULT_UNITS(configuration,
+                        locationLevelName.split("\\.")[1], units);
```
adamkorynta: I don't see validation on the location level id prior to this, the incoming string might not have any periods and would then throw and ArrayIndexOutOfBoundsException

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
zack-rma: Added a format check before calling methods using the split string

adamkorynta: you're splitting the string twice (and not using a compiled Pattern) - should reuse the split for efficiency

zack-rma: Added a pattern and cleaned up usage

adamkorynta (approved):

PR #909, opened 2024-10-10
Issue 693 - Rating curve retrieval

### LINKED ISSUE #693: Add most recent or currently effective to cwms-data/rating/{rating-id}
add a parameter or a way to grab the most recent or currently active rating curve from the cwms-data/rating/{rating-id} endpoint

currently to get the most recent I can run this end point using method=Lazy to get all of the effective and creation dates.  then run it again with the effective date in the Begin field to rating curve with the most current effective date.  Would be nice to have this done internally or in a more efficient manor.  

This would be based on most recent according to effective date.  As in if someone was creating a rating for a time of now what rating curve would be used.

zack-rma:
Fixes #693 - Adds support for retrieving a rating closest to a specified date.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
@@ -25,32 +25,7 @@
 package cwms.cda.api;
 
 import static com.codahale.metrics.MetricRegistry.name;
-import static cwms.cda.api.Controllers.AT;
-import static cwms.cda.api.Controllers.BEGIN;
-import static cwms.cda.api.Controllers.CREATE;
-import static cwms.cda.api.Controllers.DATE_FORMAT;
-import static cwms.cda.api.Controllers.DATUM;
-import static cwms.cda.api.Controllers.DELETE;
-import static cwms.cda.api.Controllers.END;
-import static cwms.cda.api.Controllers.EXAMPLE_DATE;
-import static cwms.cda.api.Controllers.FORMAT;
-import static cwms.cda.api.Controllers.GET_ALL;
-import static cwms.cda.api.Controllers.GET_ONE;
-import static cwms.cda.api.Controllers.METHOD;
-import static cwms.cda.api.Controllers.NAME;
-import static cwms.cda.api.Controllers.OFFICE;
-import static cwms.cda.api.Controllers.RATING_ID;
-import static cwms.cda.api.Controllers.RESULTS;
-import static cwms.cda.api.Controllers.SIZE;
-import static cwms.cda.api.Controllers.STATUS_200;
-import static cwms.cda.api.Controllers.STATUS_404;
-import static cwms.cda.api.Controllers.STATUS_501;
-import static cwms.cda.api.Controllers.STORE_TEMPLATE;
-import static cwms.cda.api.Controllers.TIMEZONE;
-import static cwms.cda.api.Controllers.UNIT;
-import static cwms.cda.api.Controllers.UPDATE;
-import static cwms.cda.api.Controllers.VERSION_DATE;
-import static cwms.cda.api.Controllers.addDeprecatedContentTypeWarning;
+import static cwms.cda.api.Controllers.*;
 import static cwms.cda.data.dao.JooqDao.getDslContext;
 
 import com.codahale.metrics.Histogram;
@@ -95,6 +70,7 @@ import org.jooq.DSLContext;
 public class RatingController implements CrudHandler {
     private static final Logger logger = Logger.getLogger(RatingController.class.getName());
     private static final String TAG = "Ratings";
+    private static final String ENABLE = "enable";
 
     private final MetricRegistry metrics;
 
@@ -339,6 +315,14 @@ public class RatingController implements CrudHandler {
                         + "otherwise specified), as well as the time zone of any times in the"
                         + " response. If this field is not specified, the default time zone "
                         + "of UTC shall be used."),
+                @OpenApiParam(name = DATE, description = "Specifies the "
+                        + "date to find the closest match to for retrieving a specific rating curve. Uses a time window"
+                        + " of 24 hours before and after the specified date. The format for this field is ISO 8601 extended,"
+                        + " The enable flag must be set for this parameter to be used. If the date is not specified,"
+                        + " the rating curve closest the current time will be used."),
+                    @OpenApiParam(name = ENABLE, description = "Specifies whether the date parameter is enabled. " +
+                        "If this parameter is not set, the date parameter will not be used, even if a value is provided."
+                            + " Default is `false`."),
                 @OpenApiParam(name = METHOD, description = "Specifies "
                         + "the retrieval method used.  If no method is provided EAGER will be used.",
                         type = RatingSet.DatabaseLoadMethod.class),
@@ -356,6 +340,14 @@ public class RatingController implements CrudHandler {
         try (final Timer.Context ignored = markAndTime(GET_ONE)) {
             String officeId = ctx.queryParam(OFFICE);
             String timezone = ctx.queryParamAsClass(TIMEZONE, String.class).getOrDefault("UTC");
+            boolean enable = ctx.queryParamAsClass(ENABLE, Boolean.class).getOrDefault(false);
+            Instant date = null;
+            String specificDate = ctx.queryParam(DATE);
+            if (specificDate != null && enable) {
+                date = DateUtils.parseUserDate(specificDate, timezone).toInstant();
+            } else if (enable) {
+                date = Instant.now();
+            }
 
             Instant beginInstant = null;
             String begin = ctx.queryParam(BEGIN);
@@ -373,7 +365,7 @@ public class RatingController implements CrudHandler {
                     RatingSet.DatabaseLoadMethod.class)
                     .getOrDefault(RatingSet.DatabaseLoadMethod.EAGER);
 
-            String body = getRatingSetString(ctx, method, officeId, rating, beginInstant, endInstant);
+            String body = getRatingSetString(ctx, method, officeId, rating, beginInstant, endInstant, date);
             if (body != null) {
                 ctx.result(body);
                 ctx.status(HttpCode.OK);
@@ -385,7 +377,7 @@ public class RatingController implements CrudHandler {
     @Nullable
     private String getRatingSetString(Context ctx, RatingSet.DatabaseLoadMethod method,
                                       String officeId, String rating, Instant begin,
-                                      Instant end) {
+                                      Instant end, Instant date) {
         String retval = null;
 
         try (final Timer.Context ignored = markAndTime("getRatingSetString")) {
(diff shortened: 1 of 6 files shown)


cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
```diff
@@ -354,8 +357,13 @@ public void getAll(@NotNull Context ctx) {
     public void getOne(@NotNull Context ctx, @NotNull String rating) {
 
         try (final Timer.Context ignored = markAndTime(GET_ONE)) {
-            String officeId = ctx.queryParam(OFFICE);
+
             String timezone = ctx.queryParamAsClass(TIMEZONE, String.class).getOrDefault("UTC");
+            Instant date = null;
+            String specificDate = ctx.queryParam(EFFECTIVE_DATE);
```
MikeNeilson: use `effectiveDate` throughout. it's confusing having this go from EFFECTIVE_DATE, to specificDate, to just "date" in the later code.

MikeNeilson (changes requested):
Looks good. nitpick about variable/parameter names.

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
zack-rma: Updated to use effectiveDate and effectiveDateParam for consistency

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSetDao.java
```diff
@@ -110,9 +113,43 @@ public RatingSet retrieve(RatingSet.DatabaseLoadMethod method, String officeId,
...
-                    RatingJdbcFactory.ratingSet(finalMethod, new RatingConnectionProvider(c), officeId,
-                        specificationId, start, end, false));
+            if (effectiveDateParam != null) {
+
+                String ratingId = null;
+                Long difference = null;
+                Instant startInstant = null;
+                Instant endInstant = null;
+                Result<?> results = connectionResult(dsl, c ->
+                    dsl.select(view.RATING_ID, view.EFFECTIVE_DATE, view.RATING_CODE).from(view)
```
rma-psmorris: It should be possible to have this query do the limit rather than pulling all records and do tinghe limit in the CDA code

zack-rma: Rewrote the query and removed the difference logic from the DAO

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
```diff
@@ -339,6 +338,10 @@ public void getAll(@NotNull Context ctx) {
                         + "otherwise specified), as well as the time zone of any times in the"
                         + " response. If this field is not specified, the default time zone "
                         + "of UTC shall be used."),
+                @OpenApiParam(name = EFFECTIVE_DATE, description = "Specifies the "
```
adamkorynta: This is a very confusing parameter. 
- three optional date parameters are confusing
- what happens when all three are input?
- the ask was for the current effective date, that means the client needs to always set this parameter to now. Why not just have another endpoint that does it automatically? /{rating-id}/latest-effective
- the "closest" effective date shouldn't return a newer effective date than what is supplied as the newer effective date is not yet in effect, though this is made more complicated by temporal interpolations and transition dates. 
- how do active ratings play an effect? I presume they would be skipped, but that's missing from the docs (even from the start/end parameters)

Additionally, you are passing this confusion into the DAO by expanding the existing DAO method instead of creating a new one, so now you have three dates in the DAO function. This makes test cases far more complicated.

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
adamkorynta: While consistency is good, I think this illuminates that the parameter is confusing on this endpoint. I recommend using a separate endpoint. I might be missing a use case where a the closest effective date to a parameter is needed, but given temporal interpolation and transitional dates, that seems to be too complicated.

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
zack-rma: Moved functionality to its own endpoint to reduce confusion. New endpoint returns closest effective date rating where the date is not beyond the current date.

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
```diff
@@ -437,13 +436,18 @@ private String getRatingSetString(Context ctx, RatingSet.DatabaseLoadMethod meth
 
     private RatingSet getRatingSet(Context ctx, RatingSet.DatabaseLoadMethod method,
                                    String officeId, String rating, Instant begin,
-                                   Instant end) throws IOException, RatingException {
-        RatingSet ratingSet;
+                                   Instant end, boolean latest) throws IOException, RatingException {
```
adamkorynta: Don't add a boolean to a method to change the implementation, just create a new method

zack-rma: Created a new method for latest rating retrieval

zack-rma:
Build failures appear to be unrelated to this PR

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
```diff
@@ -369,23 +367,24 @@ public void getOne(@NotNull Context ctx, @NotNull String rating) {
...
             }
 
+            String officeId = ctx.queryParam(OFFICE);
+
             RatingSet.DatabaseLoadMethod method = ctx.queryParamAsClass(METHOD,
                     RatingSet.DatabaseLoadMethod.class)
                     .getOrDefault(RatingSet.DatabaseLoadMethod.EAGER);
 
-            String body = getRatingSetString(ctx, method, officeId, rating, beginInstant, endInstant);
+            String body = getRatingSetString(ctx, method, officeId, rating, beginInstant, endInstant, false);
```
adamkorynta: The changes to this controller should be gone right? the latest option is moved to `RatingLatestController`

cwms-data-api/src/main/java/cwms/cda/api/RatingLatestController.java
```diff
@@ -0,0 +1,133 @@
...
+import io.javalin.plugin.openapi.annotations.OpenApi;
+import io.javalin.plugin.openapi.annotations.OpenApiContent;
+import io.javalin.plugin.openapi.annotations.OpenApiParam;
+import io.javalin.plugin.openapi.annotations.OpenApiResponse;
+import java.time.Instant;
+import org.jetbrains.annotations.NotNull;
+import org.jooq.DSLContext;
+
+
+public class RatingLatestController extends RatingController implements Handler {
```
adamkorynta: The inheritance doesn't make sense here, this controller is calling a different DAO method

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSetDao.java
```diff
@@ -83,11 +101,96 @@ private static String extractOfficeId(String ratingSet) throws JsonProcessingExc
         return office;
     }
 
+    public RatingSet retrieveLatest(RatingSet.DatabaseLoadMethod method, String officeId,
```
adamkorynta: Isn't this what `cwms_rating.retrieve_eff_ratings_xml` does? Can you verify with Mike Perryman?

adamkorynta (changes requested):

MikeNeilson:
> Build failures appear to be unrelated to this PR

Yep, it's the usual failures from other end points.

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
zack-rma: Correct, I moved any remnants to the RatingLatestController and cleaned up the original controller

cwms-data-api/src/main/java/cwms/cda/api/RatingLatestController.java
zack-rma: Removed unnecessary inheritance

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSetDao.java
zack-rma: I'll double check with Perryman

perrymanmd: ### DISREGARD THE PREVIOUS COMMENT

### CORRECT COMMENT

Yes you can use `cwms_rating.retrieve_eff_ratings_xml_f()` to get the most recent rating by setting both `p_start_date `and `p_end_date` to `sysdate` (or whatever time you want the most recent rating for).

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
adamkorynta, line 1: I don't think there are any meaningful edits to this file, can you revert?

cwms-data-api/src/main/java/cwms/cda/api/RatingLatestController.java
```diff
@@ -0,0 +1,133 @@
...
+    @OpenApi(
+        pathParams = {
+            @OpenApiParam(name = RATING_ID, required = true, description = "The rating-id of the effective "
+                    + "dates to be retrieve. "),
+        },
+        queryParams = {
+            @OpenApiParam(name = OFFICE, required = true, description =
+                "Specifies the owning office of the ratingset to be included in the "
+                + "response."),
+            @OpenApiParam(name = TIMEZONE, description = "Specifies "
```
adamkorynta: There are no begin/end fields

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/RatingController.java
zack-rma: Reverted

cwms-data-api/src/main/java/cwms/cda/api/RatingLatestController.java
zack-rma: Removed timezone

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSetDao.java
```diff
@@ -83,11 +89,45 @@ private static String extractOfficeId(String ratingSet) throws JsonProcessingExc
...
+    public String retrieveLatestXML(String officeId, String specificationId) {
+        return connectionResult(dsl, c -> {
+            DSLContext context = getDslContext(c, officeId);
+            return CWMS_RATING_PACKAGE.call_RETRIEVE_EFF_RATINGS_XML_F(context.configuration(), specificationId,
+                    Timestamp.from(Instant.now()), Timestamp.from(Instant.now()), null, officeId);
+        });
+    }
+
+    @Override
+    public String retrieveLatestJSON(RatingSet.DatabaseLoadMethod method, String officeId,
```
adamkorynta: What is the purpose for this method? The `JsonRatingUtils` should be able to convert the XML version to JSON. Doing it this way is retrieving the same rating twice.

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSetDao.java
zack-rma: Removed unnecessary method, added `JsonRatingUtils` conversion

adamkorynta (approved):

MikeNeilson (approved):

PR #913, opened 2024-10-11
Bugfix/910 missing data

MikeNeilson:
Moved to JOOQ call. Applied similar logic as what the cwms-data-access library did.

Updated description on the Timeseries.Record class to provide clarity.

Expanded existing test to include missing data of both supported types.


rma-rripken (approved):
lgtm

msweier:
Thanks Mike. What are the ways to write null with this update? Looks like -340282346638528859811704183484516925440 will be one way. Is there a shorthand way like 'nan' or 'null' that would work too? If not maybe worth adding something in cwms-python so we don't have to remember that long number. I don't know of a good way to get to -Float.MAX_VALUE in python easily.

adamkorynta (approved):

MikeNeilson:
None in python is the equivalent to null and the python json library *should* correctly set it. If not, open an issue on the cwms-python library as it needs to be dealt with there now.

msweier:
Tested behavior with new release. Missing value of -340282346638528859811704183484516925440  and missing quality code of 5 are still both required.

MikeNeilson:
null or none did not work?

MikeNeilson:
also was this directly or with cwms-js or cwms-python?

MikeNeilson:
if cwms-python likely related to https://github.com/HydrologicEngineeringCenter/cwms-python/issues/105

msweier:
> also was this directly or with cwms-js or cwms-python?

cwms-python, let me test null or none.

msweier:
Ok so using None without quotes and "null" as a string works only if you assign the quality-code too. The following all work.
`{'name': 'LockDam_04-Tailwater.Stage.Inst.15Minutes.0.rev',
 'office-id': 'MVP',
 'units': 'ft',
 'values': [['2015-12-11T07:30:00+00:00', None, 5]],
 'version-date': None}`

or 
`{'name': 'LockDam_04-Tailwater.Stage.Inst.15Minutes.0.rev',
 'office-id': 'MVP',
 'units': 'ft',
 'values': [['2015-12-11T07:30:00+00:00', 'null', 5]],
 'version-date': None}`
or
`{'name': 'LockDam_04-Tailwater.Stage.Inst.15Minutes.0.rev',
 'office-id': 'MVP',
 'units': 'ft',
 'values': [['2015-12-11T07:30:00+00:00', -3.4028234663852886e+38, 5]],
 'version-date': None}`

PR #927, opened 2024-10-22
Fixes #634 - Implemented data entry date option for TS data retrieval

### LINKED ISSUE #634: Need Data Entry Date for Time Series Data
I have a report that extracts forecast time series for the next three days along with the forecast date (data_entry_date) 

I need the "data_entry_date" for all values when using the TimeSeries API. 

SQL Example
```sql
select cwms_util.change_timezone(date_time, 'UTC', 'CST6CDT') as date_time
    ,value
    ,cwms_ts_id
    ,cwms_util.split_text(cwms_ts_id, 1, '.') as location_id
    ,unit_id
    ,to_char(cwms_util.change_timezone(data_entry_date, 'UTC', 'CST6CDT'), 'mm/dd HH24:MI') as data_entry_date
from CWMS_20.AV_TSV_DQU
where cwms_ts_id = 'St Charles-Missouri.Stage.Inst.6Hours.0.RVFShef-FF'
    and unit_id = 'ft'
    and date_time = to_date('05-07-2024 ' || '12:00' ,'mm-dd-yyyy hh24:mi');
```  

right now,  CDA returns this now.  
**Current Data:**
```json
{
"values": [
    [
      1715018400000,
      25.3,
      0
    ],
    [
      1715040000000,
      25.199999999999996,
      0
    ],
    [
      1715061600000,
      25.1,
      0
    ]
  ]
}
```
I would like it to include the data entry date along side. Although I am not sure if this is the best way to go about it, but just as an example:  

**New Data:**
Proposed payload:
```json
{
"values": [
    [
      1715018400000,
      25.3,
      0,
     **data_entry_date**
    ],
    [
      1715040000000,
      25.199999999999996,
      0,
     **data_entry_date**
    ],
    [
      1715061600000,
      25.1,
      0,
     **data_entry_date**
    ]
  ]
}

### PR DESCRIPTION
**Pull request by zack-rma**
Fixes #634 - Implements data entry date as option for TimeSeries data retrieval - Serialization bug in progress

### CODE REVIEW

File: cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java [hunk 1]
```diff
@@ -84,7 +83,7 @@
 
 public class TimeSeriesController implements CrudHandler {
     private static final Logger logger = Logger.getLogger(TimeSeriesController.class.getName());
-
+    private static final String INCLUDE_ENTRY_DATE = "include-entry-date";
```
MikeNeilson: From a purely pedantic standpoint this should really be a parameter of the content-type, but I may have to accept the reality of this being easier for everyone. 

@krowvin, we were just discussing this conceptually.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -248,6 +256,11 @@ public static class Record {
         @JsonProperty(value = "quality-code", index = 2)
         int qualityCode;
```
MikeNeilson, line 300: This might be better as a sub class. I know that adds some complex but eventually we may also want to include the version date and text notes that may be attached and that would be a lot of logic in this one class.

zack-rma: Changed to a subclass. I added a custom deserializer to handle the different classes.

MikeNeilson:
@jbkolze what are your thoughts on this? At least how it's done. Some appears it may be a breaking change which we want to avoid.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -623,7 +622,7 @@ private static TimeSeries buildTimeSeries(ILocationLevelRef levelRef, Interval i
             if (qualityCode != null) {
                 quality = qualityCode.intValue();
             }
-            timeSeries.addValue(dateTime, value, quality);
+            timeSeries.addValue(dateTime, value, quality, null);
```
MikeNeilson: Are we sure this isn't a breaking change?

jbkolze:
> @jbkolze what are your thoughts on this? At least how it's done. Some appears it may be a breaking change which we want to avoid.

Conceptually, I don't personally have any qualms with it.  You had mentioned in the typing discussion that you all were trying to leave flexibility for dynamically adjusting the time series values array, and this seems like a prime use case for that.

That being said, I am a little concerned about the part you marked as a breaking change.  I don't know the CDA source that well, but is this indicating that the response would include a fourth null value even if the include_entry_date is false?  Because that definitely would not be ideal -- we've written a lot of CDA code already (and I think other districts have as well) that would have to be updated.  Not as big of a deal if an API version were included in the path (`.../v3/...`), but somewhat cumbersome in the current setup.

My understanding from previous conversations is that you'd get the normal 3-value array if this were set to false, but receive a 4-value array if include_entry_date is true.  And the "value-columns" object would be updated to match.  That'd be my "ideal" implementation.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
zack-rma: I'll double check with additional test cases, but this change should not break any existing functionality. A null data entry date parameter is treated as if a standard Time-Value-Quality data entry was provided. The implementation of `addValue` will use a TimeSeries data record with only three input fields under normal circumstances: 

` if (dataEntryDate != null) {
                values.add(new TimeSeriesRecordWithDate(dateTime, value, qualityCode, dataEntryDate));
            } else {
                values.add(new Record(dateTime, value, qualityCode));
            }`

The existing use cases of TimeSeries should be unaffected, as they will be handled exactly as they were before these changes.

zack-rma:
You are correct that setting the include_entry_date parameter to true would result in a four-value array, whereas setting it to false would return a three-value array. Your "ideal" implementation is what I was aiming for to retain backwards compatibility and avoid breaking any of the other endpoints that rely on the TimeSeries implementation.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
adamkorynta: Path parameters are the "what" I am retrieving, query parameters are the "how" am I adding to or filtering that data, and content-type is the "shape" that is returned. This change effects the how and given the way the column names and data array are paired together to already give this flexibility, I don't see a change in shape.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -391,14 +393,55 @@ public TimeSeries getTimeseries(String page, int pageSize, String names, String
...
+            SelectConditionStep<Record3<Timestamp, Double, BigDecimal>> finalQuery = query;
+            logger.fine(() -> finalQuery.getSQL(ParamType.INLINED));
 
-            query.forEach(tsRecord -> timeseries.addValue(
-                            tsRecord.getValue(dateTimeCol),
-                            tsRecord.getValue(valueCol),
-                            tsRecord.getValue(qualityNormCol).intValue()
-                    )
-            );
+            if (includeEntryDate) {
```
adamkorynta, line 416: This query is doubling the time it takes to retrieve time series. Can this replace the `retrieve_ts_out_tab` calls?

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -159,7 +164,8 @@ public ZonedDateTime getEnd() {
     }
 
     // Use the array shape to optimize data transfer to client
-    @JsonFormat(shape=JsonFormat.Shape.ARRAY)
+    @JsonFormat(shape = JsonFormat.Shape.ARRAY)
+    @JsonDeserialize(contentUsing = TimeSeriesRecordDeserializer.class)
```
adamkorynta: This method is overwritten for XML using a Mixin, did you verify that behavior works as intended still?

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -186,30 +191,35 @@ public ZonedDateTime getVersionDate() {
...
-    public VersionType getDateVersionType() { return dateVersionType; }
+    public VersionType getDateVersionType() {
+        return dateVersionType;
+    }
 
     @JsonProperty(value = "value-columns")
     @Schema(name = "value-columns", accessMode = AccessMode.READ_ONLY)
     public List<Column> getValueColumnsJSON() {
-        return getColumnDescriptor();
+        return getColumnDescriptor((values != null && !values.isEmpty())
```
adamkorynta: This seems like behavior for a subclass

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -218,7 +228,16 @@ private List<Column> getColumnDescriptor() {
                 columns.add(new TimeSeries.Column(fieldName, fieldIndex + 1, f.getType()));
             }
         }
-
+        if (includeDataEntryDate) {
```
adamkorynta: This could also be accomplished better with a subclass

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -240,6 +259,16 @@ private List<Column> getColumnDescriptor() {
                                 + "placeholder which can be important in irregular and psuedo regular timeseries."
             )
     )
+
+    // This class is used to deserialize the time-series data JSON into an object
+    // Solves the issue of the deserializer getting stuck in a loop
+    // and throwing a StackOverflowError when trying to handle the Record class directly
```
adamkorynta: This seems sketchy to me, why is your custom deserializer causing this?

cwms-data-api/src/main/java/cwms/cda/formatters/json/adapters/TimeSeriesRecordDeserializer.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+public final class TimeSeriesRecordDeserializer extends JsonDeserializer<TimeSeries.Record> {
+    @Override
+    public TimeSeries.Record deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
+        JsonNode node = jsonParser.readValueAsTree();
+        if (node.get("data-entry-date") != null) {
+            return jsonParser.getCodec().treeToValue(node, TimeSeriesRecordWithDate.class);
+        }
+        String nodeString = node.toString();
+        if (nodeString.startsWith("[")) {
```
adamkorynta: A mixin doesn't solve the need for this custom parsing? All this logic looks like we're circumventing jackson too much

cwms-data-api/src/main/java/cwms/cda/formatters/json/adapters/TimeSeriesRecordDeserializer.java
```diff
@@ -0,0 +1,77 @@
...
+            nodeString = nodeString.substring(1, nodeString.length() - 1);
+            String[] valList = nodeString.split(",");
+            if (valList.length == 4) {
+                Timestamp dateTime = Timestamp.from(Instant.ofEpochMilli(Long.parseLong(valList[0])));
+                double value = Double.parseDouble(valList[1]);
+                int quality = Integer.parseInt(valList[2]);
+                Timestamp entryDate = Timestamp.from(Instant.ofEpochMilli(Long.parseLong(valList[3])));
```
adamkorynta: no need to convert from Instant to Timestamp, Timestamp's constructor takes in epoch millis.

cwms-data-api/src/main/java/cwms/cda/formatters/json/adapters/TimeSeriesRecordDeserializer.java
```diff
@@ -0,0 +1,77 @@
...
+ * Requires {@link cwms.cda.data.dto.TimeSeries.RecordChild} class to avoid
+ * a StackOverflowError when deserializing JSON data. This issue can be caused by the custom serializer
+ * getting stuck in a loop if the Record class is used directly.
+ * Allows for use of subclass with additional fields {@link TimeSeriesRecordWithDate}.
+ */
+public final class TimeSeriesRecordDeserializer extends JsonDeserializer<TimeSeries.Record> {
+    @Override
+    public TimeSeries.Record deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
+        JsonNode node = jsonParser.readValueAsTree();
+        if (node.get("data-entry-date") != null) {
```
adamkorynta: This should be a constant

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
adamkorynta: I recommend subclassing TimeSeries instead

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
MikeNeilson: Adding a column is definitely a change in shape. The what is a time series, the query parameters specify exactly which time series, or at least which portion of a given time series (I guess if we're being *really* pedantic begin and end should be in the fragment... but I digress). 

While the flexibility *is* there, it's flexibility to change the shape. I don't totally disagree with you but given we haven't communicated that portion of the contract very well we are introducing a breaking change. We already have more than one downstream library dependent on these types.

I'm going to type up something on the wiki, or maybe discussion, for the philosophy I'm going for with these, hopefully my argument makes more sense in regards to query vs content type. Especially as it relates to some of the challenges we're currently seeing.

cwms-data-api/src/main/java/cwms/cda/formatters/json/adapters/TimeSeriesRecordDeserializer.java
MikeNeilson: should probably also be ignored, data-entry-date is *always* set by the database itself, an external system/user isn't allowed to change it.

MikeNeilson:
> 
> My understanding from previous conversations is that you'd get the normal 3-value array if this were set to false, but receive a 4-value array if include_entry_date is true. And the "value-columns" object would be updated to match. That'd be my "ideal" implementation.

That is correct for the location indicated, which is location levels backed by a time series. I don't know how much it would affect what you're currently using, but it's definitely not ideal. At the least it definitely *shouldn't* be null, may as well provide the data, but seems like a parameter should be added to match on the location level end point.

But it sounds like your on the same page with Adam about the shape being already explicitly flexible so I'm okay with that section now; not "ideal", but what is, definitely something to better document though.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
adamkorynta: Yeah, it is not conveyed very well that one should use the columns array to determine which index to grab from the data array. I don't think that adding information to the swagger docs to communicate that should trigger a content-type change though (maybe version=2.5 but that seems like a huge headache) especially since not including the parameter to retrieve entry date keeps the array intact for backwards compatibility.

Also, I've never seen an API where the content-type changes how much extra (or less) data gets returned to the client. I'd like to see some examples. I also don't see the reason (pedantic or not) for adding begin/end as path parameters as those are filters on the time series. Everything for the identifier of the time series encompasses the time series (which is also why the date version shouldn't be in the path and is a query parameter).

MikeNeilson: NOTE: not arguing for it, but explanation of my logic on the fragement.
a time series is:
`/timeseries/Alder Springs.Precip-INC.Total.1Hour.1Hour.Calc?version_date=unversioned` identifies a specific time series which is a mathematical series of data - and technically all of it. The entire series could be consider a "document", a fragment is a section within a document; traditionally we would think of lines in a file, but it applies to the time series as well. A file, after all, is just a series of lines.

As I said, extremely pedantic.... admittedly almost to the point of being useless because literally no one does it that way, nor would they even if it could be proven objectively correct.

Technically the units are also representation and not identification but given how limited the use of content-type features are used it would be incredibly difficult to get people to use it; I don't even think the Swagger-UI has a mechanism to slightly tweak the content-type.

But back on the topic of what's correct for us, it seems we're all in agreement that @DanielTOsborne 's initial design is already sufficiently flexible in the current scheme and our failure was in how we documented that for the general end user.

So we leave the inclusion as a query parameter unless a better way actually comes along.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
MikeNeilson: I think it would break CWMS.js, The javascript openapi generator already has trouble with our TimeSeries class given some specific assumptions the generator chose to make.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
zack-rma: While it could replace the `retrieve_ts_out_tab` call above, doing so would require implementing trim support into the query, as that is currently handled by the `retrieve_ts_out_tab` call. I haven't quite figured out the best way to do so, so maybe we can discuss this in more detail

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -391,59 +435,29 @@ public TimeSeries getTimeseries(String page, int pageSize, String names, String
...
-                    timeseries.addValue(
-                        tsRecord.getValue(dateTimeCol),
-                        tsRecord.getValue(valueCol),
-                        tsRecord.getValue(qualityNormCol).intValue(),
-                        tsRecord.getValue(dataEntryDate)
-                    );
-                });
+                logger.fine(() -> query2.getSQL(ParamType.INLINED));
+                final TimeSeriesWithDate timeSeries = new TimeSeriesWithDate(timeseries);
+                query2.forEach(tsRecord -> timeSeries.addValue(
```
MikeNeilson, line 419: we don't need to solve this now, but definitely before we add requesting any text notes as well. There has to be a better way to handle this with the builders.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
zack-rma: I added a Mixin test that verifies that the XML tags for the value records have the appropriate labels. There are also a couple serialization/deserialization tests that verify that this works as intended.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
zack-rma: Added to TimeSeries subclass

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
zack-rma: Moved into subclass

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
zack-rma: Removed custom deserializer

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -351,27 +350,72 @@ public TimeSeries getTimeseries(String page, int pageSize, String names, String
...
+                        .lessOrEqual(CWMS_UTIL_PACKAGE.call_TO_TIMESTAMP__2(
+                                DSL.val(endTime.toInstant().toEpochMilli())))
+                        .and(AV_TSV_DQU.AV_TSV_DQU.CWMS_TS_ID.equalIgnoreCase(names))
+                        .and(AV_TSV_DQU.AV_TSV_DQU.OFFICE_ID.eq(office))
+                        .and(AV_TSV_DQU.AV_TSV_DQU.UNIT_ID.equalIgnoreCase(unit)));
 
+        TimeSeries retVal = null;
         if (pageSize != 0) {
+            if (versionDate != null) {
+                whereCond = whereCond.and(AV_TSV_DQU.AV_TSV_DQU.VERSION_DATE.eq(versionDate == null ? null :
```
MikeNeilson: Does this logic handle max version? or will AV_TSV_DQU always return every version? or only the specifically requested version?

cwms-data-api/src/main/java/cwms/cda/formatters/json/adapters/TimeSeriesRecordDeserializer.java
zack-rma: Removed custom serializer

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
MikeNeilson: Okay, so I said this, Adam said this, but after reading the code and our other discussions above timeseries does it make more sense to just make TimeSeries more generic?

e.g.

A builder where you manually add column names, index, and type, and and functions in the row builder to set such?
something like

```
withColumn(int index, String name, String description, Class<T> type) {
"logic"
}

... Record:

<T> setColumn(int index, T value, Class<T> type) {
  "logic"
}
```
Or something like that, it would prevent the need in TimeSeriesDaoImpl have two different loops doing almost 90% the same work. just a check for "I have this column requested, let's also add it."

Basically instead of hard coding the columns at all (okay, maybe time... is a time series) , the user of the given TimeSeries object (after set by builder) can define them at run-time.


Sorry, I know you did a lot here, that just came to me now looking through the current PR.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
jbkolze: > I think it would break CWMS.js, The javascript openapi generator already has trouble with our TimeSeries class given some specific assumptions the generator chose to make.

Just chiming back in here to add some context... We're already modifying the OpenAPI spec in the cwmsjs build process to adjust a few things, so we can just handle this as well if need be.  I don't think it will cause an issue (not a new one, at least), though, because at this point cwmsjs just anticipates an array of numbers in the response and leaves it up to the user to determine what is what based on either the request or the response metadata.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
zack-rma: Updated logic to handle max version in line with previous implementation. If the version date is not specified, the max version will be set to true and the most recent version date will be used for retrieval. This does prevent retrieval of multiple versions in one call, so that could be a problem if that is a desired feature.

zack-rma:
Build errors appear to be unrelated to these changes

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -382,6 +381,8 @@ public void delete(@NotNull Context ctx, @NotNull String timeseries) {
                         + "\n* `xml`"
                         + "\n* `wml2` (only if name field is specified)"
                         + "\n* `json` (default)"),
+                @OpenApiParam(name = INCLUDE_ENTRY_DATE, type = Boolean.class, description = "Specifies "
+                    + "whether to include the data entry date in the response.  Default is false."),
```
adamkorynta: this should very clearly specify that it increases the data array size to include the date

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
adamkorynta, line 1: The `value-columns` and `records` properties needs much more description. It needs need to describe how they relate to each other and how the records array can be variable length, so using the column index is key.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
+public final class TimeSeriesWithDate extends TimeSeries {
```
adamkorynta: Should be `TimeSeriesWithEntryDate` or `TimeSeriesWithDataEntryDate`. Time series implies there are dates already

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private List<TimeSeriesWithDate.Record> values;
+
+	// list of TimeSeriesWithDate.Record, uses raw to avoid typing errors
+    @Override
+    public List getValues() {
```
adamkorynta: typing was left off

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Objects;
+
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
+public final class TimeSeriesWithDate extends TimeSeries {
+
+    private List<TimeSeriesWithDate.Record> values;
```
adamkorynta: should be able to reuse the parent's values since there is inheritance

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
```diff
@@ -0,0 +1,177 @@
...
+    public List getValues() {
+        return values;
+    }
+
+    TimeSeriesWithDate() {
+        super();
+        values = new ArrayList<>();
+    }
+
+    public TimeSeriesWithDate(TimeSeries timeSeries) {
```
adamkorynta: Why so many constructors? Isn't there only one endpoint returning only one time series? Also constructors should go at the top of the class

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        return columns;
+    }
+
+    public static final class Record extends TimeSeries.Record {
+        @JsonProperty(value = "data-entry-date", index = 3)
+        @Schema(implementation = Long.class, description = "Milliseconds since 1970-01-01 (Unix Epoch), always UTC")
+        @JsonInclude(JsonInclude.Include.NON_DEFAULT)
+        Timestamp dataEntryDate;
```
adamkorynta: private

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
adamkorynta, line 1: Post/Patch should throw an error if the client is sending a time series with a data-entry-date as that field isn't editable

cwms-data-api/src/test/resources/cwms/cda/api/lrl/1day_offset_version_date.json
adamkorynta, line 1: Where's the example JSON with the data entry date included? Serialization unit tests should test against the file.

cwms-data-api/src/test/java/cwms/cda/data/dao/TimeSeriesDaoTest.java
```diff
@@ -171,7 +169,7 @@ public void testCreateWithData() throws Exception
 			int count = 60 / 15; // do I need a +1?  ie should this be 12 or 13?
 			// Also, should end be the last point or the next interval?
 
-			TimeSeries ts = new TimeSeries(null, -1, 0, tsId, officeId, start, end, "m", Duration.ofMinutes(minutes));
+			TimeSeriesWithDate ts = new TimeSeriesWithDate(null, -1, 0, tsId, officeId, start, end, "m", Duration.ofMinutes(minutes));
```
adamkorynta: Shouldn't be able to store time series with a data-entry-date, no need for this test to change.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -242,6 +349,91 @@ void test_lrl_1day_malicious_units() throws Exception {
...
+            .body("values.size()", equalTo(4))
+            .body("values[0][1]", equalTo(4.0F))
+            .body("values[0].size()", equalTo(4));
+
+        // get it back without the data entry date
+        given()
+            .log().ifValidationFails(LogDetail.ALL, true)
+            .accept(Formats.JSONV2)
+            .header("Authorization", user.toHeaderValue())
+            .queryParam(Controllers.OFFICE, officeId)
```
adamkorynta, line 535: where is the query param to include the data entry date?

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
zack-rma: Added a check for data entry date values

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
zack-rma: Added better descriptions, hopefully that clears up the usage

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
zack-rma: I added this request to check that we can get the same data back without the data entry date, the request above gets it with the entry date. I moved the query parameter variable to the Controllers class for better accessibility

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
zack-rma: Added comment reflecting the change in response format

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
zack-rma: Renamed for clarity

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
zack-rma: Added typing

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
zack-rma: Needed to add the local variables to get the typing correct. Removing it causes loss of type within the List, which drops the data entry date value.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
zack-rma: Removed unnecessary constructors, swapped usages where needed

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
zack-rma: Made private

cwms-data-api/src/test/resources/cwms/cda/api/lrl/1day_offset_version_date.json
zack-rma: Added example JSON files and a test for serialization against them

cwms-data-api/src/test/java/cwms/cda/data/dao/TimeSeriesDaoTest.java
zack-rma: Reverted test, test is disabled due to deprecated connection usage

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -200,7 +163,8 @@ public void create(@NotNull Context ctx) {
             DSLContext dsl = getDslContext(ctx);
 
             TimeSeriesDao dao = getTimeSeriesDao(dsl);
-            TimeSeries timeSeries = deserializeTimeSeries(ctx);
+            TimeSeriesWithDataEntryDate timeSeries = deserializeTimeSeries(ctx);
```
adamkorynta: This should fail deserialization if the time series doesn't include the data entry date since you are typing to the subclass.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -79,7 +79,8 @@ public class TimeSeries extends CwmsDTOPaginated {
 
     @ArraySchema(
             schema = @Schema(
-                    description = "List of retrieved time-series values",
+                    description = "List of retrieved time-series values. Contains [dateTime, value, qualityCode]. "
```
adamkorynta: The purpose for the description is to allow the array to be unspecified, the description should not identify what the columns are, rather the value-columns property is what tells the client what the columns are. ex, if we add a boolean "include-quality" in the future, we want to allow the array to be of size 2.

cwms-data-api/src/test/resources/cwms/cda/api/lrl/1day_offset_version_date_roundtrip.json
adamkorynta, line 1: this file name should specify that it includes data entry dates

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
adamkorynta: if the schema needs to provide more, we can update the schema call

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
zack-rma: I thought so too, but it deserializes with null values for the data entry dates. This seems to be the most reliable way to check for those values, since using the body content as a String or InputStream to check for them would consume the context and close the data stream. I found that the default behavior when passing in a TS with data entry values was to drop/ignore the values and treat it as a standard TS, so typing to the subclass doesn't appear to cause issues since it is cast to the base class anyway for storage. 

I'm open to suggestions for a better way to approach this.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
zack-rma: I removed the column data description and added a note that the value-columns describe the value array data structure.

cwms-data-api/src/test/resources/cwms/cda/api/lrl/1day_offset_version_date_roundtrip.json
zack-rma: Renamed to better identify the data and it's usage

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
zack-rma: After more digging and some testing, I've managed to find a solution that does not close the data stream. I've removed the deserialization to the child class

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
zack-rma: I've created a PR with a new schema call that returns timeseries data with the data entry dates: https://bitbucket.hecdev.net/projects/CWMS/repos/cwms_database/pull-requests/412/overview

zack-rma:
Build will fail until database schema changes are merged and a new release is available. Changes can be found here: https://bitbucket.hecdev.net/projects/CWMS/repos/cwms_database/pull-requests/412/overview

MikeNeilson:
Tests are still failing, actions output is still showing schema of "latest-dev"

inguyen314:
what is status?

zack-rma:
> what is status?

The necessary changes to the database were just merged in, we'll need an updated schema to be available before this feature can be finalized.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
adamkorynta: then you shouldn't inherit from TimeSeries as the class structure then has two sets of data

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -352,25 +346,69 @@ public TimeSeries getTimeseries(String page, int pageSize, String names, String
...
+                                tsMetadata.getValue("interval", Long.class)),
+                        verticalDatumInfo,
+                        tsMetadata.getValue(AV_CWMS_TS_ID2.INTERVAL_UTC_OFFSET).longValue(),
+                        tsMetadata.getValue(tzName),
+                        versionDate, finalDateVersionType
+                );
+            }
         });
 
+        if (includeEntryDate) {
```
MikeNeilson: why are we doing this check if the above would already have returned a TimeSeriesWithDataEntryDate?

Alternatively why have the login within the lambda for data_entry_date if we're goign to change it here?

(I'm not seeing a difference other than the type after new.)

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
MikeNeilson: Given the previous logic above, does this need a new or just a cast so you can get the desired methods?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -1127,15 +1184,15 @@ private void store(Connection connection, String officeId, String tsId, String u
...
-            Iterator<TimeSeries.Record> iter = values.iterator();
-            while (iter.hasNext()) {
-                TimeSeries.Record value = iter.next();
-                Double dataValue = value.getValue();
-                if (dataValue != null && dataValue == -Float.MAX_VALUE) {
-                    dataValue = null;
-                }
-                tsvArray.add(new ZTSV_TYPE(value.getDateTime(), dataValue, BigDecimal.valueOf(value.getQualityCode())));
-            }
+			for(TimeSeries.Record value : values)
```
MikeNeilson: incorrect curlies, but otherwise thank you for cleaning up the loop structures and formatting.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
MikeNeilson: I don't want Zach to do it in this PR, already been too long, but does anyone think the above idea is crazy?

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
```diff
@@ -253,9 +265,11 @@ public static class Record {
         int qualityCode;
 
         @SuppressWarnings("unused") // required so JAXB can initialize and marshal
-        private Record() {}
+        private Record() {
+        }
 
         public Record(Timestamp dateTime, Double value, int qualityCode) {
+            super();
```
MikeNeilson: ... what super?

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesRecordWithDate.java
```diff
@@ -0,0 +1,81 @@
...
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.sql.Timestamp;
+import java.util.Objects;
+
+/**
+ * TimeSeriesRecordWithDate is a subclass of TimeSeries.Record that includes a data entry date.
+ * The data entry date is the date that the data was entered into the database.
+ */
+@JsonDeserialize(using = JsonDeserializer.None.class)
+public final class TimeSeriesRecordWithDate extends TimeSeries.Record {
```
MikeNeilson: This confused me. Should be `TimeSeriesRecordWithEntryDate`

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -75,10 +78,9 @@ void test_lrl_timeseries_psuedo_reg1hour() throws Exception {
                 .config(RestAssured.config().jsonConfig(jsonConfig().numberReturnType(JsonPathConfig.NumberReturnType.DOUBLE)))
                 .log().ifValidationFails(LogDetail.ALL,true)
                 .accept(Formats.JSONV2)
-//                .body(tsData)
                 .header("Authorization",user.toHeaderValue())
                 .queryParam("office",officeId)
-                .queryParam("units","cfs")
+                .queryParam("unit","cfs")
```
MikeNeilson: It would seem we should really start using the constants we've defined. (Not a ding, pretty sure we're all guilty of that.)

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
```diff
@@ -224,24 +444,148 @@ void test_lrl_1day_malicious_units() throws Exception {
...
+        //     1675335600000 is Thursday, February 2, 2023 11:00:00 AM
+        // fyi 1675422000000 is Friday, February 3, 2023 11:00:00 AM
+
+        // get it back with the data entry date
+        ExtractableResponse<Response> response = given()
+            .log().ifValidationFails(LogDetail.ALL, true)
+            .accept(Formats.JSONV2)
+            .header("Authorization", user.toHeaderValue())
+            .queryParam(Controllers.OFFICE, officeId)
+            .queryParam(Controllers.UNIT, "CFS")
```
MikeNeilson, line 509: Yes, this is definitely the superior way to set those parameters.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
MikeNeilson: Might be good to set the parameter explicitly to false. There's always the possibility we change the default behavior.

cwms-data-api/src/test/java/cwms/cda/data/dao/TimeSeriesDaoTest.java
MikeNeilson, line 1: thank you for updating the formatting within this file while you were at it.

MikeNeilson (changes requested):
I have some minor nitpicks on formatting.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
MikeNeilson: This lends itself to the other suggestion I made with the more generic structure. Though I think that would be too much for this PR.

The Time Series class Daniel created is already generic enough (in concept at least) to handle adding additional columns.

cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java
zack-rma: Replaced with Controllers constant where applicable

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesRecordWithDate.java
zack-rma: Renamed as suggested

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeries.java
zack-rma: Leftover from when I had created a shared parent class to resolve some serialization issues. Removed

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
zack-rma: Fixed brackets

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
zack-rma: Remnant from before the above line, perhaps. I've removed it and tested that all works as intended without it.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesWithDate.java
zack-rma: Cleaned up the duplicate value list data from the subclass. Required adding custom deserializer back to properly handle the subtypes of TimeSeries.Record.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -573,17 +543,24 @@ public void update(@NotNull Context ctx, @NotNull String id) {
...
-        } catch (IOException | DataAccessException ex) {
+        } catch (DataAccessException | IOException ex) {
             CdaError re = new CdaError("Internal Error");
             logger.log(Level.SEVERE, re.toString(), ex);
             ctx.status(HttpServletResponse.SC_INTERNAL_SERVER_ERROR).json(re);
         }
     }
 
-    private TimeSeries deserializeTimeSeries(Context ctx) throws IOException {
+    private TimeSeries deserializeTimeSeries(Context ctx) throws IOException
```
adamkorynta: brackets don't follow standards

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/formatters/json/adapters/TimeSeriesRecordDeserializer.java
```diff
@@ -0,0 +1,91 @@
...
+import java.sql.Timestamp;
+
+/**
+ * A time-series record deserializer for properly deserializing JSON data.
+ * Requires {@link cwms.cda.data.dto.TimeSeries.StandardRecord} class to avoid
+ * a StackOverflowError when deserializing JSON data. This issue can be caused by the custom serializer
+ * getting stuck in a loop if the Record class is used directly.
+ * Allows for use of subclass with additional fields {@link TimeSeriesRecordWithEntryDate}.
+ */
+public final class TimeSeriesRecordDeserializer extends JsonDeserializer<TimeSeries.Record> {
```
adamkorynta: We don't deserialize the time series with data entry date since the client is no allowed to store those entry dates. We shouldn't need this class. 

Taking this logic though, it seems like we can create a very simple custom _serializer_, remove the subclassing of TimeSeries and TimeSeries.Record, move the entry date to the TimeSeries.Record, and conditionally add to the array if the data entry date is not null. There should be no case where the data entry date was requested, but then it is returned as null in the array. This is not the case if we add text in the future, but I'm sure we can figure out a reasonable solution when that feature is requested.

adamkorynta (changes requested):
I think we're almost there

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
zack-rma: Fixed brackets

cwms-data-api/src/main/java/cwms/cda/formatters/json/adapters/TimeSeriesRecordDeserializer.java
zack-rma: Removed deserializer class and added custom serializer to handle records. Removed subclass of TimeSeries.Record.

adamkorynta (approved):

PR #937, opened 2024-10-31
add missing office where clause in Clob getAll query

### LINKED ISSUE #939: Clob getAll missing office filter in query
Peter and I noticed when checking for REGI specific clobs that more than the requested office was coming back in the payload. I started a PR, but it needs more work for regex checking and integration testing: https://github.com/USACE/cwms-data-api/pull/937

adamkorynta:


rma-rripken:
Since we (probably me) got this wrong at least once there should probably be tests that go along with these changes.
This pr still isn't right.  The count sub-query uses JooqDao.caseInsensitiveLikeRegex but the data retrieval part uses 'DSL.upper(v_office.OFFICE_ID).like(officeLike.toUpperCase()))'   those should align -right? at least I think they should.  
The paging also doesn't look right. I guess it was broken before you got here but it seems to me that this needs to order by officeId and clobId

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
```diff
@@ -134,10 +134,10 @@ public Clobs getClobs(String cursor, int pageSize, String officeLike,
                         includeValues ? v_clob.VALUE : DSL.inline("").as(v_clob.VALUE)
                 )
                 .from(v_clob)
-                //.innerJoin(forLimit).on(forLimit.field(v_clob.ID).eq(v_clob.ID))
                 .join(v_office).on(v_clob.OFFICE_CODE.eq(v_office.OFFICE_CODE))
                 .where(JooqDao.caseInsensitiveLikeRegex(v_clob.ID,idRegex))
                 .and(DSL.upper(v_clob.ID).greaterThan(clobCursor))
+                .and(officeLike == null ? noCondition() : DSL.upper(v_office.OFFICE_ID).like(officeLike.toUpperCase()))
```
rma-rripken: This should be caseInsensitiveRegexLike -right?
Paging should also order by office and clobId.
Probably need test
Is there an approved issue for these changes?

rma-rripken (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
adamkorynta: No issue, I'll create one, good call on the regex.

adamkorynta: added tests, made office id support regex, ordered the page by office id and clob id.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -138,11 +138,6 @@ public static AuthDao getInstance(DSLContext dsl) {
         return getInstance(dsl, null);
     }
 
-    @Override
-    public List<DataApiPrincipal> getAll(String limitToOffice) {
-        throw new UnsupportedOperationException("Unimplemented method 'getAll'");
-    }
-
```
rma-rripken, line 145: Is this related to the other changes?

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
```diff
@@ -88,28 +65,24 @@ public Optional<Clob> getByUniqueName(String uniqueName, String office) {
...
-
     public Clobs getClobs(String cursor, int pageSize, String officeLike,
                           boolean includeValues, String idRegex) {
         int total = 0;
         String clobCursor = "*";
         AV_CLOB v_clob = AV_CLOB.AV_CLOB;
         AV_OFFICE v_office = AV_OFFICE.AV_OFFICE;
 
+        Condition whereClause = JooqDao.caseInsensitiveLikeRegex(v_clob.ID, idRegex)
+            .and(officeLike == null ? noCondition() : JooqDao.caseInsensitiveLikeRegex(v_office.OFFICE_ID, officeLike));
```
rma-rripken: Didn't you add a caseInsensitiveLikeRegexNullTrue that does exactly this?

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
```diff
@@ -128,17 +101,17 @@ public Clobs getClobs(String cursor, int pageSize, String officeLike,
...
-                .orderBy(v_clob.ID).limit(pageSize);
+                v_office.OFFICE_ID,
+                v_clob.ID,
+                v_clob.DESCRIPTION,
+                includeValues ? v_clob.VALUE : DSL.inline("").as(v_clob.VALUE)
+            )
+            .from(v_clob)
+            .join(v_office).on(v_clob.OFFICE_CODE.eq(v_office.OFFICE_CODE))
+            .where(whereClause)
+            .and(DSL.upper(v_clob.ID).greaterThan(clobCursor))
```
rma-rripken: This still isn't quite right. At least in my opinion.  If you are ordering by office and clobId then I think the paging has to do the sameOffice/nextOffice thing.  See buildPagingCondition in TimeSeriesDaoImpl for what I'm talking about

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
rma-rripken: Just removing unused.

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
MikeNeilson: I'm fairly certain Ryan is correct. *especially* if you have more than one office possibly coming through, but even just limiting to a single you want that sorting to be consistent.

rma-rripken (changes requested):
I think the paging cursor needs to deal with the office and clobId of the cursor

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
adamkorynta: wow, I have no recollection of writing that, but I guess I did!

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
adamkorynta: yup, thanks for catching this. Updated to match how you handled it in the projects endpoint

rma-rripken (changes requested):
The paging conditions still don't look right.    From your two part cursor you want to find items in the same office that have later clobIds AND you also want things that are in latter offices. Are you doing that and I'm not seeing it?

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
rma-rripken: Is there another commit? I don't see it in these changes

rma-rripken:
I made a new PR with these changes.  #1024

PR #951, opened 2024-11-15
CWMSVue-594 Bug Fix

zack-rma:
Updated schema version to address build errors

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
@@ -172,8 +172,11 @@ public class LocationGroupController implements CrudHandler {
             LocationGroupDao cdm = new LocationGroupDao(dsl);
             String office = requiredParam(ctx, OFFICE);
             String categoryId = requiredParam(ctx, CATEGORY_ID);
-            String groupOfficeId = requiredParam(ctx, GROUP_OFFICE_ID);
-            String categoryOfficeId = requiredParam(ctx, CATEGORY_OFFICE_ID);
+
+            // Not marked as required to maintain backwards compatibility with existing clients
+            String groupOfficeId = ctx.queryParam(GROUP_OFFICE_ID);
+            String categoryOfficeId = ctx.queryParam(CATEGORY_OFFICE_ID);
+
             String formatHeader = ctx.header(Header.ACCEPT);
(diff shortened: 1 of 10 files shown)


cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesGroupController.java
```diff
@@ -190,8 +188,8 @@ public void getOne(@NotNull Context ctx, @NotNull String groupId) {
             TimeSeriesGroupDao dao = new TimeSeriesGroupDao(dsl);
             String office = ctx.queryParam(OFFICE);
             String categoryId = ctx.queryParam(CATEGORY_ID);
-            String groupOffice = ctx.queryParam(GROUP_OFFICE_ID);
-            String categoryOffice = ctx.queryParam(CATEGORY_OFFICE_ID);
+            String groupOffice = requiredParam(ctx, GROUP_OFFICE_ID);
```
adamkorynta: changing this to be required breaks backwards compatibility

cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
```diff
@@ -100,7 +104,8 @@ public void getAll(@NotNull Context ctx) {
             DSLContext dsl = getDslContext(ctx);
             LocationGroupDao cdm = new LocationGroupDao(dsl);
 
-            String office = ctx.queryParam(OFFICE);
+            String groupOfficeId = requiredParam(ctx, OFFICE);
```
adamkorynta, line 107: While the swagger API should declare this as required, this would break backwards compatibility and require client updates. Leave swagger change and then put a comment in the code explaining why it isn't marked required.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesGroupController.java
zack-rma: Swapped to not required and added a comment explaining why.

cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
zack-rma: Swapped to not required and added a comment explaining why.

adamkorynta (approved):

PR #954, opened 2024-11-20
CWMSVue 594 - Added location office parameter to LocationGroup getAll

zack-rma:
Added location office parameter to LocationGroup getAll

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
@@ -59,6 +59,7 @@ import org.jooq.DSLContext;
 
 public class LocationGroupController implements CrudHandler {
     public static final Logger logger = Logger.getLogger(LocationGroupController.class.getName());
+    private static final String LOCATION_OFFICE_ID = "location-office-id";
 
     public static final String TAG = "Location Groups";
     private final MetricRegistry metrics;
@@ -79,7 +80,7 @@ public class LocationGroupController implements CrudHandler {
     @OpenApi(
         queryParams = {
             @OpenApiParam(name = OFFICE, description = "Specifies the owning office of the "
-                + "locations assigned to the group whose data is to be included in the response. If this "
+                + "location group whose data is to be included in the response. If this "
                 + "field is not specified, matching location groups information from all "
                 + "offices shall be returned."),
             @OpenApiParam(name = INCLUDE_ASSIGNED, type = Boolean.class, description = "Include"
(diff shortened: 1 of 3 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -273,10 +279,21 @@ public List<LocationGroup> getLocationGroups(String groupOfficeId, String catego
                 .on(alcg.LOC_CATEGORY_ID.eq(alga.CATEGORY_ID)
                         .and(alcg.LOC_GROUP_ID.eq(alga.GROUP_ID)));
 
-        if (groupOfficeId != null) {
-            connectBy = onStep.where(DSL.upper(alcg.GRP_DB_OFFICE_ID).eq(groupOfficeId.toUpperCase()).and(condition));
+        if (officeId != null) {
```
adamkorynta: Seems like these conditions can be simplified. If group office id is not null, then filter group. Else if location office is not null then filter location office. Not sure why there are four branches here

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
```diff
@@ -352,18 +370,24 @@ public List<LocationGroup> getLocationGroups(String groupOfficeId,
             condition = condition.and(alcg.CAT_DB_OFFICE_ID.eq(categoryOfficeId.toUpperCase()));
         }
 
-        if (groupOfficeId != null) {
-            if (CWMS.equalsIgnoreCase(groupOfficeId)) {
+        if (officeId != null) {
```
adamkorynta: for consistent and clarified naming, use "locationOfficeId" (including naming in controller and other methods)

cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
```diff
@@ -76,9 +76,10 @@ private Timer.Context markAndTime(String subject) {
         return Controllers.markAndTime(metrics, getClass().getName(), subject);
     }
 
-    @OpenApi(queryParams = {
+    @OpenApi(
+        queryParams = {
             @OpenApiParam(name = OFFICE, description = "Specifies the owning office of the "
```
adamkorynta, line 82: Changing the usage of a parameter breaks backwards compatibility

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
zack-rma: Simplified conditioning

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationGroupDao.java
zack-rma: renamed where appropriate

cwms-data-api/src/main/java/cwms/cda/api/LocationGroupController.java
zack-rma: reverted functionality and moved location office to additional query parameter

adamkorynta (approved):

PR #959, opened 2024-11-22
Creation of initial decision records.

MikeNeilson:
In reference to #942, this is the initial creation of decision records for CDA.

In the current state, as little feedback has been provided on #942 at this time, this is to get feedback on the format of the decision records themselves, but anyone looking at this PR should comment, maybe open PR to this PR to add opinions.

I do not intent to merge these until an actual decision is reached with appropriate feedback; however, barring feedback and discussion such decisions will be made.

### DIFF
--- a/docs/decisions/0001-api-versioning.md
+++ b/docs/decisions/0001-api-versioning.md
@@ -1,0 +1,44 @@
+# API Versioning and Reporting
+
+## Summary
+
+API should provide a reason version and list/matrix of capabilities for a given instance of CDA
+
+## Opinions
+
+### Opinion 1 Calendar versioning
+
+@MikeNeilson
+
+Summary: Calendar versioning is easier to support and automate with this API
+
+As the one whose has been making the "official" releases Semantic Version versioning has basically been useless.
+We have been making so many feature additions that if I was doing it right we'd never have a minor version change between releases.
+It's also caused me to, I think, release too slowly.
+
+While we can automate SemVer it is an additional step.
+
+With Calendar Versioning automation tools can just pick the current date when appropriately triggered,
+perhaps by merged into a particular branch.
+
+
+### Opinion 2 Users
+
+Summary: It has been asked more than once that a version be provided
+
+Having a version allows client to better respond to what's available instead of failing in obtuse ways.
+
+
+## Decision Status
+
+accepted
+
+1. Provide endpoint to retrieve current API version.
+2. Likely include capability list or matrix.
+
+## References
+
+
+## Related decisions
+
+- data-versioning
--- a/docs/decisions/0002-data-versioning.md
+++ b/docs/decisions/0002-data-versioning.md
@@ -1,0 +1,55 @@
+# Data Types use Calendar Versioning
+
+## Summary
+
+Instead of versioning the entire API, though the API has a version, we version the data types.
+
+## Opinions
+
+### Opinion 1
+
+Summary: Versioning the API itself, at the level of the path, will lead to too many paths to manage and be awkward for the clients
+
+@MikeNeilson
+
+By versioning the data, and using the Content-Type and Accept headers and the full features of MIME types we appropriately 
+separate the concern of "what data we are retrieving/storing" from the presentation of data.
+
+e.g /timeseries/Alder Springs.Temp-Air.Inst.15Minutes.0.GOES-raw?begin=PT0&end=PT-1D&units=C, granted with a reasonable
+ exception to the units, defines *what* we want.
+
+The header, Accept, informs the API what format, or formats, we are willing to accept the data in. 
+
+
+
+
+## Decision Status
+
+proposed
+
+Data, by content-types, are versioned. In the past there was some severe confusion on this part and it was treated as anything new was "version=2" in the content-type. To allow this design but reduce confusion going forward
+
+1. The initial version of a data set *SHALL* be the date it was finalized (e.g. PR about to be merged.)
+2. *IF* it is the first version of this data the plain content-type "e.g. application/json" will point to this data.
+3. *IF* is it not the version version of this data, it will be discussed and announced when it becomes the new default data.
+4. Downstream systems *SHOULD* use the specific version regardless of when implemented, and this behavior should be well documented.
+5. If a given data set includes definitions of its shape within the type there should be sufficient documentation for downstream
+   developers to properly account for any changes over time. (See our TimeSeries type and discussions within #927).
+
+Version format is `YYYY-MM-DD`
+
+[comment:] <> (Status: request for comments | proposed | accepted | rejected | deprecated | superseded)
+
+## References
+
+I have several, I will dig them up likely next week
+
+## Notes
+
+The initial idea in CDA was that the first version of any data type was, we'll just stick with JSON for each of message,
+"application/json;version=1" with "application/json" being the alias to the latest format version. However, this was not
+correctly communicated and several brand new data transfer objects were created as ";version=2" under the impression that
+this was the version for the new system. Attempting to use a simple number of this has clearly caused confusion in general.
+
+We also failed to create the initial alias system which caused even more confusion when users attempted to test things 
+directly in a browser instead of the provided swagger-ui.
--- a/docs/decisions/0003-searchability-and-catalogs.md
+++ b/docs/decisions/0003-searchability-and-catalogs.md
@@ -1,0 +1,42 @@
+# Data should be readily searchable
+
+## Summary
+
+It's not just a good idea, it's technically the law. While CDA currently expose a fair amount of information to search
+it's never entirely clear.
+
+## Opinions
+
+### Opinion 1
+
+Summary: Each data time should support it's own /catalog end point.
+
+@MikeNeilson
+
+The original CDA has the say, `/timeseries` end point provide a catalog if no data is set. I created a /catalog end point
+to attempt to consildate search query parameters. For TimeSeries and Locations this works reasonably well since there
+is parity between the concepts.
+
+However, if we tried to add ratings into the mix, the list of query parameters grows, and it would rather difficult to 
+document which is for what or what changes for each.
+
+To make 'catalog' operations clear, we should create /catalog for each data type that provide for discoverability of that data.
+
+### Opinion 2
+
+Summary: Each datatype under "catalog" should be a full path"
+
+@MikeNeilson
+
+If it makes sense to group all catalogs under catalog, perhaps for grouping in the SWAGGER-UI, making each catalog it's own
+path under `/catalog` instead of the current path parameter is a better approach.
+
+We would maintain the grouping, but each catalog can have it's appropriate search criteria.
+
+## Decision Status
+
+proposed
+
+[comment:] <> (Status: request for comments | proposed | accepted | rejected | deprecated | superseded)
+
+## References
--- a/docs/source/decisions/0001-api-versioning.rst
+++ b/docs/source/decisions/0001-api-versioning.rst
@@ -1,12 +1,17 @@
-# API Versioning and Reporting
+############################
+API Versioning and Reporting
+############################
 
-## Summary
+Summary
+=======
 
 API should provide a reason version and list/matrix of capabilities for a given instance of CDA
 
-## Opinions
+Opinions
+========
 
-### Opinion 1 Calendar versioning
+Opinion 1 Calendar versioning
+-----------------------------
 
 @MikeNeilson
 
@@ -22,23 +27,27 @@ With Calendar Versioning automation tools can just pick the current date when ap
 perhaps by merged into a particular branch.
 
 
-### Opinion 2 Users
+Opinion 2 Users
+---------------
 
 Summary: It has been asked more than once that a version be provided
 
 Having a version allows client to better respond to what's available instead of failing in obtuse ways.
 
 
-## Decision Status
+Decision Status
+===============
 
 accepted
 
 1. Provide endpoint to retrieve current API version.
 2. Likely include capability list or matrix.
 
-## References
+References
+==========
 
 
-## Related decisions
+Related decisions
+=================
 
 - data-versioning
--- a/docs/source/decisions/0001-api-versioning.rst
+++ b/docs/source/decisions/0001-api-versioning.rst
@@ -26,6 +26,13 @@ While we can automate SemVer it is an additional step.
 With Calendar Versioning automation tools can just pick the current date when appropriately triggered,
 perhaps by merged into a particular branch.
 
+@adamkorynta 
+
+ Summary: Calendar versioning more directly aligns with consumers needs
+ 
+Deprecation, desupport, and removal notices usually involve timelines rather than predicting x number of major/minor releases into the future.
+ 
+Given the endpoints themselves are already versioned, following semver becomes obtuse with only a real use-case of structural changes such as supported JDK changes, etc. 
 
 Opinion 2 Users
 ---------------
--- a/docs/source/decisions/0002-data-versioning.rst
+++ b/docs/source/decisions/0002-data-versioning.rst
@@ -6,6 +6,8 @@ Summary
 =======
 
 Instead of versioning the entire API, we version the data types if appropriate.
+This versioning only applies to simple non-breaking changes and it primarily presented as a hint in the returned
+data.
 
 Opinions
 ========
@@ -26,22 +28,30 @@ e.g /timeseries/Alder Springs.Temp-Air.Inst.15Minutes.0.GOES-raw?begin=PT0&end=P
 
 The header, Accept, informs the API what format, or formats, we are willing to accept the data in. 
 
-
-
 Opinion 2
+---------
 
 Summary: Remove data versioning in all endpoints.
 
 @adamkorynta 
 
-The accept/content-type headers have provided sufficient confusion to downstream clients and deviate from industry standards. I have yet to see an endpoint where this versioning solves the proposed problem of needing to different shapes of data (other than json vs xml). The concept was introduced to solve the straight-to-db queries, which did not have any OpenAPI documentation moving to in-app DTO's which now have documentation. When moving away from the straight-to-db queries, we needed to thoroughly expand queries parameters and make other backwards-incompatible changes unrelated to the data shape. Given that context in hindsight a path version would have worked better.
+The accept/content-type headers have provided sufficient confusion to downstream clients and deviate from industry standards.
+I have yet to see an endpoint where this versioning solves the proposed problem of needing to different shapes of data (other than json vs xml).
+The concept was introduced to solve the straight-to-db queries, which did not have any OpenAPI documentation moving to in-app DTO's which now have documentation. When moving away from the straight-to-db queries, we needed to thoroughly expand queries parameters and make other backwards-incompatible changes unrelated to the data shape. Given that context in hindsight a path version would have worked better.
 
-The closest we've gotten to needing new shapes on the same endpoint is the data-entry date on TimeSeries, but this was more appropriately solved via query parameter and data arrays. I think if we had really wanted to be a stickler on the "what format" we could have easily added another endpoint path instead. Even if/when we add text annotations, using a content type is obtuse given the lack of discoverability as we would then need `application/json`, `application/json+data-entry+text-annotations`, `application/json+text-annotations`, `application/json+data-entry` which seems like just another type of bloat that is more hidden from clients.
+The closest we've gotten to needing new shapes on the same endpoint is the data-entry date on TimeSeries, but this was
+more appropriately solved via query parameter and data arrays. I think if we had really wanted to be a stickler on the 
+"what format" we could have easily added another endpoint path instead. Even if/when we add text annotations,
+using a content type is obtuse given the lack of discoverability as we would then need `application/json`, 
+`application/json+data-entry+text-annotations`, `application/json+text-annotations`, `application/json+data-entry`
+which seems like just another type of bloat that is more hidden from clients.
 
 Decision Status
 ---------------
 
-partial acceptance
+`rejected` - the descriptions in this proposal are awkward and it is not clear how to fix them. Additionally as we have
+decided to adopt path based versioning and we've made `application/json` or `application/xml` default to the latest desired 
+the requirement is now moot.
 
 Data, by content-types, are versioned. In the past there was some severe confusion on this part and it was treated as anything 
 new was "version=2" in the content-type. To allow this design but reduce confusion going forward
@@ -58,6 +68,10 @@ new was "version=2" in the content-type. To allow this design but reduce confusi
 
 References
 ==========
+
+1. https://www.youtube.com/watch?v=jmoxGJ_sLgU
+2. https://newsletter.systemdesign.one/p/api-versioning
+3. https://www.speakeasy.com/api-design/versioning
 
 
 Notes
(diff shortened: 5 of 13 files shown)


MikeNeilson:
I've tagged some specific people for initial review, please feel free to bring in anyone else that may have an opinion. This is a discussion of both the decision record format as well as the actual decisions being recorded.

More will follow, I'm starting small...ish.

MikeNeilson:
@jvanaalsburg considering you started with with cwms-python, probably good to get your feedback about how I'm starting these as well.

docs/decisions/0001-api-versioning.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Opinions
+
+### Opinion 1 Calendar versioning
+
+@MikeNeilson
+
+Summary: Calendar versioning is easier to support and automate with this API
+
+As the one whose has been making the "official" releases Semantic Version versioning has basically been useless.
```
adamkorynta, line 20: Definitely agree!

docs/decisions/0002-data-versioning.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Opinions
+
+### Opinion 1
+
+Summary: Versioning the API itself, at the level of the path, will lead to too many paths to manage and be awkward for the clients
+
+@MikeNeilson
+
+By versioning the data, and using the Content-Type and Accept headers and the full features of MIME types we appropriately
```
adamkorynta, line 23: I think REST principal accuracy is a good goal, but I think we should consider the following:
 
- API clutter
- industry standards
- browser support
- clear documentation
 
I would argue that accept header versioning provides the same amount of clutter both in implementation and downstream usage as path parameter versioning. However, it fails at the other three points.
 
While I would not want to have a tool (Swagger webpage) dictate how we implement an API, I do think it is a non-trivial point when it is our only source of API documentation.
The OpenAPI does not provide a meaningful way to distinguish between different versions on the version marker itself and the Swagger webpage does not make it immediately obvious that there are multplie versions of an endpoint.
Adding/removing/editing behavior of query parameters require extra documentation on the parameter (e.g. "This parameter is not supported by application/json;version=1, application/json;version=2 but is supported by application/json;version=2025-05-01. Now make sure you check what the default is to make sure the parameter is supported by your version."). Why do that to ourselves and users?
 
Path versioning on the other hand, communicates very clearly to end users what different versions' behaviors will be, both in query parameter behaviors and in what formats are accepted/returned.
 
Switching to calendar versioning on the accept header gives away the API clutter benefit as now downstream usages will have to manage many different versions beyond easy-to-grok integers, in which case, might as well put the version in the path or as a query parameter to at least provide easier documentation and browser support.
 
I could see a slight mitigation for the explosion of versions to manage by looping it into the release cycle so they are at least versioned along with the REST API itself, rather than at PR creation date. At that point though, we're just adding more maintenance burden and over time with faster release cycles this becomes moot.

docs/decisions/0003-searchability-and-catalogs.md
```diff
@@ -0,0 +1,42 @@
...
+is parity between the concepts.
+
+However, if we tried to add ratings into the mix, the list of query parameters grows, and it would rather difficult to 
+document which is for what or what changes for each.
+
+To make 'catalog' operations clear, we should create /catalog for each data type that provide for discoverability of that data.
+
+### Opinion 2
+
+Summary: Each datatype under "catalog" should be a full path"
```
adamkorynta: trailing quote

docs/decisions/0003-searchability-and-catalogs.md
```diff
@@ -0,0 +1,42 @@
...
+
+@MikeNeilson
+
+If it makes sense to group all catalogs under catalog, perhaps for grouping in the SWAGGER-UI, making each catalog it's own
+path under `/catalog` instead of the current path parameter is a better approach.
+
+We would maintain the grouping, but each catalog can have it's appropriate search criteria.
```
adamkorynta: This does seem in opposition to the accept header reasoning. We already implement getAll for most (all?) data types. However, most of them return the full data object rather than a listing of refs/ids. The catalog endpoint is supposed to be the lightweight alternative to a getAll, which seems to be a change in the shape of the data. A `application/json;catalog=true` or something similar.

docs/decisions/0002-data-versioning.md
MikeNeilson: Note: we are expanding the API documentation, Actually I need to move these to rst and that read-the-docs infrastructure.

But I'm still not convinced path versioning is correct. I can see the point about not using dates since they will be somewhat all over the place. But while path version may be an industry standard... it seems a really sloppy industry standard.

Over time, while our outputs have varied slightly, the inputs have only been expanded, not broken in a backwards incompatible way (at least intentionally)

And the few places were they have drastically change, when ended up with a better name of the endpoint anyways, usually something more refined and specific that `/cwms-data/v1/locations` -> `/cwms-data/v2/locations` (just for a concrete example).

I have found some arguments and discussion in favor of versioning data, I will take the time to look them up.

Another option, instead of `;version=` could just be more expanded data types like `application/vnd.json-ts-yada yada` (I can't remember off the top of my end the right vnd syntax but I think that gets the idea across)

docs/decisions/0003-searchability-and-catalogs.md
MikeNeilson: Fair point, I hadn't thought of just using a different content-type of content-type parameter in that situation.

adamkorynta: I will say, while I brought up using the header, I still don't like that solution given my arguments above about discoverability and lack of clear Swagger UI documentation....

MikeNeilson:
NOTE: I will be forcibly merge this in as-is next Friday barring any additional input.

That said this is "initial decision records" should any one decide they think our design goals and intentions are working in the future, you are of course free, and encouraged, to propose changes. But we have to start someone and I don't want this to linger anymore.

docs/source/decisions/0001-api-versioning.rst
```diff
@@ -0,0 +1,53 @@
...
+
+As the one whose has been making the "official" releases Semantic Version versioning has basically been useless.
+We have been making so many feature additions that if I was doing it right we'd never have a minor version change between releases.
+It's also caused me to, I think, release too slowly.
+
+While we can automate SemVer it is an additional step.
+
+With Calendar Versioning automation tools can just pick the current date when appropriately triggered,
+perhaps by merged into a particular branch.
+
```
adamkorynta, line 28:
```suggestion

@adamkorynta 

 Summary: Calendar versioning more directly aligns with consumers needs
 
Deprecation, desupport, and removal notices usually involve timelines rather than predicting x number of major/minor releases into the future.
 
Given the endpoints themselves are already versioned, following semver becomes obtuse with only a real use-case of structural changes such as supported JDK changes, etc. 
```

docs/source/libraries/java.rst
```diff
@@ -0,0 +1,3 @@
+####
```
adamkorynta: Is this a decision record for determining which version(s) of java to support?

docs/source/decisions/index.rst
```diff
@@ -0,0 +1,23 @@
+################
+Design Decisions
+################
+
+
+Overview
+========
+
+
+Below are agree upon decision choices regarding the usage of the API.
```
adamkorynta:
```suggestion
Below are agreed upon decision choices regarding the usage of the API.
```

docs/source/decisions/0003-searchability-and-catalogs.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Decision Status
+===============
+
+proposed
+
+[comment:] <> (Status: request for comments | proposed | accepted | rejected | deprecated | superseded)
+
+References
+==========
```
adamkorynta, line 76: recommend adding a resource referencing the law: `It's not just a good idea, it's technically the law.`

docs/source/decisions/0003-searchability-and-catalogs.rst
```diff
@@ -0,0 +1,50 @@
...
+It's not just a good idea, it's technically the law. While CDA currently expose a fair amount of information to search
+it's never entirely clear.
+
+Opinions
+========
+
+Opinion 1
+---------
+
+Summary: Each data time should support it's own /catalog end point.
```
adamkorynta:
```suggestion
Summary: Each data type should support it's own /catalog end point.
```

docs/source/decisions/0003-searchability-and-catalogs.rst
```diff
@@ -0,0 +1,50 @@
...
+---------
+
+Summary: Each datatype under "catalog" should be a full path"
+
+@MikeNeilson
+
+If it makes sense to group all catalogs under catalog, perhaps for grouping in the SWAGGER-UI, making each catalog it's own
+path under `/catalog` instead of the current path parameter is a better approach.
+
+We would maintain the grouping, but each catalog can have it's appropriate search criteria.
```
adamkorynta:
```suggestion
We would maintain the grouping, but each catalog can have its appropriate search criteria.
```

docs/source/decisions/0003-searchability-and-catalogs.rst
```diff
@@ -0,0 +1,50 @@
...
+To make 'catalog' operations clear, we should create /catalog for each data type that provide for discoverability of that data.
+
+Opinion 2
+---------
+
+Summary: Each datatype under "catalog" should be a full path"
+
+@MikeNeilson
+
+If it makes sense to group all catalogs under catalog, perhaps for grouping in the SWAGGER-UI, making each catalog it's own
```
adamkorynta, line 52: Groups in the swagger-ui are created using the TAG annotation configuration, not by the paths themselves.

I find catalog grouped with the data type easier for discoverability:
- `timeseries/catalog` 
- `timeseries/<name>` 
- `location/catalog` 
- `location/<name>` 

It looks like swagger-ui does support multiple tags and allowing the same endpoint to show up under both groups in the UI.

docs/source/decisions/0004-versioning.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Summary
+=======
+
+Maintaining backwards compatibility while improving future difficulty has proven sufficiently difficulty that change
+is required.
+
+The API as a whole will retain the calendar based versioning for formal releases.
+Data *SHOULD* be versioned, if appropriate/needed, with otherwise backwards compatible changes to query parameters.
+Endpoints will be places under a new "api version" for backwards incompatible or confusing parameter changes.
```
adamkorynta:
```suggestion
Endpoints will be placed under a new "api version" for backwards incompatible or confusing parameter changes.
```

docs/source/decisions/0004-versioning.rst
```diff
@@ -0,0 +1,61 @@
...
+Opinion 1
+---------
+
+Summary: Current scheme is not working
+
+Author MikeNeilson, on behalf of others
+
+We have failed to properly handle existing usages while attempting to improve the overall design of the api
+and have been breaking various downstream usages due to the confusion. Allowing the endpoints to be versioned allows
+an easier time keeping existing behavior while at allowing more drastic improvements in usages to happen.
```
adamkorynta:
```suggestion
an easier time keeping existing behavior while also allowing more drastic improvements in usages to happen.
```

docs/source/decisions/0004-versioning.rst
```diff
@@ -0,0 +1,61 @@
...
+.. code-block:: bash
+
+    # new time series becomes
+    cwms-data/v2/timeseries
+    # the new location becomes
+    cwms-data/v3/locations
+
+.. NOTE::
+
+    Or is that confusing and we should just allows add a new endpoint to the highest endpoint version?
```
adamkorynta, line 40: Worth calling out how various methods would get versioned with path versioning:

I presume we would want `cwms-data/v3/locations GET`, `cwms-data/v3/locations/<name> GET`, `cwms-data/v3/locations/<name> DELETE`, `cwms-data/v3/locations/<name> POST` would all get updated to the v3 even if only the `cwms-data/v3/locations/<name> GET` became backwards incompatible?

docs/source/decisions/0003-searchability-and-catalogs.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Summary: Each data time should support it's own /catalog end point.
+
+@MikeNeilson
+
+The original CDA has the say, `/timeseries` end point provide a catalog if no data is set. I created a /catalog end point
+to attempt to consildate search query parameters. For TimeSeries and Locations this works reasonably well since there
+is parity between the concepts.
+
+However, if we tried to add ratings into the mix, the list of query parameters grows, and it would rather difficult to
```
adamkorynta, line 40: Can we define what a "catalog" request means under this concept? We already have Javalin CrudHandler getAll implementations under the data types for every data type. There is a redundant /locations getAll already that does similar cataloging as the `/catalog/locations`. If we are adding an explicit `/locations/catalog` would there be a difference in implementation? Would the returned results be a smaller payload than the other CRUD endpoints? We would want that to be clear and explicit in this decision doc to avoid confusion.

More questions on this: if we are separating out the CrudHandler getAll endpoints from a /catalog endpoint, would one handle aliases, and would one return a smaller subset of metadata? Ex. /locations/catalog would return office+loc id only but /locations (CrudHandler getAll) would return all physical location metadata and all aliases for each location?

docs/source/decisions/0002-data-versioning.rst
```diff
@@ -0,0 +1,66 @@
...
+
+Decision Status
+---------------
+
+partial acceptance
+
+Data, by content-types, are versioned. In the past there was some severe confusion on this part and it was treated as anything 
+new was "version=2" in the content-type. To allow this design but reduce confusion going forward
+
+1. The initial content-type of a data set *SHALL* be be the plain content-type and *SHOULD* include an additional expanded content-type
```
adamkorynta, line 59: For clarification: this is what we are doing now, correct? the first version is `application/json` and `application/json;version=1`

docs/source/decisions/0002-data-versioning.rst
```diff
@@ -0,0 +1,66 @@
...
+3. *IF* is it not the first version of this data, additional information will be set in the content-type as 
+   appropriate to the to the data. (e.g. `application/json+<something>` or `application/json;<something>`)
```
adamkorynta, line 61: I think this needs a more concrete example than `<something>`, i.e. to demonstrate the problem we are trying to solve

docs/source/decisions/0002-data-versioning.rst
```diff
@@ -0,0 +1,66 @@
...
+   1. It will be discussed and announced when it becomes the new default data, if that decision is made.
+5. Downstream systems *SHOULD* use the specific version regardless of when implemented, and this behavior should be well documented.
```
adamkorynta, line 63: The swagger-ui does not provide a mechanism for describing the content types, only providing examples. Where should the current and upcoming new defaults get documented? We would also want to document which version the default version points to, which doesn't look possible to do in the OpenAPI schema.

docs/source/decisions/0002-data-versioning.rst
```diff
@@ -0,0 +1,66 @@
...
+
+@MikeNeilson
+
+By versioning the data, and using the Content-Type and Accept headers and the full features of MIME types we appropriately 
+separate the concern of "what data we are retrieving/storing" from the presentation of data.
+
+e.g /timeseries/Alder Springs.Temp-Air.Inst.15Minutes.0.GOES-raw?begin=PT0&end=PT-1D&units=C, granted with a reasonable
+ exception to the units, defines *what* we want.
+
+The header, Accept, informs the API what format, or formats, we are willing to accept the data in.
```
adamkorynta, line 29: The example above needs to expand here to describe how to do this with the time series above

docs/source/decisions/0002-data-versioning.rst
```diff
@@ -0,0 +1,66 @@
...
+
+By versioning the data, and using the Content-Type and Accept headers and the full features of MIME types we appropriately 
+separate the concern of "what data we are retrieving/storing" from the presentation of data.
+
+e.g /timeseries/Alder Springs.Temp-Air.Inst.15Minutes.0.GOES-raw?begin=PT0&end=PT-1D&units=C, granted with a reasonable
+ exception to the units, defines *what* we want.
+
+The header, Accept, informs the API what format, or formats, we are willing to accept the data in. 
+
+
```
adamkorynta, line 37:
```suggestion


Opinion 2

Summary: Remove data versioning in all endpoints.

@adamkorynta 

The accept/content-type headers have provided sufficient confusion to downstream clients and deviate from industry standards. I have yet to see an endpoint where this versioning solves the proposed problem of needing to different shapes of data (other than json vs xml). The concept was introduced to solve the straight-to-db queries, which did not have any OpenAPI documentation moving to in-app DTO's which now have documentation. When moving away from the straight-to-db queries, we needed to thoroughly expand queries parameters and make other backwards-incompatible changes unrelated to the data shape. Given that context in hindsight a path version would have worked better.

The closest we've gotten to needing new shapes on the same endpoint is the data-entry date on TimeSeries, but this was more appropriately solved via query parameter and data arrays. I think if we had really wanted to be a stickler on the "what format" we could have easily added another endpoint path instead. Even if/when we add text annotations, using a content type is obtuse given the lack of discoverability as we would then need `application/json`, `application/json+data-entry+text-annotations`, `application/json+text-annotations`, `application/json+data-entry` which seems like just another type of bloat that is more hidden from clients.
```

docs/decisions/0002-data-versioning.md
rma-rripken: I'd like to see some examples of how industry does versioning.  What is done by Stripe, Twitter, OpenAI etc?  Do they version data-types too?

docs/decisions/0003-searchability-and-catalogs.md
rma-rripken: In my opinion we haven't done a great job of explaining what a catalog is vs just a bulk-retrieve end-point.  I think they are (or can be) different things.   In some places we've added flags to enable the controllers to returned reduced data fields or added new *Identifier endpoints but that hasn't been done consistently

docs/source/libraries/java.rst
MikeNeilson: No, this is a placeholder that was missing from the original sphinx setup that was putting errors in the build output.

docs/source/decisions/0002-data-versioning.rst
MikeNeilson: Yeah, that's what we started doing.

docs/source/decisions/0002-data-versioning.rst
MikeNeilson: The schemas are the content-types and are listed individual in the response section and the accepted input sections. Where the examples fall flat is having overly generic example text, but that's getting fixed in other efforts.

docs/source/decisions/0003-searchability-and-catalogs.rst
MikeNeilson: Fair point. "Catalog" here is "I want to search for data", not necessarily I want to get all the data.

My thought behind having `<data>/catalog/` vs `/catalog/<data>` was to try and simplify things; however, since that was written I think we also consider moving the `/catalog/<data>` to have the `<data>` part be it's own independent Handler vs a single data type (the amount of query parameters on the `/catalog` endpoint are getting a bit... nuts.

docs/source/decisions/0003-searchability-and-catalogs.rst
MikeNeilson: Good to know.

docs/source/decisions/0003-searchability-and-catalogs.rst
MikeNeilson: Fair point, that one should actually be that hard to dig up.

docs/source/decisions/0004-versioning.rst
MikeNeilson: Probably best to duplicate

MikeNeilson (commented):
Thanks, definitely going to have to consider some of those to fill out some explinations.

docs/source/decisions/0002-data-versioning.rst
adamkorynta: examples are different than documentation though. using the example of including annotations in the time series retrieval - specifying `application/json;text-annotations=true`, there isn't a way to discover or describe what that option does. 

i.e. what happens if there is no annotation time series? do you still get a 3-dimensional data array with nulls or would an error get thrown?

docs/decisions/0003-searchability-and-catalogs.md
MikeNeilson: > In my opinion we haven't done a great job of explaining what a catalog is vs just a bulk-retrieve end-point. I think they are (or can be) different things. In some places we've added flags to enable the controllers to returned reduced data fields or added new *Identifier endpoints but that hasn't been done consistently

We also haven't done a great job of decide if we should even have a "catalog" endpoint or just treat the `<dataset>/` as the catalog for each.

docs/source/decisions/0002-data-versioning.rst
MikeNeilson: I've marked this is rejected and it's now left in for posterity and to show that it was considered.

docs/source/decisions/0002-data-versioning.rst
MikeNeilson: Document now only exists for posterity.

MikeNeilson (commented):
Updated have been provided based the the given feedback.

docs/source/decisions/0003-searchability-and-catalogs.rst
```diff
@@ -0,0 +1,80 @@
...
+It's not just a good idea, it's technically the law, see reference 1 and 2. While CDA currently expose a fair amount 
+of information to search it's never entirely clear. We *MUST* adopt a standard method of searchability.
+
+Additional Information
+----------------------
+
+Catalog - a complete list of items, for examples of things that people can look at or buy [3]
+
+By having a well defined structure of information users can more easily discover what they are looking for. While the
+Sagger-UI, if used, presented all of the types of data that can be found. The `catalog` for each type should present
```
adamkorynta:
```suggestion
Swagger-UI, if used, presented all of the types of data that can be found. The `catalog` for each type should present
```

docs/source/decisions/0003-searchability-and-catalogs.rst
```diff
@@ -0,0 +1,80 @@
...
+a clear way to find the available data of each type.
+
+The catalog of each data set would include only metadata associated with each data type. For example a time series
+catalog would include support to discover primary timeseries names, aliases, extends, and the like but not actual time
```
adamkorynta:
```suggestion
catalog would include support to discover primary timeseries names, aliases, extents, and the like but not actual time
```

adamkorynta (dismissed):

docs/source/decisions/0003-searchability-and-catalogs.rst
MikeNeilson: and having it specifically named "catalog" what to indicate that it's more like a library card catalog than for actual data.

docs/source/decisions/0003-searchability-and-catalogs.rst
MikeNeilson: References were added.

PR #965, opened 2024-12-02
Devops/build automation

MikeNeilson:
Use the new "ready-database" to run the integration tests from github actions.

Updates versioning in build.gradle, calendar versioning, but also supporting tag based so we have some control.

Added an additional Annotation that uses the full schema version (versions, it's a list) to decided if a test should run.


MikeNeilson:
Current failures in the GH build are the same as the teamcity build, tests that we know don't pass; several due to needing that staging version.

I will appropriately disable those for now.

MikeNeilson:
I may have, by label, incorrectly disabled a few tests, let me know.

The ones disabled due to multiple connections definitely need to be fixed, but they do correctly close the connections so in productions not a huge deal. That said the integration tests *intentionally* run with a pool size of 1 to expose connection usages issues.

.github/workflows/build.yml
```diff
@@ -27,6 +26,8 @@ jobs:
       - name: build and test
         id: thebuild
         run: ./gradlew build --info --init-script init.gradle
+      - name: integration tests
```
adamkorynta, line 29: Should there be a second step that runs on latest image?

MikeNeilson: Yes, but the "ready database" is currently only available for the latest-dev and eventually next schema. Install scripts required changes for the Oracle Free database.

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
```diff
@@ -90,7 +90,7 @@ public Outlet retrieveOutlet(String officeId, String locationId) {
             LOCATION_REF_T locRef = LocationUtil.getLocationRef(locationId, officeId);
             Configuration config = DSL.using(conn).configuration();
             PROJECT_STRUCTURE_OBJ_T outletStruct = CWMS_OUTLET_PACKAGE.call_RETRIEVE_OUTLET(config, locRef);
-            
+            // TODO: This may be causing tests to fail with pool size = 1.
```
adamkorynta: Should be fixed now with PR: https://github.com/USACE/cwms-data-api/pull/970

build.gradle
```diff
@@ -1,10 +1,26 @@
...
+    // tag is returned as is. Branch may need cleanup
+    // our primary versioning is branch+date, the tag will be used for the final "release"
+    return branch == null ? tag : new Date().format("yyyy.MM.dd-") + branch.replace("/", "-")
 }
+
 allprojects {
     apply plugin: 'cda.java-conventions'
     group = 'mil.army.usace.hec.cwms'
-    version = '3.1.6-SNAPSHOT' // ApiServlet.VERSION should be updated to match MAJOR.MINOR changes.
+    version = versionLabel(versionDetails()) // TODO: Figure out ApiServlet
```
adamkorynta: TODO

cwms-data-api/src/main/java/cwms/cda/data/dao/location/kind/OutletDao.java
MikeNeilson: Yep tests passed.

build.gradle
MikeNeilson: No longer todo, determined method of embedding version... or more correctly yanking the version that was already embedded in.

adamkorynta (approved):

PR #975, opened 2024-12-12
CWMVUE-611 - Adding in endpoint to get time range map for measurments

rma-bryson:
Updated measurement retrieval to use XML api. Fixed issues with quality. Introduced TimeExtents object.


cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -128,6 +130,16 @@ public static DSLContext getDslContext(Context ctx) {
         return retVal;
     }
 
+    protected static Timestamp buildTimestamp(Instant date)
+    {
+        Timestamp retval = null;
+        if(date != null)
+        {
```
MikeNeilson: The curlies on this are wrong.

Also this would probably read easier just doing a `return date !=null ? Timestamp.from(date) : null;`

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesExtents.java
```diff
@@ -16,33 +39,27 @@
...
-    @JsonFormat(shape = Shape.STRING)
-    ZonedDateTime latestTime;
-
     @Schema(description = "Last update in the timeseries")
     @JsonFormat(shape = Shape.STRING)
     ZonedDateTime lastUpdate;
 
     @SuppressWarnings("unused") // required so JAXB can initialize and marshal
     private TimeSeriesExtents() {
+        super(new Builder());
```
MikeNeilson: This is just... odd.

The constructor should just take a builder and thus both constructors should be private. We appear to be mixing builder/non-builder patterns here and a rather odd way..

MikeNeilson (changes requested):
Overall structure looks good. There a few "readability" areas that I think should be addressed.

cwms-data-api/src/main/java/cwms/cda/data/dto/TimeSeriesExtents.java
rma-bryson: I agree. I tried to preserve the existing TimeSeriesExtents object as much as possible, while making the new TimeExtents object follow the builder pattern of our other dtos. I would be in favor of updating the TimeSeriesExtents to use a builder pattern.

rma-bryson: I updated the TimeSeriesExtents to use a builder pattern for consistency

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
rma-bryson: Updated to fix brackets and used a ternary

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -130,14 +130,8 @@ public static DSLContext getDslContext(Context ctx) {
...
 
-    protected static Timestamp buildTimestamp(Instant date)
-    {
-        Timestamp retval = null;
-        if(date != null)
-        {
-            retval = Timestamp.from(date);
-        }
-        return retval;
+protected static Timestamp buildTimestamp(Instant date) {
```
MikeNeilson:
```suggestion
    protected static Timestamp buildTimestamp(Instant date) {
```

MikeNeilson (dismissed):
That looks a lot better.

MikeNeilson (approved):

Issue #818, opened 2024-07-30
Swagger Specification Issues


krowvin:
# CWMSjs 
*Note package.json requires semantic numbering of the format (##.##.##). I can't put RC in the version*

## I'm working on releasing the next version of [`CWMSjs`](https://hydrologicengineeringcenter.github.io/cwms-data-api-client-javascript/) 
*To include all (most) of the new endpoints*

Running the openapi generator for TypeScript and targeting the swagger.json spec json located [here](https://cwms-data.usace.army.mil/cwms-data/swagger-docs) 
gives a handful of errors (seen below)


 I thought it would be worth bringing it up here especially if the java and jython generators have not been ran yet.
 
## Background on how i'm getting the errors
TypeScript will do the type checking when you run `tsc` and transpile to JavaScript so that you don't have as many runtime errors in the browser (types/etc are checked on compile). Below is the output of that.  

In some cases I was able to make the changes to the `swagger schema` and then run the generator + tsc again and the errors would be fixed. I believe this confirms it to be an issue with the swagger schema (annotations?)


## Attempted error fixes to get the generation to complete

- Renaming all TimeSeries to Timeseries 
  - Did a CTRL +F  (case sensitive) TimeSeries and replace ALL with Timeseries
    - *This removed all TS errors*

- As for the remaining errors I tried doing the following:
  - Changed `AbstractRatingMetadata > discriminator > propertyName` from `"rating-type"` to `"ratingType"`
    - No Change to remaining errors
    
- Ended up removing all endpoints and references to the other errors for now. 


## Related to
- #351 
- #587 

## Error output running `tsc` against generated `.ts` files
```typescript
src/apis/LevelsApi.ts:22:3 - error TS2724: '"../models/index"' has no exported member named 'TimeSeries'. Did you mean 'Timeseries'?

22   TimeSeries,
     ~~~~~~~~~~

src/apis/LevelsApi.ts:33:5 - error TS2724: '"../models/index"' has no exported member named 'TimeSeriesFromJSON'. Did you mean 'TimeseriesFromJSON'?

33     TimeSeriesFromJSON,
       ~~~~~~~~~~~~~~~~~~

  src/models/TimeSeries.ts:171:17
    171 export function TimeseriesFromJSON(json: any): Timeseries {
                        ~~~~~~~~~~~~~~~~~~
    'TimeseriesFromJSON' is declared here.

[...]

PR #1003, opened 2024-12-30
Clean up API urls by eliminating basepath from the wrong location.

MikeNeilson:
Will be handling camel case of "TimeSeries" next as I believe the OpenAPI corrections should be one PR; will squash them to two different commits though before merge.

Solution was quite simple, don't mix up OpenAPI 2.x and OpenAPI 3.x configuration. After reading https://swagger.io/docs/specification/v3_0/api-host-and-base-path/ I realized we've just been doing it wrong because we were keeping old information around; that wasn't even being used anymore.

Partial fix of #818


MikeNeilson:
Actually the camel case thing might take a bit more. The provided plugin requires all paths to be lower case, which defeats our purpose of making sure downstream functions and such end up camel case for client generators.

MikeNeilson:
well, the latest version of Javalin has a simple "caseInsenstiveRoutes" options, but requires java 11.

rma-rripken (approved):

PR #1007, opened 2025-01-10
Feature/354 openid auth

MikeNeilson:
Adds support for initial user creation from a trusted JWT.

NOTES:
- Requires "execute on cwms_upass" for the web_user role.
- Still some work to be done on mapping claims, some will be ignored but the test keycloak just doesn't have several configured at the moment.

Uses the "principle_name" (note: spelled incorrectly in database) field to store the combination of JWT issuer and subject claim . The subject is a UUID and so the combination of issue+subject is always unique. This was done after discovering that the one of the test infrastructure OIDC providers doesn't include the CAC EDIPI anywhere and it could not be used for user lookup.

Using a full identity provider principal is better anyways, so may as well start here.

Additionally updates the docker-compose to require no external setup. This should make testing and development far easier for newcomers to the project.


MikeNeilson:
Still thinking of ways to automate testing this without going overboard.

compose_files/keycloak/realm.json
```diff
@@ -26,7 +26,7 @@
   "oauth2DeviceCodeLifespan": 600,
   "oauth2DevicePollingInterval": 5,
   "enabled": true,
-  "sslRequired": "external",
+  "sslRequired": "none",
```
rma-rripken, line 29: what does sslRequired: none do? why wouldn't you want ssl? aren't the urls https ?

Dockerfile
```diff
@@ -44,7 +44,7 @@ ENV cwms.dataapi.access.providers "KeyAccessManager,OpenID"
 ENV cwms.dataapi.access.openid.wellKnownUrl "https://identity-test.cwbi.us/auth/realms/cwbi/.well-known/openid-configuration"
 ENV cwms.dataapi.access.openid.issuer "https://identity-test.cwbi.us/auth/realms/cwbi"
 ENV cwms.dataapi.access.openid.timeout "604800"
-ENV cwms.dataapi.access.openid.altAuthUrl "https://identityc-test.cwbi.us/auth/realms/cwbi"
+#ENV cwms.dataapi.access.openid.altAuthUrl "https://identityc-test.cwbi.us/auth/realms/cwbi"
```
rma-rripken, line 47: I see its commented out but was identityc a typo?  why the c char?

MikeNeilson: "CAC". without the C you can get to the .well-known, with the c you get the mtls version.

Not that this mechanism actually worked, unfortunately.

compose_files/keycloak/realm.json
MikeNeilson: For the automated tests ssl isn't needed and just adds complexity. This may not be required with the change to start-dev but I want to to eliminate all posibilities of an HTTPs redirect for testing.

cwms-data-api/src/main/java/cwms/cda/security/OpenIDAccessManager.java
```diff
@@ -42,6 +43,8 @@
 public class OpenIDAccessManager extends CdaAccessManager {
     private static final FluentLogger log = FluentLogger.forEnclosingClass();
     public static final String AUTHORIZATION = "Authorization";
+    public static final String CREATE_USERS_KEY = "cda.openid.create_users";
+    private static final boolean createUsers = true;// Boolean.parseBoolean(System.getProperty(CREATE_USERS_KEY,"false"));
```
rma-rripken: I'd say either leave the property in but default to true or removed the commented out property bit

cwms-data-api/src/main/java/cwms/cda/security/OpenIDAccessManager.java
```diff
@@ -69,15 +72,26 @@ public void manage(Handler handler, @NotNull Context ctx, @NotNull Set<RouteRole
     private DataApiPrincipal getUserFromToken(Context ctx) throws CwmsAuthException {
         try {
             Jws<Claims> token = jwtParser.parseClaimsJws(getToken(ctx));
-            String username = token.getBody().get("preferred_username",String.class);
+            Claims claims = token.getBody();
+            final String issuer = claims.getIssuer();
+            final String subject = claims.getSubject();
+            final String username = issuer + "::" + subject;
```
rma-rripken: is there any risk of issuer or subject already containing "::" ?  would it matter if they did?

cwms-data-api/src/main/java/cwms/cda/security/OpenIDAccessManager.java
```diff
@@ -69,15 +72,26 @@ public void manage(Handler handler, @NotNull Context ctx, @NotNull Set<RouteRole
...
+            if (principal.isPresent()) {
+                return principal.get();
+            } else if (createUsers) {
+                final String preferredUserName = claims.get("preferred_username", String.class);
+                final String givenName = claims.get("given_name", String.class);
+                final String email = claims.get("email", String.class);
+                return dao.createUser(preferredUserName,username,givenName, email);
+            } else {
+                throw new CwmsAuthException("Not Authorized",HttpServletResponse.SC_UNAUTHORIZED);
+            }
```
rma-rripken: some spaces after commas here would make this look better.   Those string literals should get moved into constants

cwms-data-api/src/test/java/cwms/cda/api/auth/OpenIdConnectTestIT.java
```diff
@@ -0,0 +1,61 @@
...
+
+        given()
+            .log().ifValidationFails(LogDetail.ALL, true)
+            .header("Authorization", "Bearer " + token.get())
+            .queryParam("name-mask","asdf")
+        .when()
+            .get("/properties")
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+            // 403 here means the user was created, but has no privileges as the user was just created.
```
rma-rripken: the comment here was probably copy/paste - right?

cwms-data-api/src/test/java/fixtures/KeyCloakExtension.java
```diff
@@ -0,0 +1,155 @@
...
+                given()
+                    .log().ifValidationFails(LogDetail.ALL,true)
+                    .contentType(ContentType.URLENC)
+                    .formParam("client_id","cwms")
+                    .formParam("grant_type","password")
+                    .formParam("client_secret","")
+                    .formParam("scope","openid profile email")
+                    .formParam("username",username)
+                    .formParam("password",password)
+                    .formParam("response_type","token")
```
rma-rripken, line 143: The keys here could probably be constants too

rma-rripken:
I'm not in a position to test this on my own.  The changes look ok but there are several things here that I don't entirely understand.   Not in a "I'm confused about it" sense but in the sense that I don't know any of the details so I don't have an opinion.

rma-rripken (dismissed):

cwms-data-api/src/main/java/cwms/cda/security/OpenIDAccessManager.java
MikeNeilson: I don't think so. :: wouldn't be valid in a URL, and subject is either a UUID or some form of username. It's possible it could be an issue in the future, but it wouldn't be difficult to change later. Long term I want them to be their own columns, but we're kinda working on sort timelines so this particular form is more of a stop gap.

cwms-data-api/src/test/java/fixtures/KeyCloakExtension.java
MikeNeilson: Probably, but at this time this is the only place they're used. If we end up duplicating this, say for testing AuthCode + PKCE, that's probably the better time to do that. (And yes, I'm kinda just being a bit lazy.)

cwms-data-api/src/test/java/cwms/cda/api/auth/OpenIdConnectTestIT.java
MikeNeilson: No, it was explaining why 403 instead of 401. Updated the language to make that comment on assertion more clear... hopefully.

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
```diff
@@ -144,6 +145,7 @@ public void beforeAll(ExtensionContext context) throws Exception {
 
             // OIDC properties
             System.setProperty("cwms.dataapi.access.providers","KeyAccessManager,OpenID,CwmsAccessManager");
+            System.getProperty(OpenIDAccessManager.CREATE_USERS_KEY,":true");
             System.setProperty(OpenIDAccessManagerProvider.WELL_KNOWN_PROPERTY,KeyCloakExtension.getOidcWellKnown());
```
rma-rripken: getProperty line here doesn't look right.  Is this supposed to be setProperty?  Is the value supposed to be ":true"?

cwms-data-api/src/test/java/cwms/cda/api/auth/OpenIdConnectTestIT.java
rma-rripken: The test right above this one has the same comment but that test checks for FORBIDDEN (403) and not UNAUTHORIZED (401) .  This test looks for OK so seemed weird to be explaining what a 403 means when its not supposed to happen here.

MikeNeilson: You are correct, It would help if I paid attention to exactly which line you were commenting on and not just the text itself.

rma-rripken (changes requested):
The random getProperty call still needs to be fixed - right?

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
MikeNeilson: Yep, now it's fixed. also wasn't supposed to have the `:` in there either.

docker-compose.yml
```diff
@@ -1,55 +1,34 @@
...
 volumes:
   oracle_data_1:
   auth_data:
 services:
   db:
-    # if using gvenzl/oracle-xe you'll need to adjust the environment below
-    image: ${ORACLE_DOCKER_IMAGE}
-    volumes:
-      - oracle_data_1:/opt/oracle/oradata
+    image: registry-public.hecdev.net/cwms/database-ready-ora-23.5:latest-dev
```
rma-rripken, line 6: You are slipping in the ready database change too?  I want that change but it seems like it should have been its own pr.  oh well.  move fast and break things I guess.

rma-rripken (approved):

docker-compose.yml
MikeNeilson: I thought about that, but it was required to make developing this fast (having killed by existing database) so I felt it made sense to just do.

PR #1016, opened 2025-01-27
Removed TS Code from TimeSeriesGroup getOne method

### LINKED ISSUE #978: TS_code exposed
I was able to get the ts_code with this endpoint. 

https://wm.mvs.ds.usace.army.mil:8243/mvs-data/timeseries/group/Stage?office=MVS&category-id=Commerce-Mississippi

```
{
  "office-id": "MVS",
  "id": "Stage",
  "time-series-category": {
    "office-id": "MVS",
    "id": "Commerce-Mississippi"
  },
  "assigned-time-series": [
    {
      "office-id": "MVS",
      "timeseries-id": "Commerce-Mississippi.Stage.Inst.30Minutes.0.lrgsShef-rev",
      "ts-code": 4182018
    }
  ]
}
```

_Originally posted by @inguyen314 in https://github.com/USACE/cwms-data-api/issues/964#issuecomment-2540513096_
            

_code fields should not be exposed to the end user.

zack-rma:
Fixes #978.

Adds an ignore annotation to the TS code field of the AssignedTimeSeries data encompassed by the TimeSeriesGroup DTO.

Now, the TS code is not returned as part of the response when serialized. Adds check to existing tests.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
@@ -199,7 +199,7 @@ public class TimeSeriesGroupDao extends JooqDao<TimeSeriesGroup> {
             if (attrBD != null) {
                 attr = attrBD.intValue();
             }
-            retval = new AssignedTimeSeries(officeId, timeseriesId, tsCode, aliasId, refTsId, attr);
+            retval = new AssignedTimeSeries(officeId, timeseriesId, aliasId, refTsId, attr);
         }
 
         return retval;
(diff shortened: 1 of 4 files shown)


adamkorynta:
I don't see a reason to include the ts code in the DTO at all.

MikeNeilson:
> I don't see a reason to include the ts code in the DTO at all.

I'd have to check but I believe it's used internally for lookup of the actual data. Maybe not by the this endpoint, but I think it was used. Sorry, vaguely remember some conversations about it. Though not difficult to just remove and test, or lookup usages.

MikeNeilson (approved):

PR #1024, opened 2025-02-13
Bugfix/939 clob getall office filter

rma-rripken:
I'm redoing pr #937 to address #939.  

develop had moved significantly since #937 had been started that it seemed cleaner to just grab the handful of changes.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
@@ -44,49 +44,46 @@ public class ClobDao extends JooqDao<Clob> {
 
     @Override
     public Optional<Clob> getByUniqueName(String uniqueName, String office) {
-        AV_CLOB ac = AV_CLOB.AV_CLOB;
-        AV_OFFICE ao = AV_OFFICE.AV_OFFICE;
+        AV_CLOB vClob = AV_CLOB.AV_CLOB;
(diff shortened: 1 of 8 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
```diff
@@ -269,11 +240,7 @@ public void getClob(String clobId, String officeId, ClobConsumer clobConsumer) {
                 try (ResultSet resultSet = preparedStatement.executeQuery()) {
                     if (resultSet.next()) {
                         java.sql.Clob clob = resultSet.getClob("VALUE");
-                        if (clob != null) {
-                            clobConsumer.accept(clob);
-                        } else {
-                            clobConsumer.accept(null);
-                        }
+                        clobConsumer.accept(clob);
```
MikeNeilson, line 243: are we sure null here isn't valid?

rma-rripken: I think it is valid.  Its just that when clob == null the variable clob already has null in it so clobConsumer.accept(clob) is the right line in either case.   I didn't spot this myself, the IDE suggested it.  More of a - 'Huh - its right. ok.'

MikeNeilson: oh right we're just passing null anyways.

MikeNeilson (approved):

PR #1040, opened 2025-03-14
Update release actions to deploy to shared dev environment.

MikeNeilson:
Set container to push to dev on the nightly build.

This will keep the container up-to-date for security and catch new changes on a reasonable interval.

Other option is to deploy on tag based versions; however, I believe that more appropriate for the test and prod environments, with prod definitely needing more than just a tag to approve.

Also updated some job names that were copy pasted from other projects.

Unfortunately due to how many secrets and package systems this all uses, there's not a good way to test it even with nektos/act.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,55 @@
+---
+name: Deploy
+on:
+  workflow_call:
+    inputs:
+      api_image:
+        type: string
+        required: true
+      region:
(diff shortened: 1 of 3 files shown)


adamscarberry (approved):
As far as I can tell, this should work.

MikeNeilson:
> As far as I can tell, this should work.

The joy of github actions.

PR #1055, opened 2025-03-21
Move schema migration to release action to get better feedback.

MikeNeilson:


MikeNeilson:
Okay, this time I think it'll work. trying to manually type out those local action references always has me adding an extra / or something.

krowvin (approved):

PR #1063, opened 2025-03-24
Create app user, or just update password.

MikeNeilson:
@adamscarberry if you're curious.

Definitely easier and more structured with flyway and the afterMigrate scripts. But same affect. 
Technically this would also allow automated updates of the app user password; at least if the particular container could be triggered in the appropriate way.


MikeNeilson:
This was tested locally and expected users were created and configured.

PR #1082, opened 2025-04-22
Feature/rate function

### LINKED ISSUE #641: Create a Rate Endpoint
# Rate Endpoint - GET
*Creation of an endpoint that allows you to rate various curves in Oracle and returns the result(s)*

This is more efficient than pulling that curves/specs and _attempting_ to rate on the client in various different methods than what is already established in the procedures/functions in Oracle. 

Especially given some ratings (EXSA / expanded) have thousands of points and would create IO/Bandwidth/time issues on clients.

## Starting Ideas :
Gate Settings -> Flow
Stage -> Regulated Flow

## District Endpoints - i.e. Custom
As a means of attempting a district specific endpoint for 1 off's I can create this in a 
`/cwms-data/swt/rate` endpoint that could then be bubbled up to the national once complete. 

Given this *is* something we need as a national endpoint anyways. 
*see* #393 

@adamkorynta  @rma-rripken 
I will report here as I run into roadblocks. But do you have any references you can email/share here as to how you/REGI/CMWS expects these payloads? Ideally I will keep it data in - data out from the procedures. But if you have specifics in mind i'm all ears. Otherwise I will start on this and we can build the payloads up/down based on needs.

adamkorynta:
Fixes #641


cwms-data-api/src/main/java/cwms/cda/data/dto/rating/RateInput.java
```diff
@@ -0,0 +1,107 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import cwms.cda.data.dto.CwmsDTOBase;
+import cwms.cda.formatters.Formats;
+import cwms.cda.formatters.annotations.FormattableWith;
+import cwms.cda.formatters.json.JsonV1;
+import io.swagger.v3.oas.annotations.media.Schema;
+import java.time.Instant;
+import java.util.Optional;
+
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class, aliases = {Formats.DEFAULT, Formats.JSON})
```
MikeNeilson, line 41: FYI, I know this hasn't been official merged in yet, but I don't think anyone had a major issue with the following as the general rule for "data type versioning": https://github.com/USACE/cwms-data-api/pull/959/files#diff-9f37666d47127d069db9e7ef0915049a7d4f4cfafe03f30ddc015ef3c2f3f1ddR1

the aliases are still correct with the above, we've just been overusing JSONV1 vs JSONV2

MikeNeilson:
I don't see rate limiting at this time. I assume that will be done before the PR is marked as ready for review?

cwms-data-api/src/main/java/cwms/cda/api/rating/RatingTemplateController.java
```diff
@@ -1,7 +1,7 @@
 /*
  * MIT License
  *
- * Copyright (c) 2023 Hydrologic Engineering Center
+ * Copyright (c) 2025 Hydrologic Engineering Center
```
MikeNeilson, line 4: in the future, we probably don't need to bother updating the year if we're just tweaking imports and packages.

cwms-data-api/src/main/java/cwms/cda/data/dao/RateDao.java
```diff
@@ -0,0 +1,186 @@
...
+    private <R> R connectionResult(ConnectionCallable<R> callable) {
+        try {
+            return connectionResult(dsl, callable);
+        } catch(DataAccessException ex) {
+            throw handleRateDbError(ex);
+        }
+    }
+
+    static RuntimeException handleRateDbError(DataAccessException ex) {
+        RuntimeException retval = ex;
```
MikeNeilson, line 170: See Ryan's message on another PR: https://github.com/USACE/cwms-data-api/pull/1093#issuecomment-2873710523

This may be appropriate as-is, but possible it should be organized with the other JOOQ/Database derived exception handling.

MikeNeilson (changes requested):
My issue with the content-type (see the RateInput.java comment) needs to be addressed.


Otherwise looks reasonable. I do think the districts wanted this to be public. But definitely easier to add rate limiting after it's all in-place than to deal with denial of services issues while also trying to work out any quirks that are found.

cwms-data-api/src/main/java/cwms/cda/data/dao/RateDao.java
adamkorynta: I wanted to specifically target the exception codes thrown by the rate functions in the cwms_rating package, I don't necessarily want to globally accept database error messages from other packages. I should add a comment here explaining why it isn't part of the base class exception handling though.

cwms-data-api/src/main/java/cwms/cda/data/dto/rating/RateInput.java
MikeNeilson: I'll let this go for now. Given that's how we've been doing it and we haven't formally accepted the new documentation of the "right way."

cwms-data-api/src/main/java/cwms/cda/data/dao/RateDao.java
MikeNeilson: Yeah, that makes sense.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dto/rating/RateInput.java
adamkorynta: works for me

cwms-data-api/src/main/java/cwms/cda/api/rating/RatingTemplateController.java
adamkorynta: yeah, my IDE is setup to update copyright on modified files as part of pre-commit

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -320,6 +325,7 @@ public void init() {
         PolicyFactory sanitizer = new HtmlPolicyBuilder().disallowElements("<script>").toFactory();
         APP_CONTEXT = this.getServletContext().getContextPath();
         javalin = Javalin.createStandalone(config -> {
+                    config.enableDevLogging();
```
rma-rripken: Did you mean to include Dev Logging?

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -402,6 +408,11 @@ public void init() {
                     logger.atInfo().withCause(e).log(re.toString());
                     ctx.status(HttpServletResponse.SC_NOT_FOUND).json(re);
                 })
+                .exception(RateException.class, (e, ctx) -> {
```
rma-rripken, line 410: There is no parent class that RateException can extend?  Otherwise what keeps ApiServlet from filling up with implementation specific exception types?  I would think that if we really can't think of anything more generic at the very least we should be able to make a class for Controller exceptions and another for DaoExceptions depending on where the issue is.  I don't see ApiServlet knowing how to handle RateException in any helpful or meaningfully different way than other exceptions.

MikeNeilson: Though I didn't see it before. I completely agree with Ryan on this one.

Though based on the suggestion of DAO vs Controller exception, would it be better to accept for now and create a follow up issue to review the exception handling.

This particular block has already been getting a bit large and it's possible that just a few marker interfaces may help sort that out.

rma-rripken (dismissed):

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
adamkorynta: I agree, I'd like to move that refactor into a subsequent effort as my ideas for implementing something standard are either insufficient or too large of a change for this PR.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
adamkorynta: I most definitely did not! Thank you for catching

MikeNeilson (approved):

PR #1086, opened 2025-04-28
Add support for virtual location levels

### LINKED ISSUE #469: Add support for virtual location levels
the levels endpoint should be updated to support storage and retrieval of virtual location levels

zack-rma:
Fixes #469

Adds support for virtual location levels to the levels endpoint. Integration testing included.


MikeNeilson:
@zack-rma merge develop in for the nicer test report.

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
```diff
@@ -733,6 +749,579 @@ void testRetrievalInvalidLevelName()
...
+            .extract();
+
+        assertThat(response.path("constituents.size()"), is(2));
+        assertThat(response.path("constituents[0].name"), equalTo(virtualLoc +  ".Elev.Ave.1Day.Regulating"));
+        assertThat(response.path("constituents[0].type"), equalTo("LOCATION_LEVEL"));
+        assertThat(response.path("constituents[0].attribute-id"), equalTo("Elev"));
+        assertThat(response.path("constituents[1].name"), equalTo(existingSpec));
+        assertThat(response.path("constituents[1].type"), equalTo("RATING"));
+
+        CwmsDataApiSetupCallback.getDatabaseLink().connection(c -> {
```
MikeNeilson: These won't get called on failure which can hinder repeatability of tests.

If the location is registered with create location the cascade delete should take care of the ratings so these shouldn't be required.

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
```diff
@@ -733,6 +749,579 @@ void testRetrievalInvalidLevelName()
...
+        ZonedDateTime time = ZonedDateTime.of(2023, 6, 1, 0, 0, 0, 0, ZoneId.of("America/Los_Angeles"));
+        String levelJson = readResourceFile("cwms/cda/api/virtuallevels/virtual_level_1.json");
+        String ratingXml = readResourceFile("cwms/cda/api/Zanesville_Stage_Flow_COE_Production.xml");
+        ratingXml = ratingXml.replaceAll("Zanesville", existingLoc);
+        RatingSetContainer container = RatingSetContainerXmlFactory.ratingSetContainerFromXml(ratingXml);
+        RatingSpecContainer specContainer = container.ratingSpecContainer;
+        String specXml = RatingSpecXmlFactory.toXml(specContainer, "", 0, true);
+        String setXml = RatingContainerXmlFactory.toXml(container, "", 0, true, false);
+
+        CwmsDataApiSetupCallback.getDatabaseLink().connection(c -> {
```
MikeNeilson: This large block is repeated in a near identical way. Some sort of helper function (similar to createLocation) should be created.

MikeNeilson (changes requested):
Generally looks reasonable. Only issue is with the level of duplication within the test code.

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
zack-rma: Removed ratings cleanup methods

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
zack-rma: Added helper methods to reduce code duplication

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
```diff
@@ -808,4 +1266,63 @@ enum GetAllTestNewAliases
             _expectedContentType = expectedContentType;
         }
     }
+
+    private String createRatingSpec(String locationName) throws Exception
```
MikeNeilson: Preferably these would've gone in the DataApiTest class, but this at least makes it easier for the next person to yank up and make more generic.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -360,14 +378,24 @@ String.class, null, metrics, name(LevelsController.class.getName(),
...
+
         try (final Timer.Context ignored = markAndTime(GET_ONE)) {
             DSLContext dsl = getDslContext(ctx);
             ZonedDateTime unmarshalledDateTime = DateUtils.parseUserDate(dateString, timezone);
 
             LocationLevelsDao levelsDao = getLevelsDao(dsl);
-            LocationLevel locationLevel = levelsDao.retrieveLocationLevel(levelId,
-                    units, unmarshalledDateTime, office);
-            ctx.json(locationLevel);
+            if (virtual) {
```
adamkorynta: This isn't right, each effective date can return a level type of timeseries, constant, seasonal, or virtual. You can't return a virtual level for a timeseries effective date, or vice versa.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -240,12 +243,16 @@ public void delete(@NotNull Context ctx, @NotNull String levelId) {
                         + "request you are. This is an opaque value, and can be obtained from "
                         + "the 'next-page' value in the response."),
                 @OpenApiParam(name = PAGE_SIZE, type = Integer.class, description = "How "
-                        + "many entries per page returned. Default " + DEFAULT_PAGE_SIZE + ".")},
+                        + "many entries per page returned. Default " + DEFAULT_PAGE_SIZE + "."),
+                @OpenApiParam(name = VIRTUAL, type = Boolean.class, description = "Should virtual levels be returned." +
```
adamkorynta: This is a filter? Why not filter out constant, timeseries, and seasonal as well?

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -240,12 +243,16 @@ public void delete(@NotNull Context ctx, @NotNull String levelId) {
...
+                        " Default: False.")
+            },
             responses = {
                 @OpenApiResponse(status = STATUS_200, content = {
                     @OpenApiContent(type = Formats.JSON),
                     @OpenApiContent(type = ""),
-                    @OpenApiContent(from = LocationLevels.class, type = Formats.JSONV2)
+                    @OpenApiContent(from = LocationLevels.class, type = Formats.JSONV2),
+                    @OpenApiContent(from = VirtualLocationLevel.class, type = Formats.JSONV2),
```
adamkorynta: This will blow up Javalin (in our version at least) as you can't have classes with the same type.

cwms-data-api/src/main/java/cwms/cda/data/dto/VirtualLocationLevel.java
```diff
@@ -0,0 +1,432 @@
...
+import rma.util.RMAConst;
+
+@JsonRootName("VirtualLocationLevel")
+@JsonDeserialize(builder = VirtualLocationLevel.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class)
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
+public final class VirtualLocationLevel extends LocationLevel {
```
adamkorynta: This isn't right, LocationLevel has timeseries, seasonal, timeseriesid parameter that aren't appropriate for virtual. The LocationLevel class should get subclassed into 4 types.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -122,7 +102,9 @@ private Timer.Context markAndTime(String subject) {
             requestBody = @OpenApiRequestBody(
                     content = {
                         @OpenApiContent(from = LocationLevel.class, type = Formats.JSON),
-                        @OpenApiContent(from = LocationLevel.class, type = Formats.XML)
+                        @OpenApiContent(from = LocationLevel.class, type = Formats.XML),
+                        @OpenApiContent(from = VirtualLocationLevel.class, type = Formats.JSON),
```
adamkorynta: This doesn't work and blows up the swagger page

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dto/VirtualLocationLevel.java
```diff
@@ -0,0 +1,432 @@
...
+import hec.data.level.ISeasonalValues;
+import rma.util.RMAConst;
+
+@JsonRootName("VirtualLocationLevel")
+@JsonDeserialize(builder = VirtualLocationLevel.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class)
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
```
adamkorynta: XML does not needed to be supported (and you didn't test either)

zack-rma: LevelControllerTest tests XML deserialization and has been updated for the new classes of LocationLevel. Should I disable those tests and remove the XML support instead?

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -0,0 +1,434 @@
...
+
+@JsonRootName("LocationLevel")
+@JsonDeserialize(builder = LocationLevel.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class)
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class LocationLevel extends CwmsDTO
```
MikeNeilson: @adamkorynta considering this new class hierarchy can we ever meaningfully have a plain "LocationLevel"? if not seems like this should be marked abstract.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Removed extra type, verified swagger page loads without issue.

cwms-data-api/src/main/java/cwms/cda/data/dto/VirtualLocationLevel.java
zack-rma: Subclassed into 4 types as suggested.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Removed extra type.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: It was, but I've removed it as part of the subclassing. It should no longer be necessary.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevels.java
```diff
@@ -15,7 +42,8 @@
 
 @JsonRootName("location-levels")
 @FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
-public class LocationLevels extends CwmsDTOPaginated {
+public class LocationLevels extends CwmsDTOPaginated
```
MikeNeilson: we do keep the curly on the same line in this project.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Updated per subclassing change.

cwms-data-api/src/main/java/cwms/cda/data/dto/VirtualLocationLevel.java
adamkorynta: New data types don't need to support XML, we should leave LocatoinLevel, but not add support to VirtualLocationLevel

cwms-data-api/src/main/java/cwms/cda/data/dto/VirtualLocationLevel.java
adamkorynta: How is backwards compatibility effected?

zack-rma: The serialization/deserialization to and from the subclasses all happens in the controller/DAO level, so existing input data is still interpreted correctly without alterations. Some fields may not be present (e.g. constant-value for a seasonal location level) in the output and would need to be ignored if they are required in any downstream usages. Any direct usages of the LocationLevel class will need to be updated to use the specific subclass that corresponds with the input data.

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -133,15 +115,33 @@ private Timer.Context markAndTime(String subject) {
...
+            throw new IllegalArgumentException("Unable to parse the request body", e);
+        }
+    }
+
+    private LocationLevel deserializeLocationLevel(Context ctx) throws IOException {
+        String formatHeader = ctx.req.getContentType();
+        ContentType contentType = Formats.parseHeader(formatHeader, LocationLevel.class);
+        StringWriter writer = new StringWriter();
+        IOUtils.copy(ctx.bodyAsInputStream(), writer, StandardCharsets.UTF_8);
+        if (writer.toString().contains("constituent")) {
```
adamkorynta: Calling `writer.toString()` on each condition statement as well as within the condition block. Assign to a variable

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
```diff
@@ -0,0 +1,428 @@
...
+import rma.util.RMAConst;
+
+@JsonRootName("SeasonalLocationLevel")
+@JsonDeserialize(builder = SeasonalLocationLevel.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class)
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
+public class SeasonalLocationLevel extends LocationLevel
```
adamkorynta: final

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
```diff
@@ -0,0 +1,278 @@
...
+import hec.data.level.JDomLocationLevelImpl;
+
+@JsonRootName("TimeSeriesLocationLevel")
+@JsonDeserialize(builder = TimeSeriesLocationLevel.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class)
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
+public class TimeSeriesLocationLevel extends LocationLevel
```
adamkorynta: final

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
```diff
@@ -0,0 +1,277 @@
...
+@JsonRootName("ConstantLocationLevel")
+@JsonDeserialize(builder = ConstantLocationLevel.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class)
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
+public final class ConstantLocationLevel extends LocationLevel
+{
+	@Schema(description = "Single value for this location level. Mutually exclusive with "
```
adamkorynta: Update schema comment

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -0,0 +1,450 @@
...
```
adamkorynta: Should be abstract

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -133,15 +115,33 @@ private Timer.Context markAndTime(String subject) {
...
+            return Formats.parseContent(contentType, writer.toString(), VirtualLocationLevel.class);
+        } else if (writer.toString().contains("seasonal-timeseries-id")) {
+            return Formats.parseContent(contentType, writer.toString(), TimeSeriesLocationLevel.class);
+        } else if (writer.toString().contains("seasonal-values")) {
+            return Formats.parseContent(contentType, writer.toString(), SeasonalLocationLevel.class);
+        } else if (writer.toString().contains("constant-value")) {
+            return Formats.parseContent(contentType, writer.toString(), ConstantLocationLevel.class);
+        } else {
+            return Formats.parseContent(contentType, writer.toString(), LocationLevel.class);
```
adamkorynta: This case should throw an IllegalArgumentException or format exception

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -168,13 +179,63 @@ public LocationLevels getLocationLevels(String cursor, int pageSize,
...
+                .where(whereCondition)
+                .orderBy(DSL.upper(virtView.OFFICE_ID), DSL.upper(virtView.LOCATION_LEVEL_ID),
+                        virtView.EFFECTIVE_DATE_UTC
+                )
+                .offset(offset)
+                .limit(pageSize);
+
+        final SelectLimitPercentAfterOffsetStep<Record> virtQueryFinal = query;
+
+        logger.info(() -> "getLocationLevels query: " + virtQueryFinal.getSQL(ParamType.INLINED));
```
adamkorynta: Shouldn't log at info

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -168,13 +179,63 @@ public LocationLevels getLocationLevels(String cursor, int pageSize,
...
+                    Timestamp.from(endZdt.toInstant())));
+        }
+
+        query = dsl.selectDistinct(getAddVirtualFields())
+                .from(virtView)
+                .where(whereCondition)
+                .orderBy(DSL.upper(virtView.OFFICE_ID), DSL.upper(virtView.LOCATION_LEVEL_ID),
+                        virtView.EFFECTIVE_DATE_UTC
+                )
+                .offset(offset)
```
adamkorynta, line 222: This doesn't seem right, you're looking only at virtual levels, but the offset/limit are for all levels

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
adamkorynta, line 1: For future developer sanity, break each store/retrieve method down into 4 separate method calls, each handling the different level types separately. Trying to read through a bunch of if/else is very obtuse

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -401,8 +591,29 @@ private Collection<TableField<?,?>> getAddSeasonalValueFields() {
         return retval;
     }
 
+    // These are all the fields that we need to pull out of jOOQ record for addVirtualValue
+    private Collection<TableField<?,?>> getAddVirtualFields() {
```
adamkorynta: Why make a method? At this point might as well have a statically initialized constant

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
```diff
@@ -0,0 +1,277 @@
...
+			IParameterTypedValue constantLevel = copyFrom.getConstantLevel();
+			if (constantLevel != null) {
+				withConstantValue(constantLevel.getSiParameterUnitsValue());
+			}
+			buildPropertyFunctions();
+		}
+
+
+		@JsonIgnore
+		private void buildPropertyFunctions() {
```
adamkorynta: What is this for?

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
```diff
@@ -0,0 +1,277 @@
...
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class)
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
+public final class ConstantLocationLevel extends LocationLevel
+{
+	@Schema(description = "Single value for this location level. Mutually exclusive with "
+			+ "seasonableTimeSeriesId and seasonValues.")
+
+	private final Double constantValue;
+
+	ConstantLocationLevel(ConstantLocationLevel.Builder builder) {
```
adamkorynta: private

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
```diff
@@ -0,0 +1,277 @@
...
+		public ConstantLocationLevel.Builder withConstantValue(Double value) {
+			if (value != null && RMAConst.isUndefinedValue(value)) {
+				value = null;
+			}
+			this.constantValue = value;
+			return this;
+		}
+
+		@Override
+		public ConstantLocationLevel.Builder withSpecifiedLevelId(String specifiedLevelId) {
```
adamkorynta: All of these common builder methods should be in the base abstract class

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
```diff
@@ -0,0 +1,428 @@
...
+
+	public Integer getIntervalMinutes() {
+		return intervalMinutes;
+	}
+
+	@JsonPOJOBuilder
+	@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+	@JsonIgnoreProperties(ignoreUnknown = true)
+	public static class Builder extends LocationLevel.Builder {
+		List<SeasonalValueBean> seasonalValues;
```
adamkorynta: Why are these package private instead of just private?

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
```diff
@@ -0,0 +1,428 @@
...
+	}
+
+	public Integer getIntervalMinutes() {
+		return intervalMinutes;
+	}
+
+	@JsonPOJOBuilder
+	@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+	@JsonIgnoreProperties(ignoreUnknown = true)
+	public static class Builder extends LocationLevel.Builder {
```
adamkorynta: final

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
```diff
@@ -0,0 +1,428 @@
...
+		public SeasonalLocationLevel.Builder withIntervalMinutes(Integer minutes) {
+			if (minutes != null && RMAConst.isUndefinedValue(minutes)) {
+				minutes = null;
+			}
+			this.intervalMinutes = minutes;
+			return this;
+		}
+
+		@Override
+		public SeasonalLocationLevel.Builder withSpecifiedLevelId(String specifiedLevelId) {
```
adamkorynta: Duplicates that should be in the base abstract class

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
```diff
@@ -0,0 +1,428 @@
...
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+@FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class)
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
+public class SeasonalLocationLevel extends LocationLevel
+{
+	@Schema(description = "The start point of provided seasonal values")
+	@JsonFormat(shape = JsonFormat.Shape.STRING)
+
+	private final ZonedDateTime intervalOrigin;
```
adamkorynta, line 73: I believe all these parameters required? make sure to check the schema

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
```diff
@@ -0,0 +1,278 @@
...
+{
+	@JsonProperty(required = true)
+	@Schema(description = "Timeseries ID (e.g. from the times series catalog) to use as the "
+			+ "location level. Mutually exclusive with seasonalValues and "
+			+ "siParameterUnitsConstantValue")
+
+	private final String seasonalTimeSeriesId;
+
+	TimeSeriesLocationLevel(TimeSeriesLocationLevel.Builder builder) {
```
adamkorynta: private

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	public String getSeasonalTimeSeriesId() {
+		return seasonalTimeSeriesId;
+	}
+
+	@JsonPOJOBuilder
+	@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+	@JsonIgnoreProperties(ignoreUnknown = true)
+	public static class Builder extends LocationLevel.Builder {
```
adamkorynta: final

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
```diff
@@ -0,0 +1,278 @@
...
+		String seasonalTimeSeriesId;
```
adamkorynta: fix scoping

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
```diff
@@ -0,0 +1,278 @@
...
+			return this;
+		}
+
+		public TimeSeriesLocationLevel.Builder withSeasonalTimeSeriesId(String seasonalTimeSeriesId) {
+			this.seasonalTimeSeriesId = seasonalTimeSeriesId;
+			return this;
+		}
+
+		@Override
+		public TimeSeriesLocationLevel.Builder withSpecifiedLevelId(String specifiedLevelId) {
```
adamkorynta: duplicates

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/VirtualLocationLevel.java
```diff
@@ -0,0 +1,352 @@
...
+	private final String constituentConnections;
+
+	private VirtualLocationLevel(Builder builder) {
+		super(builder);
+		this.expirationDate = builder.expirationDate;
+		this.constituents = builder.constituents;
+		this.constituentConnections = builder.constituentConnections;
+	}
+
+	public ZonedDateTime getExpirationDate()
```
adamkorynta: expiration date is common to all level types

cwms-data-api/src/test/resources/cwms/cda/api/virtuallevels/virtual_level_3.json
```diff
@@ -0,0 +1,32 @@
...
+      "attribute-value": 78.99
+    },
+    {
+      "abbr": "R1",
+      "type": "RATING",
+      "name": "LevelsControllerTestIT.Stage;Flow.COE.Production"
+    }
+  ],
+  "constituent-connections": "L1=R1I1",
+  "seasonal-values" : [
```
adamkorynta: seasonal values on a virtual level?

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
```diff
@@ -733,6 +751,447 @@ void testRetrievalInvalidLevelName()
...
+            .body("level-units-id", equalTo("m3"));
+
+        given()
+            .log().ifValidationFails(LogDetail.ALL, true)
+            .accept(Formats.JSONV2)
+            .contentType(Formats.JSONV2)
+            .queryParam(Controllers.OFFICE, OFFICE)
+            .queryParam(UNIT, "SI")
+            .queryParam(EFFECTIVE_DATE, time.toInstant().toString())
+            .queryParam(VIRTUAL, true)
```
adamkorynta: This is no longer a query parameter right?

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/VirtualLocationLevel.java
```diff
@@ -0,0 +1,352 @@
...
+			public Builder(@JsonProperty(value = "abbr", required = true) String abbr,
+					@JsonProperty(value = "type", required = true) String type,
+					@JsonProperty(value = "name", required = true) String name)
+			{
+				this.abbr = abbr;
+				this.type = type;
+				this.name = name;
+			}
+
+			public Builder withAttributeId(String constituentAttributeId)
```
adamkorynta: Attributes are common to all levels, and this is missing all the other common data

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/VirtualLocationLevel.java
```diff
@@ -0,0 +1,352 @@
...
+		{
+			return new VirtualLocationLevel(this);
+		}
+	}
+
+	@JsonPOJOBuilder
+	@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+	@JsonDeserialize(builder = Constituent.Builder.class)
+	@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+	public static final class Constituent
```
adamkorynta: There's the rating and level types of constituents, there should be two subclasses

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
adamkorynta: and some of the parameters are required.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -0,0 +1,450 @@
...
+            withLevelUnitsId(copyFrom.getLevelUnitsId());
+            withOfficeId(copyFrom.getOfficeId());
+            withParameterId(copyFrom.getParameterId());
+            withParameterTypeId(copyFrom.getParameterTypeId());
+            withSpecifiedLevelId(copyFrom.getSpecifiedLevelId());
+            buildPropertyFunctions();
+        }
+
+        @JsonIgnore
+        private void buildPropertyFunctions()
```
adamkorynta: What is the purpose of the propertyFunctionMap?

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Assigned to local variable

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Now throws a UnsupportedFormatException

cwms-data-api/src/test/resources/cwms/cda/api/virtuallevels/virtual_level_3.json
zack-rma: Removed extraneous test data

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
zack-rma: Finalized

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
zack-rma: Privatized

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
zack-rma: Finalized

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
zack-rma: Made private and added required annotations

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
zack-rma: Updated comment

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
zack-rma: Made private

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
zack-rma: Split into separate methods.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
zack-rma: I believe this may be used by something downstream, but I am not sure precisely how it is used.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/VirtualLocationLevel.java
zack-rma: Moved to base class

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/VirtualLocationLevel.java
zack-rma: Attributes are present in the base class. This attribute data is for the constituents, which are made up of type, name, and abbreviation. Location level constituents also include attribute ID and value per the specification.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/VirtualLocationLevel.java
zack-rma: Converted into two classes

zack-rma (commented):
Some issues resolved, more fixes to come

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -581,20 +648,76 @@ private LocationLevel updatedClearedFields(String body, String format,
...
+                    String propertyName = propertyDefinition.getName();
+                    JsonNode propertyValue = root.findValue(propertyName);
+                    if (propertyValue != null && "".equals(propertyValue.textValue())) {
+                        retVal = new ConstantLocationLevel.Builder(retVal)
+                                .withProperty(propertyName, null).build();
+                    }
                 }
+            } catch (NullPointerException e) {
+                //gets thrown if required field is null
+                throw new IllegalArgumentException(e.getMessage());
```
MikeNeilson: we should include the exception so that the stacktrace can hit the log and people can trace issues, should they happen.

also what is throwing null that we can't check for directly?

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
MikeNeilson: You can look at https://bitbucket.hecdev.net/projects/CWMS/repos/cwms-database-testcontainers/browse/src/main/java/mil/army/usace/hec/test/database/CwmsDatabaseContainer.java for an example of how builders can work within an inheritance hierarchy.

cwms-data-api/src/test/java/cwms/cda/data/dto/LocationLevelTest.java
```diff
@@ -1,84 +1,82 @@
...
+	void test_serialization_formats_Virtual()
 	{
 		ZonedDateTime zdt = ZonedDateTime.parse("2021-06-21T08:00:00-07:00[PST8PDT]");
-		final LocationLevel level = new LocationLevel.Builder("Test", zdt).build();
+		final VirtualLocationLevel level = new VirtualLocationLevel.Builder("Test", zdt).build();
 
 		ContentType contentType = Formats.parseHeader(Formats.JSONV2, LocationLevel.class);
 		String jsonStr = Formats.format(contentType, level);
 
-		// If JSONv2 isn't setup correctly it will serialize the level like:
```
MikeNeilson, line 72: even though these comments are duplicate, they should be retained, and probably even added to the one above.

anyone of these tests should be reasonable in isolation so it is reasonable to see duplication we would otherwise avoid in normal code.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
zack-rma: Finalized

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
zack-rma: Updated internal validation, the interval months and interval minutes are mutually exclusive and one is required.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: The only usage I can find is using it to set empty string values to null for an updated location level. It appears to allow for updating the values of a location level once it has been built.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Made abstract

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Made abstract

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Updated throw to include the exception. 

Looking into the NPE catch now.

cwms-data-api/src/test/java/cwms/cda/data/dto/LocationLevelTest.java
zack-rma: Added comments to other similar methods.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/TimeSeriesLocationLevel.java
zack-rma: Removed duplicated methods

cwms-data-api/src/main/java/cwms/cda/data/dto/VirtualLocationLevel.java
zack-rma: Removed support for XML from VirtualLocationLevel

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
zack-rma: Swapped to logging at config

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
zack-rma: Changed to statically initialized constant

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
zack-rma: Made private

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
zack-rma: Made final

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
zack-rma: Correct, I've removed it from the tests

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -430,44 +432,55 @@ public void update(@NotNull Context ctx, @NotNull String oldLevelId) {
                 //only store (update) if level does exist
                 LocationLevel updatedLocationLevel = getUpdatedLocationLevel(existingLevelLevel,
                     levelFromBody);
-                updatedLocationLevel = new LocationLevel.Builder(updatedLocationLevel)
-                    .withLevelDate(unmarshalledDateTime).build();
-                levelsDao.storeLocationLevel(updatedLocationLevel);
+                if (updatedLocationLevel instanceof SeasonalLocationLevel) {
```
adamkorynta: This factory pattern should live with the DTO's

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -430,44 +432,55 @@ public void update(@NotNull Context ctx, @NotNull String oldLevelId) {
...
+                        .withLevelDate(unmarshalledDateTime);
+                    levelsDao.storeLocationLevel(virtualLevel);
+                } else {
+                    throw new IllegalArgumentException("Unsupported location level type");
+                }
                 ctx.status(HttpServletResponse.SC_OK).json("Updated Location Level");
             }
         } catch (JsonProcessingException ex) {
             throw new FormattingException("Failed to format location level update request", ex);
+        } catch (IOException ex) {
```
adamkorynta, line 439: I don't see why this extra catch was added

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -430,44 +432,55 @@ public void update(@NotNull Context ctx, @NotNull String oldLevelId) {
...
+            throw new IllegalArgumentException("Unable to parse the request body", ex);
         }
     }
 
     private LocationLevel getUpdatedLocationLevel(LocationLevel existingLevel,
```
adamkorynta: This whole method belongs in a level factory within the DTO space

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -553,20 +652,76 @@ private LocationLevel updatedClearedFields(String body, String format,
...
+            }
+            return retVal;
+        } else if (existingLevel instanceof SeasonalLocationLevel) {
+            SeasonalLocationLevel seasonalLevel = (SeasonalLocationLevel) existingLevel;
+            SeasonalLocationLevel retVal = new SeasonalLocationLevel.Builder(seasonalLevel).build();
+            try {
+                for (BeanPropertyDefinition propertyDefinition : properties) {
+                    String propertyName = propertyDefinition.getName();
+                    JsonNode propertyValue = root.findValue(propertyName);
+                    if (propertyValue != null && "".equals(propertyValue.textValue())) {
```
adamkorynta: `propertyValue.textValue().isEmpty()`

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    public String getParameterId()
+    {
+        return parameterId;
+    }
+
+    @JsonPOJOBuilder
+    @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+    public static class Builder {
```
adamkorynta: should be abstract

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -0,0 +1,429 @@
...
+        String attributeDurationId;
+        String attributeComment;
+        String locationId;
+        String officeId;
+        String interpolateString;
+        ZonedDateTime expirationDate;
+        final Map<String, Consumer<Object>> propertyFunctionMap = new HashMap<>();
+
+        @JsonCreator
+        public Builder(@JsonProperty(value = "location-level-id", required = true) String name,
```
adamkorynta: protected

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
adamkorynta: your new class is inheriting the behavior so you'll have to figure that out

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/VirtualLocationLevel.java
```diff
@@ -0,0 +1,260 @@
...
+	@JsonPOJOBuilder
+	@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+	@JsonDeserialize(builder = Constituent.Builder.class)
+	@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
+	@JsonSubTypes({
+			@JsonSubTypes.Type(value = LocationLevelConstituent.class, name = "LOCATION_LEVEL"),
+			@JsonSubTypes.Type(value = Constituent.class, name = "RATING")
+	})
+	@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "type", visible = true)
+	public static class Constituent {
```
adamkorynta: Class should be named `RatingConstituent` to match

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
adamkorynta: Can't the builders handle this behavior?

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -199,12 +199,12 @@ public LocationLevels getLocationLevels(String cursor, int pageSize,
...
 
-        logger.info(() -> "getLocationLevels query: " + queryFinal.getSQL(ParamType.INLINED));
+        logger.fine(() -> "getLocationLevels query: " + queryFinal.getSQL(ParamType.INLINED));
 
         query.stream().forEach(r -> parseLevels(r, builderMap));
 
-        int newPageSize = pageSize - builderMap.size();
-        int newOffset = offset - builderMap.size(); // TODO: check if this is correct, this may cause issues with paging
+        int virtualPageSize = pageSize - builderMap.size();
+        int virtualOffset = offset - builderMap.size(); // TODO: check if this is correct, this may cause issues with paging
```
MikeNeilson: this todo seems fairly important.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
```diff
@@ -574,49 +569,44 @@ private VirtualLocationLevel buildVirtualLocationLevel(LOCATION_LEVEL_T level, S
 
     private ConstantLocationLevel buildConstantLocationLevel(LOCATION_LEVEL_T level, String officeId, String units,
             String locationLevelName, ZonedDateTime realEffectiveDate, Double constantValue) {
-        return new ConstantLocationLevel.Builder(locationLevelName, realEffectiveDate)
+        return ((ConstantLocationLevel.Builder) new ConstantLocationLevel.Builder(locationLevelName, realEffectiveDate)
```
MikeNeilson: This cast seems odd.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
```diff
@@ -423,5 +316,15 @@ protected void validateInternal(CwmsDTOValidator validator) {
...
 		validator.required(getLocationLevelId(), "location-level-id");
 		validator.required(getSeasonalValues(), "seasonal-values");
+		if (getIntervalMonths() == null) {
+			validator.required(getIntervalMinutes(), "interval-minutes");
+		} else if (getIntervalMinutes() == null) {
+			validator.required(getIntervalMonths(), "interval-months");
+		} else {
+			throw new RequiredFieldException("Either interval-minutes or interval-months must be set, but not both");
+		}
+		validator.mutuallyExclusive("Only one of the following can be defined at once for a seasonal location level: "
```
MikeNeilson, line 265: this duplicates the login of the above if/else if/else section.

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
```diff
@@ -155,10 +155,10 @@ void test_retrieve_effective_date() throws Exception {
         String levelId = "level_with_effect.Flow.Ave.1Day.Regulating";
         ZonedDateTime time = ZonedDateTime.of(2023, 6, 1, 0, 0, 0, 0, ZoneId.of("America/Los_Angeles"));
         CwmsDataApiSetupCallback.getDatabaseLink().connection(c -> {
-            LocationLevel level = new ConstantLocationLevel.Builder(levelId, time)
+            LocationLevel level = ((ConstantLocationLevel.Builder) new ConstantLocationLevel.Builder(levelId, time)
```
MikeNeilson: Hmm, the amount of casting you're doing really shouldn't be required.

Was it the compiler or the IDE complaining about this? Because somehow this behaves right for the things we do in testcontainers-cwms.. I'll go back and look at the root location level builder.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
MikeNeilson: Ah, there it is. It should also be parameterized on a Type, with each derived builder passing itself in.

That then becomes the return type of each setter method and you then don't have to deal with all the casting.

MikeNeilson (changes requested):
Some nitpicks, and an change to the builder pattern.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/SeasonalLocationLevel.java
zack-rma: Removed duplicate logic above

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/VirtualLocationLevel.java
zack-rma: Renamed to `RatingConstituent`

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Moved to LocationLevel class

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: There was an update to the `deserializeLocationLevel()` method above that requires copying the data from the context input stream. This can throw an IOException, so a catch was added here

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Moved to LocationLevel class

cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
zack-rma: Fixed

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Abstracted and parameterized.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Fixed

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
zack-rma: Removed unnecessary casts.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Yes, they should be able to do so. This method just seemed to allow for alterations without creating a new object, which somewhat defeats the purpose of the builder. I've removed these methods, as they do not appear to have any usages that I've been able to find.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
zack-rma: Updated query to combine tables for consistent paging results. In order to get the VirtualLocationLevel constituents list a subquery would be needed, along with some more logic that is currently in a PL/SQL method for a single retrieval. Retrieving a single VirtualLocationLevel would include the constituent data, currently.

zack-rma:
@MikeNeilson It seems this PR and https://github.com/USACE/cwms-data-api/pull/1109 are failing from the StreamDAOTestIT failing in the same way. It seems to be inconsistent, as it hasn't been failing on all PRs...

cwms-data-api/src/test/java/cwms/cda/api/LevelsControllerTestIT.java
zack-rma: Removed unnecessary casts

MikeNeilson (dismissed):
My concerns have been address.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationLevelsDaoImpl.java
zack-rma: Resolved in rewrite of retrieval query

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
zack-rma: Moved common builder methods to the base abstract class

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/ConstantLocationLevel.java
zack-rma: Removed as part of class hierarchy cleanup

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -283,139 +283,290 @@ public Builder(JDomLocationLevelImpl copyFrom) {
...
+                    .withIntervalOrigin(intervalOrigin)
+                    .withIntervalMinutes(intervalMinutes)
+                    .withIntervalMonths(intervalMonths)
+                    .build();
+        } else {
+            throw new UnsupportedFormatException("Unsupported location level type");
+        }
+    }
+
+    public static <T extends LocationLevel> T castLocationLevel(LocationLevel level, ZonedDateTime unmarshalledDateTime) {
```
adamkorynta: This is not doing what you think it is doing...

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -283,139 +283,290 @@ public Builder(JDomLocationLevelImpl copyFrom) {
...
+        String locationId = (updatedLevel.getLocationLevelId() == null
+                ? existingLevel.getLocationLevelId() : updatedLevel.getLocationLevelId());
+        String officeId = (updatedLevel.getOfficeId() == null
+                ? existingLevel.getOfficeId() : updatedLevel.getOfficeId());
+
+        if (existingLevel.getAttributeValue() == null) {
+            attributeUnitsId = null;
+        }
+
+        if (existingLevel instanceof VirtualLocationLevel) {
```
adamkorynta: All of the common builder attributes should get extracted

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Removed method and refactored logic

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Extracted common builder attributes

MikeNeilson (commented):
This looks reasonable to me.
@adamkorynta still has request changes marked so I'll let him have the final say.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -138,86 +138,73 @@ public abstract class LocationLevel extends CwmsDTO {
...
 
-    public String getParameterId()
-    {
+    public String getParameterId() {
         return parameterId;
     }
 
     @JsonPOJOBuilder
     @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
-    public abstract static class Builder<T extends Builder<T>> {
```
adamkorynta: Should stay abstract

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -427,79 +412,55 @@ public static LocationLevel getUpdatedLocationLevel(LocationLevel existingLevel,
             attributeUnitsId = null;
         }
 
-        if (existingLevel instanceof VirtualLocationLevel) {
+        LocationLevel.Builder builder = new LocationLevel.Builder(locationId, unmarshalledDate)
```
adamkorynta: This is targeting the instantiating base class

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -427,79 +412,55 @@ public static LocationLevel getUpdatedLocationLevel(LocationLevel existingLevel,
...
-                    .withLevelComment(levelComment)
-                    .withDurationId(durationId)
-                    .withAttributeValue(attributeValue)
-                    .withAttributeUnitsId(attributeUnitsId)
-                    .withAttributeParameterTypeId(attributeParameterTypeId)
-                    .withAttributeParameterId(attributeParameterId)
-                    .withAttributeDurationId(attributeDurationId)
-                    .withAttributeComment(attributeComment)
-                    .withOfficeId(officeId)
+            return ((VirtualLocationLevel.Builder) builder)
```
adamkorynta: This is casting to the subclass. This will throw a ClassCastException. You should have unit test coverage of this that is failing, if not please add.

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Reverted to abstract

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Refactored common builder methods

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
zack-rma: Added unit test coverage

adamkorynta (approved):

PR #1090, opened 2025-05-01
Added Blob DELETE and PATCH endpoints

### LINKED ISSUE #1050: Add DELETE Function to Blob Endpoint
I can't seem to find a way to delete a BLOB. I'm not finding the DELETE endpoint. Thanks

Image

zack-rma:
Fixes #1050 

Adds delete and update endpoints to blob controller. Added integration test for update and deletion.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
@@ -151,11 +151,11 @@ public class BlobDao extends JooqDao<Blob> {
     }
 
     public void update(Blob blob, boolean ignoreNulls) {
-        String pFailIfExists = OracleTypeMap.formatBool(false);
-        String pIgnoreNulls = OracleTypeMap.formatBool(ignoreNulls);
+        String pFailIfExists = formatBool(false);
+        String pIgnoreNulls = formatBool(ignoreNulls);
 
         if(blob == null){
-            throw new NotFoundException("null blob provided to update");
+            throw new NotFoundException("Null blob provided to update");
         }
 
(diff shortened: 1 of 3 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -150,6 +150,55 @@ public void create(Blob blob, boolean failIfExists, boolean ignoreNulls) {
...
 
+    public void update(Blob blob, boolean ignoreNulls) {
+        String pFailIfExists = formatBool(false);
+        String pIgnoreNulls = formatBool(ignoreNulls);
+
+        if(blob == null){
+            throw new NotFoundException("Null blob provided to update");
+        }
+
+        if (!blobExists(blob.getOfficeId(), blob.getId())) {
```
MikeNeilson, line 161: So we're ignoring the FailIfExists parameter and just requiring that the Blob already exists?

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
zack-rma: We're requiring that the Blob exists to update it. I have yet to see an update endpoint that takes a FailIfExists parameter as input.

MikeNeilson: Yeah, I'm blind. I saw the formatBool above and somehow thought it was pulling an actual query parameter.

MikeNeilson (approved):

PR #1093, opened 2025-05-06
Return better error message when POST request value too large

### LINKED ISSUE #870: Return better error message when POST request value too large
Not sure about all the other endpoints. But need to catch when a POST request value is too large and let the user know. 

In this example case I was gunning for it just to see what happens. 

```bash
 Caused by: java.sql.SQLException: ORA-12899: value too large for column "CWMS_20"."AT_API_KEYS"."KEY_NAME" (actual: 235, maximum: 64)
 ```
 
 ```bash
 curl -X 'POST' \
  'T7IP:TCport/swf-data/auth/keys' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "user-id": "myuserid",
  "key-name": "Lorem ipsum dolor sit amet, consectetur adipisicing elit. Quaerat beatae dolore perferendis pariatur reiciendis temporibus? Maxime, aperiam perferendis, modi, reiciendis qui expedita repellat ea adipisci omnis laboriosam in! Reprehenderit, modi?
Impedit, quidem earum ducimus, sequi reprehenderit voluptate magni omnis, mollitia harum excepturi nesciunt nobis amet doloribus voluptatum? Dignissimos facere, voluptatum excepturi id, ipsam dolore asperiores, quibusdam ducimus doloribus magnam aliquid?"
}'
```

This error message is too generic. What if you have 65 instead of 64?

```json
{
  "message": "System Error",
  "incidentIdentifier": "-8721545769261696166",
  "details": {}
}
```

zack-rma:
Fixes #870 

Added value length checking to auth endpoint, more in progress


rma-rripken:
ApiServlet should already be catching IllegalArgumentExceptions and turning them into 400 error codes.  You probably don't have to do that part.  
JooqDao has a method called wrapException that could be made to automatically turn a SQLException -> a ValueTooLongException

Those two things should simplify your changes.

The piece I'm not sure about is how you tell users that some columns are 64 and others are 32 or 1024 or whatever the limit is.  There might be a way to introspect the schema.  Or we could hard code the known limits on the fields.   Or maybe both - hardcode the limit but also have a unit test that verifies them against the schema ? I'm sure if we think about it a while we can come up with smart ways of handling and verifying

MikeNeilson:
You can look at the meta data of a table directly, but JOOQ *may* have already extracted that and put it somewhere.

adamkorynta:
jOOQ's DataType's include length/precision/scale (https://www.jooq.org/javadoc/dev/org.jooq/org/jooq/DataType.html#length()) so you should be able to do something like `AV_CWMS_TS_ID.AV_CWMS_TS_ID.BASE_LOCATION_ID.getDataType().length();` though last we had talked about this, we wanted to pull from the table directly and we're still waiting on a codegen with table access.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -410,6 +410,11 @@ public void init() {
                     CdaError re = new CdaError(e.getMessage());
                     ctx.status(HttpServletResponse.SC_BAD_REQUEST).json(re);
                 })
+                .exception(ValueTooLongException.class, (e, ctx) -> {
+                    CdaError re = new CdaError("One or more provided values exceeds the maximum length for the parameter.");
```
adamkorynta: We're not providing which value was too long or what the correct size is?

cwms-data-api/src/main/java/cwms/cda/api/auth/ApiKeyController.java
```diff
@@ -89,6 +90,8 @@ public void create(Context ctx) {
         } catch (CwmsAuthException ex) {
             if (ex.getMessage().equals(AuthDao.ONLY_OWN_KEY_MESSAGE)) {
                 ctx.json(new CdaError(ex.getMessage(), true)).status(ex.getAuthFailCode());
+            } else if (ex.getMessage().contains("too large for column")) {
```
adamkorynta: Should be able to use an oracle code instead of text matching: https://docs.oracle.com/en/error-help/db/ora-12899/?r=23ai

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
```diff
@@ -0,0 +1,46 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package cwms.cda.api.errors;
+
+public class ValueTooLongException extends IllegalArgumentException
```
adamkorynta: final, and what is the benefit of extending `IllegalArgumentException`? We wouldn't want to accidentally catch this using its super type.

adamkorynta: Would be great if this exception could include parameter, parameter size, and accepted max length and then standardize the message format.

cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterContractCreateController.java
```diff
@@ -104,8 +104,12 @@ public void handle(@NotNull Context ctx) {
             boolean ignoreNulls = Boolean.parseBoolean(ctx.queryParam(IGNORE_NULLS));
             String newContractName = ctx.pathParam(WATER_USER);
             WaterContractDao contractDao = getContractDao(dsl);
-            contractDao.storeWaterContract(waterContract, failIfExists, ignoreNulls);
-            ctx.status(HttpServletResponse.SC_CREATED).json(newContractName + " created successfully");
+            try {
+                contractDao.storeWaterContract(waterContract, failIfExists, ignoreNulls);
+                ctx.status(HttpServletResponse.SC_CREATED).json(newContractName + " created successfully");
+            } catch (ValueTooLongException ex) {
```
adamkorynta: You're handling this in ApiServlet, no need to handle in the controller

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -436,6 +435,11 @@ public ApiKey createApiKey(DataApiPrincipal p, ApiKey sourceData) throws CwmsAut
                     }
                     createKey.execute();
                 }
+                catch (Exception e)
```
adamkorynta: this repo using end of line brackets.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -436,6 +435,11 @@ public ApiKey createApiKey(DataApiPrincipal p, ApiKey sourceData) throws CwmsAut
...
+                {
+                    RuntimeException re = new RuntimeException(e);
```
adamkorynta: Why not catch SQLException and use `DataAccessException` to follow the other patterns?

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -139,15 +139,23 @@ public List<Blob> getAll(String  officeId, String like) {
...
-                blob.getValue(),
-                blob.getId(),
-                blob.getMediaTypeId(),
-                blob.getDescription(),
-                pFailIfExists,
-                pIgnoreNulls,
-                blob.getOfficeId()));
+
+        dsl.connection(c -> {
+            try
```
adamkorynta: watch your brackets

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -139,15 +139,23 @@ public List<Blob> getAll(String  officeId, String like) {
...
-                getDslContext(c, blob.getOfficeId()).configuration(),
-                blob.getValue(),
-                blob.getId(),
-                blob.getMediaTypeId(),
-                blob.getDescription(),
-                pFailIfExists,
-                pIgnoreNulls,
-                blob.getOfficeId()));
+
+        dsl.connection(c -> {
```
adamkorynta: if you're updating this, might as well update to use the connection/connectionResult methods from the base class so that you automatically get `JooqDao.wrapException`

adamkorynta: this goes for all the other ones you did too

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationCategoryDao.java
```diff
@@ -86,7 +86,11 @@ public void create(LocationCategory category) {
         String office = category.getOfficeId();
         connection(dsl, conn -> {
             DSLContext dslContext = getDslContext(conn, office);
-            CWMS_LOC_PACKAGE.call_CREATE_LOC_CATEGORY(dslContext.configuration(), category.getId(), category.getDescription(), office);
+            try {
+                CWMS_LOC_PACKAGE.call_CREATE_LOC_CATEGORY(dslContext.configuration(), category.getId(), category.getDescription(), office);
+            } catch (RuntimeException e) {
+                throw JooqDao.wrapException(e);
```
adamkorynta: This should already be happening in the base class through the connection method

adamkorynta: This goes for several other that you updated here

cwms-data-api/src/test/java/cwms/cda/api/LocationGroupControllerTestIT.java
```diff
@@ -1426,4 +1426,31 @@ void test_CWMS_permissions_with_replacement() throws Exception {
             .statusCode(is(HttpServletResponse.SC_OK))
             .body("assigned-locations.size()", equalTo(expectedSize));
     }
+
+    @Test
+    void test_ID_too_long()
+    {
+        String officeId = user.getOperatingOffice();
+        String invalidId = "ThisIsAnInvalidIdThatIsWayTooLongForTheDatabaseSinceTheMaximumLengthIs65Characters";
```
adamkorynta: While these messages are funny, I recommend creating a utility class that can generate a string for a given length.

cwms-data-api/src/test/java/cwms/cda/api/auth/ApiKeyControllerTestIT.java
```diff
@@ -355,6 +354,29 @@ public void test_api_key_delete_key(String authType, TestAccounts.KeyUser theUse
         assertTrue(keys.size() < firstReturnedKeys.size(), "Keys were not deleted.");
     }
 
+    @Order(8)
```
adamkorynta: order shouldn't matter for this test

cwms-data-api/src/main/java/cwms/cda/api/auth/ApiKeyController.java
adamkorynta: you already wrote the logic, call into JooqDao for this check

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
MikeNeilson: I took a look at the link you provided below, the database itself below 23ai doesn't actually report that information.

We'd have to manually get the metadata out of the database and validate ourselves... which might be important enough to do here.

cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterContractTypeCreateController.java
```diff
@@ -85,8 +86,12 @@ public void handle(@NotNull Context ctx) {
             ctx.contentType(contentType.toString());
             LookupType contractType = Formats.parseContent(contentType, ctx.body(), LookupType.class);
             WaterContractDao contractDao = getContractDao(dsl);
-            contractDao.storeWaterContractType(contractType, failIfExists);
-            ctx.status(HttpServletResponse.SC_CREATED).json("Contract type successfully stored to CWMS.");
+            try {
+                contractDao.storeWaterContractType(contractType, failIfExists);
+                ctx.status(HttpServletResponse.SC_CREATED).json("Contract type successfully stored to CWMS.");
+            } catch (ValueTooLongException ex) {
+                ctx.status(HttpServletResponse.SC_BAD_REQUEST).json(ex.getMessage());
```
MikeNeilson: Same as Adam's comment in WaterContractCreateController, let the handlers in ApiServlet deal with it.

Not as critical in these cases but it simplifies the logic of when to print the stacktrace to the logs.

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
MikeNeilson: It should also derive from the CdaException and include support for suppressing the incident identifier or not. (granted almost always true for what this is used for.)

cwms-data-api/src/test/java/cwms/cda/api/LocationGroupControllerTestIT.java
MikeNeilson: Feel free to have that utility class generated silly names. Nothing wrong with some levity in the mix.

MikeNeilson: (Silly but appropriate anyways.)

cwms-data-api/src/test/java/cwms/cda/api/WaterContractControllerTestIT.java
```diff
@@ -69,7 +68,7 @@
 
 @Tag("integration")
 class WaterContractControllerTestIT extends DataApiTestIT {
-    private static final String OFFICE_ID = "SWT";
+    private static final String OFFICE_ID = "SPK";
```
MikeNeilson: Did these fail with SWT? Several of these tests have the OFFICE set to a different user specifically to test multiple office operations.

MikeNeilson (commented):
As @adamkorynta already review with request changes I will just comment for tracking purpose.

My additions were follow up comments on his except one.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
zack-rma: I've updated the error handling to parse the error message coming from the database. One of the two possible errors reports the problem parameter, as well as the actual and expected lengths. It's also possible to manually provide values to the ValueTooLongException constructor, but that would require knowing precisely which parameter is the issue. I can either use the most likely parameter and provide it's max length, or adjust the validator in the relevant classes to throw a ValueTooLongException. The second option may be better for when we are able to gather the max length from JOOQ.

cwms-data-api/src/main/java/cwms/cda/api/auth/ApiKeyController.java
zack-rma: Removed text comparison, called into JooqDao

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
zack-rma: That's a good point about `IllegalArgumentException`. I've added parameter, input size, and max size as values in the new exception class.

I would derive from CdaError, except the error handling in ApiServlet that processes the ValueTooLongException requires the error derive from RuntimeException. I've added a suppression parameter to the ValueTooLongException class, so it will be passed through to the generated CdaError.

cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterContractCreateController.java
zack-rma: Removed extraneous handling

cwms-data-api/src/main/java/cwms/cda/api/watersupply/WaterContractTypeCreateController.java
zack-rma: Removed error catch, ApiServlet will handle this.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
zack-rma: Went through and corrected where I found issues

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
zack-rma: Now catching SQLException, swapped RuntimeException for DataAccessException

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
zack-rma: Updated connection methods

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
zack-rma: Fixed

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationCategoryDao.java
zack-rma: Removed unnecessary catch

cwms-data-api/src/test/java/cwms/cda/api/LocationGroupControllerTestIT.java
zack-rma: Added a string generation class. It doesn't maintain the whimsical nature, however.

cwms-data-api/src/test/java/cwms/cda/api/WaterContractControllerTestIT.java
zack-rma: They did fail with SWT locally (a GRANT would likely fix that, but my Docker database instance has been acting up, so I can't verify at the moment), but I can swap these back.

cwms-data-api/src/test/java/cwms/cda/api/auth/ApiKeyControllerTestIT.java
zack-rma: Removed order annotation

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -410,6 +410,18 @@ public void init() {
                     CdaError re = new CdaError(e.getMessage());
                     ctx.status(HttpServletResponse.SC_BAD_REQUEST).json(re);
                 })
+                .exception(ValueTooLongException.class, (e, ctx) -> {
+                    CdaError re;
+                    if (e.hasParameter()) {
+                        re = new CdaError(String.format("One or more provided values exceeds the "
```
adamkorynta: Move this message into the `ValueTooLongException`

cwms-data-api/src/test/java/helpers/StringGenerator.java
```diff
@@ -0,0 +1,45 @@
...
+	{
+		throw new AssertionError("Utility class");
+	}
+
+	public static String createString(int length)
+	{
+		StringBuilder sb = new StringBuilder(length);
+		for (int i = 0; i < length; i++)
+		{
+			sb.append((char) (Math.random() * 26 + 'a'));
```
adamkorynta: This is not fun at all...

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
adamkorynta: 23ai doesn't report the information to the codegen? or doesn't report in the error message? (I think it is the latter, but Zack's response sounds like the former?)

MikeNeilson: 23ai will report the column and size restriction in the Oracle Error. Below 23ai all we get is "there's a problem writing to a column"

zack-rma: It looks like I was incorrect and the length is available via the codegen, at least for views. The codegen would help with knowing the length of a parameter, but it wouldn't assist with knowing which parameter is causing the issue. I could add this to the internal validation for the appropriate DTOs to throw an error based on the max lengths from the codegen, but I'm not sure it would be possible to get that information from the returned database error.

adamkorynta: I think it is worth creating a follow-up ticket - that is contingent on a codegen that includes tables - that would cover length validation on the DTO's. Should be easy enough to add to timeseries ids and location ids for a start.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
zack-rma: Moved into ValueTooLongException constructor

cwms-data-api/src/test/java/helpers/StringGenerator.java
zack-rma: Removed custom class and replaced with Apache Commons string generator. Not as fun, but cleans up the code.

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
```diff
@@ -36,11 +40,47 @@ public ValueTooLongException(String message, Throwable cause) {
...
+		this.maxLength = maxLength;
+		this.suppressIncidentId = suppressIncidentId;
+	}
+
+	public boolean hasParameter()
+	{
+		return parameter != null && !parameter.isEmpty() && length > 0 && maxLength > 0;
+	}
+
+	public static ValueTooLongException fromString(String messageToParse, Throwable cause, boolean suppressIncidentId) {
```
adamkorynta: I don't like having both public constructors and factory methods. Are they all needed?

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
```diff
@@ -36,11 +40,47 @@ public ValueTooLongException(String message, Throwable cause) {
...
+		String[] parts = messageToParse.split("\"");
+		String parameter = parts[parts.length - 2];
```
adamkorynta: There is no format checking here, very dangerous

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
zack-rma: @MikeNeilson Adam pointed me at this declined database PR that could be helpful to add codegen coverage of the tables. This could be used in the DTO validation to provide the necessary parameter length information: https://bitbucket.hecdev.net/projects/CWMS/repos/cwms_database/pull-requests/390/overview

MikeNeilson: I'll have to revisit that. At time it was created it didn't quite behave right, but it might have just been during build and several other changes may have helped along the way.

But don't wait for it on this PR.

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
zack-rma: I could convert the `fromString` method into a constructor instead, though I believe it could be easily confused for a non-parsing constructor.

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
zack-rma: Added a check for the error message code and format checking for parameters to be parsed

zack-rma: Added format error checking

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
zack-rma: Moved parsing to JooqDao, dropped factory methods from ValueTooLargeException class

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -410,6 +410,16 @@ public void init() {
                     CdaError re = new CdaError(e.getMessage());
                     ctx.status(HttpServletResponse.SC_BAD_REQUEST).json(re);
                 })
+                .exception(ValueTooLongException.class, (e, ctx) -> {
+                    CdaError re;
+                    if (e.hasParameter()) {
```
adamkorynta: The message should be in the ValueTooLongException, that way you don't need to check for a parameter here

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -524,6 +533,43 @@ private static InvalidItemException buildInvalidUnits(RuntimeException input) {
...
+        String localizedMessage = cause.getLocalizedMessage();
+        if (localizedMessage != null) {
+            String[] parts = localizedMessage.split("\n");
+            if (parts.length > 1) {
+                localizedMessage = parts[0];
+            }
+        }
+        if (hasCodeOrMessage((SQLException) input.getCause(), Collections.singletonList(12899),
+                Collections.singletonList("value too large for column"))) {
+            if (!localizedMessage.startsWith("ORA-12899:")) {
```
adamkorynta: IllegalArgumentException should be when a user supplies an illegal argument, not when the database does. If the message format doesn't match what you are expecting, then throw the other `ValueTooLongException` message.

Additionally, use a regex so you can do all the checks at once

adamkorynta (changes requested):

MikeNeilson (dismissed):
My primary concerns have been address, It appears some of Adam's comments still need to be addressed.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
zack-rma: Moved message to ValueTooLongException class and cleaned up ApiServlet usage

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
zack-rma: Added regex for error message content matching. Removed IllegalArgumentException usages and replaced with standard ValueTooLongException throw

zack-rma (commented):
Updated per feedback

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -558,25 +560,15 @@ private static ValueTooLongException buildValueTooLongException(RuntimeException
                 localizedMessage = parts[0];
             }
         }
-        if (hasCodeOrMessage((SQLException) input.getCause(), Collections.singletonList(12899),
-                Collections.singletonList("value too large for column"))) {
-            if (!localizedMessage.startsWith("ORA-12899:")) {
-                throw new IllegalArgumentException("The provided message does not appear to be an ORA-12899 error message.");
-            }
+        if (localizedMessage != null && VALUE_TOO_LONG.matcher(localizedMessage).matches()) {
```
adamkorynta: Don't throw away your matcher, extract relevant data with groups instead of text splits

zack-rma: Updated to use matcher groups instead of text splits

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -80,11 +80,11 @@ public abstract class JooqDao<T> extends Dao<T> {
...
 
     static ExecuteListener listener = new ExceptionWrappingListener();
-    private static Pattern INVALID_OFFICE_ID = Pattern.compile(
+    private static final Pattern INVALID_OFFICE_ID = Pattern.compile(
         "INVALID_OFFICE_ID: \"([^\"]+)\" is not a valid CWMS office id");
-    private static Pattern VALUE_TOO_LONG = Pattern.compile(
-            "(^ORA-12899: value too large for column )(\".+\"\\.\".+\"\\.\".+\")" +
-                    "( \\(actual: )(\\d+)(, maximum: )(\\d+)(\\)(\\R*)$)");
+    private static final Pattern VALUE_TOO_LONG = Pattern.compile(
+            "^ORA-12899: value too large for column \".+\"\\.\".+\"\\.\"(.+)\" "
```
MikeNeilson, line 86: doesn't this only work in Oracle 23.x?

E.g. some of this needs to be optional and account for this lack of data. Or is there already a fall-back I'm not seeing at the moment?

MikeNeilson (changes requested):
I like the Regex for value too long, but only Oracle 23.x provides all of that. Unfortunately none of the production systems we use allow for 23.x at the moment.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
zack-rma: There is an initial check in the `isValueTooLargeException` method that looks for ORA-06502, ORA-12899, and ORA-20041, which all can be thrown by the same underlying error of a value that is too long for the parameter. The regex match looks for the specific error message we can extract details from, and defaults to a generic error message of `One or more provided values exceeds the maximum length for the parameter.`

MikeNeilson (approved):

adamkorynta (approved):

PR #1100, opened 2025-05-14
Switch to buildx to create sbom.

### LINKED ISSUE #1099: Setup docker buildx with creation of an sbom.
At least one scanner (AWS) goes crazy and calls out dependencies list in pom.xml files embedded in jars that *aren't* actually on the system.

That same scanner will respect a provided SBOM and not go completely crazy.

MikeNeilson:
Attempting to test manually before merging this in.


MikeNeilson:
System works. only the sha tag was generated during the build, but I suspect that is a side effect of how the test was run.

MikeNeilson:
The generated SBOM also include only information about artifacts *actually* present in the container.

PR #1103, opened 2025-05-16
Composite Time Series Design document.

MikeNeilson:
Initial design work to meet the needs of https://github.com/USACE/cwms-data-api/discussions/956 as previously discussed.

### DIFF
--- a/docs/source/design/composite-time-series.rst
+++ b/docs/source/design/composite-time-series.rst
@@ -1,0 +1,220 @@
+#####################
+Composite Time Series
+#####################
+
+Purpose
+=======
+
+It is a challenge for users to identity what the correct authoritative time series is for a given measurement at a location. Additionally these time series often change over time, either being completely new or changing their interval as newer technologies become available.
+
+Gather an entire Period of Record for the value at a location is also rather difficult. And the POR record and "authoritative timeseries" may be one-in-the same.
+
+
+Need
+====
+
+#. CWMS and Access-2-Water require a simple mechanism to allow users of data to retrieve the Authoritative Period of Record data for a given measurement without having to  understand all of the possible component time series that may be involved. 
+#. Period-of-Record time series *should* not be created by duplicating data from the component time series and merging them into a new one.
+#. The naming of the time series should fit within the excepting CWMS Time Series Identifier design and not unreasonably interfere with existing usages.
+
+
+Caveats
+=======
+
+#. It is assumed that CWMS-Vue will, as-always, require updates to handle what is created here.
+   #. e.g. we're not going to let any current limitations of CWMS-Vue hinder our design.
+
+
+Proposal
+========
+
+Description
+-----------
+
+CDA should handle a concept of a "Composite Time Series". Whether a Time Series is considered composite will be determined by a specific element of the Time Series Identifier.
+Data Administrators will configure which Time Series, and the range there-in, are part of the composite time series.
+CDA will use this stored information to build the Time Series per the question.
+
+Additional names not used
+-------------------------
+
+#. Virtual Time Series
+#. Period of Record Time Series
+
+Both names have been discarded. We use "Virtual" in too many other places with a more direct meaning of that word. 
+For Period-of-Record, while that is the primary use-case, the concept is useful in other situations as well.
+
+Hence generically have have a "composite time series"
+
+Axioms
+------
+
+#. Composite Time Series are Irregular
+#. The definition of the composite time series is stored within the CWMS database
+#. The members of a composite time series define a continuous range
+   #. The date ranges of a composite time series *MUST* not overlap
+   #. The date ranges of a composite time series *MUST* have any gaps
+   #. Data may have gaps, an explanation range should be provided.
+#. The members of a composite time measure the same thing. (e.g. all members are Elevation, not some are elevation and some are stage.)
+
+
+Time Series Naming
+------------------
+
+Option 1
+~~~~~~~~
+
+`<Location Id>.<Parameter>.<Parameter Type>.Composite.var.<version>`
+
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+| Element              | Description                                                                                                            |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Location Id           |As the normal CWMS TS ID, the location for this measure                                                                 |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter             |As the normal CWMS TS ID, the measurement (e.g. Stage, Precip, Elevation, flow, etc)                                    |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter Type        |As Normal CWMS TS ID, Instantaneous, average, total, etc                                                                |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval -\> Composite| Marker that this time series does not have a fix information and is build of various member time series.               |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration -\> var      |Duration of average or total may change over time with new members, duration will be indicated in the member definition |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version               |As Normal CWMS TS ID                                                                                                    |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+
+
+Option 2
+~~~~~~~~
+
+`<Location Id>.<Parameter>.Composite.0.0.<version>`
+
+
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+| Element                | Description                                                                                                            |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Location Id             |As the normal CWMS TS ID, the location for this measure                                                                 |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter               |As the normal CWMS TS ID, the measurement (e.g. Stage, Precip, Elevation, flow, etc)                                    |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter Type Composite|Marker that this time series does not have a fix information and is build of various member time series.                |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval -\> 0          |Interval of data elements. may change over time with new members, duration will be indicated in the member definition   |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration -\> 0          |Duration of average or total. may change over time with new members, duration will be indicated in the member definition|
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version                 |As Normal CWMS TS ID                                                                                                    |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+
+The zero's could also be var
+
+Composite Time Series Definition
+================================
+
+.. code-block::jsonc
+
+    {
+    "office": "<string>",   
+    "name": "<ts id name>",
+    "is-period-of-record": true, // or is authoritative. to distinguish between other possible use-cases?
+    "members": [
+        {
+            "time-series-id": "TS ID for this range",
+            "start": "start date of this",
+            "end": "end date of this range",
+            "notes": "text",
+            "version", "version date", // maybe not? could just use POR or period-of-record in the ts id version
+            // if values that equals the start or end timestamp are included
+            "start-inclusive": true,
+            "end-inclusive": false
+            // suggest default of "start-inclusive": true, "end-inclusive": false
+            // it may also make sense to just make this *always* the above and not let the user set it.
+            // alternatively if this is always [) then only start is required.
+            // however, the class would required and end field as the actual Time Series output needs to know the actual end.
+        }
+    ]
+    // array above *should* be sorted by start when provided to user.
+    }
+
+
+Operations required:
+
+* Create
+* Remove member (ts id + range)
+* Add member
+* List members
+* Replace all members?
+* Delete
+
+
+Composite Time Series Response
+==============================
+
+.. code-block::jsonc
+
+    {
+    // ... as current TimeSeries JSON
+    "composite-members-present": [
+        // member definition from above
+    ] 
+    }
+
+
+Supported Operations:
+
+* Get, through existing TimeSeries classes.
+
+Does it make sense to support writing directly to a composite time series. While the write of each element *could* be sent to the underlying member, this seems ripe for error when editing or updating any data. It is likely that any edits would always be to the most recent time series, and configured in some other system.
+
+Storage of member information
+================================
+
+#. Store in Clob as we refine the design
+#. Create appropriate tables once the design is stable.
+
+System responsibility for "knowing" to process composite.
+=========================================================
+
+Time Series Catalog
+-------------------
+
+Time Series Catalog should show composite time series and allow searching by "authoritative"
+
+TimeSeries DTO
+--------------
+
+Add nullable "members" property.
+
+TimeSeriesDao
+-------------
+
+If the system sees the "Composite" marker retrieve the members for the range and build the time series.
+
+NOTE: considering the user may request the *entire* Period-of-record, this is a good opportunity to see that, start the retrieval in a job queue, and return a status URL to the user for future download. I have see such mechanism for bulk data in other systems. Maybe return an "I'm working on it variant" that the controller can know how to format.
+
+Error handling and other conditions.
+====================================
+
+Versioned (date) time series
+----------------------------
+
+As the composite time series is comprised of multiple other time series should this always be an error to specify?
+The marker for always latest or always first may make sense to allow, however, at the time series is supposed to be authoritative, that would add ambiguity.
+
+Datum conversions
+-----------------
+
+Retrievers of the Period-of-Record *SHOULD* be able to retrieve the data as a single datum. Composite retrieval should respond as https://github.com/USACE/cwms-data-api/issues/1102 and convert each member as appropriate
+
+
+On the saving of a composite definition
+---------------------------------------
+
+The even if only a single member is added, the full definition needs to be check to ensure the ranges are still overlapping and continuous.
+
+References
+==========
+
+#. https://github.com/USACE/cwms-data-api/discussions/956
+#. https://github.com/USACE/cwms-data-api/issues/955
+#. https://www.hec.usace.army.mil/confluence/spaces/CWMS/pages/290456000/Virtual+Timeseries
+#. https://discourse.hecdev.net/t/period-of-record-timeseries/3859/2
--- a/docs/source/design/composite-time-series.rst
+++ b/docs/source/design/composite-time-series.rst
@@ -44,7 +44,7 @@ Additional names not used
 Both names have been discarded. We use "Virtual" in too many other places with a more direct meaning of that word. 
 For Period-of-Record, while that is the primary use-case, the concept is useful in other situations as well.
 
-Hence generically have have a "composite time series"
+Hence generically we have a "composite time series"
 
 Axioms
 ------
@@ -52,10 +52,11 @@ Axioms
 #. Composite Time Series are Irregular
 #. The definition of the composite time series is stored within the CWMS database
 #. The members of a composite time series define a continuous range
-   #. The date ranges of a composite time series *MUST* not overlap
-   #. The date ranges of a composite time series *MUST* have any gaps
+   #. The date ranges of a member *MUST* not overlap
+   #. The date ranges of a member *MUST* not have any gaps
    #. Data may have gaps, an explanation range should be provided.
 #. The members of a composite time measure the same thing. (e.g. all members are Elevation, not some are elevation and some are stage.)
+#. The interval and duration of each member *MAY* be different.
 
 
 Time Series Naming
@@ -107,6 +108,87 @@ Option 2
 
 The zero's could also be var
 
+
+Option 3
+~~~~~~~~
+
+`<Location Id>.<Parameter>.<Parameter Type>.<Interval>.<Duration>.Composite`
+
+
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+| Element                | Description                                                                                                            |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Location Id             |As the normal CWMS TS ID, the location for this measure                                                                 |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter               |As the normal CWMS TS ID, the measurement (e.g. Stage, Precip, Elevation, flow, etc)                                    |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter Type          |Marker that this time series does not have a fix information and is build of various member time series.                |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval                |Interval of data elements. may change over time with new members, duration will be indicated in the member definition   |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration                |Duration of average or total. may change over time with new members, duration will be indicated in the member definition|
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version                 |Composite or POR ... or check for composite at the front/back?                                                          |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+
+From Daniel 
+
+Argument Against: the "Version" field it freeform and we often encode other information in it. 
+Argument Against above argument: That said, perhaps forcing the version to be "clean" is the right choice here.
+
+
+Option 4
+~~~~~~~~
+
+`<Location Id>.<Parameter>[Composite].<Parameter Type>.<Interval>.<Duration>.<Version>`
+
+
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+| Element                | Description                                                                                                            |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Location Id             |As the normal CWMS TS ID, the location for this measure                                                                 |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter               |As the normal CWMS TS ID, the measurement (e.g. Stage, Precip, Elevation, flow, etc)                                    |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter Type          |Marker that this time series does not have a fix information and is build of various member time series.                |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval                |Interval of data elements. may change over time with new members, duration will be indicated in the member definition   |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration                |Duration of average or total. may change over time with new members, duration will be indicated in the member definition|
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version                 |As Normal CWMS TS ID                                                                                                    |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+
+
+This form with something in [] has been discussed for embedded TimeZone and Offset information into the interval. Arguably this code go in any field.
+
+
+Option 4
+~~~~~~~~
+
+`<Location Id>.<Parameter>.<Parameter Type>.<Interval>.<Duration>.<Version>` and/or arbitrary TS "alias"
+
+
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+| Element                | Description                                                                                                            |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Location Id             |As the normal CWMS TS ID, the location for this measure                                                                 |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter               |As the normal CWMS TS ID, the measurement (e.g. Stage, Precip, Elevation, flow, etc)                                    |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter Type          |Marker that this time series does not have a fix information and is build of various member time series.                |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval                |Interval of data elements. may change over time with new members, duration will be indicated in the member definition   |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration                |Duration of average or total. may change over time with new members, duration will be indicated in the member definition|
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version                 |As Normal CWMS TS ID                                                                                                    |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+
+However, on request for the timeseries the list of composite time series is consulted and used if present, otherwise passthrough to normal
+time series retrieval.
+
+
 Composite Time Series Definition
 ================================
 
@@ -115,21 +197,13 @@ Composite Time Series Definition
     {
     "office": "<string>",   
     "name": "<ts id name>",
-    "is-period-of-record": true, // or is authoritative. to distinguish between other possible use-cases?
+    "is-authoritative": true, // or is authoritative. to distinguish between other possible use-cases?
     "members": [
         {
             "time-series-id": "TS ID for this range",
-            "start": "start date of this",
-            "end": "end date of this range",
-            "notes": "text",
-            "version", "version date", // maybe not? could just use POR or period-of-record in the ts id version
-            // if values that equals the start or end timestamp are included
-            "start-inclusive": true,
-            "end-inclusive": false
-            // suggest default of "start-inclusive": true, "end-inclusive": false
-            // it may also make sense to just make this *always* the above and not let the user set it.
-            // alternatively if this is always [) then only start is required.
-            // however, the class would required and end field as the actual Time Series output needs to know the actual end.
+            "start": "start date of this", // Inclusive
+            "end": "end date of this range", // Exclusive
+            "notes": "text",            
         }
     ]
     // array above *should* be sorted by start when provided to user.
@@ -163,13 +237,12 @@ Supported Operations:
 
 * Get, through existing TimeSeries classes.
 
-Does it make sense to support writing directly to a composite time series. While the write of each element *could* be sent to the underlying member, this seems ripe for error when editing or updating any data. It is likely that any edits would always be to the most recent time series, and configured in some other system.
 
 Storage of member information
 ================================
 
-#. Store in Clob as we refine the design
-#. Create appropriate tables once the design is stable.
+#. Store in Clob as we refine the design - cache appropriately in member to avoid any major performance issues.
+#. Create appropriate tables once the design is stable - still cache things.
 
 System responsibility for "knowing" to process composite.
 =========================================================
@@ -187,9 +260,14 @@ Add nullable "members" property.
 TimeSeriesDao
 -------------
 
-If the system sees the "Composite" marker retrieve the members for the range and build the time series.
+If the system sees the "Composite" marker/determines is composite retrieve the members for the range and build the time series.
 
-NOTE: considering the user may request the *entire* Period-of-record, this is a good opportunity to see that, start the retrieval in a job queue, and return a status URL to the user for future download. I have see such mechanism for bulk data in other systems. Maybe return an "I'm working on it variant" that the controller can know how to format.
+.. NOTE:: 
+    Considering the user may request the *entire* Period-of-record, this is a good opportunity to see that,
+    start the retrieval in a job queue, and return a status URL to the user for future download. I have see such mechanism 
+    for bulk data in other systems. Maybe return an "I'm working on it variant" that the controller can know how to format.
+
+    Perhaps we do this for data beyond "x amount"?
 
 Error handling and other conditions.
 ====================================
@@ -197,13 +275,13 @@ Error handling and other conditions.
 Versioned (date) time series
 ----------------------------
 
-As the composite time series is comprised of multiple other time series should this always be an error to specify?
-The marker for always latest or always first may make sense to allow, however, at the time series is supposed to be authoritative, that would add ambiguity.
+It is an error to specify a Version (date) when requesting composite data.
 
 Datum conversions
 -----------------
 
-Retrievers of the Period-of-Record *SHOULD* be able to retrieve the data as a single datum. Composite retrieval should respond as https://github.com/USACE/cwms-data-api/issues/1102 and convert each member as appropriate
+Retrievers of the Period-of-Record *SHOULD* be able to retrieve the data as a single datum. Composite retrieval should respond
+ as https://github.com/USACE/cwms-data-api/issues/1102 and convert each member as appropriate
 
 
 On the saving of a composite definition
--- a/docs/source/design/composite-time-series.rst
+++ b/docs/source/design/composite-time-series.rst
@@ -286,6 +286,7 @@ Composite Time Series Definition
 
 
 Operations required:
+--------------------
 
 * Create
 * Remove member (ts id + range)
@@ -299,7 +300,13 @@ Immutable fields:
 
 Fields marked immutable above cannot be updated. At this time no field are thought to be immutable.
 
+Operations Prohibited:
+----------------------
 
+* Any direct manipulation of the underlying time series members.
+
+Example: one cannot `POST` values to a composite time series. Doing so will result in an HTTP 405 - Method Not Allowed
+error.
 
 Composite Time Series Response
 ==============================
--- a/docs/source/design/composite-time-series.rst
+++ b/docs/source/design/composite-time-series.rst
@@ -267,7 +267,7 @@ time series retrieval.
 Composite Time Series Definition
 ================================
 
-.. code-block::jsonc
+.. code-block:: jsonc
 
     {
     "office": "<string>",   
--- a/docs/source/design/composite-time-series.rst
+++ b/docs/source/design/composite-time-series.rst
@@ -58,15 +58,15 @@ only one of which is valid for certain conditions.
 Period of Record
 ~~~~~~~~~~~~~~~~
 
-The Period of Record (POR, period-of-record) for a measurement (such as the Stage at a river or the pool elevation of a 
-dam) is *ALL* available (time,value) pairs since recording began until either recording as ended or the most recent available
-regardless of changes in intervals and, if unavoidable changes in averaging.
+The Period of Record (POR, period-of-record) for a measurement (such as the Stage at a river or the pool elevation of 
+a dam) is ALL available (time,value) pairs since recording began until recording has ended or the most recent available 
+pair, regardless of changes in intervals or unavoidable changes in averaging.
 
 The POR is the "best" available combined dataset that would be desired for studies requiring all values for a given
 location.
 
 However, "best" is subjective. The POR of data used to make a given decision may not be the same as data that has formal
-validation. Additionally having a POR of the raw, unedited data may be what 
+validation. Additionally having a POR of the raw, unedited data, may be what a user studying data validation requires.
 
 
 Naming Option 3, below, is selected for the path forward. A future design document will develop appropriate naming
@@ -78,7 +78,7 @@ Authoritative
 An authoritative time series is a period-of-record time series with the additional constraint that is contains the best
 official data. In other words the data that is determined to be "correct" by appropriate methods of validation.
 
-The data provide will have validated or corrected after events when additional information become available.
+The data provide will have been validated or corrected after events when additional information become available.
 The data may not match what was used at the moment a decision was made.
 
 .. NOTE::
@@ -127,9 +127,11 @@ Axioms
    #. Each member *MUST* have a start date
    #. The last member *MAY* have an end date indicating no more data will be available for this location and measure.
    #. Data may have gaps, an explanation range *SHOULD* be provided. For data with regular gaps, e.g. season gauges
-      a description should be provided in the notes. A Link to a Location Level can be provided if the timing is well known.
-#. The members of a composite time measure the same thing. (e.g. all members are Elevation, not some are elevation and some are stage.)
-#. The interval and duration of each member *MAY* be different.
+      a description should be provided in the notes. 
+      Example: A Link to a Location Level can be provided if the seasonal timing is well known. This would let users
+      of the data now if the gap is missing data "an error" or if just out-of-service.
+#. The members of a composite time series measure the same thing. (e.g. all members are Elevation, not some are elevation and some are stage.)
+#. The parameter type (e.g. Instantaneous; Average; etc), interval, and duration of each member *MAY* be different.
 
 
 Time Series Naming
@@ -296,6 +298,17 @@ Operations required:
 * Update member
 * Delete
 
+
+Events Requires:
+----------------
+
+* Composite Time Series created
+* Composite Time Series deleted
+* Composite Time Series modified
+  * Member added
+  * Member modified
+  * Member removed
+
 Immutable fields:
 
 Fields marked immutable above cannot be updated. At this time no field are thought to be immutable.
@@ -363,7 +376,8 @@ Error handling and other conditions.
 Versioned (date) time series
 ----------------------------
 
-It is an error to specify a Version (date) when requesting composite data.
+It is an error to specify a Version (date) when requesting composite data. Only the latest data of each member will
+be returned.
 
 Datum conversions
 -----------------
--- a/docs/source/design/composite-time-series.rst
+++ b/docs/source/design/composite-time-series.rst
@@ -267,7 +267,8 @@ time series retrieval.
 Composite Time Series Definition
 ================================
 
-.. code-block:: jsonc
+.. code-block:: json
+    :caption: Schema
 
     {
     "office": "<string>",   
@@ -278,7 +279,7 @@ Composite Time Series Definition
     "members": [
         {
             "time-series-id": "TS ID for this range",
-            "start": "start date of this", // Inclusive
+            "start": "start date of this", // Inclusive, ISO 8601 Full
             "end": "end date of this range", // Exclusive, can be null
             "notes": "text",            
         }
@@ -286,6 +287,38 @@ Composite Time Series Definition
     // array above *should* be sorted by start when provided to user.
     }
 
+
+.. code-block:: json
+    :caption: Example (Using Option 5)
+
+    // NOTE: While a specific location was used, the dates and time series are arbitrary and may not exist.
+    {
+    "office": "SPK",   
+    "name": "Alder Springs.Precip-Cumulative.Inst.0.0.Best",
+    "is-authoritative": true,
+    "seasonal-information": null, // This location's measurement is not seasonal
+    "members": [
+        {
+            "time-series-id": "Alder Springs.Precip-Cumulative.Inst.1DayLocal.0.Best",
+            "start": "1948-04-01T14:14:00Z", // Inclusive
+            "end": null, // Exclusive, can be null
+            "notes": "Drum Chart Recorder, Catch Tube",
+        },
+        {
+            "time-series-id": "Alder Springs.Precip-Cumulative.Inst.1Hour.0.Best",
+            "start": "1962-04-01T14:14:00Z", // Inclusive
+            "end": null, // Exclusive, can be null
+            "notes": "Punched Tape Recorder, Catch Tube",
+        },
+        {
+            "time-series-id": "Alder Springs.Precip-Cumulative.Inst.1DayLocal.0.Best",
+            "start": "1990-04-01T14:14:00Z", // Inclusive
+            "end": null, // Exclusive, can be null
+            "notes": "Digital Logger, Catch Tube",
+        },
+    ]
+    // array above *WILL* be sorted by start when provided to user.
+    }
 
 Operations required:
 --------------------
--- a/docs/source/design/composite-time-series.rst
+++ b/docs/source/design/composite-time-series.rst
@@ -7,7 +7,7 @@ Purpose
 
 It is a challenge for users to identity what the correct authoritative time series is for a given measurement at a location. Additionally these time series often change over time, either being completely new or changing their interval as newer technologies become available.
 
-Gather an entire Period of Record for the value at a location is also rather difficult. And the POR record and "authoritative timeseries" may be one-in-the same.
+Gather an entire Period of Record for the value at a location is also rather difficult. And the Period of Record (POR) and "authoritative timeseries" may be one-in-the same.
 
 
 Need
@@ -15,7 +15,7 @@ Need
 
 #. CWMS and Access-2-Water require a simple mechanism to allow users of data to retrieve the Authoritative Period of Record data for a given measurement without having to  understand all of the possible component time series that may be involved. 
 #. Period-of-Record time series *should* not be created by duplicating data from the component time series and merging them into a new one.
-#. The naming of the time series should fit within the excepting CWMS Time Series Identifier design and not unreasonably interfere with existing usages.
+#. The naming of the time series should fit within the existing CWMS Time Series Identifier design and not unreasonably interfere with existing usages.
 
 
 Caveats
@@ -31,9 +31,9 @@ Proposal
 Description
 -----------
 
-CDA should handle a concept of a "Composite Time Series". Whether a Time Series is considered composite will be determined by a specific element of the Time Series Identifier.
+CWMS-Data-API (CDA) should handle a concept of a "Composite Time Series". Whether a Time Series is considered composite will be determined by some means (see naming options below).
 Data Administrators will configure which Time Series, and the range there-in, are part of the composite time series.
-CDA will use this stored information to build the Time Series per the question.
+CDA will use this composite time series definition to build an expand Time Series per query for the range of time requested.
 
 Additional names not used
 -------------------------
@@ -53,8 +53,10 @@ Axioms
 #. The definition of the composite time series is stored within the CWMS database
 #. The members of a composite time series define a continuous range
    #. The date ranges of a member *MUST* not overlap
-   #. The date ranges of a member *MUST* not have any gaps
-   #. Data may have gaps, an explanation range should be provided.
+   #. Each member *MUST* have a start date
+   #. The last member *MAY* have an end date indicating no more data will be available for this location and measure.
+   #. Data may have gaps, an explanation range *SHOULD* be provided. For data with regular gaps, e.g. season gauges
+      a description should be provided in the notes. A Link to a Location Level can be provided if the timing is well known.
 #. The members of a composite time measure the same thing. (e.g. all members are Elevation, not some are elevation and some are stage.)
 #. The interval and duration of each member *MAY* be different.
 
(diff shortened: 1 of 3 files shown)


docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+
+Axioms
+------
+
+#. Composite Time Series are Irregular
+#. The definition of the composite time series is stored within the CWMS database
+#. The members of a composite time series define a continuous range
+   #. The date ranges of a composite time series *MUST* not overlap
+   #. The date ranges of a composite time series *MUST* have any gaps
+   #. Data may have gaps, an explanation range should be provided.
```
msweier: Is this user defined? Will it allow for definition of a timeseries that may not have extents that cover that time period (e.g. there's a ~2 month gap between timeseries A that ends at 2014-01-03 12:00 and timeseries B which starts at 2014-03-14- 12:00). What does an explanation range look like (e.g. "no data, start 2014-01-03 12:00, end 2014-03-14- 12:00)? Is that assigned automatically if there is a gap in the timeseries?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+Additional names not used
+-------------------------
+
+#. Virtual Time Series
+#. Period of Record Time Series
+
+Both names have been discarded. We use "Virtual" in too many other places with a more direct meaning of that word. 
+For Period-of-Record, while that is the primary use-case, the concept is useful in other situations as well.
+
+Hence generically have have a "composite time series"
```
jbkolze: I do think having "virtual" somewhere in the name is somewhat helpful in immediately indicating to the user that the time series doesn't actually exist.  If I saw "composite time series" without any additional context my first thought would probably be a merged copy of other time series data.

Possibly "virtual composite time series" or "composite virtual time series"?  Although there may be some advantage to keeping the name more succinct as in your recommendation.  It will likely be in standard enough usage that users will catch on quickly to whatever the terminology is, so I'm not too worried either way.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+|Parameter Type Composite|Marker that this time series does not have a fix information and is build of various member time series.                |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval -\> 0          |Interval of data elements. may change over time with new members, duration will be indicated in the member definition   |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration -\> 0          |Duration of average or total. may change over time with new members, duration will be indicated in the member definition|
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version                 |As Normal CWMS TS ID                                                                                                    |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
```
msweier, line 182: I like these options, but it would be nice to differentiate between a POR timeseries that includes all best available intervals (e.g. daily inst, 4 hr, 1 hr, 15 minute) and a POR timeseries that includes the best available on a daily interval (e.g. 8 am inst or daily avg). MVP's merged TS denotes these as ~15Minutes and ~1Day but maybe there's a better way. I'm thinking some way like the USGS makes it easy to pick instantaneous value data vs. daily data.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+        // member definition from above
+    ] 
+    }
+
+
+Supported Operations:
+
+* Get, through existing TimeSeries classes.
+
+Does it make sense to support writing directly to a composite time series. While the write of each element *could* be sent to the underlying member, this seems ripe for error when editing or updating any data. It is likely that any edits would always be to the most recent time series, and configured in some other system.
```
msweier: I agree that seems like a bad idea. It's probably better to write to the underlying timeseries.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter Type        |As Normal CWMS TS ID, Instantaneous, average, total, etc                                                                |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval -\> Composite| Marker that this time series does not have a fix information and is build of various member time series.               |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration -\> var      |Duration of average or total may change over time with new members, duration will be indicated in the member definition |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version               |As Normal CWMS TS ID                                                                                                    |
```
msweier, line 158: Could a common version name denote it as authoritative? Or does just the existence of a composite timeseries imply that?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+Hence generically have have a "composite time series"
+
+Axioms
+------
+
+#. Composite Time Series are Irregular
+#. The definition of the composite time series is stored within the CWMS database
+#. The members of a composite time series define a continuous range
+   #. The date ranges of a composite time series *MUST* not overlap
+   #. The date ranges of a composite time series *MUST* have any gaps
```
jbkolze: Is this missing a "not"?  If so, I think it's reasonable to allow for gaps in the data.  For example, if a gage gets washed out and isn't replaced for a month (and the new gage uses a different interval) then there simply wouldn't be data for the missing month.  The user could theoretically extend the range of preceding or following record to include the gap, but that wouldn't feel very intuitive to me.  I would expect that the interpreter could simply return gaps as missing data?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval -\> Composite| Marker that this time series does not have a fix information and is build of various member time series.               |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration -\> var      |Duration of average or total may change over time with new members, duration will be indicated in the member definition |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version               |As Normal CWMS TS ID                                                                                                    |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+
+
+Option 2
```
jbkolze, line 162: Option 2 makes more sense to me.  I'm unsure how Option 1 would deal with potentially varying parameter types among a set of composite data.

I'm a little curious about the implications of doing something like `<Location Id>.<Parameter>.Var.0.0.Composite`.  That would lose the flexibility of using the version as a sort of unique identifier, but I can't really think of a use case for needing multiple composites for one location/parameter off of the top of my head.  Using Option 2's style I'm not sure what I would use for the version besides "Composite" or "POR"  -- I think it's probably somewhat rare to have a consistent source for a full set of POR data.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+
+Composite Time Series Definition
+================================
+
+.. code-block::jsonc
+
+    {
+    "office": "<string>",   
+    "name": "<ts id name>",
+    "is-period-of-record": true, // or is authoritative. to distinguish between other possible use-cases?
```
jbkolze: I'm a little unclear on the relationship between the "composite time series" and the "authoritative time series".  It makes sense that the authoritative TS will generally be a composite TS, but will the functionality for both be handled within this one construct?  What would be the implications of setting "is-period-of-record" to true here?  I feel like this would be more appropriate as "isAuthoritative" based on my assumptions.

Under the CMA paradigm a separate construct exists to link the "authoritative" TS with the parameter for a location.  It seems like this will handle that automatically?  e.g. if I have a Buckhorn.Flow-Outflow composite record with this set to true, this will automatically be returned when a user requests Flow-Outflow data for Buckhorn (also, is there another endpoint created for this)?

Just to clarify -- I don't have strong feelings either way, just trying to understand the intent.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+    "members": [
+        {
+            "time-series-id": "TS ID for this range",
+            "start": "start date of this",
+            "end": "end date of this range",
+            "notes": "text",
+            "version", "version date", // maybe not? could just use POR or period-of-record in the ts id version
```
jbkolze: Not sure what this would be for -- seems like the version for an individual TSID member would just be the version of that TSID?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+            // if values that equals the start or end timestamp are included
+            "start-inclusive": true,
+            "end-inclusive": false
+            // suggest default of "start-inclusive": true, "end-inclusive": false
+            // it may also make sense to just make this *always* the above and not let the user set it.
```
jbkolze: I think just choosing a standard for this would suffice -- seems like a toggle would over-complicate things

docs/source/design/composite-time-series.rst
jbkolze: I'm against allowing writes to the composite time series.  That sounds like a recipe for disaster (and likely a pain to implement while handling edge cases).  Writes should, in my opinion, be done to the underlying time series.

docs/source/design/composite-time-series.rst
MikeNeilson: That's a possibility. I had put in a place older "is-authoritative" flag in the composite definition. 

Though I agree the version is technically a good place for that, it does seem to get a bit... overused at times. 

There are certainly arguments to be me in either case, so we'll wait for commentary from others to tip the scales.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+NOTE: considering the user may request the *entire* Period-of-record, this is a good opportunity to see that, start the retrieval in a job queue, and return a status URL to the user for future download. I have see such mechanism for bulk data in other systems. Maybe return an "I'm working on it variant" that the controller can know how to format.
+
+Error handling and other conditions.
+====================================
+
+Versioned (date) time series
+----------------------------
+
+As the composite time series is comprised of multiple other time series should this always be an error to specify?
+The marker for always latest or always first may make sense to allow, however, at the time series is supposed to be authoritative, that would add ambiguity.
```
jbkolze: I can't personally think of any realistic use cases for creating a composite of versioned time series data.  Maybe not worth supporting unless someone can give an example?

jbkolze (dismissed):
Good summary of the design -- thanks for putting that together.  Noted a few points of clarity / contention.

docs/source/design/composite-time-series.rst
MikeNeilson: That seems like the type of information to put in the version, if desired.

My assumption with the POR time series was that it should be suitable for "this is the full record as we have it" knowing that over time the official record has improved methods of measurement.

So like the first few decades could be daily instantaneous, and the next decade 12 hours, then 1 hour, then 15 minutes, and *maybe* things would change to an average or not. But if you go further down in the document you'll see that the returned time series values also includes the members with their definition.

So yes, you *could* make a composite time series that only included certain intervals and durations, but to the composite system itself it wouldn't care.


That said we could open up the definition to allow the interval and duration to be set, we would then need to decide if that is enforced.

For example:
1. if the composite is 1Day, do we limit every member to 1Day. (This applies to Local Regular as well)
2. if the composite is ~1Day, do we limit every member to ~1Day or allow others since ~1Day means most likely 1 day but could be different


I'm not opposed, I don't think that adds too much complexity, but other one of those more feedback from the group would be good type things.

docs/source/design/composite-time-series.rst
MikeNeilson: ah, I definitely did not describe this clearly enough in the sample data.
couple of things
1. a time series with missing in it still counts as complete

A gap here means that you have an end date for one member and then... and not I've realized if you don't have a defined interval it's hard to determine this.

perhaps that should change to *SHOULD* since I don't think the system can meaningfully define what a "gap" in member is. Does the next start have to be the smallest time unit after the previous end (e.g. nano seconds), if not what is acceptable?


Here's what I was thinking, how do we handle known gaps in service? be it accidental destruction (2 different SPK/SPN gauges have suffered alcohol related removals from service). One site at SPK is removed during most of a year due to no water and it kept suffering vandalism.

So intent is "there's oddly large amount of missing data, how do we report that."

docs/source/design/composite-time-series.rst
MikeNeilson: You are correct, the *NOT* is missing.

I wasn't thinking about the processing returning the missings, but that is a good idea, and would required the gap to be defined... and probably some other information. So it may just be easier to put the information into the time series itself (e.g. in notes) and just let the missings be returned.

docs/source/design/composite-time-series.rst
MikeNeilson: I think @msweier made a fairly decent case for why you might have more than one composite for a location+measure. It does makes sense to me to have a "single authoritative" time series followed by "all data with interval X". Really depends on exactly what you're doing with the data.

docs/source/design/composite-time-series.rst
MikeNeilson: I disagree. adding both composite and virtual would be redundant.

Additionally, we use virtual in the context of Location and Location levels to mean something calculated that doesn't physically exist somewhere (e.g. a "virtual" gauge between two others on a river.)

Whereas this really is just a composite. It's is made of other time series that physically measure something.

@ktarbet came up with the Composite name, he may be able to add more to the argument.

Though as with other things, if enough users say I'm wrong and virtual makes more sense I'll accept the group think.

But I am a little worried about everyone eventually wanting a virtual time series that's more akin to the virtual location levels and then what do we do?

MikeNeilson:
> Good summary of the design -- thanks for putting that together. Noted a few points of clarity / contention.

I say this to anyone, if you think a few things need clarity use the "request changes" feature of the pull-request if you have the authority to do so. If not (e.g. the option just doesn't exist) go ahead make it blindingly obvious (bold text, big red letters like in Hitch Hikers Guide to the Galaxy, etc) something should change first.

Especially in documents like this that will define things for years to come.

docs/source/design/composite-time-series.rst
MikeNeilson: This is exactly why I'm weary of using "virtual" anywhere in the naming.

This is the "date based version" that some districts use, not the last portion of the time series id. Though I really should've called that field "version-date" instead of version.

My though behind this is that we technically have two rather important periods-of-record that may exist
1. The data as it was actually used to make a decision which may either be not corrected, corrected in real-time, etc.
2. The data as corrected, say, using more detailed analysis or improved knowledge - e.g. in the moment you didn't know a tree had fallen in the river or something.

1 can certainly qualify as an official "record" in the full sense of that legal term.
2 seems to be what a student doing a research project would actually want

well, and depending on what the student is doing they might want both.

docs/source/design/composite-time-series.rst
MikeNeilson: I think `is-authoritative` is probably better. Especially after the ideas @msweier provided in his responses. 

Though idea behind this flag is that it makes it easier to decide what should be rendered. and now that I think about it I suspect A2W would always want to render the "authoritative" so it makes sense to actually use that language.

@adamscarberry ?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+#. Store in Clob as we refine the design
+#. Create appropriate tables once the design is stable.
+
+System responsibility for "knowing" to process composite.
+=========================================================
+
+Time Series Catalog
+-------------------
+
+Time Series Catalog should show composite time series and allow searching by "authoritative"
```
adamkorynta, line 387: just FYI: we'd have to teach CWMS-Vue a new Interval definition or a new Parameter Type in order to show up properly. I know you called up inevitable CWMS-Vue updates elsewhere, but wanted to be explicit.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+
+#. Virtual Time Series
+#. Period of Record Time Series
+
+Both names have been discarded. We use "Virtual" in too many other places with a more direct meaning of that word. 
+For Period-of-Record, while that is the primary use-case, the concept is useful in other situations as well.
+
+Hence generically have have a "composite time series"
+
+Axioms
```
msweier, line 120: Is there any way to add a comment or remark to a timeseries (e.g. staff gage readings, gate computations, tailwater rating etc...) or to the composite timeseries itself? If not this could be separately managed in a CLOB, but it would be neat if you could optionally comment on the timeseries as you added them.

MikeNeilson: I'm expecting that when we add the more direct support for extracting the text timeseries along with a value time series that the time series "notes" will just come along.

Alternatively one can also just take the member time series and go retrieve that (definitely not ideal though.)

As for the 2nd part of that. There is a "notes" field for each member.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Location Id           |As the normal CWMS TS ID, the location for this measure                                                                 |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter             |As the normal CWMS TS ID, the measurement (e.g. Stage, Precip, Elevation, flow, etc)                                    |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter Type        |As Normal CWMS TS ID, Instantaneous, average, total, etc                                                                |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval -\> Composite| Marker that this time series does not have a fix information and is build of various member time series.               |
```
DanielTOsborne, line 154: I'm of the opinion that:
1) duration MUST be the same for all composite members. If you change the window size of an averaging function, then you are producing different data. It's not the same as simply swapping out a sensor.
2) interval MUST be the same for all composite members. Trying to retrieve a composite time series where you have no idea what you're going to get is a nightmare to code for. If you want to combine different intervals, create a computation to create the requisite interval data from other intervals.

Tying back into an earlier section, that means the composite doesn't have to be irregular, and could potentially be regular, lrts, or prts depending on the sources.
If all timeseries in a composite are hourly regular, for example, empty values could be generated to fill in the gaps.

Now, that leads to the section I highlighted:
I say make the naming somewhat arbitrary, like it is now. Allow it to operate like an alias, so the user creates the name they want the timeseries to labeled, then say "this is a composite".
For example, that would allow the user to just specify .Composite in the version, and prevent confusion by overloading the other parameters.

`<Location Id>.<Parameter>.<Parameter Type>.<Interval>.<Duration>.Composite`

If that's not feasible, then perhaps adding to the interval, like lrts did:
Something like 1HourComp:
`<Location Id>.<Parameter>.<Parameter Type>.1HourComp.<Duration>.<Version>`

Otherwise, how would you specify composite data for different intervals and types, and keep them separate?

At SPK we have period-of-record data like this:
New Bullards Bar.Elev.Inst.1Hour.0.POR
New Bullards Bar.Elev.Inst.~1Day.0.POR

Currently that's a separate timeseries with duplicate data. But it doesn't have to be.
They could be changed internally to be composites, then maintain the same names, and everything else matches the rest of the system (parameter type, interval, duration).

Otherwise, you end up with something like:
New Bullards Bar.Elev.Inst.Composite.0.POR - What is that? Hourly, daily? What if I only want daily data?

New Bullards Bar.Elev.Composite.~1Day.0.POR - Is that averaged data, or instantaneous?

docs/source/design/composite-time-series.rst
DanielTOsborne: You have to define what "authoritative" means in order to properly answer this question.
At SPK we have two different types of data: real time, and analyzed/processed data. The analyzed data isn't available in our CWMS system until at least 6 months after the water year ends.

Do you want the data as it was in the system when operational decisions were made, or what the 'true' system state that data represents was?

DanielTOsborne: See my comment on this (below?).

docs/source/design/composite-time-series.rst
DanielTOsborne: > Here's what I was thinking, how do we handle known gaps in service? be it accidental destruction (2 different SPK/SPN gauges have suffered alcohol related removals from service). One site at SPK is removed during most of a year due to no water and it kept suffering vandalism.
> 
> So intent is "there's oddly large amount of missing data, how do we report that."

Well, this is something that doesn't currently exist in CWMS at all, as I realized when developing a system to read in punch tapes. There are notes on the tapes for station maintenance, but I have no where to save that information in a meaningful, easy to access way.
I'd argue this is out of scope, since this is a useful feature beyond this discussion.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,220 @@
...
+
+Hence generically have have a "composite time series"
+
+Axioms
+------
+
+#. Composite Time Series are Irregular
+#. The definition of the composite time series is stored within the CWMS database
+#. The members of a composite time series define a continuous range
+   #. The date ranges of a composite time series *MUST* not overlap
```
DanielTOsborne: > The members of a composite time series define a continuous range

This could be worked around by using versioning in the underlying implementation. If there's ambiguity (which I 100% guarantee will happen), then pick whichever is the most recent version.
Alternatively, for a simpler implementation, it could use the "archived" flag on the timeseries. If a timeseries is archived, then we can decide on a behavior based on the start-end dates of that timeseries.
Or, let the creator specify priority order.

docs/source/design/composite-time-series.rst
DanielTOsborne: As I commented elsewhere, I say make it work like aliases, so if you fetch a composite, it checks the composite list first, if not found there, then regular timeseries, or something like that.
Then you can have it arbitrarily named, such as `<Location Id>.<Parameter>.Var.0.0.Composite` or `<Location Id>.<Parameter>.Var.0.0.Composite;Lowess-SPK`

docs/source/design/composite-time-series.rst
DanielTOsborne: > That seems like the type of information to put in the version, if desired.
> 
> My assumption with the POR time series was that it should be suitable for "this is the full record as we have it" knowing that over time the official record has improved methods of measurement.

This is completely different from my understanding of what we were going to have for POR. Honestly, I can't think of any use case where having everything jumbled into a single time series is remotely useful. As it is, I have a hard time wanting to classify readings from two different sensor types (e.g. bubbler vs shaft-encoder) into a single POR. It's not the same data. Yes, it represents the same real-word measurement, but how useful is it to have them together? You can't run any worthwhile scientific/mathematical analysis on the data, since difference sensors respond in different ways and can throw off expectations.
Imagine training a ML model off bubbler recorded data, then trying to have it work with shaft-encoder data. It won't return expected results.

Also, what if we're actively recording the same measurement with two different sensors? Which do we put into the POR?

docs/source/design/composite-time-series.rst
msweier: @MikeNeilson that's an interesting take too. I was thinking most people would be interested in the best available which would include the data from your second point as well as recent realtime data. But would you differentiate the two? The USGS uses "Approved" and "Preliminary" qualifiers, but our qualifiers don't quite translate as well. There's a discourse on the qualifier discussion. https://discourse.hecdev.net/t/best-practices-of-cwms-data-qualifier-codes/3805

docs/source/design/composite-time-series.rst
MikeNeilson: > I'd argue this is out of scope, since this is a useful feature beyond this discussion.

It's not well documented but if you write to a text timeseries  with a TS ID that matches the data TS ID, that affectively becomes the "NOTES".

CDA itself should be setup to allow those entries and retrieve those entries in the current interface. But as you said, not in scope of this.

docs/source/design/composite-time-series.rst
MikeNeilson: > interval MUST be the same for all composite members. Trying to retrieve a composite time series where you have no idea what you're going to get is a nightmare to code for. If you want to combine different intervals, create a computation to create the requisite interval data from other intervals.

The intent is to avoid duplicating data while providing a simple name to the entire range of the Period of Record for the measurement at a location (that's where the this started from was  the period of record)

That said, I think I agree with the Duration. Or at least that if a duration is specified everything must match. But as a matter of historical record and how things changed, the durations do change and would be indicated in each.

I think one thing to point out, the results of this aren't meant for the *entire* PoR to be cleaning put into a display, but to provide the data as it was used at that was used at that time. Downstream users, say doing a study, would need to determine how they would interpolate/correct any gaps.

MikeNeilson: what about 

`New Bullards Bar.Elev[Composite]...`

As you said, it's an alias, that will be processed before further work and we don't use [] for anything so far as I'm aware. Certainly other options.

That said, since it is an alias. maybe we should just add yet-another .

- `New Bullards Bar.Elev.Inst.0.0.POR.Composite` basically the full history regardless of interval
- `New Bullards Bar.Elev.Inst.~1Day.0.Calc-val.Composite` useful for situations of multiple sensors and setting an "active".

That said, it would be more work to use in say, OpenDCS and any client software, but it is an option. But I do agree that fixing it to one location does limit things that appear rather useful.

docs/source/design/composite-time-series.rst
MikeNeilson: The ranges are stored within the composite time series "object" so the creator/curator is always specifying the "this TS ID is for (start,end]".

But reading this the 4th time, the language isn't clear. I should've said "member" instead of composite time series in those.

docs/source/design/composite-time-series.rst
MikeNeilson: oh, that's interesting. technically you wouldn't even need Composite in the name.... will have to think about that. Going to type it up though.

docs/source/design/composite-time-series.rst
MikeNeilson: > You can't run any worthwhile scientific/mathematical analysis on the data, since difference sensors respond in different ways and can throw off expectations.
Imagine training a ML model off bubbler recorded data, then trying to have it work with shaft-encoder data. It won't return expected results.

not automatically, no.

> Also, what if we're actively recording the same measurement with two different sensors? Which do we put into the POR?

For a generic composite time series, whatever you want. For "Period of Record" that's intended to be what was used to make any decisions.

docs/source/design/composite-time-series.rst
MikeNeilson: Agreed.

docs/source/design/composite-time-series.rst
MikeNeilson: My inclination for Period of Record, and thus authoritative, is the data when the operational decisions were made.

perhaps `is-authorized` should be, instead of boolean, `<some name>: (authoritative|raw|por|analyzed)` or something like that.

MikeNeilson (commented):
I've updated the document with the things we seem to agree on.

For the other elements I've added some of the discussed options for historical reference and made some changes to the language of the text to avoid ambiguity.

MikeNeilson:
From Ruth (with a few responses):

> In the Axioms there is a discussion of the composite time-series not have any gaps.
Have you considered the case (water quality usually) where seasonal equipment is deployed or where a study collected data for one year and then stopped?

That is a fair point, no gaps makes that difficult, the more important issue is no overlap.

Would allowing gaps, but, so far as standards of use go, requiring information in the notes about seasons, or perhaps a link to a location level for seasonal placement?

Though reading that again, if it's only in place for a year, isn't that just the complete record? Or are you saying sometimes that same collection will start again in some another year sometimes?

> Is there an assumption that a composite time-series must be defined through today?

Perhaps each member should only be defined by a start date? that would make it clear that whatever the last member is valid until stated otherwise. I'll review the that section, allowing end to be null/not set for the last member seems a reasonable compromise.

> Having to enter explanations every time there is a range with no data is a pretty big burden.

We should be able to create a standard way to set that to reduce the burden, alternatively perhaps just require we have such information in the notes section for the whole composite time series?

For a period of time, say when a station was catastrophically removed from service I think it does make sense to 

> Operations required
> Might want to add the ability to edit a member range, specifically the begin and end date.

That would be covered under the update operations as an editable field. However, the question does helpfully reiterate that we should describe exactly which fields are updateable or not.

> Can a composite time-series include a versioned time-series as a member?

A versioned time-series yes, depending on a specific version, we're so far thinking now, that when queried by "the composite time series" the data should always be the latest.

> Can one edit member time-series when reviewing the composite time-series? (CWMS-VUE I think this is to be determined)

That will have to be determined by CWMSVue itself. We have decided that one cannot write *to* the composite time series. E.G. the system won't go figure out which member it would need to write to and then do so. However, CWMSVue and other interfaces, are free to behave in which ever way they desire.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
+#####################
+Composite Time Series
+#####################
+
+Purpose
+=======
+
+It is a challenge for users to identity what the correct authoritative time series is for a given measurement at a location, when there are multiple time series at the same location. Additionally these time series often change over time, either being completely new or changing their interval as newer technologies become available.
+
+Gather an entire Period of Record for the value at a location is also rather difficult. And the POR record and "authoritative timeseries" may be one-in-the same.
```
krowvin: > Period of Record

Establish Period of Record (POR)

> POR record

nit; just say POR not Period of Record record

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
...
+
+
+Need
+====
+
+#. CWMS and Access-2-Water require a simple mechanism to allow users of data to retrieve the Authoritative Period of Record data for a given measurement without having to understand all of the possible component time series that may be involved.
```
krowvin: > Access-2-Water

Are we still calling the new site A2W? I only see mention of this branding in the FAQ. It might not be immediately obvious that https://water.usace.army.mil is A2W if users did not have prior knowledge. 

Could at least add a link here since this doc has nothing to do with that nomenclature!

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
...
+
+Gather an entire Period of Record for the value at a location is also rather difficult. And the POR record and "authoritative timeseries" may be one-in-the same.
+
+
+Need
+====
+
+#. CWMS and Access-2-Water require a simple mechanism to allow users of data to retrieve the Authoritative Period of Record data for a given measurement without having to understand all of the possible component time series that may be involved. 
+#. Period-of-Record time series *should* not be created by duplicating data from the component time series and merging them into a new one.
+#. The naming of the time series should fit within the excepting CWMS Time Series Identifier design and not unreasonably interfere with existing usages.
```
krowvin: > excepting

nit: I believe this is "accepting"

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
...
+
+
+Caveats
+=======
+
+#. It is assumed that CWMS-Vue will, as-always, require updates to handle what is created here.
+   #. e.g. we're not going to let any current limitations of CWMS-Vue hinder our design.
+
```
krowvin, line 26: Other Caveats to consider documenting here:

And these maybe more for the implementation phase and not fit here. But seems you might as well list any gotchas in this section? As a form of Q/A? (or make a Q/A section)
2. Will require each district to establish these Composite timeseries and define them as such
3. Will require updating A2W to possibly use these new Composite timeseries *replacing the legacy system*
4. Documentation on how to make these will need created 
5. The Standard Naming will need updated to include "composite timeseries"

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
...
+
+Proposal
+========
+
+Description
+-----------
+
+CDA should handle a concept of a "Composite Time Series". Whether a Time Series is considered composite will be determined by a specific element of the Time Series Identifier.
```
krowvin: > CDA

nit: CWMS Data API (CDA) - I realize this is probably a rough draft, so maybe just treat this as a reminder for the official doc?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
...
+Data Administrators will configure which Time Series (members), and the date-time range there-in, to define the composite time series.
+CDA will use this stored information to build the Composite Time Series during a query.
```
krowvin: I might elaborate more on "stored information", while I think it's insinuated here it might be worth saying this is not a duplicate of the data but a "reference to data that exists in their members that, when changed, will also update the composite TSID with such data"?

docs/source/design/composite-time-series.rst
krowvin: > time series "notes" will just come along.

Seamless notes associated with a given composite for, possibly, why it exists would be great. For us to not have to also query CLOB/text-timeseries/etc if the same call returned said meta notes.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
...
+
+Axioms
+------
+
+#. Composite Time Series are Irregular
+#. The definition of the composite time series is stored within the CWMS database
+#. The members of a composite time series define a continuous range
+
+   #. The date ranges of members *MUST* not overlap
+   #. The date ranges of members *MUST* not have any gaps
```
krowvin: To which I expect an error would be thrown

That is the data itself can overlap but the member date definitions cannot overlap or <insert error> is thrown. 

This leads more into the design but I imagine, if using CDA to build the composite, a `400` of some sort would get returned.

Enovotny:
> Perhaps each member should only be defined by a start date? that would make it clear that whatever the last member is valid until stated otherwise. I'll review the that section, allowing end to be null/not set for the last member seems a reasonable compromise.

I think this is the way to go.  That would take care of the issue with checking for gaps/overlap.  We should allow for a timeseries to be listed more then once to allow instances where Timeseies1, timeseries2, then timeseries 1 again.  If there are gaps in the timeseries like Ruth's examples those would just be displayed in the composite timeseries as it would be when looking at the original timeseries.  Just because the composite is grabbing from a timeseries between Date 1 and Date2 doesn't mean that timeseries needs to have values consistently between Date1 and Date 2.  The gaps would just pull as null.

docs/source/design/composite-time-series.rst
krowvin: > So intent is "there's oddly large amount of missing data, how do we report that."

# Some thoughts

Is it possible to build off the known extents of a given timeseries? 
I'm expecting if you are able to dynamically provide an interval to a composite timeseries, as some of the naming options suggest, then the interval for the gaps would be the same. 

## Enforcing what the user requests in the interval/etc
`<Location Id>.<Parameter>.<Parameter Type>.1Hour.<Duration>.Composite`
I.e. user provided they wanted a 1hour composite
1. A 1 hour timeseries May->Dec
2. A 15min timeseries Jan->Feb
And for the months of Mar, Apr that are missing you would get 1 hour null values back

## Segments? The "Don't know, Don't care" of Missing Intervals
Perryman, in version 1, of the timeseries had a segment concept that would build TS and leave literal gaps between segments. Doing something like this rules out the question of what interval you need on the response and leaves the client to decide what that gap can be. (Makes it a pain to plot and takes longer client-side, why v2 is so nice)

Whatever you decide +1 to just having gaps in a composite timeseries that are "just handled" and return `null`/etc for those values. (or nothing at all/segments)

Enovotny:
> > Can a composite time-series include a versioned time-series as a member?
> 
> A versioned time-series yes, depending on a specific version, we're so far thinking now, that when queried by "the composite time series" the data should always be the latest.

In this case I think the most likely scenario is that it would be grabbing from the aggregate not the actual dated versions.  That should probably be the default behavior.  But I guess having that ability to grab from a version might make sense in some cases I just can't really think of one at the moment.

docs/source/design/composite-time-series.rst
krowvin: > create a computation to create the requisite interval data from other intervals.

Tossing this out there, but *what if*, this computation was done (not by OpenDCS?) and cached to [(insert cache solution)](https://github.com/USACE/cwms-data-api/discussions/532) and the composite is pulled from that. 

Implementing this could be a pain but it lends into making *any* composite timeseries immediate when it's all well defined in the backend on what the data should be. Then you add additional metadata/fields to establish, much like the spec/templates for ratings, what is to be done with these composite timeseries given one of these situations.

>  just add yet-another .

Boy another part, I disagree there but then I wouldn't be doing all of that work!

Could instead add another sub-version, i.e. `calc-val-composite` and look for `composite` in the version. Issue there is if *anyone* already has a TS with that in the version...

docs/source/design/composite-time-series.rst
krowvin: > "is-authoritative" flag

To me this seems more explicit, because if what happens in a world when someone hears about composite and thinks it's okay to just have that in the version? 

I expect the UI would throw an error saying "You can't make a composite TS manually this way, you must XYZ" or "<Popup Appears>" prompting for proper composite creation when a user tries to make one with this standard name in the version?

docs/source/design/composite-time-series.rst
krowvin: Daniel is on it! Sane defaults for TS so that you have less of an awkward transition for districts to start adopting this too.

docs/source/design/composite-time-series.rst
krowvin: > I can't think of any use case where having everything jumbled into a single time series is remotely useful.

But then if you need both of these timeseries to complete a period of record to then present to (non-technical-group). You might present them as two lines on the same plot -as-is-.

Unless the scale is wildly different (sensor anomalies aside) a layperson will appreciate having it in one place for easy visuals, tabulation, and other data dissemination needs. I'm making an assumption people are using `Rev-Ccp` or other type `Rev` TS for this and not `RAW` TS...

The issue is compounded if we are asked to build POR timeseries with existing data and duplicate, breaking the earlier `*should* not do` classification.

In other words, i think the spirit of this is to make a tool available for the situations where it does make sense. And it's up to you, the arbiter of data, to decide where and *if* that makes sense for what you need. 

Perhaps the composite just points to your existing *singular* POR data and you're golden! Others are not so lucky to have that done already.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
...
+
+
+Operations required:
+
+* Create
+* Remove member (ts id + range)
+* Add member
+* List members
+* Replace all members?
+* Delete
```
krowvin, line 332: Should we add `Update`?

I.e. you want to update members, TSIDs, or any other such operations. And then when an update is issued do any triggers, or recalculations (mentioned above), need to fire? I.e. send a message down the que and anything subscribed knows to respond. (Report generation?!?)

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
...
+
+Supported Operations:
+
+* Get, through existing TimeSeries classes.
+
+
+Storage of member information
+================================
+
+#. Store in Clob as we refine the design - cache appropriately in member to avoid any major performance issues.
```
krowvin: If there's no preference, might we go for Blob, to allow existing libraries/frameworks to make potential web pages easy to make for this. 

i.e. if you store it in a JSON in BLOB we could build web interfaces with inputs to change them. 

Doable with CLOB too, but Blob is the one being heavily looked at.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,300 @@
...
+Add nullable "members" property.
+
+TimeSeriesDao
+-------------
+
+If the system sees the "Composite" marker/determines is composite retrieve the members for the range and build the time series.
+
+.. NOTE:: 
+    Considering the user may request the *entire* Period-of-record, this is a good opportunity to see that,
+    start the retrieval in a job queue, and return a status URL to the user for future download. I have see such mechanism
```
krowvin, line 401: I still think focusing on an edge cache solution, ValKey/Redis/etc, is worth the effort over piggybacking on a job queue and expecting the client to query when it's available. 

If you're a user wanting to see a plot of POR data. You might not want to wait 1,2,3,5+ minutes to see it. Whereas the edge cache can have it precomputed/rounded up for you with optional TTL. 

Perhaps even trigger the queue to compute this if it's stale to replace it in the cache.

krowvin (commented):
Added off-the-cuff thoughts. Nothing I said is a deal breaker. Any variation of what this becomes is better than the current proposed solutions that this would solve.

docs/source/design/composite-time-series.rst
MikeNeilson: That is the official name in the various repos and how it's referred to.

docs/source/design/composite-time-series.rst
MikeNeilson: actually that should be `existing`.

docs/source/design/composite-time-series.rst
MikeNeilson: Arguably both solutions are required as the very first request needs to update the cache. (if something happened to the cache). However, we have other situtations were the job queue is useful. 

Ex.

Should we really cache the Pine Flat Elev->Stor rating. It's 900 ft expanded to .01 ft. So large, and likely not queried often.
(Though I'll note, we should probably just store the ratings as XML/JSON not decompose them into tables, the tables do making the actual rating lookup a bit faster so. 6 of one, half dozen of the other.)

MikeNeilson:
> > > Can a composite time-series include a versioned time-series as a member?
> > 
> > 
> > A versioned time-series yes, depending on a specific version, we're so far thinking now, that when queried by "the composite time series" the data should always be the latest.
> 
> In this case I think the most likely scenario is that it would be grabbing from the aggregate not the actual dated versions. That should probably be the default behavior. But I guess having that ability to grab from a version might make sense in some cases I just can't really think of one at the moment.

Considering one gets the list of member time series, it's not difficult, where that need exists, to then retrieve a specific version. Given this will generate a time series that would by definition include multiple versioned time series, it would be near impossible to even establish "correct" behavior, much less implement it.

MikeNeilson:
> > Perhaps each member should only be defined by a start date? that would make it clear that whatever the last member is valid until stated otherwise. I'll review the that section, allowing end to be null/not set for the last member seems a reasonable compromise.
> 
> I think this is the way to go. That would take care of the issue with checking for gaps/overlap. We should allow for a timeseries to be listed more then once to allow instances where Timeseies1, timeseries2, then timeseries 1 again. If there are gaps in the timeseries like Ruth's examples those would just be displayed in the composite timeseries as it would be when looking at the original timeseries. Just because the composite is grabbing from a timeseries between Date 1 and Date2 doesn't mean that timeseries needs to have values consistently between Date1 and Date 2. The gaps would just pull as null.

I'm changing the end to "can be null." There's a situation were a given measure may be completely discontinued (station permanently removed, river routed, whatever) and setting that last one would make it clear.

docs/source/design/composite-time-series.rst
MikeNeilson: And the job queue is easier to implement. yes we can store the entire POR in the cache, but if the user doesn't request it, we still have to retrieve it, decompose it to the amount desired, and then return. Or store it decomposed. Either way, it's a fair amount of effort to implement.

MikeNeilson: (Not saying the effort isn't worth it, just that it's not really a simple determination.)

docs/source/design/composite-time-series.rst
msweier, line 1: _Overall looking good!_

It is a challenge for users to identity what the correct authoritative time series is for a given measurement at a location. 
_You may want to consider defining the authoritative timeseries up front._ 

_Regarding Option 5_
Parameter Type 	Marker that this time series does not have a fix information and is build of various member time series.
_How is this queried then if the parameter type is not fixed?_

MikeNeilson: > _Regarding Option 5_
> Parameter Type Marker that this time series does not have a fix information and is build of various member time series.
> _How is this queried then if the parameter type is not fixed?_

presumably the end user at least knows if they want say, elavation or storage?

> It is a challenge for users to identity what the correct authoritative time series is for a given measurement at a location.
You may want to consider defining the authoritative timeseries up front.

unfortunate the definitions depends on the previous information. yay language paradoxes!

msweier: Isn't parameter type "Inst", "Ave", "Max", etc...? Or am I thinking about this wrong?

Edit: Oh are you saying it's fixed for those standard options? If that's so, that makes sense.

MikeNeilson: In this context "fixed" would mean that all members would need to have the same type.

But that's just if used. I need to review that section and likely allow for a 0 or something to indicate that particular composite time series doesn't enforce that constraint (most likely POR would not given changes to data collection.)

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,379 @@
...
+
+The Period of Record (POR, period-of-record) for a measurement (such as the Stage at a river or the pool elevation of a 
+dam) is *ALL* available (time,value) pairs since recording began until either recording as ended or the most recent available
+regardless of changes in intervals and, if unavoidable changes in averaging.
+
+The POR is the "best" available combined dataset that would be desired for studies requiring all values for a given
+location.
+
+However, "best" is subjective. The POR of data used to make a given decision may not be the same as data that has formal
+validation. Additionally having a POR of the raw, unedited data may be what
```
adamkorynta: trailing statement

docs/source/design/composite-time-series.rst
adamkorynta: Is the idea that ValKey/Redis/etc would exist on local district servers and cache misses would go back to CDA (cloud)? Event queuing would be done to perform cache refreshes?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,379 @@
...
+
+    Perhaps we do this for data beyond "x amount"?
+
+Error handling and other conditions.
+====================================
+
+Versioned (date) time series
+----------------------------
+
+It is an error to specify a Version (date) when requesting composite data.
```
adamkorynta: Meaning composite time series always represent the latest version date for all member time series?

docs/source/design/composite-time-series.rst
MikeNeilson: resolving this comment. While the statement is correct. This is a data design document. Caching solutions are an implementation detail that would be handled in a separate decision (if it even needs a decision vs just doing it.)

docs/source/design/composite-time-series.rst
MikeNeilson: No, we don't allow saving directly to the member time series from this "object".

However, the messaging does provide a concern. Should saves to the underlying member time series also trigger a TS_STORED message for the composite. And if so how would it happen.

As there is a work around to that at the moment, simply pull the composite definition and filter client side for any TS_STORE message that for a member. That will be determined and handled at a later date.

docs/source/design/composite-time-series.rst
MikeNeilson: Correct

docs/source/design/composite-time-series.rst
MikeNeilson: oof, thanks.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+Hence generically we have a "composite time series"
+
+Definitions
+-----------
+
+Composite Time Series
+~~~~~~~~~~~~~~~~~~~~~
+
+A Time Series that is comprised of multiple same measure time series. For example, a river gage that has two sensors
+only one of which is valid for certain conditions.
```
rma-psmorris, line 56: This statement is fairly broad and conditions could be restricted to time if that is the goal here. Conditions could be an evaluation of the values, quality, or some independent data.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+
+However, on request for the timeseries the list of composite time series is consulted and used if present, otherwise passthrough to normal
+time series retrieval.
+
+
+Composite Time Series Definition
+================================
+
+.. code-block::jsonc
```
adamkorynta: The markup view in GitHub doesn't show this code block.... Heads up in case anyone is only viewing the rich diff instead of the source

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+
+
+
+Axioms
+------
+
+#. Composite Time Series are Irregular
+#. The definition of the composite time series is stored within the CWMS database
+#. The members of a composite time series define a continuous range
+   #. The date ranges of a member *MUST* not overlap
```
adamkorynta, line 126: Definition of the JSON should define whether the start/end dates are open or closed

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+~~~~~~~~~~~~~~~~~~~~~
+
+A Time Series that is comprised of multiple same measure time series. For example, a river gage that has two sensors
+only one of which is valid for certain conditions.
+
+Period of Record
+~~~~~~~~~~~~~~~~
+
+The Period of Record (POR, period-of-record) for a measurement (such as the Stage at a river or the pool elevation of a 
+dam) is *ALL* available (time,value) pairs since recording began until either recording as ended or the most recent available
```
rma-psmorris: I found this statement to be confusing to read, perhaps:
"The Period of Record (POR, period-of-record) for a measurement (such as the Stage at a river or the pool elevation of a dam) is ALL available (time,value) pairs since recording began until recording has ended or the most recent available pair, regardless of changes in intervals or unavoidable changes in averaging."

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+Naming Option 3, below, is selected for the path forward. A future design document will develop appropriate naming
+to communicate the intent of any given composite time series, include period-of-record.
+
+Authoritative
+~~~~~~~~~~~~~
+
+An authoritative time series is a period-of-record time series with the additional constraint that is contains the best
+official data. In other words the data that is determined to be "correct" by appropriate methods of validation.
+
+The data provide will have validated or corrected after events when additional information become available.
```
rma-psmorris: will have **been** validated?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+The data may not match what was used at the moment a decision was made.
+
+.. NOTE::
+
+    At time of design we are considering a boolean flag to indicate whether a time series is "the authoritative correct"
```
rma-psmorris, line 86: If this were an enum/state value instead of a boolean, it would support more than just authoritative vs non-authoritative.

docs/source/design/composite-time-series.rst
rma-psmorris, line 126: Given the time series is irregular, you could go with setting start/end to opposing open/closed dates. That way you dont have the end at 11:59 and the start at 12:00 with a possible miss of data at 11:59:30. You would be able to set both to 12:00 with convention stating that one is < while the other is >= (if end is open and start is closed).

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+------
+
+#. Composite Time Series are Irregular
+#. The definition of the composite time series is stored within the CWMS database
+#. The members of a composite time series define a continuous range
+   #. The date ranges of a member *MUST* not overlap
+   #. Each member *MUST* have a start date
+   #. The last member *MAY* have an end date indicating no more data will be available for this location and measure.
+   #. Data may have gaps, an explanation range *SHOULD* be provided. For data with regular gaps, e.g. season gauges
+      a description should be provided in the notes. A Link to a Location Level can be provided if the timing is well known.
```
rma-psmorris: Is this suggesting that location levels can be used within the composite?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+#. The members of a composite time measure the same thing. (e.g. all members are Elevation, not some are elevation and some are stage.)
```
rma-psmorris: Missed a word: composite time **series** measure

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+#. The interval and duration of each member *MAY* be different.
```
rma-psmorris: A statement should be made here in regards to parameter type as it is implied given allowing mixed durations.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter             |As the normal CWMS TS ID, the measurement (e.g. Stage, Precip, Elevation, flow, etc)                                    |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Parameter Type        |As Normal CWMS TS ID, Instantaneous, average, total, etc                                                                |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Interval -\> Composite| Marker that this time series does not have a fix information and is build of various member time series.               |
++----------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration -\> var      |Duration of average or total may change over time with new members, duration will be indicated in the member definition |
```
rma-psmorris, line 156: I would suggest writing this out as Variable as that is already proposed in standard naming. This is of course dependent on variable durations being present given the want to have a consistent duration as noted above.

docs/source/design/composite-time-series.rst
rma-psmorris: To me, the composite time series should be the most granular series of time value data pairs for a given site/measurement. From this most granular time series, math operations can be used to derive a common interval, etc.

docs/source/design/composite-time-series.rst
rma-psmorris: To consider authoritative as part of the identifier, we would want to compare two composites for the same site/parameter one of which is authoritative and the other non-authoritative. How would those two time series identifiers be differentiated from one another? Does authoritative uniquely define the time series vs non-authoritative?

docs/source/design/composite-time-series.rst
rma-psmorris: This Option forces usage of a common duration of 0? Meaning the composed time series would not include any non-zero durations?

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+|Interval                |Interval of data elements. may change over time with new members, duration will be indicated in the member definition   |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Duration                |Duration of average or total. may change over time with new members, duration will be indicated in the member definition|
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version                 |Composite or POR ... or check for composite at the front/back?                                                          |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+
+From Daniel 
+
+Argument Against: the "Version" field it freeform and we often encode other information in it.
```
rma-psmorris, line 211: This is a problem inherent with the identifier design pattern. Once set, it becomes hard to capture new concepts/metadata/attributes. Use of an identifying data structure (XML/JSON) would capture this better.

docs/source/design/composite-time-series.rst
rma-psmorris: From reviewing all of the options, this option 2 is my preferred.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+|Duration                |Duration of average or total. may change over time with new members, duration will be indicated in the member definition|
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+|Version                 |As Normal CWMS TS ID                                                                                                    |
++------------------------+------------------------------------------------------------------------------------------------------------------------+
+
+However, on request for the timeseries the list of composite time series is consulted and used if present, otherwise passthrough to normal
+time series retrieval.
+
+
+Composite Time Series Definition
```
rma-psmorris, line 267: Note that this is not showing up in the marked up view of this file.

docs/source/design/composite-time-series.rst
```diff
@@ -0,0 +1,386 @@
...
+        }
+    ]
+    // array above *should* be sorted by start when provided to user.
+    }
+
+
+Operations required:
+--------------------
+
+* Create
```
rma-psmorris, line 326: A filtered catalog of composites will be useful capturing a set of composites including the list members content.

docs/source/design/composite-time-series.rst
rma-psmorris: If a composite is depended on externally, shouldnt a change in the definition of the composite trigger an event/message for dependents to be able to re-retrieve the definition and react accordingly?

rma-psmorris (commented):
Feedback on design

docs/source/design/composite-time-series.rst
MikeNeilson: The idea primarily stems from having a "period of record" whose intervals will have changed over time as measurement technology improved. So the granularity of those would be variable. Additionally the exact usage is unknown and further interval manipulations should be sent to left to the end user of the data set.

docs/source/design/composite-time-series.rst
MikeNeilson: All fair points, but beyond scope of this initial document.

docs/source/design/composite-time-series.rst
MikeNeilson: The statement is intentionally broad, it is in fact just a grouping of time series in some meaningful way. Meaningful ways vary.

docs/source/design/composite-time-series.rst
MikeNeilson: Thanks, the end of that description definitely got away from me.

docs/source/design/composite-time-series.rst
MikeNeilson: thanks.

docs/source/design/composite-time-series.rst
MikeNeilson: Yeah, we're all just not entirely sure how that will work out. I may just make it a "state" field or something. Once we have some solid example we'll likely refine the design with additional discussions.

docs/source/design/composite-time-series.rst
MikeNeilson: No, that was more link to a seasonal location level so that downstream users can decide if a gap is "expected or not."

docs/source/design/composite-time-series.rst
MikeNeilson: thanks.

docs/source/design/composite-time-series.rst
MikeNeilson: That is correct, I've now stated it specifically.

docs/source/design/composite-time-series.rst
MikeNeilson: No, not specifically. Not enough that we should rely on the time series id version field anyways.

docs/source/design/composite-time-series.rst
MikeNeilson: will make sure that shows in the render output though.

docs/source/design/composite-time-series.rst
MikeNeilson: That's a fair point. Added events to document.

docs/source/design/composite-time-series.rst
MikeNeilson: I added that to the text.

PR #1105, opened 2025-05-19
Fixes #1051: Blob and Clob ID case sensitivity

### LINKED ISSUE #1051: Case senstivity in BLOB/CLOB ids?
Seems like BLOBs and CLOBs are changing all characters in the ID to upper case when stored and can't be retrieved without calling them in all upper case. Is that intentional?

zack-rma:
Fixes #1051 

Added toUpper on clob and blob IDs, added test coverage

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
@@ -115,7 +115,7 @@ public class BlobController implements CrudHandler {
             ContentType contentType = Formats.parseHeader(formatHeader, Blobs.class);
 
             BlobDao dao = new BlobDao(dsl);
-            List<Blob> blobList = dao.getAll(office, like.toUpperCase());
+            List<Blob> blobList = dao.getAll(office, like);
 
             Blobs blobs = new Blobs.Builder(cursor, pageSize, 0).addAll(blobList).build();
             String result = Formats.format(contentType, blobs);
@@ -156,9 +156,9 @@ public class BlobController implements CrudHandler {
                 }
             };
             if (office.isPresent()) {
-                dao.getBlob(blobId.toUpperCase(), office.get(), tripleConsumer);
+                dao.getBlob(blobId, office.get(), tripleConsumer);
             } else {
-                dao.getBlob(blobId.toUpperCase(), tripleConsumer);
+                dao.getBlob(blobId, tripleConsumer);
             }
         }
     }
@@ -186,9 +186,8 @@ public class BlobController implements CrudHandler {
             boolean failIfExists = ctx.queryParamAsClass(FAIL_IF_EXISTS, Boolean.class).getOrDefault(true);
             ContentType contentType = Formats.parseHeader(formatHeader, Blob.class);
             Blob blob = Formats.parseContent(contentType, ctx.bodyAsInputStream(), Blob.class);
-            Blob blobUpper = new Blob(blob.getOfficeId(), blob.getId().toUpperCase(), blob.getDescription(), blob.getMediaTypeId(), blob.getValue());
             BlobDao dao = new BlobDao(dsl);
-            dao.create(blobUpper, failIfExists, false);
+            dao.create(blob, failIfExists, false);
             ctx.status(HttpCode.CREATED);
         }
     }
@@ -217,7 +216,6 @@ public class BlobController implements CrudHandler {
 
             ContentType contentType = Formats.parseHeader(formatHeader, Blob.class);
             Blob blob = Formats.parseContent(contentType, ctx.bodyAsInputStream(), Blob.class);
-            Blob blobUpper = new Blob(blob.getOfficeId(), blob.getId().toUpperCase(), blob.getDescription(), blob.getMediaTypeId(), blob.getValue());
 
             if (blob.getOfficeId() == null) {
                 throw new FormattingException("An officeId is required when updating a blob");
@@ -233,7 +231,7 @@ public class BlobController implements CrudHandler {
             }
 
             BlobDao dao = new BlobDao(dsl);
-            dao.update(blobUpper, false);
+            dao.update(blob, false);
             ctx.status(HttpServletResponse.SC_OK);
         }
     }
@@ -256,7 +254,7 @@ public class BlobController implements CrudHandler {
             DSLContext dsl = getDslContext(ctx);
             String office = requiredParam(ctx, OFFICE);
             BlobDao dao = new BlobDao(dsl);
-            dao.delete(office, blobId.toUpperCase());
+            dao.delete(office, blobId);
             ctx.status(HttpServletResponse.SC_NO_CONTENT);
         }
     }
--- a/cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
@@ -96,7 +96,6 @@ public class BlobController implements CrudHandler {
             DSLContext dsl = getDslContext(ctx);
             String office = ctx.queryParam(OFFICE);
 
-
             String cursor = queryParamAsClass(ctx, new String[]{PAGE, CURSOR},
                     String.class, "", metrics, name(BlobController.class.getName(), GET_ALL));
 
@@ -116,7 +115,7 @@ public class BlobController implements CrudHandler {
             ContentType contentType = Formats.parseHeader(formatHeader, Blobs.class);
 
             BlobDao dao = new BlobDao(dsl);
-            List<Blob> blobList = dao.getAll(office, like);
+            List<Blob> blobList = dao.getAll(office, like.toUpperCase());
 
             Blobs blobs = new Blobs.Builder(cursor, pageSize, 0).addAll(blobList).build();
             String result = Formats.format(contentType, blobs);
@@ -157,9 +156,9 @@ public class BlobController implements CrudHandler {
                 }
             };
             if (office.isPresent()) {
-                dao.getBlob(blobId, office.get(), tripleConsumer);
+                dao.getBlob(blobId.toUpperCase(), office.get(), tripleConsumer);
             } else {
-                dao.getBlob(blobId, tripleConsumer);
+                dao.getBlob(blobId.toUpperCase(), tripleConsumer);
             }
         }
     }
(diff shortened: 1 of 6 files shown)


adamkorynta:
The responsibility of knowing the database forces ID's to uppercase belongs to the DAO, not the controller. If we were validating user input to require upper case then that would be a controller responsibility.

cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
```diff
@@ -116,7 +115,7 @@ public void getAll(@NotNull Context ctx) {
             ContentType contentType = Formats.parseHeader(formatHeader, Blobs.class);
 
             BlobDao dao = new BlobDao(dsl);
-            List<Blob> blobList = dao.getAll(office, like);
+            List<Blob> blobList = dao.getAll(office, like.toUpperCase());
```
MikeNeilson: do a search for "caseInsenstiveRegex" (I think that's it.) vs just doing an upper case.

MikeNeilson (changes requested):
The tests are good.

In the future though, when you find yourself retypeing the name simple thing over and over again. Take a step back, breath, and ask yourself:

1. Why am I doing that?
2. Is this the right place to do this?
3. Should something else be responsible for this behavior?

If you can't answer those you may just be in brand new territory and there's not a right answer. Or you may just be missing something because often some vagaries are hidden (like runtime exceptions already being taken care of so most of the code can look "clean").

In either case reach out to other. It may make sense to do a small amount of work to say "I think I need to do this" but unfortunately that's really just a judgement call you have to make.

cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
zack-rma: I added `caseInsensitiveRegex` to the clob DAO, looks like the blob DAO was already calling `upper()` within the queries

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -43,9 +43,9 @@ public Optional<Blob> getByUniqueName(String id, String limitToOffice) {
         ResultQuery<Record> query;
         if (limitToOffice != null && !limitToOffice.isEmpty()) {
             queryStr = queryStr + " and CWMS_OFFICE.OFFICE_ID = ?";
-            query = dsl.resultQuery(queryStr, id, limitToOffice);
+            query = dsl.resultQuery(queryStr, id.toUpperCase(), limitToOffice);
```
adamkorynta: The queries do not have an upper on the ID, but you're uppering the input. Since this was already missed once, someone is likely going to come by and wonder why there is a mismatched upper. Save me/them/us the hassle and add a comment.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -158,14 +158,14 @@ public void update(Blob blob, boolean ignoreNulls) {
...
-            throw new NotFoundException("Unable to find blob with id " + blob.getId() + " in office " + blob.getOfficeId());
+        if (!blobExists(blob.getOfficeId(), blob.getId().toUpperCase())) {
+            throw new NotFoundException("Unable to find blob with id " + blob.getId().toUpperCase() + " in office " + blob.getOfficeId());
         }
 
         dsl.connection(c -> CWMS_TEXT_PACKAGE.call_STORE_BINARY(
                 getDslContext(c, blob.getOfficeId()).configuration(),
                 blob.getValue(),
-                blob.getId(),
+                blob.getId().toUpperCase(),
```
MikeNeilson: while all the comparisons should be upper case, we do want the user to be able to set the casing to whatever they want for visualization and readability.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -158,14 +158,14 @@ public void update(Blob blob, boolean ignoreNulls) {
             throw new NotFoundException("Null blob provided to update");
         }
 
-        if (!blobExists(blob.getOfficeId(), blob.getId())) {
-            throw new NotFoundException("Unable to find blob with id " + blob.getId() + " in office " + blob.getOfficeId());
+        if (!blobExists(blob.getOfficeId(), blob.getId().toUpperCase())) {
```
MikeNeilson: blob exists already does the upper, it makes sense to have blobExists do the upper.

Perhaps just add javadoc to blobExists that an upper case comparison will be done.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -142,7 +142,7 @@ public void create(Blob blob, boolean failIfExists, boolean ignoreNulls) {
         dsl.connection(c -> CWMS_TEXT_PACKAGE.call_STORE_BINARY(
                 getDslContext(c, blob.getOfficeId()).configuration(),
                 blob.getValue(),
-                blob.getId(),
+                blob.getId().toUpperCase(),
```
MikeNeilson: Same as below comment. The casing can vary, only the comparison/lookup is case insensistive.

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
```diff
@@ -170,7 +170,7 @@ public void create(Clob clob, boolean failIfExists) {
         dsl.connection(c -> CWMS_TEXT_PACKAGE.call_STORE_TEXT(
             getDslContext(c, clob.getOfficeId()).configuration(),
             clob.getValue(),
-            clob.getId(),
+            clob.getId().toUpperCase(),
```
MikeNeilson: Same as comment in blow, you're taking the user control away here.

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
```diff
@@ -235,14 +235,14 @@ public void getClob(String clobId, String officeId, ClobConsumer clobConsumer) {
...
-                preparedStatement.setString(2, clobId);
+                preparedStatement.setString(2, clobId.toUpperCase());
 
                 try (ResultSet resultSet = preparedStatement.executeQuery()) {
                     if (resultSet.next()) {
                         java.sql.Clob clob = resultSet.getClob("VALUE");
                         clobConsumer.accept(clob);
                     } else {
-                        throw new NotFoundException("Unable to find clob with id " + clobId + " in office " + officeId);
+                        throw new NotFoundException("Unable to find clob with id " + clobId.toUpperCase() + " in office " + officeId);
```
MikeNeilson: we shouldn't upper case the user message.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
MikeNeilson: honestly, I think a lot of the solution should just be changing the `where id=?` to `where upper(id)=upper(?)`.

I believe there's even an upper index on that column which would help the planner.

MikeNeilson (changes requested):
I'm not sure how we didn't really see this before, but you have gone somewhat overboard with the .toUpperCase calls.

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
zack-rma: Reverted

cwms-data-api/src/main/java/cwms/cda/data/dao/ClobDao.java
zack-rma: Reverted

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
zack-rma: Reverted

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
zack-rma: Reverted. Added relevant Javadoc

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
zack-rma: Reverted

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
zack-rma: Added `upper()` to queries

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -23,11 +23,11 @@ public class BlobDao extends JooqDao<Blob> {
             + "FROM CWMS_20.AT_BLOB \n"
             + "join CWMS_20.CWMS_MEDIA_TYPE on AT_BLOB.MEDIA_TYPE_CODE = CWMS_MEDIA_TYPE.MEDIA_TYPE_CODE \n"
             + "join CWMS_20.CWMS_OFFICE on AT_BLOB.OFFICE_CODE=CWMS_OFFICE.OFFICE_CODE \n"
-            + "WHERE ID = ? and CWMS_OFFICE.OFFICE_ID = ?";
+            + "WHERE upper(ID) = upper(?) and CWMS_OFFICE.OFFICE_ID = ?";
```
MikeNeilson: Put the office id clause first.

It might sound counter intuitive to have to deal with it but SQL Planners need a lot of hints to properly plan. One of the hints CDA queries need to help with providing is "what narrows this down the most soonest." In our case limiting by office first almost invariably does that.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
```diff
@@ -179,15 +179,26 @@ public void delete(String office, String id) {
...
+     * @param office the office associated with the blob
+     * @param id     the unique identifier for the blob
+     * @return true if the blob exists; false otherwise
+     */
     private boolean blobExists(String office, String id) {
         String existsQuery = "select 1 "
                 + "from CWMS_20.AT_BLOB \n"
                 + "join CWMS_20.CWMS_OFFICE on AT_BLOB.OFFICE_CODE = CWMS_OFFICE.OFFICE_CODE \n"
-                + "WHERE ID = ? AND upper(CWMS_OFFICE.OFFICE_ID) = upper(?)";
+                + "WHERE upper(ID) = upper(?) AND upper(CWMS_OFFICE.OFFICE_ID) = upper(?)";
```
MikeNeilson: same as previous query.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
MikeNeilson: (it does mostly depend on the indexes, so far as I know we haven't mistakenly *not* put the office field first anywhere.)

MikeNeilson (changes requested):
Okay, there's one subtle nitpick on some queries, but otherwise this all looks reasonable.

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
zack-rma: Moved office ID clause into first position

cwms-data-api/src/main/java/cwms/cda/data/dao/BlobDao.java
zack-rma: Moved office ID clause into first position

MikeNeilson (approved):

adamkorynta (approved):

Issue #1080, opened 2025-04-15
Failed Locations Post Returns 200 for CDA 2025.03.05


msweier:
I'm trying to write a new location and I get a 200 response that it can't parse the JSON and that it created the location. It did not create the location.

```bash
curl -X 'POST' \
  'SERVER/mvp-data/locations' \
  -H 'accept: */*' \
  -H 'Authorization: apiKey myKey' \
  -H 'Content-Type: application/json' \
  -d '{
    "office-id": "MVP",
    "name": "BallClub",
    "latitude": 47.33181227,
    "longitude": -93.8929429,
    "active": true,
    "public-name": "Ball Club River Near Ball Club, MN",
    "long-name": "Ball Club River Near Ball Club, MN",
    "timezone-name": "US/Central",
    "location-kind": "SITE",
    "nation": "US",
    "state-initial": "MN",
    "county-name": "Itasca",
    "nearest-city": "Ball Club",
    "horizontal-datum": "NAD83",
    "vertical-datum": "LOCAL",
    "map-label": "Ball Club River Near Ball Club, MN",
    "bounding-office-id": "MVP",
    "elevation-units": "ft"
}'
```

![Image](https://github.com/user-attachments/assets/d314b3e3-4c6d-4351-9645-4d60ca1828c9)

msweier:
So `BallClub` was an existing location alias. What happened was the location details clobbered the existing location `WILM5` fields with the specified fields in the above json. Not sure if that behavior is expected. I would expect it would require the patch endpoint for an existing alias?

Pushing the above json to `BallClubTest` works as expected; however, I still get the `can't parse JSON` return which is odd.

MikeNeilson:
Hmm, that's going to take some looking into, Kinda wonder would happen if you tried something similar in CWMSVue (assuming there was a edit all of location at once, I suspect we'd see the same behavior.)

The can't parseJson expected for the data returned `Created Location` isn't valid json, `"Created Location"` would be. But clearly we need some better feedback.

Oh, I think there is a fail if exists query parameter. Though maybe not as it should be the default.

msweier:
I did the CWMS-VUE test already. It does not produce the same behavior where it overwrites the the data of the original location. A little detail here:
https://discourse.hecdev.net/t/cwmsvue-4-0-rc58-cant-save-new-location/4046

zack-rma:
[...]

PR #1111, opened 2025-05-27
Updated response for location creation

zack-rma:
Addresses an issue identified in #1080. Broader issue described in #1112. 

Updates response message to be proper JSON format.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/LocationController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/LocationController.java
@@ -63,7 +63,6 @@ import cwms.cda.data.dao.LocationsDaoImpl;
 import cwms.cda.data.dto.Location;
 import cwms.cda.formatters.ContentType;
 import cwms.cda.formatters.Formats;
-import cwms.cda.formatters.FormattingException;
 import cwms.cda.formatters.UnsupportedFormatException;
 import io.javalin.apibuilder.CrudHandler;
 import io.javalin.core.util.Header;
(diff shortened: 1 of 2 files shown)


cwms-data-api/src/main/java/cwms/cda/api/LocationController.java
```diff
@@ -310,7 +300,7 @@ public void create(@NotNull Context ctx) {
             ContentType contentType = Formats.parseHeader(formatHeader, Location.class);
             Location locationFromBody = Formats.parseContent(contentType, ctx.body(), Location.class);
             locationsDao.storeLocation(locationFromBody);
-            ctx.status(HttpServletResponse.SC_OK).json("Created Location");
+            ctx.status(HttpServletResponse.SC_OK).json("{\"Created Location\"}");
```
MikeNeilson: Should this be an object or just a plain string without the { and }?

MikeNeilson (changes requested):
Just want to clarify the reason for object vs plain string, seems odd to have an object that only contains a string.

cwms-data-api/src/main/java/cwms/cda/api/LocationController.java
MikeNeilson: According to jsonlint.com what you have is still invalid json. (note: put the string between json() as-is except the escaping.)

zack-rma: Removed the { and }. Should be properly formatted now.

PR #1115, opened 2025-05-29
Feature/1092 1098 basic user management

### LINKED ISSUE #1092: Create `auth/user` end point for current user information.
Came up during https://github.com/USACE-WaterManagement/groundwork-water/pull/107

Application have no meaningful way to retrieve current user information or roles.
Additionally there is some information it would be useful to allow users to change (if we can also edit any upstream provider).

### LINKED ISSUE #1098: Create basic endpoints for account management
We are at the point of needed to allow for the management of users in environments where the original serveradmin can't be used.

Additionally, we are about to revamp the entire authorization system.

However, as that will not be ready soon enough for some on-going work, we need at least a basic ability to assign current privileges from CDA.

MikeNeilson:
I think there's still a bit more to do, but ready for discussion. Plus the tests just started failing on my laptop after adding one more (like fail to even start correctly) so checking on github.

closes #1092 
closes #1098


MikeNeilson:
NOTE: these are a stop gap until further work is done on authentication and so I'm not super worried about getting all the timing information in and such. 

I did on one, but only a handful of people are going to be using this, and even then not super often.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -670,6 +678,20 @@ protected void configureRoutes() {
...
+        addUserManagementHandlers();
+    }
+
+    private void addUserManagementHandlers() {
+        RouteRole[] adminRoles = new RouteRole[] { new Role("CWMS User Admins")};
+        RouteRole[] userRoles = new RouteRole[] {new Role("CWMS Users")};
+        crud("/users/{user-name}", new UsersController(metrics), adminRoles);
+        get("/roles", new GetRolesController(metrics), adminRoles);
+        get("/user/profile", new UserProfileController(metrics), userRoles);
+        post("/user/{user-name}/roles/{office-id}", new AddRoleController(metrics), adminRoles);
```
krowvin, line 698: Does this suggest that a user could have a space in multiple offices?


As opposed to 
`/user/{office-id}/roles/{user-name}`?


I would just think that office would be the base and a user would be requested from that office

This probably doesn't really matter since either way the values get passed in. It's just, to me an office seems like it would come first.

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UserProfileController.java
```diff
@@ -0,0 +1,73 @@
...
+        responses = @OpenApiResponse(
+                    content = {
+                        @OpenApiContent(from = User.class, type = Formats.JSON)
+                    },
+                    status = STATUS_200
+        ),
+        security = {
+            @OpenApiSecurity(name = "gets overridden allows lock icon.")
+        },
+        description = "View users's own information",
```
krowvin: `user's`* I believe

krowvin: Unless you meant multiple users then disregard!

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UsersController.java
```diff
@@ -0,0 +1,177 @@
...
+        throw new UnsupportedOperationException("Unimplemented method 'delete'");
+    }
+
+
+    @OpenApi(
+        queryParams = {
+            @OpenApiParam(allowEmptyValue = true, name = OFFICE, type = String.class,
+                    description = "Show only users with active privileges in a given office" ),
+            @OpenApiParam(name = PAGE,
+                    description = "This end point can return a lot of data, this "
```
krowvin, line 83: > This end point can return a lot of data

I don't know if this needs to be said, but I get the spirit of it

"Lots of data, don't forget about the page option"

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UsersController.java
```diff
@@ -0,0 +1,177 @@
...
+            String result = Formats.format(contentType, users);
+
+            ctx.result(result);
+            ctx.contentType(contentType.toString());
+        }
+    }
+
+    @OpenApi(
+        pathParams = {
+            @OpenApiParam(name = "user-name", required = true,
```
krowvin, line 140: `username` or `user-name` - to hyphen or not to hyphen

cwms-data-api/src/main/java/cwms/cda/api/auth/users/roles/AddRoleController.java
```diff
@@ -0,0 +1,61 @@
...
+    private final MetricRegistry metrics;
+
+    public AddRoleController(MetricRegistry metrics) {
+        this.metrics = metrics;
+    }
+
+    @OpenApi(
+        pathParams = {
+            @OpenApiParam(name = "office-id", required = true,
+            description = "Office for these roles"),
```
krowvin: I still think this should just be generic

I.e. in one place create a definition with a link to the offices endpoint 

```java
String OFFICE_DESCRIPTION = "Office Identifier 3/4 letter as returned by the office endpoint: " + OFFICE_ENDPOINT
```

Throwing this idea out there again

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
```diff
@@ -0,0 +1,224 @@
...
+import cwms.cda.security.CwmsAuthException;
+import cwms.cda.security.DataApiPrincipal;
+import usace.cwms.db.jooq.codegen.tables.AV_SEC_USERS;
+
+public class UserDao extends JooqDao<User> {
+    public static final FluentLogger logger = FluentLogger.forEnclosingClass();
+
+    private static final String GET_USER =
+        "select ut.userid as username,ut.email, ut.principle_name,groups.db_office_id as office,groups.user_group_id as role " +
+        " from cwms_20.at_sec_cwms_users ut " +
```
krowvin, line 47: I know we are not making more plsql packages/etc (raw SQL)

But how much longer will `cwms_20` really be around (usage of cwms_20)

Gotta do something to move forward I suppose

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
```diff
@@ -0,0 +1,224 @@
...
+            }
+        });
+        logger.atInfo().log("Roles '%s' removed for user '%s' and office '%s'", String.join(",", roles), user, office);
+    }
+
+    public List<String> getRoles() {
+        return dsl.connectionResult(c -> {
+            AuthDao.setSessionForAuthCheck(c);
+            final ArrayList<String> roles = new ArrayList<>();
+            try (PreparedStatement getUser = c.prepareStatement("select distinct user_group_id from cwms_20.at_sec_user_groups order by user_group_id asc");
```
krowvin, line 132: Is there not a jooq equivalent for this?

i.e.

```java
Result<Record1<String>> result = DSL.using(configuration)
    .selectDistinct(AT_SEC_USER_GROUPS.USER_GROUP_ID)
    .from(AT_SEC_USER_GROUPS)
    .orderBy(AT_SEC_USER_GROUPS.USER_GROUP_ID.asc())
    .fetch();
```

Or does it fall short because you need a `PreparedStatement`?

krowvin (commented):
Just tossing my thoughts out there!

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UserProfileController.java
MikeNeilson: actually it should be `users'`, but technically `users's` is valid. in that sentence it's both plural and "owned."

I think. I'm certainly not going to claim I got excellent marks in English classes, but I often remember that esoteric stuff fairly well.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
MikeNeilson: For most of the other data you would be correct, but in this case I don't think so.
The user is a the root of the infrastructure, and permissions to an office data are secondary.
E.g. the user "belongs" to the system as a whole.

Versus say a Time Series which "belongs" to a specific office.

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UsersController.java
MikeNeilson: That is the question.

cwms-data-api/src/main/java/cwms/cda/api/auth/users/roles/AddRoleController.java
MikeNeilson: Fair point, and it might already exist and I should use it.

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
MikeNeilson: "CWMS_20" forever, or at least until we've moved to Postgres (but to be fair maybe even then).

But as for the creation of PLSQL package, only when it improves/simplifies usage or is required for specific permissions issues. 
In this case there's no benefit to doing so.

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
MikeNeilson: There's no code gen for these tables, and it's too much work to make that happen just for this change.

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UsersController.java
```diff
@@ -0,0 +1,177 @@
...
+                            + " value, and can be obtained from the 'next-page' value in "
+                            + "the response."),
+            @OpenApiParam(name = PAGE_SIZE,
+                    type = Integer.class,
+                    description = "How many entries per page returned. Default "
+                            + DEFAULT_PAGE_SIZE + ".")
+        },
+        responses = @OpenApiResponse(
+                    content = {
+                        @OpenApiContent(from = User[].class, type = Formats.JSON)
```
rma-rripken: Doesn't User[] mean were returning a array of User ?   That doesn't seem right.  Usually we return a plural object so that there is a place to hang the paging attributes and a place for the builder to calculate next-page.

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
```diff
@@ -0,0 +1,224 @@
...
+            int total = 0;
+            String cursorUserId = null;
+            int pageSizeTmp = pageSize;
+
+            Condition whereClause = office == null ? DSL.noCondition()
+            : dsl.select(count(asterisk()))
+                 .from(vUserGroups)
+                 .where(upper(vUserGroups.DB_OFFICE_ID).eq(upper(office)))
+                 .and(vUserGroups.IS_MEMBER.eq("T")).asField().gt(1)
+            ;
```
rma-rripken: select(count(asterisk())) doesn't seem like something I'd expect in a whereClause.  Is this supposed to be just the where portion?

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
```diff
@@ -0,0 +1,224 @@
...
+                if (parts.length > 1) {
+                    cursorUserId = parts[0];
+                    total = Integer.parseInt(parts[1]);
+                    pageSizeTmp = Integer.parseInt(parts[2]);
+                }
+            }
+
+            Condition pagingCondition = cursorUserId == null
+                ? noCondition()
+                : userId.greaterThan(cursorUserId);
```
rma-rripken, line 203: The orderby is userId and officeId.   I would think that the cursor needs to include both

rma-rripken (changes requested):
Are there unit tests for User and Users?  JSON test resources make it really obvious what the input/outputs look like.

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UsersController.java
MikeNeilson: You are correct, I typed out the response before thinking I was going to simplify and not bother with any pagination.

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
MikeNeilson: For this particular query it's deriving some information from another table to decide if the user row should be returned or not. I'll try to figure out a comment to clarify that. well, and review and make sure i'm not over doing it.

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
MikeNeilson: Hmm, there was a reason I didn't, will experiment and review.

MikeNeilson:
> Are there unit tests for User and Users? JSON test resources make it really obvious what the input/outputs look like.

I'll add them, may help with sorting out the pagination cursor anyways.

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
rma-rripken: oh.  I missed the gt(1) at the end - that >1 applies to the count?  that makes a little more sense.

MikeNeilson: It's not very obvious, didn't look like there was a `greaterThan` format when I was writing it.
I did add a comment to explain, should help future readers identify that particular element in the future.

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
MikeNeilson: Since other fixes are taking too long. The root level element is the user which is independent of an office but may have permissions in one. So we are only paginating on the Users. the addition of the order by of the office just to maintain some consistency for the joining operations.

rma-rripken: That still doesn't seem right.  If a user has perms in 3 offices but the pagesize is 2 you won't ever get back the third office.  Maybe the pagesize needs to apply to an inner select that chooses the userId to include in the results and you return every office for the users in the page?

MikeNeilson: Yeah, I got the overall query wrong and am working on it now. It needs to be limited with an initial list of users and then have the roles added back in.

Basically the same flow at the Location catalog and how it adds in the aliases.

MikeNeilson:
@rma-rripken FYI, I have hit a snag with some insanity in the database tables. I'm going to take a break before I throw the database itself off a cliff.... I don't know how I would do it, I just know I would try to make it happen.

MikeNeilson:
@rma-rripken there's likely some slop you'll have to point out to be, but it all actually behaves now. Had some definite issues in the paging design itself.

rma-rripken (approved):
Looks good to me.  I'd still like to see unit test for User and Users serialization with json test resource.  I find those helpful to see what the actual json looks like.

MikeNeilson:
> Looks good to me. I'd still like to see unit test for User and Users serialization with json test resource. I find those helpful to see what the actual json looks like.

I'll at least try to muster myself to do that in a follow up. I had enough issues that I see the value better, even for something this simple, but at the moment I want to get this out for at least use within the CWBI setup.

PR #1119, opened 2025-06-06
Fixes 1017: TimeSeries retrieval unit conversion error handling

### LINKED ISSUE #1017: Timeseries retrieval should provide a reasonable error message on invalid unit conversions.
Entirely my own fault on selecting the wrong units, but "System error" isn't really helpful here.:
```
curl -X 'GET' \
  'https://<host>/cwms-data/timeseries?name=Black%20Butte-Pool.Stor.Inst.1Hour.0.Calc-val&office=SPK&unit=ft&begin=2025-01-27T00%3A00%3A00-00%3A00&trim=false' \
```

zack-rma:
Fixes #1017

Adds global exception handling for unit conversion errors to JOOQ DAO base class. Adds tests for timeseries, levels, and basins.

Updates documentation for Ratings and Locations, as controllers were mislabeled to allow units beyond just unit systems as input.

Swaps string query parameters in test for constants.


cwms-data-api/src/test/java/cwms/cda/api/rating/RatingsControllerTestIT.java
```diff
@@ -267,7 +286,7 @@ void test_get_one_latest() {
 			effectiveDate = response.path("simple-rating.effective-date");
 		}
 		assertNotNull(effectiveDate);
-		assertEquals("2016-06-06T00:00:00Z", effectiveDate);
+		assertEquals("2025-06-06T00:00:00Z", effectiveDate);
```
MikeNeilson: Not that we need to fix this in this test but it's odd that you needed to adjust the effective date, this should just be a fundamental part of test data and not change over time, except explicitly, which implies the source data would also change.

MikeNeilson (dismissed):

adamkorynta (dismissed):

cwms-data-api/src/test/java/cwms/cda/api/rating/RatingsControllerTestIT.java
zack-rma: I believe I understand what happened here. The original test had 2016 considered as the most recent date, since the 2025 date was in the future, and would thus be ignored as a most recent date. Now that this "future" date has passed, it is considered the most recent. This was an oversight in the design of the test, and it should have had a "future" date *much* farther out.

MikeNeilson: or configurable. But that makes sense.

adamkorynta (dismissed):

adamkorynta (approved):

PR #1126, opened 2025-06-09
CWMS-2024: New LRTS identifier support

zack-rma:
CWMS-2024: New LRTS identifier support through header and integration testing


cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -267,6 +265,7 @@ public class ApiServlet extends HttpServlet {
     public static final String DATA_SOURCE = "data_source";
     public static final String RAW_DATA_SOURCE = "data_source";
     public static final String DATABASE = "database";
+    public static final String IS_NEW_LRTS = "X-NEW_LRTS";
```
MikeNeilson: `X-CWMS-LRTS-Formatting`

or something like that, but at a minimum: `X-CWMS-...`

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesIdentifierDescriptorControllerTestIT.java
```diff
@@ -109,6 +110,107 @@ public void test_create_delete() throws JsonProcessingException, SQLException {
...
+                .log().ifValidationFails(LogDetail.ALL,true)
+                .accept(Formats.JSONV2)
+                .contentType(Formats.JSONV2)
+                .queryParam("office", OFFICE)
+                .queryParam(Controllers.METHOD,JooqDao.DeleteMethod.DELETE_ALL)
+                .header("Authorization", user.toHeaderValue())
+            .when()
+                .redirects().follow(true)
+                .redirects().max(3)
+                .delete("/timeseries/identifier-descriptor/" + tsId)
```
MikeNeilson:
```suggestion
                .delete("/timeseries/identifier-descriptor/{tsID}", tsId)
```

There's a minor improvement in the log output when using this form. Also makes it easier to look at when there's more than one parameter.

cwms-data-api/src/test/java/cwms/cda/api/rating/RatingsControllerTestIT.java
```diff
@@ -68,7 +68,7 @@ static void beforeAll() throws Exception
 		String ratingXml = readResourceFile("cwms/cda/api/Zanesville_Stage_Flow_COE_Production.xml");
 		ratingXml = ratingXml.replaceAll("Zanesville", EXISTING_LOC);
 		String ratingXml2 = ratingXml.replaceAll("2002-04-09T13:53:01Z", "2016-06-06T00:00:00Z");
-		String ratingXml3 = ratingXml.replaceAll("2002-04-09T13:53:01Z", "2025-06-06T00:00:00Z");
+		String ratingXml3 = ratingXml.replaceAll("2002-04-09T13:53:01Z", "2065-06-06T00:00:00Z");
```
MikeNeilson: Let's put this in a separate PR. You're already going to have to update several otherwise complete branches.... and I have one that needs it as well.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
adamkorynta: I'd argue that we don't want `LRTS` as we're talking about the interval, so something like `X-CWMS-Use-LocalRegularInterval`. But then again maybe client usage will only be thinking in context of time series so LRTS is clearer. Thoughts?

cwms-data-api/src/test/java/cwms/cda/api/rating/RatingsControllerTestIT.java
zack-rma: https://github.com/USACE/cwms-data-api/pull/1149 Created a PR here

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesIdentifierDescriptorControllerTestIT.java
zack-rma: Adjusted path format as suggested.

zack-rma:
The current endpoints that return the legacy LRTS interval identifier despite the database being set to use the new format are:

- BinaryTimeSeries - GET
- Levels - GET
- TimeSeriesGroups - GET

This will be updated if more problematic endpoints are discovered.

An issue has been created for cwms-database, and is visible here: https://github.com/HydrologicEngineeringCenter/cwms-database/issues/24

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
```diff
@@ -267,11 +288,135 @@ void test_create_get_delete_get() throws IOException {
...
+            .header("Authorization", user.toHeaderValue())
+            .header(ApiServlet.IS_NEW_LRTS, false)
+        .when()
+            .redirects().follow(true)
+            .redirects().max(3)
+            .post("/timeseries/binary/")
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_INTERNAL_SERVER_ERROR));
```
adamkorynta: Should be a client error

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
```diff
@@ -267,11 +288,135 @@ void test_create_get_delete_get() throws IOException {
...
+            .body("binary-values.size()", equalTo(0))
+        ;
+    }
+
+    @Test
+    void test_create_get_delete_get_lrts() throws Exception {
+
+        // Structure of test:
+        //
+        // 1)Try to create the binary time series without LRTS
```
adamkorynta:
```suggestion
        // 1)Try to create the binary time series using new LRTS format without the new LRTS format header
```

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
```diff
@@ -267,11 +288,135 @@ void test_create_get_delete_get() throws IOException {
...
+        // Retrieve the binary time series and assert that it exists
+
+        given()
+            .log().ifValidationFails(LogDetail.ALL,true)
+            .accept(Formats.JSONV2)
+            .queryParam(Controllers.OFFICE, OFFICE)
+            .queryParam(Controllers.NAME, tsIdentifier)
+            .queryParam(Controllers.BEGIN, "2004-05-01T12:00:00Z")
+            .queryParam(Controllers.END, "2007-05-19T16:00:00Z")
+            .header(ApiServlet.IS_NEW_LRTS, true)
```
adamkorynta, line 444: Add a test right after this that tests without the new header to assert that the old interval id is returned

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
```diff
@@ -267,11 +288,135 @@ void test_create_get_delete_get() throws IOException {
...
+        .when()
+            .redirects().follow(true)
+            .redirects().max(3)
+            .get("/timeseries/binary/")
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_OK))
+            .body("binary-values.size()", equalTo(3))
```
adamkorynta, line 454: Need to test that the new identifier is returned to the client as part of the `name`.

cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
```diff
@@ -331,6 +337,28 @@ protected static void createTimeseries(String office, String timeseries, int off
...
+    protected static void createTimeseries(String office, String timeseries, boolean requireNewLRTSIntervalFormat) throws SQLException {
+        CwmsDatabaseContainer<?> db = CwmsDataApiSetupCallback.getDatabaseLink();
+        db.connection((c)-> {
+            String requireBool = requireNewLRTSIntervalFormat ? "T" : "F";
+            int requireIntValue = requireNewLRTSIntervalFormat
+                ? REQUIRE_NEW_LRTS_ID_FORMAT : REQUIRE_OLD_LRTS_ID_FORMAT;
+            org.jooq.Configuration configuration = DSL.using(c).configuration();
+            CWMS_UTIL_PACKAGE.call_SET_SESSION_INFO(configuration,
+                SESSION_USE_LRTS_ID_FORMAT, requireBool, requireIntValue);
+            try(PreparedStatement stmt = c.prepareStatement(createTimeseriesQuery)) {
```
adamkorynta: This query does not distinguish between PRTS and LRTS using the old identifier

adamkorynta (changes requested):
Integration tests should test both new and old format and assert that the returned identifier/name matches the request old/new format.

If the client passes in the new identifier without the requisite header, that should be a client error 400, not an internal server error 500.

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
zack-rma: Added check for name value

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
zack-rma: Added test to confirm old interval ID is returned

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
zack-rma: Updated comment as suggested

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
zack-rma: Added error checking and regex matching, now throws CONFLICT client error.

cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
zack-rma: Refactored method to only be used for new interval identifier. Should an alternative query specifying LRTS vs PRTS be used here?

cwms-data-api/src/main/java/cwms/cda/api/BinaryTimeSeriesController.java
```diff
@@ -208,8 +213,20 @@ public void create(@NotNull Context ctx) {
...
             boolean replaceAll = ctx.queryParamAsClass(REPLACE_ALL, Boolean.class).getOrDefault(false);
 
-            dao.store(tts, maxVersion, replaceAll);
-            ctx.status(HttpServletResponse.SC_CREATED);
+            try {
+                dao.store(tts, maxVersion, replaceAll);
+                ctx.status(HttpServletResponse.SC_CREATED);
+            } catch (DataAccessException ex) {
+                Matcher matcher = TS_ID_NOT_FOUND.matcher(ex.getMessage());
+                if (matcher.find()) {
```
adamkorynta: What is this special catch block for?

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesIdentifierDescriptorControllerTestIT.java
```diff
@@ -109,6 +110,107 @@ public void test_create_delete() throws JsonProcessingException, SQLException {
...
+        String likePattern = "Alder Springs\\.Precip-Cumulative\\.Inst\\.12HoursLocal\\.0\\.DescriptorTEST_LRTS*";
+
+        // Check that we don't have any ts like this in the catalog.
+        List<String> names = getIdsLike(OFFICE, likePattern);
+        Assertions.assertTrue(names.isEmpty());
+
+        ObjectMapper om = JsonV2.buildObjectMapper();
+        TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
+
+        // Create a bunch of ts and store them.
```
adamkorynta: Rule: comments should explain why, not just what

I don't understand why an iteration in the test is useful here. If it is useful, then I'd recommend using a ParameterizedTest

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesIdentifierDescriptorControllerTestIT.java
```diff
@@ -109,6 +110,107 @@ public void test_create_delete() throws JsonProcessingException, SQLException {
...
+                .header(ApiServlet.IS_NEW_LRTS, false)
+                .queryParam("office",OFFICE)
+            .when()
+                .redirects().follow(true)
+                .redirects().max(3)
+                .post("/timeseries/identifier-descriptor/")
+            .then()
+                .log().ifValidationFails(LogDetail.ALL,true)
+            .assertThat()
+                .statusCode(is(HttpServletResponse.SC_INTERNAL_SERVER_ERROR));
```
adamkorynta: I can't think of a reason any test would assert that an internal server error was thrown. If we're testing that the time series can't be found, then it should be throwing a 404 error code, not a 500.

cwms-data-api/src/test/resources/cwms/cda/api/spk/bin_ts_create_lrts.json
```diff
@@ -0,0 +1,28 @@
+{
+  "office-id": "SPK",
+  "name": "TsBinTestLoc.Binary.Inst.1HourLocal.0.lrts",
```
adamkorynta: In the hec-nucleus PR we determined that anything < 2HourLocal just shouldn't be supported. The tests should use regular use-case data, I recommend 6HourLocal or 1DayLocal

adamkorynta (changes requested):

cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
adamkorynta: The SQL query doesn't set the interval offset so I'd guess by default it will create it as PRTS. The overload for cwms_ts.create_ts has an offset parameter

cwms-data-api/src/test/resources/cwms/cda/api/spk/bin_ts_create_lrts.json
zack-rma: I've swapped this and a few other usages to 1DayLocal.

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesIdentifierDescriptorControllerTestIT.java
zack-rma: I borrowed the structure of the test above it, I've updated it to store and retrieve just one ID to verify functionality.

cwms-data-api/src/main/java/cwms/cda/api/BinaryTimeSeriesController.java
zack-rma: That was designed to handle the error when a user attempts to store a timeseries using the new LRTS identifier without the necessary header. I've discovered the error handling is an issue elsewhere, so I've moved this logic to the JooqDao.

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesIdentifierDescriptorControllerTestIT.java
zack-rma: This was intended to test whether we can store the new interval format with the header set to false. I've updated the error handling and the expected response code.

cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
zack-rma: I've updated the timeseries creation method to take in an offset value.

adamkorynta (dismissed):

cwms-data-api/src/test/resources/cwms/cda/api/spk/bin_ts_create_irr.json
```diff
@@ -0,0 +1,29 @@
+{
+  "office-id": "SPK",
+  "name": "TsBinTestLoc.Binary.Inst.IR-Day.0.lrts",
```
MikeNeilson: Isn't that a DSS interval?

cwms-data-api/src/test/resources/cwms/cda/api/timeseriesprofile/ts_profile_timeseries.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Alder Springs.Precip-Cumulative.Inst.1HourLocal.0.cda-lrts",
+  "name": "Sacramento River.Elev.Total.1DayLocal.0.Raw",
```
MikeNeilson: Elev.Total doesn't make any sense. This one should remain Precip. (or change to Elev.Inst if that's better suited to the dataset in question.)

MikeNeilson (changes requested):
Have a few nitpicks,

Otherwise looks reasonable.

cwms-data-api/src/test/resources/cwms/cda/api/timeseriesprofile/ts_profile_timeseries.json
zack-rma: Reverted to `Precip-Cumulative`

cwms-data-api/src/test/resources/cwms/cda/api/spk/bin_ts_create_irr.json
zack-rma: It would appear so, I was referencing the wrong list of intervals. I've fixed it.

MikeNeilson (approved):

PR #1158, opened 2025-06-18
Added TimeSeriesFilteredController for filtered time series API, supp…

### LINKED ISSUE #620: Add Filtering Parameters for Greater Than, Less Than, and Order to Timeseries Endpoint
Districts have a need for their public water management websites to filter data (**values**) on the server before it is returned. 

**Suggest adding**: Greater Than, Less Than, and Order (desc/asc) to the Timeseries Endpoint for the return values
example: ```?filter=greater-than&filter-value=25```

Otherwise you find yourself getting full period of record data via the client and having to filter out data you do *NOT* need after the fetch. 

This makes for a very slow query and longer processing to filter on the client. 

I do not think this merits a v3 of the timeseries if it can backport to the existing codebase (i.e. url parameters to add functionality)

Some districts are running their own SQL queries in Cold Fusion. Having this feature implemented now would ensure they can release their website without the use of Cold Fusion (or any other backend processing scripts that effectively do this).

Other districts are using python to run SQL for filtering to create `JSON` files with this data.

rma-rripken:
…orting query parameters like value range, null filtering, and pagination.


cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
```diff
@@ -0,0 +1,302 @@
...
+                    @OpenApiParam(name = Controllers.TRIM, type = Boolean.class, description = "Specifies "
+                            + "whether to trim missing values from the beginning and end of the "
+                            + "retrieved values. "
+                            + "Only supported for:" + Formats.JSONV2 + " and " + Formats.XMLV2 + ". "
+                            + "Default is true."),
+                    @OpenApiParam(name = INCLUDE_ENTRY_DATE, type = Boolean.class, description = "Specifies "
+                            + "whether to include the data entry date of each value in the response. Including the data entry "
+                            + "date will increase the size of the array containing each data value from three to four, "
+                            + "changing the format of the response. Default is false."),
+                    @OpenApiParam(name = "asc", type = Boolean.class, description = "Specifies "
```
rma-rripken: Would it make sense to add the "asc" parameter to the normal TimeSeries end-point?   It doesn't seem to be a filter and more about presentation.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDao.java
```diff
@@ -25,10 +26,31 @@ void store(TimeSeries timeSeries, boolean createAsLrts,
...
+     * @param includeEntryDate
+     * @return requested TimeSeries
+     * @deprecated Use {@link #getTimeseries(String, int, TimeSeriesRequestParameters)}
+     *             instead.  Create a {@link TimeSeriesRequestParameters} instance and
+     *             call that overload.
+     */
+    @Deprecated
     TimeSeries getTimeseries(String cursor, int pageSize, String names, String office,
                              String unit, ZonedDateTime begin, ZonedDateTime end,
                              ZonedDateTime versionDate, boolean trim, boolean includeEntryDate);
```
rma-rripken, line 49: The number of parameters on this dao method are getting out of hand.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
MikeNeilson: I feel like if the call *actually* wants the data in reverse order, they should just deal with it client side.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDao.java
MikeNeilson: Yep, though I think what you've done is probably the only practical solution. At least it groups things.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
```diff
@@ -0,0 +1,302 @@
...
+
+            String office = requiredParam(ctx, OFFICE);
+
+            boolean ascending = ctx.queryParamAsClass("asc", Boolean.class).getOrDefault(true);
+            Double minValue = ctx.queryParamAsClass("min-value", Double.class).getOrDefault(null);
+            Double maxValue = ctx.queryParamAsClass("max-value", Double.class).getOrDefault(null);
+            boolean filterNulls = ctx.queryParamAsClass("filter-nulls", Boolean.class).getOrDefault(false);
+
+            // Having all the parameters in a RequestParameters class will make changing/extending the args easier.
+            FilteredTimeSeriesParameters ftsParams = new FilteredTimeSeriesParameters.Builder()
```
MikeNeilson: it seems like it would make sense to have a `fromContext` method in these builders. Not that they shouldn't have the individual methods, but that it does make sense for a "...Parameters" to know about the `Context:queryParam...` functions.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+
+    private void addLinkHeader(@NotNull Context ctx, FilteredTimeSeries fts, ContentType contentType) {
+        // Send back the link to the next page in the response header
+        try {
+            // Send back the link to the next page in the response header
+            StringBuilder linkValue = new StringBuilder(600);
+            String pageUrl = buildRequestUrl(ctx, fts.getPage());
+            linkValue.append(String.format("<%s>; rel=self; type=\"%s\"",
```
MikeNeilson, line 258: why aren't these just embedded in the opaque next-page cursor value?
That was the point of handling the cursor in the way we do, to be able to encode such information without the user needing to worry about those particulars.

MikeNeilson (commented):
Looks reasonable to me so far.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
rma-rripken: Does anyone or anything use the Link header value?  I was only doing it b/c the TimeSeriesController was doing it and that has been doing it from the start.

rma-rripken: Some of the page and next-page tokens encode more information than others.  I don't think any of them include every query parameter - they mostly include the cursor part.  Its mostly been up to the caller to take the token and add it as a parameter to the url.   At least that is how I understand it.   I think this Link header is only done on these two controller but its kinda nice in that is builds the full url.

MikeNeilson: it's supposed to be a single thing to said to the system so it *should* encode every parameter relevant to retrieving the next page of results.

I can see the value in both for some usages. but the page/cursor should be "all of it."

MikeNeilson: As far as the link header, I think only the time series. and I think @DanielTOsborne set that up original. I think it was before we settled on the cursor as well.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
rma-rripken: Should I remove the asc/desc flag from the Filtered end-point?  It was in the original request but I'm not sure how important that specific feature was.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
rma-rripken: That would let FilteredTimeSeriesParameters find all the parameters it normally takes from the context by itself and the Controller wouldn't have to do it?   That could work.  Would clean up the controller code.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
rma-rripken: A downside of the page containing every parameter is that we have to remember whenever we add Controller arguments that we also have to change the page parsing code.  If we switch to using Parameter objects there might be a better pattern.   Like if the user passed "page" we call fromPage(str) to build the parameter and if they didn't we call fromContext(context) to build it.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
MikeNeilson: I say remove it. But give @krowvin an hour or two to chime in.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
MikeNeilson: hmm. I think I like that idea. but it's definitely not the way we've, or at least I've, been thinking about and doing it. So a bit beyond the scope of this PR perhaps.

Given that this particular end point was already a bit of a special case, I think we leave what you have for know and we do a little investigation on the paging design after.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -676,6 +688,20 @@ protected void configureRoutes() {
 
         addProjectLocksHandlers("/project-locks/{name}", requiredRoles);
         addProjectLockRightsHandlers("/project-lock-rights/{project-id}", requiredRoles);
+
+
+        addUserManagementHandlers();
```
MikeNeilson, line 705: I think you're source branch needs updating. these are already in the target branch

cwms-data-api/src/main/webapp/rsql.html
```diff
@@ -0,0 +1,118 @@
+<!DOCTYPE html>
```
MikeNeilson, line 1: I'm not going to worry about it at the moment, and frankly this will be useful anyways, but we do now have the Read-The-Docs documentation setup in the docs folder.

cwms-data-api/src/main/webapp/times.html
```diff
@@ -0,0 +1,122 @@
+<!DOCTYPE html>
```
MikeNeilson, line 1: this one probably should've gone in the docs folder. But again, better to have something now, we'll worry about it later.

cwms-data-api/src/test/java/cwms/cda/data/dao/rsql/RSQLConditionBuilderTest.java
```diff
@@ -0,0 +1,187 @@
...
+        // Create a field resolver using the JooqFieldResolver
+        JooqFieldResolver resolver = new JooqFieldResolver(AV_TSV_DQU.AV_TSV_DQU);
+        RSQLConditionBuilder builder = new RSQLConditionBuilder(resolver);
+
+        Condition condition = builder.buildCondition("unit_id==EN;value>25");
+
+        assertNotNull(condition);
+        String conditionString = condition.toString();
+
+        assertTrue(conditionString.contains("\"AV_TSV_DQU\".\"UNIT_ID\" = 'EN'"));
```
MikeNeilson, line 36: do these end up as bind variables by the time it gets to the database?

MikeNeilson: Based on the visitor above it seems like they should, so I'm surprised you can test this way, unless it's just part of the `Condition::toString`?

MikeNeilson (commented):
I have one paranoid question relating to the generated queries.

But otherwise this look great.

Though something is wonky with the pull request, when looking at the full commits it's claiming to add the new usermanagement things we merged in recently, not sure what's going on there.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
rma-rripken: Tried this for FilteredTimeSeriesParameters.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
rma-rripken: Ok, I like the idea of the Controller or even the Parameter object parsing the paging cursor into a new Parameter object and the Dao not having to know anything about the paging token.  Might have to draw some pictures or have a call to make this really clear.  I think we can come up with a better pattern

cwms-data-api/src/test/java/cwms/cda/data/dao/rsql/RSQLConditionBuilderTest.java
rma-rripken: This test is relying on jOOQs Condition::toString.  Its not actually talking to the database, just using the view that it knows about from the jOOQ codegen.   I have two "Resolvers" that take the RSQL selectors and connect them to jOOQ fields.   I was hoping the JooqFieldResolver used in this test was going to make the timeseries impl really nice and clean.  It works when the jOOQ "from" clause is a simple view.  In reality the timeseries query is more complicated and I ended up using a different Resolver in the actual Dao code.  The JooqFieldResolver does work and makes the unit test cleaner.

rma-rripken: This example is operating on AV_TSV_DQU.UNIT_ID from the view.    JooqFieldResolver can find unit_id b/c it looks at all the table/view fields.    The Resolver I used in the actual Dao impl only knows about the fields you tell it about and I didn't include UNIT_ID b/c I figured it didn't make real-world sense.  Why would people give the end-point a timeseries_id and then ask it to filter by unit_id.  Didn't make real-world sense to me.   But RSQLConditionBuilder is written to potentially be more generic than just the timeseries usage so I just left the test in there.  It demonstrates that the RSQL syntax can have String comparisons.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
rma-rripken: Yes.  thank you for catching this, I will revisit.  I think the user management was merged after my branch started.  I did rebase but maybe this was missed.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesFilteredController.java
MikeNeilson: Oh, that makes sense. the DAO just handle actual values query values and some other component is responsible for the actual interpretation. Would allow either an opaque cursor or a navlink setup to be used with minimal difference, none of which is in the DAO itself.

PR #1160, opened 2025-06-21
CWMS-2096 - Updates ratings to allow for replacing base curve. 

### LINKED ISSUE #624: How to use CDA to update a single rating curve
```
Hi Adam and Ryan,

Can either of you go over ratings for me and Krishna with CDA. I am looking on the swagger page and it looks like posting ratings are all in xml? Is that correct? Will there also be a json format or is it all using the existing xml? If we needed to post a new rating curve to a rating spec what would the call be in CDA? We are starting to work on a script to store ratings in CDA form the USGS and I am unclear what format we need to use to post a new rating curve to a rating spec. Do you have an example from posting in cwms-vue?

Thanks

Eric Novotny, PhD
```

Got the following questions:
- Is the only way to POST ratings in xml?  
- Will there also be a json format or is it all using the existing xml?  
- If we needed to post a new rating curve to a rating spec what would the call be in CDA?  
- We are starting to work on a script to store ratings in CDA from the USGS and I am unclear what format we need to use to post a new rating curve to a rating spec.  
- Do you have an example from posting in cwms-vue?

rma-bryson:
Fixes https://github.com/USACE/cwms-data-api/issues/624

Fixes issue with errors not propagating. Fixes issue with wrong query parameter being added in for replace base curve. Adds logic to remove template if store-template flag is false.


cwms-data-api/src/test/java/cwms/cda/api/rating/RatingsControllerTestIT.java
rma-rripken, line 1: I think it would be fine to split RatingsControllerTestIT into multiple files.   It seems like a large number of these tests do an initial store and then a cleanup in a finally{}.   Maybe all those could be moved into one file and @BeforeEach @AfterEach used to simplify the pattern?

cwms-data-api/src/test/java/cwms/cda/api/rating/RatingsControllerTestIT.java
```diff
@@ -191,106 +335,501 @@ static void afterAll()
...
+				.map(sr -> (String) sr.get("effective-date"))
+				.filter(Objects::nonNull)
+				.collect(Collectors.toList());
+	}
 
-		effectiveDate = response.path("ratings.simple-rating[0].effective-date");
-		if (effectiveDate == null) {
-			effectiveDate = response.path("simple-rating.effective-date");
+	@NotNull
+	private static String getNormalizedNumberString(String val) {
```
rma-rripken: Curious.  I would have thought you'd parse into double and then compare those.

cwms-data-api/src/test/java/cwms/cda/api/rating/RatingsControllerTestIT.java
rma-bryson: Split into two classes, one which uses @BeforeAll and @AfterAll, and one which has various store/update calls but uses @AfterEach ... removed all try-finally blocks

cwms-data-api/src/test/java/cwms/cda/api/rating/RatingsControllerTestIT.java
rma-bryson: yeah, that's probably more robust. I updated to remove this method and just compare double values in each test case

rma-rripken (approved):

PR #1161, opened 2025-06-24
Issue 1107: Add rate limiting to rate endpoints

### LINKED ISSUE #1107: Add rate limiting to rate endpoints
Currently the rate endpoints require auth, @krowvin requested that we allow a mechanism for unauthenticated access, but we do not want to allow for a denial-of-service attack. Add a rate limiter as requested in: https://github.com/USACE/cwms-data-api/issues/641#issuecomment-2605468283. Another thought was based on the size of a curve, but that is likely more complicated and resource intensive than a rate limiter.

zack-rma:
Fixes #1107

Adds rate limiting to ratings endpoints. Designed for future generalized application to unauthorized endpoints.


cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -709,27 +706,48 @@ private void addRatingHandlers(RouteRole[] requiredRoles) {
...
+     * @param requiredRoles the user roles required to access the path.
+     */
     private void addRateLimitCheckAuth(String path, BaseHandler handler, RouteRole[] requiredRoles) {
         Set<RouteRole> roles = new HashSet<>(Arrays.asList(requiredRoles));
         staticInstance().before(path, ctx -> {
             try {
-                cdaAccessManager.manage(handler, ctx, roles);
-            } catch (CwmsAuthException ex) {
                 NaiveRateLimit.requestPerTimeUnit(ctx, REQUEST_LIMIT, REQUEST_LIMIT_UNIT);
+            } catch (HttpResponseException ex) {
```
MikeNeilson: FYI, using try-catch blocks for flow control in Java is generally a performance nightmare.

zack-rma: The performance hit only occurs when the rate limit is reached. The impact from exception handling should be overshadowed by the authorization database call. Exception handling would add approximately 17ms according to these benchmarks: https://www.baeldung.com/java-exceptions-performance. 

All that said, we could write our own rate limiter instead of using Javalin's, which may be useful if we never plan to have rate limits on authenticated users.

MikeNeilson: Oh sure, bring baeldung into it. :).
Anyways, fair, I think I'm still feeling rather burned from a tight loop in opendcs were that 17ms added up... alot. You're right in context here not as performance critical.

That said it *is* still awkward to see in Java; tends to be very python-esque technique. But balance of development ease and maintainability it probably makes sense to stick with until proven otherwise.

Though the user should already be authorized by the time it gets to this function; that's now also done in a before handler so an if statement checking for the principal (though with roles, not just not null) should work.

MikeNeilson: If there's an ordering issue, it may make more sense to move this to being within the AccessManager itself, now that we only have the one. rating limiting *is* an access issue.

zack-rma: I've gone ahead and moved the rate limiting logic into the CdaAccessManager. The ordering should now be resolved, as the check for rate limiting occurs in the `manage` method.

adamkorynta (approved):

MikeNeilson (approved):

PR #1163, opened 2025-06-27
Issue 1080: Location Post Update

### LINKED ISSUE #1080: Failed Locations Post Returns 200 for CDA 2025.03.05

zack-rma:
Fixes #1080

Adds `FAIL-IF-EXISTS` query parameter to locations create endpoint. Flag defaults to true.


cwms-data-api/src/test/java/cwms/cda/data/dao/basin/BasinDaoIT.java
```diff
@@ -172,7 +172,7 @@ private static void createTestLocation(String locationName, String kind) throws
         db.connection(c -> {
             try {
                 LocationsDaoImpl locationsDao = new LocationsDaoImpl(getDslContext(c, OFFICE_ID));
-                locationsDao.storeLocation(location);
+                locationsDao.storeLocation(location, false);
```
MikeNeilson, line 175: In the future consider providing a override that uses the default... though certainly debatable whether or not that's a good idea in the context of the our particular usage.

cwms-data-api/src/main/java/cwms/cda/api/LocationController.java
```diff
@@ -296,8 +307,15 @@ public void create(@NotNull Context ctx) {
             String formatHeader = ctx.req.getContentType();
             ContentType contentType = Formats.parseHeader(formatHeader, Location.class);
             Location locationFromBody = Formats.parseContent(contentType, ctx.body(), Location.class);
-            locationsDao.storeLocation(locationFromBody);
-            ctx.status(HttpServletResponse.SC_OK).json("Created Location");
+            boolean failIfExists = ctx.queryParamAsClass(FAIL_IF_EXISTS, Boolean.class).getOrDefault(true);
+            locationsDao.storeLocation(locationFromBody, failIfExists);
+            ctx.status(HttpServletResponse.SC_CREATED).json("Created Location");
+        } catch (AlreadyExists e) {
```
MikeNeilson: This controller has the error handling within as it's one of the first and was created before we established the error handling within the ApiServlet exception definition.

New code should follow the new scheme for error handling.

MikeNeilson (changes requested):

cwms-data-api/src/test/java/cwms/cda/data/dao/basin/BasinDaoIT.java
zack-rma: I was debating the same thing, I figured I'd update the usages now to avoid keeping around the previous method. I've added a deprecated override that sets the `FAIL-IF-EXISTS` flag to false. This should allow for compatibility for downstream applications that use the locations DAO directly.

cwms-data-api/src/main/java/cwms/cda/api/LocationController.java
zack-rma: I moved the error handling out into ApiServlet and JooqDao.

MikeNeilson (approved):

cwms-data-api/src/main/java/cwms/cda/api/LocationController.java
```diff
@@ -280,6 +281,13 @@ public void getOne(@NotNull Context ctx, @NotNull String name) {
                         @OpenApiContent(from = Location.class, type = Formats.XML)
                     },
                     required = true),
```
rma-bryson, line 283: @MikeNeilson - As a general design question, does it really make sense to include a fail-if-exists parameter for a create operation? Conceptually, create implies that the resource does not already exist (at least to me, but maybe i'm wrong there) — and if it does exist, the operation should fail with exception. Conversely, update should be used when the resource already exists. So I'm questioning the need for a parameterized fail-if-exists flag for create.

MikeNeilson: in general, no, but it's a fundamental part of our our concepts have worked in the past that just isn't mapping to the HTTP verbs as well as we'd like.

PR #1165, opened 2025-06-30
Issue #1113 - Updates to use 'units' as a query parameter to match timeseries json payload 'units'.

### LINKED ISSUE #1113: timeseries GET endpoint mismatch on unit/units parameter
The /timeseries GET endpoint accepts `unit` as the input parameter, but returns `units` in the response.
It should be consistent IMO. See /units

Possibly also the same with /levels.

rma-bryson:
Fixes https://github.com/USACE/cwms-data-api/issues/1113

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
@@ -287,8 +287,8 @@ public class TimeSeriesController implements CrudHandler {
                         + "Required for:" + Formats.JSONV2 + " and " + Formats.XMLV2 + ". "
                         + "For other formats, if this field is not specified, matching location "
                         + "level information from all offices shall be returned."),
-                @OpenApiParam(name = UNIT,  description = "Specifies the "
-                        + "unit or unit system of the response. Valid values for the unit "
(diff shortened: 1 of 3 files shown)


cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -287,16 +287,16 @@ public void delete(@NotNull Context ctx, @NotNull String timeseries) {
                         + "Required for:" + Formats.JSONV2 + " and " + Formats.XMLV2 + ". "
                         + "For other formats, if this field is not specified, matching location "
                         + "level information from all offices shall be returned."),
-                @OpenApiParam(name = UNIT,  description = "Specifies the "
-                        + "unit or unit system of the response. Valid values for the unit "
```
MikeNeilson, line 291: Leave the old parameter in the openapi spec but mark it as deprecated.

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
rma-bryson: added that back with deprecated flag

MikeNeilson (approved):

PR #1168, opened 2025-07-01
Fix NPE in OpenIdConnectIdentityProvider when not using OpenID

zack-rma:
Fixes NPE in OpenIdConnectIdentityProvider when not using OpenID


MikeNeilson (approved):
Sorry, my brain kept thinking I've already accounted for this, but this still makes sense to do anyways.

PR #1169, opened 2025-07-01
Added filtered time series retrieval API.

### LINKED ISSUE #620: Add Filtering Parameters for Greater Than, Less Than, and Order to Timeseries Endpoint

rma-rripken:
Introduced `TimeSeriesRequestParameters` and 'FilteredTimeSeriesParameters' to clean up TimeSeriesDaoImpl. Added RSQL query capability


rma-rripken:
Replacement PR b/c https://github.com/USACE/cwms-data-api/pull/1158 got closed when I force-pushed a new branch that didn't have the weird self-merge in its history

rma-rripken:
@MikeNeilson Last we talked you suggested merging this and letting people give feedback

MikeNeilson:
I was hoping some feedback would be provided on the PR, but merge and let be used is likely the more practical way at this point.

rma-rripken:
@MikeNeilson   Should I remove the explicit min-value, max-value and filter-nulls parameters from the controller?  With the filter-expression-language implemented those features can be achieved in an RSQL query.  Removing them might force people to figure out how to build their query?

MikeNeilson:
That seems rational to me, doesn't really make sense to have both and need to support two different paths.

rma-rripken:
@MikeNeilson   This is ready now

MikeNeilson (approved):

PR #1187, opened 2025-07-15
CWMS-2172: Enabled LRTS tests for Forecast Spec

### LINKED ISSUE #24: New LRTS interval identifier not being returned properly
While testing the implementation of the support for the new LRTS interval identifiers within cwms-data-access, (appending `Local` to the interval instead of prepending `~`) it was discovered that endpoints that handle data associated with these new intervals are still using the previous format. 

The methods currently known to have this behavior are:
- Binary TimeSeries: storage: ~~The storage method is failing to associate binary data with the timeseries ID. Data is not being stored into the `AT_TSV_BINARY` table. There could be an underlying issue with the association between the new and old interval id: calls into `CWMS_TEXT.RETRIEVE_TS_BINARY()`. Recent testing suggests this issue occurs for all irregular timeseries, further investigation is underway.~~ See #28 for details on this issue, as it applies to all irregular time series data.
- ForecastSpec: retrieval: uses custom query against the `AV_FCST_SPEC`, `AV_FCST_LOCATION`, and `AV_FCST_TIME_SERIES` views. Tests show that this issue persists when using the `CWMS_FCST_PACKAGE.call_RETRIEVE_FCST_SPEC` procedure in CDA.

ForecastSpec has been verified to return the legacy format (e.g. `~1Hour` instead of `1HourLocal`) when the `USE_NEW_LRTS_ID_FORMAT` session variable is set to level 6 - which should require the new format input and provide it on output.

zack-rma:
Fixes https://github.com/HydrologicEngineeringCenter/cwms-database/issues/24.

Enables LRTS interval identifier integration tests for Forecast spec retrieval.

### DIFF
--- a/cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
+++ b/cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
@@ -123,7 +123,7 @@ public class CwmsDataApiSetupCallback implements BeforeAllCallback,AfterAllCallb
                             .withOfficeId("HQ")
                             .withVolumeName(TeamCityUtilities.cleanupBranchName(ORACLE_VOLUME))
                             .withSchemaImage(CWMS_DB_IMAGE);
-            cwmsDb.withImagePullPolicy(PullPolicy.alwaysPull());
+            cwmsDb.withImagePullPolicy(PullPolicy.defaultPolicy());
             cwmsDb.start();
 
             final String jdbcUrl = cwmsDb.getJdbcUrl();
--- a/cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
+++ b/cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
@@ -49,7 +49,7 @@ public class CwmsDataApiSetupCallback implements BeforeAllCallback,AfterAllCallb
 
     private static final String ORACLE_IMAGE =
         System.getProperty("CDA.oracle.database.image",
-                           "registry-public.hecdev.net/cwms/database-ready-ora-23.5:latest-dev"
+                           "ghcr.io/hydrologicengineeringcenter/cwms-database/cwms/database-ready-ora-23.5:latest-dev"
                        );
     private static final String ORACLE_VOLUME =
         System.getProperty("CDA.oracle.database.volume",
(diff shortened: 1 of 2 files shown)


adamkorynta (dismissed):

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
```diff
@@ -123,7 +123,7 @@ public void beforeAll(ExtensionContext context) throws Exception {
                             .withOfficeId("HQ")
                             .withVolumeName(TeamCityUtilities.cleanupBranchName(ORACLE_VOLUME))
                             .withSchemaImage(CWMS_DB_IMAGE);
-            cwmsDb.withImagePullPolicy(PullPolicy.defaultPolicy());
+            cwmsDb.withImagePullPolicy(PullPolicy.alwaysPull());
```
MikeNeilson: Should be "if not present"

If not present will still check if it's up-to-date or not.

MikeNeilson (changes requested):

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
zack-rma: I didn't see an available "if not present" policy, but I've reverted to the default policy. The db image link change resolved the build error, so the pull policy shouldn't be an issue.

rma-rripken: @MikeNeilson Zack reverted back to defaultPolicy.  Can you resolve your requested change so that the PR can be merged?

adamkorynta (approved):

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
MikeNeilson: Yeah, we need to figure out how to override that when I randomly go on vacation and forget to use "comment" instead of "request changes"

MikeNeilson (approved):

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
adamkorynta: I do have the ability to dismiss your review, but figured there was no pressing timeline on this simple change

adamkorynta: image

PR #1190, opened 2025-07-18
[WebUI] Fix legacy web UI mobile menu + Add menu items

krowvin:
This PR adds new header links for the client-side libraries.

I also noticed the mobile menu was not working. If you attempt to tap it currently nothing happens visually. 

It looks as if the JavaScript for the event listener was not being loaded. Updated the `load()` method to account for this. 

# Brief Summary of Changes
- Update touched files to newer const/let
- Update `load` method for html files with JS
- Update header to include github + client library links to desktop and mobile

## Testing
I ran the index.html using VSCode's live server to ensure it rendered as expected on mobile and desktop:

### Mobile Menu
image

### Desktop Menu
image

### DIFF
--- a/cwms-data-api/src/main/webapp/templates/HEC.header.html
+++ b/cwms-data-api/src/main/webapp/templates/HEC.header.html
@@ -20,11 +20,11 @@
                                     </ul>
                                 </div>
                                 <div id="mobile-menu">
-                                    <span class="burger-bar" id="burgerBtn" onclick="openNav()">&#9776;</span>
+                                    <span class="burger-bar" id="burgerBtn">&#9776;</span>
                                 </div>
                             </div>
                             <div id="mobileNav" class="overlay">
-                                <a href="javascript:void(0)" class="closebtn" onclick="closeNav()">&times;</a>
+                                <a href="javascript:void(0)" class="closeBtn">&times;</a>
                                 <div id="mobileNavContent" class="overlay-content">
                                     <a class="home mobile-top-level-menu" href="./swagger-docs">Swagger Docs</a>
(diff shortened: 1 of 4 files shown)


cwms-data-api/src/main/webapp/templates/HEC.header.html
```diff
@@ -48,7 +68,7 @@
                 </div>
 
                 <!-- Burger Bar Control -->
-                <script defer type="text/javascript" src="./js/header.js"></script>
+                <script defer type="text/javascript" src="./js/header.js?_=2"></script>
```
MikeNeilson, line 71: what is the purpose of the `?_=2`?

MikeNeilson (approved):

cwms-data-api/src/main/webapp/templates/HEC.header.html
krowvin: This makes sure if someone has loaded the header recently and has it cached that it will force the browser to refetch the file (ignore cache). I've also used `v=2` to try to make that more clear in other sites, but I wasn't intending on starting versioning for that file as someone may miss this little detail between updates.

Instead I use `_` to suggest this parameter value is not important to being passed in. 

Although technically I don't think this JavaScript was being executed/fetched in the current site so in this case it might not actually do anything!


We could certainly get more into this topic, but it won't matter when the site goes to using vite as it uses a new css/file hash for each build. (Files are hashed on the output of the bundler `rollup`): https://vite.dev/guide/build.html#advanced-base-options


Read more on the cache busting I did above with `_=` here: (they use `ver`)
https://www.keycdn.com/support/what-is-cache-busting#using-cache-busting-with-a-cdn

Issue #256, opened 2022-10-20
catalog/timeseries handle aliases consistently


adamkorynta:
CWMSVue queries both the catalog/timeseries and catalog/locations endpoints to build up the Time Series tab. By design it does not take into account the "aliases" list returned by catalog/location since that data is built up separately with the location groups endpoint. This has the unfortunate side-effect of ~RADAR~ CDA returning timeseries catalog rows that are aliases instead of concrete id's. This results in inconsistent row numbering in CWMSVue and inconsistent row counts as seen in the image below:

![image](https://user-images.githubusercontent.com/47677856/197025038-58072fd3-088d-4d15-9c89-41a473b5dd2b.png)

The row number is based on the row returned by the timeseries catalog BEFORE the resolution with the location catalog.

I can see a couple different solutions. 

1. client side: set the row count AFTER the resolution with the location catalog. 
2. server side: add a boolean parameter to catalog/ endpoints to ignoreAliases in order to minimize the amount of queried data for both timeseries and locations since CWMSVue doesn't require it (default to true for backwards compatibility?).
3. server side: make the catalog/timeseries consistent with catalog/locations by moving the alias identifier to an "aliases" JSON node under the concrete identifier.

MikeNeilson:
catalog/locations and /timeseries are supposed to only return the concrete id's and provide the aliases in an aliases node.

I thought we had done that but maybe I only did that particular operation for the ts extents data.

so option 3 was the planned solution.

adamkorynta:
Well that's good then, I was wondering how it would work with option 3 if the user entered an alias for the id filter and it sounds like it shouldn't!

MikeNeilson:
.... wait. But yeah. I think for the catalog that's okay. but we *should* probably include a "check aliases on search" option, but it would return the original with that alias wrapped inside. But that's for a latter date.

adamkorynta:
yeah I'm all in favor of that solution and eventual feature

rma-rripken:
So should we remove the aliases from the return results until the include aliases option gets added?   Or do we leave it as is until then?

rma-rripken:
Related to https://github.com/USACE/cwms-data-api/issues/107

rma-rripken:
This could probably be closed.  I believe the timeseries portion of this was handled buy https://github.com/USACE/cwms-data-api/pull/1193 and the location portion https://github.com/USACE/cwms-data-api/pull/1294

zack-rma:
[...]

PR #1193, opened 2025-07-22
CWMS-2197 Adds include alias flag for timeseries catalog

### LINKED ISSUE #107: Add flag for showing aliased items in catalog/timeseries
The current implementation uses AV_CWMS_TS_ID2, which pulls back all alias time series. For the client, non aliased items are preferred, so either always filtering them out or adding a flag to conditionally filter them.

rma-bryson:
Fixes https://github.com/USACE/cwms-data-api/issues/107

and

https://github.com/USACE/cwms-data-api/issues/256


rma-rripken:
I'd like to see a unit-test with resource file that has a demo of the serialization.

rma-rripken (approved):

PR #1194, opened 2025-07-24
allow for null designator query parameters for forecast spec/instance

### LINKED ISSUE #1108: Unable to save forecast spec
Sample data.

`POST /forecast-spec endpoint`

```
{
  "spec-id": "LRL-Daily-UKY-Test",
  "office-id": "LRL",
  "source-entity-id": "USACE",
  "designator": "designator",
  "description": "3-day forecast of pool elevation and outflow for flood control projects in the Upper Kentucky watershed",
  "time-series-ids": [
    "Buckhorn.Elev.Inst.~1Day.0.LRL-cavi-fct",
    "Buckhorn.Flow-Outflow.Inst.~1Day.0.LRL-cavi-fct",
    "CarrCreek.Elev.Inst.~1Day.0.LRL-cavi-fct",
    "CarrCreek.Flow-Outflow.Inst.~1Day.0.LRL-cavi-fct"
  ]
}
```

user attempted to save this SPEC using an APIKey and was unable to do so with the result being a permission issue.

adamkorynta:
Fixes: #1108

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
@@ -219,8 +219,10 @@ public final class ForecastInstanceDao extends JooqDao<ForecastInstance> {
                             + format(param, Controllers.NAME, URLEncoder.encode(specId, utf8))
                             + format(param, Controllers.FORECAST_DATE, URLEncoder.encode(forecastDate.toString(), utf8))
                             + format(param, Controllers.ISSUE_DATE, URLEncoder.encode(issueDate.toString(), utf8))
-                            + format(param, Controllers.DESIGNATOR, URLEncoder.encode(designator, utf8))
                             + format(param, Controllers.OFFICE, URLEncoder.encode(officeId, utf8));
+                    if(designator != null) {
+                        url += format(param, Controllers.DESIGNATOR, URLEncoder.encode(designator, utf8));
+                    }
                 } else {
                     try (InputStream is = blob.getBinaryStream()) {
                         fileData = BlobDao.readFully(is);
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
@@ -99,15 +99,15 @@ public final class ForecastInstanceDao extends JooqDao<ForecastInstance> {
 
     private static final String GET_ALL_CONDITIONS = " WHERE (? IS NULL OR spec_office_id = ?)" +
             " AND (? IS NULL OR spec_id = ?)" +
-            " AND (? IS NULL OR spec_designator = ?)";
+            " AND ((spec_designator = ?) OR (spec_designator IS NULL AND ? IS NULL))";
     private static final String GET_ONE_CONDITIONS = " WHERE (spec_office_id = ?)" +
             " AND (spec_id = ?)" +
-            " AND (spec_designator = ?)" +
+            " AND ((spec_designator = ?) OR (spec_designator IS NULL AND ? IS NULL))" +
             " AND (inst.FCST_DATE_TIME = cwms_20.cwms_util.to_timestamp(?))" +
             " AND (inst.ISSUE_DATE_TIME = cwms_20.cwms_util.to_timestamp(?))";
     private static final String FILE_CONDITIONS = " WHERE (spec.OFFICE_CODE = CWMS_UTIL.GET_DB_OFFICE_CODE(?))" +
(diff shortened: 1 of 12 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -209,7 +209,9 @@ public static Condition caseInsensitiveLikeRegex(Field<String> field, String reg
         return new CustomCondition() {
             @Override
             public void accept(org.jooq.Context<?> ctx) {
-                if (ctx.family() == ORACLE) {
+                if("*".equals(regex) || ".*".equals(regex)) {
+                    ctx.visit(DSL.trueCondition());
```
rma-rripken: I think this needs some comment about why this branch is necessary.  Why does it come first, why is trueCondition the right thing etc.   I think you are making "*" or ".*" also match null - is that what this is doing? Is that needed?  It looked like you were explicitly handling null designators another way.   Adding extra branches here makes me nervous.

rma-rripken (changes requested):
Lets discuss the JooqDao changes.  If they are needed we might make a similarly method in ForecastInstanceDao b/c I'm not sure the changes apply everywhere.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
adamkorynta: sorry, this was an initial attempt, but should get reverted in favor of actually handling null designators.

rma-bryson: reverted this back

rma-rripken (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
```diff
@@ -99,15 +99,15 @@ public final class ForecastInstanceDao extends JooqDao<ForecastInstance> {
 
     private static final String GET_ALL_CONDITIONS = " WHERE (? IS NULL OR spec_office_id = ?)" +
             " AND (? IS NULL OR spec_id = ?)" +
-            " AND (? IS NULL OR spec_designator = ?)";
+            " AND ((spec_designator = ?) OR (spec_designator IS NULL AND ? IS NULL))";
     private static final String GET_ONE_CONDITIONS = " WHERE (spec_office_id = ?)" +
             " AND (spec_id = ?)" +
-            " AND (spec_designator = ?)" +
+            " AND ((spec_designator = ?) OR (spec_designator IS NULL AND ? IS NULL))" +
```
MikeNeilson, line 105: considering own common this particular idiom is, you'd think they'd have made a shortcut in the SQL standard by now.

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
```diff
@@ -261,8 +263,9 @@ public ForecastInstance getForecastInstance(int byteLimit, ReplaceUtils.Operator
                 preparedStatement.setString(1, office);
                 preparedStatement.setString(2, name);
                 preparedStatement.setString(3, designator);
-                preparedStatement.setLong(4, forecastDate.toEpochMilli());
-                preparedStatement.setLong(5, issueDate.toEpochMilli());
+                preparedStatement.setString(4, designator);
```
MikeNeilson, line 266: Don't bother with this PR, but FYI, one can use the
`with inputs( ? as column_name from dual) select * from something where something.column = inputs.column_name` to avoid binding the same value several times.

There are some examples here for future reference: https://github.com/HydrologicEngineeringCenter/cwms-database/tree/devops/flyway_migration/schema/src/main/resources/db/data/csv

(NOTE: while those are for insert, they still apply, definitely avoid the anonymous PL/SQL block if you can, performance is terrible.)

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
```diff
@@ -123,7 +124,7 @@ public void beforeAll(ExtensionContext context) throws Exception {
                             .withOfficeId("HQ")
                             .withVolumeName(TeamCityUtilities.cleanupBranchName(ORACLE_VOLUME))
                             .withSchemaImage(CWMS_DB_IMAGE);
-            cwmsDb.withImagePullPolicy(PullPolicy.defaultPolicy());
+            cwmsDb.withImagePullPolicy(PullPolicy.ageBased(Duration.ofDays(1)));
```
MikeNeilson, line 127: Didn't even realize that was an option. Totally reasonable.

MikeNeilson (approved):

PR #1200, opened 2025-07-29
Deserialize timeZoneName while considering known aliases.

rma-rripken:


cwms-data-api/src/main/java/cwms/cda/helpers/ZoneIdHelper.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Builds a complete map of timezone aliases by combining resource-based aliases
+     * with default aliases when necessary.
+     *
+     * @return A map of timezone aliases where keys are non-standard timezone IDs and
+     *         values are their corresponding standard timezone IDs
+     */
+    public static @NotNull Map<String, String> buildTimeZoneAliases() {
+        Map<String, String> aliases = buildResourceAliases();
```
MikeNeilson, line 98: would it be better to just load the defaults into the map, and then let the resource file override any existing entries? versus the all or nothing approach?

MikeNeilson (commented):
Conceptually this looks reasonable. Wish there was a better way to handle the aliases... would just doing a .json file be cleaner than trying to map to the java properties syntax?

cwms-data-api/src/main/java/cwms/cda/helpers/ZoneIdHelper.java
rma-rripken: You could be right.  My worry was that there would be no good way to replace a default if the default was wrong or no longer wanted.

rma-rripken: It weirded me out to need jackson to read the resource when the class was going to be needed by jackson objectmappers.  Idk, maybe it would just work.

MikeNeilson: ... fair point. though the most generic instance of ObjectMapper would work.

But there's also nothing wrong with the current setup, just also seemed... awkward.

MikeNeilson (approved):

PR #1202, opened 2025-07-29
Issue #599: Added tests for location level retrieval for constant values

### LINKED ISSUE #599: Allow retrieval of constant value location levels over a date range

zack-rma:
Fixes #599 

Adds `constants-only` parameter to getAll. Includes integration test coverage.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
@@ -238,6 +238,9 @@ public class LevelsController implements CrudHandler {
                         + "\n* `xml`"
                         + "\n* `wml2` (only if name field is specified)"
                         + "\n* `json` (default)"),
+                @OpenApiParam(name = CONSTANTS_ONLY, type = Boolean.class, description = "If true, "
+                        + "only constant location levels will be returned. If false, all "
+                        + "location levels will be returned. Only available for JSON V2. Default is false."),
                 @OpenApiParam(name = PAGE, description = "This identifies where in the "
                         + "request you are. This is an opaque value, and can be obtained from "
                         + "the 'next-page' value in the response."),
@@ -265,6 +268,9 @@ public class LevelsController implements CrudHandler {
 
             String office = ctx.queryParam(OFFICE);
             String unit = ctx.queryParamAsClass(UNIT, String.class).getOrDefault(UnitSystem.SI.getValue());
+            if (!unit.equalsIgnoreCase(UnitSystem.SI.getValue()) && !unit.equalsIgnoreCase(UnitSystem.EN.getValue())) {
+                unit = UnitSystem.SI.getValue();
+            }
             String datum = ctx.queryParam(DATUM);
             String begin = ctx.queryParam(BEGIN);
             String end = ctx.queryParam(END);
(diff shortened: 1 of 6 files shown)


cwms-data-api/src/main/java/cwms/cda/api/LevelsController.java
```diff
@@ -265,6 +265,9 @@ public void getAll(@NotNull Context ctx) {
 
             String office = ctx.queryParam(OFFICE);
             String unit = ctx.queryParamAsClass(UNIT, String.class).getOrDefault(UnitSystem.SI.getValue());
+            if (!unit.equalsIgnoreCase(UnitSystem.SI.getValue()) && !unit.equalsIgnoreCase(UnitSystem.EN.getValue())) {
```
rma-rripken, line 268: This is curious - what is an example of an input that isn't EN or SI that we want to be treated as if it was SI ?    I'd think if it wasn't null, EN or SI it would probably be an exception

zack-rma: I found that it caused a 500 response when provided with a specific unit instead of a unit system. I've updated the check to throw an exception if the unit parameter input is invalid.

rma-rripken (approved):

rma-rripken (approved):

PR #1208, opened 2025-08-01
Paging blobs

rma-rripken:
Does writing DSL queries with manually created table and column references even work?  I don't think we've tried doing this anywhere else.


rma-rripken:
This needs IT that test Blob paging.

MikeNeilson:
> Does writing DSL queries with manually created table and column references even work? I don't think we've tried doing this anywhere else.

It does, but it can get a little wonky because of the way the table is assigned to the field... or the field to the table.

Here's an example: https://github.com/USACE/cwms-data-api/blob/976fad3b3e27f4cea9aa0a3016e6d134b10946b3/cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java#L142

But you seems to already be on the right track. in the above I'm doing it so JOOQ doesn't create a new unique name on every query, but same context of just knowing what a table is and manually wrapping it. Likely would've done that for more of Auth/User stuff if I was aware at the time I did it. (though the Auth queries were pretty simple.)

rma-rripken:
I've added what I think is the missing IT test.   Something in my docker config has broken so I can't run and verify these changes locally and its going to be several weeks before I'm able to work on this again.  If someone else wants to take over the PR, have at it.  I think its ready.

MikeNeilson (dismissed):

rma-rripken:
@MikeNeilson  merging develop dismissed previous reviews.

MikeNeilson (approved):

PR #1211, opened 2025-08-04
REGI-469: Filtering Parameters for Location getAll Method

zack-rma:
REGI-469: Implement AtFlowProducingLocationManager::retrieveValidLocations

Adds location kind exclusion parameter and base location exclusion parameter to Location catalog method. Allows for filtering unwanted kinds and filtering to only include sub-locations.


cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
```diff
@@ -125,6 +135,12 @@ public List<Location> getLocations(String nameRegex, String unitSystem, String d
             whereCondition = whereCondition.and(AV_LOC.VERTICAL_DATUM.equalIgnoreCase(datum));
         }
 
+        if (ignoredKinds != null) {
+            for (String kind : ignoredKinds) {
```
MikeNeilson: I would just use an "in clause" here instead. performance wise likely to be the same but the code would be cleaner.

Either way extra queries will be created based on the number of ignored elements, but more of the "sets" would map to each other at the bind level and JOOQ should be able to reasonably take care of it.

MikeNeilson (commented):
Not unreasonable, but FYI catalog/LOCATIONS has a location-kind-like and location-type-like parameter, so not really necessary either.

cwms-data-api/src/main/java/cwms/cda/api/CatalogController.java
```diff
@@ -165,6 +167,14 @@ public void getAll(Context ctx) {
                         + "by using Regular Expression OR clauses. For example: "
                         + "\"(SITE|STREAM)\""
                 ),
+            @OpenApiParam(name = FILTER_BASE_LOCATIONS, type = Boolean.class,
```
adamkorynta, line 145: Did you try to get the current `like` and `location-kind-like` to have an excludes regex?

rma-rripken: For disucssion with Adam an Zack we found out that Oracle makes it impossible to write exclusion style regex.  Negative lookahead is not supported.    In sql its trivial to add a NOT in front of a regex to achieve the same thing.   Our HTTP interfaces/parameters doesn't make it quite so easy.   We came up with 3 options and need feedback

rma-rripken: A) support a new ! syntax where any regex that starts with an exclamation has the exclamation stripped off and a SQL NOT condition added around the remaining regexp.      

B) add new not-like parameters to any end-poin that needs to support exclusion  

 C) support specific limited patterns of standard negative lookahead syntax with known equivalents.

zack-rma: I've made a discussion post here: https://github.com/USACE/cwms-data-api/discussions/1212

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -240,6 +233,24 @@ public void accept(org.jooq.Context<?> ctx) {
...
+     * an easy to read manner without having to worry about the syntax.
+     */
+    public static Condition caseInsensitiveNotLikeRegex(Field<String> field, String regex) {
+        return new CustomCondition() {
+            @Override
+            public void accept(org.jooq.Context<?> ctx) {
+                if (ctx.family() == ORACLE) {
+                    ctx.visit(DSL.condition("{not regexp_like}({0}, {1}, 'i')", field, DSL.val(regex)));
+                } else {
+                    ctx.visit(DSL.upper(field).notLikeRegex(regex.toUpperCase()));
```
MikeNeilson: I appreciate the planning for the future of other databases.

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
```diff
@@ -445,4 +442,133 @@ void test_loc_kind() {
...
+        Location location = new Location.Builder(Formats.parseContent(Formats.parseHeader(Formats.JSON, Location.class),
+            json, Location.class))
+            .withOfficeId(officeId)
+            .build();
+        String baseLocationName = location.getName();
+
+        TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
+
+        // create base location
+        createLocation(baseLocationName, true, officeId, location.getLocationKind());
```
MikeNeilson: can we not pass an entire Location DTO into this function? Like don't add it if not, I'm just surprised.

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
```diff
@@ -445,4 +442,133 @@ void test_loc_kind() {
...
+        location = new Location.Builder(Formats.parseContent(Formats.parseHeader(Formats.JSON, Location.class),
+            json, Location.class))
+            .withOfficeId(officeId)
+            .build();
+        String subLocationName = location.getName();
+
+        // create sub-location
+        createLocation(subLocationName, true, officeId, location.getLocationKind());
+
+        json = loadResourceAsString("cwms/cda/api/valid_sub_location2.json");
```
MikeNeilson: FYI, I would think it reasonable here to store the list of locations in a single json file as a list and loop over them or something.

MikeNeilson (commented):
Seems reasonably clear.

Also like Adam's "NOT:" idea.

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
zack-rma: No, there's not a `createLocation(Location)` method in the DataApiTestIT class

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
zack-rma: I put the locations and kinds in lists and removed the unnecessary test JSON files.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
```diff
@@ -503,7 +503,11 @@ private static Condition buildWhereCondition(CatalogRequestParameters params) {
 
         condition = condition.and(caseInsensitiveLikeRegexNullTrue(AV_LOC2.AV_LOC2.BOUNDING_OFFICE_ID,
                 params.getBoundingOfficeLike()));
-        if (params.isNegateLocationKindLike()) {
+        if (params.isNegateLocationKindLike()
+            || (params.getLocationKind() != null && params.getLocationKind().toUpperCase().startsWith("NOT:"))) {
+            if (params.getLocationKind().toUpperCase().startsWith("NOT:")) {
```
MikeNeilson: i would just make sure stripNot can handle no not and not bother with this if.

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
```diff
@@ -550,6 +534,26 @@ void testFilterLocations() throws Exception{
...
 
+        // get valid locations using base location, filtering out STREAM kind using NOT operator
+        given()
+            .log().ifValidationFails(LogDetail.ALL,true)
+            .accept(Formats.JSON)
+            .header("Authorization", user.toHeaderValue())
+            .queryParam(OFFICE, officeId)
+            .queryParam(LIKE, "VALID_LOC_TEST-*")
+            .queryParam(LOCATION_KIND_LIKE, "NOT:^(STREAM)*$")
+            .queryParam(UNIT_SYSTEM, UnitSystem.SI.getValue())
```
MikeNeilson, line 587: we should test using not-like and "NOT:" at same time. shouldn't invert just be pointless but not fail.

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
```diff
@@ -444,48 +444,32 @@ void test_loc_kind() {
...
         String officeId = "SPK";
-        String json = loadResourceAsString("cwms/cda/api/valid_base_location.json");
-        Location location = new Location.Builder(Formats.parseContent(Formats.parseHeader(Formats.JSON, Location.class),
-            json, Location.class))
-            .withOfficeId(officeId)
-            .build();
-        String baseLocationName = location.getName();
 
-        TestAccounts.KeyUser user = TestAccounts.KeyUser.SPK_NORMAL;
+        List<String> locationNames = Arrays.asList("VALID_LOC_TEST", "VALID_LOC_TEST-VALID_LOC",
```
MikeNeilson: i meant a json list in a file and looping that, having the full location is generally better.

zack-rma: Added JSON to a single file, changed test to loop through parsed locations.

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
zack-rma: Added test for this case.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
zack-rma: Moved startsWith check into `stripNot`

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
```diff
@@ -447,33 +448,55 @@ void testFilterLocations() throws Exception{
...
 
-        List<String> locationNames = Arrays.asList("VALID_LOC_TEST", "VALID_LOC_TEST-VALID_LOC",
-            "VALID_LOC_TEST-VALID_LOC2", "VALID_LOC_TEST-VALID_LOC3");
-        List<String> locationKinds = Arrays.asList("SITE", "STREAM", "OUTLET", "STREAM");
+        String json = loadResourceAsString("cwms/cda/api/filter_locations.json");
+        ContentType contentType = new ContentType(Formats.JSONV2);
+        List<Location> locations = Formats.parseContentList(contentType, json, Location.class);
 
-        String baseLocationName = locationNames.get(0);
+        String baseLocationName = locations.get(0).getName();
```
MikeNeilson, line 459: I think I'm clearly missing something here but since it works and you've already changed it like 3 times in response I'm just going to let it go.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/CatalogController.java
```diff
@@ -165,9 +142,18 @@ public void getAll(Context ctx) {
...
```
adamkorynta, line 145: `.*-.*` should work, seems like it does spot testing against CWBI

cwms-data-api/src/main/java/cwms/cda/api/CatalogController.java
```diff
@@ -165,9 +142,18 @@ public void getAll(Context ctx) {
...
+                description = "Specifies whether to filter the locations based on the "
+                    + "base location. Default: false. If true, only sublocations "
+                    + "locations will be returned. If false, all locations will be returned. "
+                    + "Only supported for JSON format."),
+            @OpenApiParam(name = NEGATE_LOCATION_KIND_LIKE, description = "Whether to use the location kind "
+                    + "regular expression to exclude locations with the specified kinds. Default is false."),
             @OpenApiParam(name = LOCATION_TYPE_LIKE,
                     description = "Posix <a href=\"regexp.html\">regular expression</a> matching "
-                        + "against the location type."
+                        + "against the location type. When prepended with `NOT:`, this parameter is treated as "
```
adamkorynta: This can be handled globally, and we could move the "NOT:" prefix description to the regexp.html file

cwms-data-api/src/test/resources/cwms/cda/api/filter_locations.json
```diff
@@ -0,0 +1,87 @@
+[
+  {
+    "name": "VALID_LOC_TEST",
```
adamkorynta: replace with real location ids

cwms-data-api/src/test/resources/cwms/cda/data/sql/set_test_filter_loc_kinds.sql
```diff
@@ -0,0 +1,19 @@
+begin
+cwms_loc.update_location_kind(
```
adamkorynta: Don't use this, this is a CMA back-door into bad location creation. Utilize our functions for creating outlets/streams in the outlet/stream integration tests.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
```diff
@@ -504,14 +503,47 @@ private static Condition buildWhereCondition(CatalogRequestParameters params) {
 
         condition = condition.and(caseInsensitiveLikeRegexNullTrue(AV_LOC2.AV_LOC2.BOUNDING_OFFICE_ID,
                 params.getBoundingOfficeLike()));
-        condition = condition.and(caseInsensitiveLikeRegexNullTrue(AV_LOC2.AV_LOC2.LOCATION_KIND_ID,
+        if (params.isNegateLocationKindLike()
+            || (params.getLocationKind() != null && params.getLocationKind().toUpperCase().startsWith("NOT:"))) {
```
adamkorynta: `NOT:` could be handled globally rather than just in this one API

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
```diff
@@ -101,6 +110,24 @@ private static void createProject(String id, String office) throws SQLException
         });
     }
 
+    private static void createStream(Stream stream) throws SQLException {
```
MikeNeilson: In the future, put things like this is DataApiTestIT, while created specific to your test, these could rather clearly be useful to other tests and it seems prudent to track test data at the base level.

cwms-data-api/src/test/resources/cwms/cda/api/filter_locations.json
```diff
@@ -1,9 +1,9 @@
 [
   {
-    "name": "VALID_LOC_TEST",
+    "name": "Pine Flat Area",
```
MikeNeilson, line 3: Thank you for changing to real names.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/CatalogController.java
zack-rma: Updated tests appropriately to use this regex format

cwms-data-api/src/main/java/cwms/cda/api/CatalogController.java
zack-rma: Moved prefix description, moved prefix handling to JooqDao for globalization

cwms-data-api/src/test/java/cwms/cda/api/CatalogControllerTestIT.java
zack-rma: Moved to DataApiTestIT, added data cleanup

cwms-data-api/src/test/resources/cwms/cda/data/sql/set_test_filter_loc_kinds.sql
zack-rma: Removed and replaced with methods added to DataApiTestIT

cwms-data-api/src/test/resources/cwms/cda/api/filter_locations.json
zack-rma: Replaced

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
zack-rma: Moved to JooqDao for global handling

MikeNeilson (approved):
NOTE: for reference, why I say in the future, I actually do mean you don't need to change it now. (Granted I don't know if someone else asked to you to do so.)

adamkorynta (approved):

PR #1213, opened 2025-08-05
[RFC] Data Authorizaton Middleware

vairav:
### Summary

* RFC for the Data Authorization Middleware

### DIFF
--- a/docs/source/rfc/0001-data-authorization-middleware.md
+++ b/docs/source/rfc/0001-data-authorization-middleware.md
@@ -1,0 +1,628 @@
+# Authorization Middleware Implementation
+
+| Status         | Proposed                                                                                                                   |
+| :------------- | :------------------------------------------------------------------------------------------------------------------------- |
+| **RFC #**      | 0001                                                                                                                       |
+| **Author(s)**  | Solid Logix Team<br/>• @jolitinh<br/>• @cunningryan<br/>• @vairav<br/>• @milver<br/>• @toddeboss<br/>• @ChristinaWhitehead |
+| **Sponsor**    | HEC/USACE                                                                                                                  |
+| **Date**       | 7/6/2025                                                                                                                   |
+| **Supersedes** | Initial RFC draft                                                                                                          |
+
+## Objective
+
+Implement a high-performance TypeScript-based authorization middleware service for the CWMS Data API that provides fine-grained, office-based access control to support HEC's transition from Oracle VPD to API-level authorization. The solution must support multiple user personas, maintain compatibility with existing authentication systems, and enable gradual migration from the current 32-database architecture to a single cloud-based system.
+
+**Scope**: While the CWMS Data API comprises 279 REST endpoints across 96 controller classes, this implementation focuses exclusively on timeseries resource endpoints. This targeted approach will establish the foundational authorization patterns and infrastructure that can be systematically extended to cover all remaining resources (locations, ratings, forecasts, water supply, etc.) in future phases.
+
+## Motivation
+
+### Current Limitations
+
+The existing Oracle VPD-based authorization system cannot support HEC's evolving requirements:
+
+- Limited to two Q-38 roles (modifier and admin)
+- Cannot enforce the 7 distinct user personas defined in PWS Exhibit 3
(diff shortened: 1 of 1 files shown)


docs/source/rfc/0001-data-authorization-middleware.md
```diff
@@ -0,0 +1,628 @@
...
+@RestController
+public class TimeSeriesController {
+    @Autowired
+    private CwmsAuthorizationHelper authHelper;
+
+    @GetMapping("/timeseries")
+    public ResponseEntity<List<TimeSeries>> getTimeSeries(
+            String office, Context ctx) {
+
+        QueryBuilder query = QueryBuilder.create().table("cwms_ts_data");
```
MikeNeilson, line 516: The table information would either not be known at this point or possibly meaningless.

What CDA does know is the endpoint URL and the DTOs involved. I would suspect that would be able to provide the required information checks.

Doesn't negate the point the sample is making though, same concept different source.

docs/source/rfc/0001-data-authorization-middleware.md
```diff
@@ -0,0 +1,628 @@
...
+        return ResponseEntity.ok(results);
+    }
+}
+```
+
+**Integration Options:**
+
+1. **Dependency Injection**: `@Autowired` helper in controllers
+2. **Annotation-Based**: `@CwmsAuthorized` with AOP interceptors
+3. **Filter-Based**: Servlet filter for automatic processing
```
MikeNeilson, line 531: Javalin has before handlers and a Context object:

https://github.com/USACE/cwms-data-api/blob/2c19ffccf88e3e5c2480760e3f138cfa5279763b/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java#L353C17-L353C35

https://github.com/USACE/cwms-data-api/blob/develop/cwms-data-api/src/main/java/cwms/cda/security/Authenticator.java#L15

With the before handler functionally equivalent to the Filter Based approach.

I suspect a combination of work within an CdaAccessManager, a before handler, and the specific endpoints for any data specific issues will likely be what happens.

e.g. a POST call with no `x-cwms-auth-context` should fail before it before it even gets to the Time Series Controller.

MikeNeilson (commented):
Some minor picks that will really only matter at time of implementation. Otherwise everything seems reasonable to me.

Will wait on feedback from others.

docs/source/rfc/0001-data-authorization-middleware.md
```diff
@@ -0,0 +1,628 @@
...
+- Office-based access control (primary filtering mechanism)
+- Role-based permissions for CRUD operations
+- Time-based embargo rules with override capabilities (7-day default)
+- Resource sensitivity filtering (public, internal, restricted, classified data)
+- Persona-specific constraints per PWS Exhibit 3
+
+**7 User Personas (PWS Exhibit 3):**
+
+1. **Anonymous/Public User**: Read-only access to public data after embargo period
+2. **Dam Operator**: Manual data entry only, 24-hour modification window, shift hours (6am-6pm)
```
jbkolze, line 185: For clarification, will the details of these personas (or the users themselves) be configurable?  I believe that 

> - User persona definitions with specific constraints and capabilities

 below under Policy Data Management indicates that they will be, but just want to confirm.

Our current operations workflow would require potential modification by dam operators beyond the 24-hr window and outside of set shift hours.

docs/source/rfc/0001-data-authorization-middleware.md
```diff
@@ -0,0 +1,628 @@
...
+
+Each resource type will follow the established pattern, reusing the core infrastructure while adding resource-specific policies.
+
+**Key Use Cases:**
+
+- **Emergency Override**: Water Manager bypasses 7-day embargo during flood event
+
+#### Phase 3: Administration and Optimization
+
+- React-based admin UI for policy management
```
jbkolze, line 278: I assume that direct API access for policy management will be available prior to Phase 3?

jbkolze (commented):
Looks good to me -- just had a couple quick questions for clarification.

MikeNeilson (approved):

PR #1228, opened 2025-08-12
CWMSVUE-600 Fixed issue with millisecond offset

rma-bryson:


cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecDao.java
```diff
@@ -40,6 +40,7 @@
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
```
rma-rripken, line 43: This is no longer needed

rma-rripken (approved):

PR #1230, opened 2025-08-14
REGI-470: Added lightweight catalog endpoint for location + location kind

zack-rma:
Adds support for lightweight data retrieval on locations getAll endpoint. 

Returns location ID, office ID, and location kind ID.

To be used by REGI in AtLocationManager::retrieveAllLocationTemplatesWithKinds method.


MikeNeilson:
This is getting dangerously close to graphql behavior.

The purpose of the API is not to map 1:1 to what database calls would be made but to provide a standard, if, as in this case, at times verbose, view of the data. Generally denormalized for rapid use.

zack-rma:
This endpoint modification reduces the time to retrieve the requested data from ~150-300ms to ~50ms when testing against our internal SWT database export. It also significantly reduces the payload size. 

Is there a target speed you'd like to see in a catalog endpoint? @MikeNeilson

MikeNeilson:
Don't get me wrong, that's a good improvement in response time, but if this is the method we take for everything we're going to have an explosion in endpoints that's going to become rather unmanageable especially if things change in the future.

There is likely still remove for improvement in the CDA side query, there's *definitely* room for improvement in the database side with the view. @JeremyDKellett has been working on improving that with materialized views, and the recent database change to not spuriously commit changes to the location information tables will most likely make that work even more practical.

If the overall performance can be improved, then this work around doesn't really make a lot of sense, just adds additional maintenance burden.

rma-rripken:
Its not obvious to me that Location Kind is the key other fact to include with each Location.  Which suggests that maybe this isn't just a Location catalog but more Location+LocationKind catalog.   We could have almost endless variations by adding new Accept types.  I think thats Mike's complaint.   I'm not so sure that is such a terrible thing.  Its sort of gross that it changes not just the view of the data but the content of the data that is returned.  It does increase maintenance but assuming someone needs these end-points the end-points are going to incur some maintenance cost regardless of where they get implemented.   A second option might be to have this same functionality moved into a LocationKind controller getAll, maybe with a include_locations flag or something.  It would probably return the reverse of this output - map from kind->locations instead of locations->kind.    We've talked about letting Offices add their own custom end-points.  Not sure where those would go in the Urls but a third option would be to have custom endpoints for specific applications.   I'm honestly surprised that the existing Location getAll methods return so quickly given the number of Locations in the database, the amount of data that is returned and the lack of paging on these end-points.  Even 500ms seems pretty impressive to me.    Other places we've added flags to control some of the data that  is returned.  I'm thinking data-entry-date on the Timeseries endpoint.  Or include-assigned in the Groups end-points.  I'm sure there are more examples.   I don't think we have an agreed upon strategy for when that is appropriate, its sort of been by necessity.  This new Accept header is like one of those flags but uber-flag that turns off lots of returned data.  @MikeNeilson Would it be more palatable if it was a method parameter instead of an Accept header?  If this functionality got moved into its own Controller, something new, you could implement paging and have additional mask parameters.

MikeNeilson:
I think moving it to an endpoint under "regi" e.g. regi/locations or something would make sense.

As you pointed out it's mostly that the catalog itself is already too complex. In hindsight even the time series and locations should be their own controller instead of sharing logic.

So actually even just catalog/regi-locations would make some sense, but for the "custom office" grouping my original idea was to have it's own "base" just for easier grouping, I think that makes sense for regi specific things as well. We'd also likely do that for A2W things in the future..

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -703,6 +705,7 @@ protected void configureRoutes() {
         addProjectLocksHandlers("/project-locks/{name}", requiredRoles);
         addProjectLockRightsHandlers("/project-lock-rights/{project-id}", requiredRoles);
 
+        get("/location-kinds/", new LocationKindController(metrics));
```
MikeNeilson:
```suggestion
        get("/regi/location-kinds/", new LocationKindController(metrics));
```

cwms-data-api/src/main/java/cwms/cda/api/LocationKindController.java
```diff
@@ -0,0 +1,136 @@
...
+        responses = {
+            @OpenApiResponse(status = STATUS_200,
+                content = {
+                    @OpenApiContent(isArray = true, type = Formats.JSONV2, from = Location.class),
+                })
+        },
+        description = "Returns CWMS Location Data.  The Catalog end-point is also capable of "
+            + "retrieving lists of locations and can filter on additional fields.",
+        method = HttpMethod.GET,
+        path = "/location-kinds",
```
MikeNeilson:
```suggestion
        path = "/regi/location-kinds",
```

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsIdLocationKind.java
```diff
@@ -0,0 +1,75 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import cwms.cda.formatters.Formats;
+import cwms.cda.formatters.annotations.FormattableWith;
+import cwms.cda.formatters.json.JsonV2;
+
+@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class,
+    aliases = {Formats.JSON, Formats.DEFAULT})
+@JsonDeserialize(builder = CwmsIdLocationKind.Builder.class)
+@JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
+public final class CwmsIdLocationKind extends CwmsDTOBase {
```
MikeNeilson, line 42: I wonder if this should get organized like the was it outlets DTOs, with progressively more information?

Maybe not, definitely not in this initial PR.

MikeNeilson (commented):
Overall this seems more reasonable to me.

That actual change itself isn't super complex, but the grouping seems more manageable now.

the propose `/regi` change isn't required, but I wanted to see what it would look like.

cwms-data-api/src/main/java/cwms/cda/data/dto/CwmsIdLocationKind.java
zack-rma: I can see that being a useful implementation for future endpoints that provide more information.

zack-rma:
@MikeNeilson I know the Locations getAll is missing paging, and REGI would likely not use it on this endpoint, but is that a feature you would consider worthwhile for this endpoint? With the large number of locations in some databases, it may be a useful feature that could be carried over into the Locations getAll at some point, especially since the retrieval methods share similar logic.

MikeNeilson:
No. the getAll currently goes to the database to run a function at get the results. if you wanted add paging to that you'd be making a bunch of rather pain (to you, and those reviewing it) plsql changes.

the /catalog endpoint was intended to change that behavior, though in hindsight I should've just changed it from the get go at /locations.

Not sure when we'll really get around to changing that. Most likely *after* we'd started making new "versions" under cwms-data/...

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDao.java
```diff
@@ -35,6 +36,8 @@ public interface LocationsDao {
 
     List<Location> getLocations(String names, String units, String datum, String officeId);
 
+    List<CwmsIdLocationKind> getLocationKinds(String names, String units, String datum, String officeId);
```
rma-rripken: Can you rename "names" to something else.  The old-school pl/sql used to have a field called names that was a pipe delimitied list of ids.  I think some other parameter name would be better.  Not sure what, regex, idRegex, idPattern, mask - I'm sure you can come up with something slightly cleaner

cwms-data-api/src/test/java/cwms/cda/data/dto/CwmsIdLocationKindTest.java
```diff
@@ -0,0 +1,57 @@
...
+package cwms.cda.data.dto;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import cwms.cda.formatters.ContentType;
+import cwms.cda.formatters.Formats;
+import cwms.cda.helpers.DTOMatch;
+import org.junit.jupiter.api.Test;
+
+final class CwmsIdLocationKindTest {
```
rma-rripken, line 37: I like seeing a json resource file that has an example of the serialized output.  I find it helpful to just be able to look at something concrete.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
```diff
@@ -131,6 +133,29 @@ public List<Location> getLocations(String nameRegex, String unitSystem, String d
                     .fetch(this::buildLocation);
     }
 
+    @Override
+    public List<CwmsIdLocationKind> getLocationKinds(String nameRegex, String unitSystem, String datum, String officeId) {
```
rma-rripken: How about paging? Paging is not absolutely required but adding it later seems like it would require re-coding the Regi-side

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDao.java
zack-rma: Renamed to `idRegexMask` to be clearer.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
zack-rma: Mike doesn't seem to enthusiastic about paging for this endpoint. The current REGI implementation does not use it, though it would not be too difficult to implement here, since there isn't PL/SQL usage for this retrieval. Thoughts, @MikeNeilson?

cwms-data-api/src/test/java/cwms/cda/data/dto/CwmsIdLocationKindTest.java
zack-rma: Added a unit test for serialization with example JSON.

zack-rma:
> No. the getAll currently goes to the database to run a function at get the results. if you wanted add paging to that you'd be making a bunch of rather pain (to you, and those reviewing it) plsql changes.
> 
> the /catalog endpoint was intended to change that behavior, though in hindsight I should've just changed it from the get go at /locations.
> 
> Not sure when we'll really get around to changing that. Most likely _after_ we'd started making new "versions" under cwms-data/...

The Location getAll DAO methods I'm seeing don't use PL/SQL, and instead use Jooq calls to retrieve data.

MikeNeilson:
> > No. the getAll currently goes to the database to run a function at get the results. if you wanted add paging to that you'd be making a bunch of rather pain (to you, and those reviewing it) plsql changes.
> > the /catalog endpoint was intended to change that behavior, though in hindsight I should've just changed it from the get go at /locations.
> > Not sure when we'll really get around to changing that. Most likely _after_ we'd started making new "versions" under cwms-data/...
> 
> The Location getAll DAO methods I'm seeing don't use PL/SQL, and instead use Jooq calls to retrieve data.

https://github.com/USACE/cwms-data-api/blob/14a7882a3975e84f174fd850bfdac1b3b3133661/cwms-data-api/src/main/java/cwms/cda/api/LocationController.java#L186

https://github.com/USACE/cwms-data-api/blob/14a7882a3975e84f174fd850bfdac1b3b3133661/cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java#L107

Then code find retrieve_locations_f in the database code... prepare to be agast that raw XML and JSON formatting in XML.

zack-rma:
> > > No. the getAll currently goes to the database to run a function at get the results. if you wanted add paging to that you'd be making a bunch of rather pain (to you, and those reviewing it) plsql changes.
> > > the /catalog endpoint was intended to change that behavior, though in hindsight I should've just changed it from the get go at /locations.
> > > Not sure when we'll really get around to changing that. Most likely _after_ we'd started making new "versions" under cwms-data/...
> > 
> > 
> > The Location getAll DAO methods I'm seeing don't use PL/SQL, and instead use Jooq calls to retrieve data.
> 
> https://github.com/USACE/cwms-data-api/blob/14a7882a3975e84f174fd850bfdac1b3b3133661/cwms-data-api/src/main/java/cwms/cda/api/LocationController.java#L186
> 
> https://github.com/USACE/cwms-data-api/blob/14a7882a3975e84f174fd850bfdac1b3b3133661/cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java#L107
> 
> Then code find retrieve_locations_f in the database code... prepare to be agast that raw XML and JSON formatting in XML.

Not sure how I missed that call at the top of the file...

Wow, that retrieval method in the database is... painful to look at. I can see why it's been left alone. 

We could implement paging for JSONV2 format as a stop-gap measure, but I'm not sure if that's the route we want to go down. It would only be solving the problem for a single format and would just be kicking the can down the road until someone tackles that procedure.

MikeNeilson:
> We could implement paging for JSONV2 format as a stop-gap measure, but I'm not sure if that's the route we want to go down. It would only be solving the problem for a single format and would just be kicking the can down the road until someone tackles that procedure.

That linked procedure will simply be getting dropped.

cwms-data-api/src/test/java/cwms/cda/data/dto/CwmsIdLocationKindTest.java
```diff
@@ -54,4 +55,15 @@ void test_CreateCwmsIdLocationKind_RoundTrip() {
         DTOMatch.assertMatch(kind.getLocationId(), parsedKind.getLocationId());
         assertEquals(kind.getLocationKindId(), parsedKind.getLocationKindId());
     }
+
+    @Test
+    void test_SerializeFromFile() {
```
MikeNeilson: shouldn't this be deserialize?

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
```diff
@@ -131,6 +133,29 @@ public List<Location> getLocations(String nameRegex, String unitSystem, String d
                     .fetch(this::buildLocation);
     }
 
+    @Override
+    public List<CwmsIdLocationKind> getLocationKinds(String idRegexMask, String unitSystem, String datum, String officeId) {
```
rma-rripken: I know this is carried over from when this method was located in Location but what purpose does searching with the unitSystem and datum serve here?  I'm not trying to be critical I really don't know.  You are only getting back the {office, id, kind}. Are there some locations that have different id b/c of the unitSystem or datum?   I wonder if those exra args should be dropped and the select turned into a selectDistinct?  Should there also be a regex for kind so that people can get a catalog of just Outlets or just Projects?

cwms-data-api/src/test/java/cwms/cda/data/dto/CwmsIdLocationKindTest.java
zack-rma: Yes. Fixed.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
zack-rma: You're right, that shouldn't have been carried over. That data is not returned, so filtering on it makes no sense.

I updated the retrieval query to use `selectDistinct` to avoid duplication of results by different unit systems.

I also added kind regex filtering, along with an integration test for it. That could prove useful at some point, I'd imagine.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
MikeNeilson:
```suggestion
        get("/location/by-kinds/", new LocationKindController(metrics));
```

Okay, one last nitpick. This as the name?

It's not a location-kind list, it's a list of locations with only the office, name, and kind that allows some searching.

@rma-rripken

MikeNeilson (changes requested):
Okay, this looks reasonable. The current path name doesn't seem correct, I've provided another suggestion.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
zack-rma: That more accurately represents the usage/intent. Perhaps `/location-with-kind/`?

MikeNeilson: `locations/with-kind`

This shouldn't be a root endpoint. it's behavior under locations.

zack-rma: Changed endpoint path to `/locations/with-kind/`

rma-rripken: ```/locations/with-kind/``` seems like a good place

rma-rripken (approved):

MikeNeilson (approved):

PR #1234, opened 2025-08-19
Add schema versions to matrix

MikeNeilson:
Run integration tests with current released version and latest-dev to verify operations behave gracefully for various updates.

NOTE: added release currently being deployed only, discovered that the actual database version wasn't getting embedded into the ready database or schema install images and so was always 99.99.99.


MikeNeilson:
waiting to merge even if approved. I'm fairly certain I want the database image versions to you . instead of -, getting it to work with - was just to make sure the mechanism was behaving.

rma-rripken (approved):

MikeNeilson:
Will fix version naming later. This is more important than that exacting little detail.

PR #1239, opened 2025-08-19
Karl's proof read

ktarbet:


docs/source/design/composite-time-series.rst
```diff
@@ -5,15 +5,15 @@ Composite Time Series
 Purpose
 =======
 
-It is a challenge for users to identity what the correct authoritative time series is for a given measurement at a location. Additionally these time series often change over time, either being completely new or changing their interval as newer technologies become available.
+It is a challenge for users to identity what the correct authoritative time series is for a given measurement at a location, when there are multiple time series a the same location. Additionally these time series often change over time, either being completely new or changing their interval as newer technologies become available.
```
MikeNeilson:
```suggestion
It is a challenge for users to identity what the correct authoritative time series is for a given measurement at a location, when there are multiple time series at the same location. Additionally these time series often change over time, either being completely new or changing their interval as newer technologies become available.
```

MikeNeilson (changes requested):
found a minor typo in your corrections, but otherwise looks like good changes.

PR #1244, opened 2025-08-22
Baseline integration into build system.

MikeNeilson:
NOTE: Does break the index page and swagger UI for T7 office based installations.

That will be addressed in a follow up PR in some generic manner.

If attempting to test, use context of "cwms-data" regardless of office.


krowvin (dismissed):
Looks good!

Dockerfile
```diff
@@ -3,7 +3,16 @@ USER $USER
 RUN --mount=type=cache,target=/home/gradle/.gradle
 WORKDIR /builddir
 COPY . /builddir/
-RUN  gradle prepareDockerBuild --info --no-daemon
+RUN apt update && apt install -y curl
+ENV NVM_DIR="/root/.nvm"
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.7/install.sh | bash
```
krowvin, line 8: @jbkolze is out this week but may be interested in what you've done here. Especially since you're using nvm. First i've seen it used in Dockerfile. Clever, makes it easy.

Dockerfile
```diff
@@ -3,7 +3,16 @@ USER $USER
...
+ENV NODE_VERSION=22.17.0
```
krowvin, line 9: Grabbing and installing node manually using the gradle:8.5-jdk11

Does it make more sense to actually use a Node Docker image from {approved source}? I ask because there's a chance some of this breaks/changes and it's another maintenance item. 

I imagine you could just use the NodeJS docker image to temporarily do the build and then pass the resultant files into the following layers?

krowvin (approved):
Commented on what I could. Trusting on the gradle bits. 

Nothing blocking, just thoughts you can chose if you want to act on!

Dockerfile
MikeNeilson: The gradle task expects to run the NPM command, yes we could run them separately but that a larger maintanance and configuration burden.

If using a node image, then we need java, same problem, different direction.

krowvin: Got it, didn't realize we weren't just dumping raw files in.

MikeNeilson: There likely *is* a better way to handle this, I also need to review the current prepareDockerBuild task, but this is not the time.

PR #1245, opened 2025-08-22
Correct general behavior in retrieveLocationLevel

MikeNeilson:
Location levels could only be retrieved by exact data. Expected behavior was most recent.

during #1086 the "exact" match flag was changed from true to false. Change is behavior was not correctly caught due to all tests using the exact dates and the endpoint itself not providing appropriate description.

Description has been updated.

The virtual location level tests were modified to specify exact date. In diagnosing the issue I've discovered that the order of these operations is wrong:

https://github.com/HydrologicEngineeringCenter/cwms-database/blob/f2670f46de2f41b724ac9d516505fa9a37c50cfa/schema/src/cwms/cwms_level_pkg_body.sql#L612

Virtual location levels should still support "or nearest before behavior" but at the least with match being false while providing the exact date should return the level and it does not.

I've set the default behavior of exact match to False, knowing this will break existing virtual location level behavior, however, that will be corrected with a database proceedure fix. 
... well, or we could just rewrite that whole sequence of events in CDA but that doesn't seem necessary at this time.


MikeNeilson:
Initial test had some incorrect office setup remaining. Attempting some variations.

MikeNeilson:
Okay, apparently I forgot to run all the other tests locally and also adjust the behavior. will do so after the database PR to properly fix the situation.

rma-rripken (approved):

PR #1246, opened 2025-08-23
CDA GUI Contributing Docs

krowvin:
Adds docs on how to:

1. Get started
2. Add a new page
3. Understand vite env files
4. References to parent react libraries
5. Requried requisits
6. Reference to the cda-gui README from the root contributing doc

This is marked as draft so we can review and/or add more to it after #1244 is merged.


MikeNeilson (approved):

CONTRIBUTING.md
```diff
@@ -92,13 +92,13 @@ Otherwise use the defined checkstyle format for new code.
...
-       ```sql
+1. Use the JOOQ wrapper. Generally the wrapper provides sufficiently reasonable query generation.
+   However, _DO NOT_ be afraid to say, "that looks terrible", and tweak it until it generates something better.
+   a. If the query you're making has nested queries name the queries. Example  
+    `sql
        select a.* from (select col1,col2 from a_table) a;
-       ```
-       Otherwise JOOQ creates a new name each time the query is run which can starved the shared memory.
+       `
+   Otherwise JOOQ creates a new name each time the query is run which can starved the shared memory.
```
MikeNeilson:
```suggestion
   Otherwise JOOQ creates a new name each time the query is run which can starve the shared memory.
```

CONTRIBUTING.md
```diff
@@ -92,13 +92,13 @@ Otherwise use the defined checkstyle format for new code.
 
 #### SQL Coding
 
-1. Use the JOOQ wrapper. Generally the wrapper provides sufficiently reasonable query generation. 
-However, *DO NOT* be afraid to say, "that looks terrible", and tweak it until it generates something better.
-    a. If the query you're making has nested queries name the queries. Example        
-       ```sql
+1. Use the JOOQ wrapper. Generally the wrapper provides sufficiently reasonable query generation.
+   However, _DO NOT_ be afraid to say, "that looks terrible", and tweak it until it generates something better.
+   a. If the query you're making has nested queries name the queries. Example
```
MikeNeilson:
```suggestion
   a. If the query you're making has nested queries, name the queries. Example  
```

MikeNeilson (changes requested):

MikeNeilson:
@krowvin bump. I think everyone agrees good idea, just need to finish it up.

MikeNeilson (changes requested):
Looks good, needs a conflict fix.

MikeNeilson (dismissed):
some of the changes do seem ... odd though.

The formatting changes make sense, mostly like just husky being "helpful" but there is some more.

MikeNeilson (approved):

PR #1248, opened 2025-08-25
setup to load locations and timeseries into cwms database

Enovotny:
Add user and API key, load timeseries and location into cwms-database.  currently this is a manual process.  

still need to 
-setup docker python container. 
-automate loading of data.


compose_files/sql/users.sql
```diff
@@ -1,22 +1,148 @@
...
-    
-
-    cwms_sec.add_cwms_user('l2hectest',NULL,'SPK');
-    cwms_sec.update_edipi('l2hectest',1234567890);
-    cwms_sec.add_user_to_group('l2hectest','All Users', 'SPK');
-    cwms_sec.add_user_to_group('l2hectest','CWMS Users', 'SPK');
-    cwms_sec.add_user_to_group('l2hectest','TS ID Creator','SPK');
-
-    cwms_sec.add_cwms_user('l1hectest',NULL,'SPL');
+   cwms_sec.add_user_to_group(
```
MikeNeilson: the formatting here doesn't need to be changed, and this change makes it more difficult to see at a glance what is setup.

MikeNeilson (changes requested):
Looks reason, other than the format changes in user.sql.

Can we add some sort of documentation for how to go about adding more data? I'm not very familiar with jupiter notebooks so I woudln't even know where to start.

krowvin:
I ran into this error and Eric was able to point me to this PR to solve it. 

> ORA-20048: NO_WRITE_PRIVILEGE: User doesnt have write privileges

I can overwrite the users.sql and re-run the permissions docker until this is resolved

MikeNeilson:
> I ran into this error and Eric was able to point me to this PR to solve it.
> 
> > ORA-20048: NO_WRITE_PRIVILEGE: User doesnt have write privileges
> 
> I can overwrite the users.sql and re-run the permissions docker until this is resolved

NOTE: the permission that is most often missing with this error is `TS ID CREATOR` on the individual user.

MikeNeilson (approved):

Issue #1252, opened 2025-08-26
Blob Endpoints Not Returning Data


krowvin:
When doing the following query against the national CDA instance:

Happens for any blob, MVS and SWT ran into this so far. 

```bash
curl -X 'GET' \
  'https://cwms-data.usace.army.mil/cwms-data/blobs/DAILY_DATA_DISPLAY_TABLE.TXT?office=SWT' \
  -H 'accept: */*'
```

Gives

404
```json
{
  "message": "Unable to find blob based on given parameters",
  "incidentIdentifier": "-70377708931677094",
  "details": {}
}
```

Additional Details

Query for something that truly doesn't exist

```json
{
  "message": "Not Found.",
  "incidentIdentifier": "-3680534926870662849",
  "details": {}
}
```

@Enovotny was able to login to the public database and confirm the blob data is actually present in SQL Dev

zack-rma:
I've been able to recreate this on the national instance, but have been unable to replicate the issue on the latest CDA locally. It's possible whatever caused this has been fixed since the last update to the national instance. 

@krowvin  Can you provide more details about the blob and its format? Are the error logs corresponding with the incident identifier available?

On another note: in the course of trying to recreate the issue, it was discovered that the office ID query parameter is case sensitive, which can result in a `NOT FOUND` error when attempting to retrieve data with a lower-case office ID.

krowvin:
> has been fixed since the last update?

Entirely possible. But it is odd that it did work at one point and has stopped. I didn't think anything else on the national changed? I.E. Schema is the same? 

It works fine on TEST, works fine on our T7 (running 5/8/25), possibly we wait and see if a version bump on the PROD CDA fixes this? @MikeNeilson Unless you're able to check the WAF (or access logs to see what URL was tried)? But it generates an INC which makes me think it does make it through, and possibly is clobbering the URL somehow on the way?

> Can you provide more details about the blob and its format?

It seems to do this for any blob and/or format. Here's a collection of BLOB I tried from various districts. 

| blob | office | mime-type | error | Inc# |
| :--- | :------: | :----: | :----: | ----: |
| [DATACHECK.DATACHECK.JSON](https://cwms-data.usace.army.mil/cwms-data/blobs/DATACHECK.DATACHECK.JSON?office=SWT) | SWT | application/json | `Unable to find blob based on given parameter` |
[...]

PR #1258, opened 2025-08-27
Cleaning up potential resource handling errors.

rma-rripken:
Explicitly calling blob and clob free.  Putting streams from blob and clob in try-with-resources blocks.   Calling jOOQ fetchStream in try-with-resources.

I don't think any of these is actually causing connection leaks but they seem like easy changes for better safety


rma-rripken:
@MikeNeilson  Not immediately obvious to me why this build is failing

MikeNeilson:
> @MikeNeilson Not immediately obvious to me why this build is failing

https://github.com/USACE/cwms-data-api/pull/1258/files#annotation_38309067931

Granted that still doesn't make it obvious. All of those errors should be in a log in the build directory, we should probably upload it as an action artifact that so we can review it.

rma-rripken:
Its hit-or-miss for me whether that annotations link shows anything of not.  The github ui seems to be inconsistent.

rma-rripken:
I've added to the pr and I believe it will fix #1252

MikeNeilson:
> Its hit-or-miss for me whether that annotations link shows anything of not. The github ui seems to be inconsistent.

Yeah, it tends not to show on the file unless it's a file that was edited and shows in the PR.

MikeNeilson:
It's failing on the Time Series integration test that inserts a massive amount of data.

rma-rripken:
Something has changed in the backend.  I ran the timeseries test_create_big that inserts 200,000 points and the connection gets forceably closed in 45 seconds and the insert still hasn't completed.  I'm glad the test is failing if its taking that long but the test used to pass.  Is it the change that we are now actually closing the connections?  Dialing down the number of points to 10,000 and the zstore_ts method is still taking over 30s to complete.  And that is after I moved the json serialization (400ms) and ZTSV array creation (30ms) so it happens before we get a connection to do the zstore_ts call.  Maybe I create a new issue to figure out why the timeseries store is so slow?

MikeNeilson:
> Something has changed in the backend. I ran the timeseries test_create_big that inserts 200,000 points and the connection gets forceably closed in 45 seconds and the insert still hasn't completed. I'm glad the test is failing if its taking that long but the test used to pass. Is it the change that we are now actually closing the connections? Dialing down the number of points to 10,000 and the zstore_ts method is still taking over 30s to complete. And that is after I moved the json serialization (400ms) and ZTSV array creation (30ms) so it happens before we get a connection to do the zstore_ts call. Maybe I create a new issue to figure out why the timeseries store is so slow?

Yeah, the only recent change I can think of was for retrieve data entry, at least within CDA, but it's definitely possible a database change screwed something up.

MikeNeilson (approved):

PR #1289, opened 2025-09-03
Repsond to Options request with reasonable values.

### LINKED ISSUE #1175: Ensure OPTIONS request is handled in CDA
Currently we are handling the OPTIONS requests via the WAF. This issue is a reminder (for @MikeNeilson) to handle this within CDA itself at some point so users deploying CDA do not need to handle OPTIONS requests themselves. 

Users getting a 403 Forbidden for TimeSeries or other endpoints to their local T7 may experience this.

The current solution for apache as a reverse proxy is:

## 
```

    RewriteEngine On
    RewriteCond %{REQUEST_METHOD} OPTIONS
    RewriteRule ^(.*)$ $1 [R=200,L]
 

```

MikeNeilson:
Fixes #1175.

Reported a while ago but saw @dennisgsmith implementing some workarounds (https://github.com/USACE/instrumentation-api/pull/281) which really shouldn't be necessary.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
+++ b/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
@@ -504,6 +504,12 @@ public class ApiServlet extends HttpServlet {
                     ctx.json(errResponse);
                 })
                 .routes(this::configureRoutes)
+                .options("/*", ctx -> {
+                    ctx.header("Access-Control-Allow-Origin", "*"); // Allow requests from any origin
+                    ctx.header("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, OPTIONS"); // Specify allowed methods
+                    ctx.header("Access-Control-Allow-Headers", "Content-Type, Authorization"); // Specify allowed headers
+                    ctx.status(200); // Respond with a 200 OK status
+                })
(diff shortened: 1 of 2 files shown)


cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -504,6 +504,12 @@ public void init() {
                     ctx.json(errResponse);
                 })
                 .routes(this::configureRoutes)
+                .options("/*", ctx -> {
```
krowvin, line 507: I'm assuming this wildcards to all paths and sub paths and will also include all path versions when that comes along

krowvin (approved):
Looks good. Pending the path wildcard is actually */ and not **/ or something

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
MikeNeilson: That's how it behaved in the tests. Hopefully it stays that way in future versions.

MikeNeilson:
@dennisgsmith this should be deployed in -dev now if you want to try testing against https://water.dev.cwbi.us/cwms-data/.

PR #1291, opened 2025-09-04
CDA-28: Add endpoint for CDA software version that requires auth

### LINKED ISSUE #594: Add endpoint for CDA software version that requires auth
It is fairly difficult to determine what version you are on if you did not take a note of this prior. This goes for both the public and internal instances of CDA. 

Admittedly this only really applies to developers and/or maintainers of the software. 

Proposing we add an endpoint that returns the software version and potentially any other metadata about the install. 

To prevent leaking version numbers it was suggested this be behind an endpoint that requires auth.

zack-rma:
Fixes #594 

Added endpoint for CDA software version that requires authentication.

Includes unit and integration tests.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/CdaVersionHandler.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/CdaVersionHandler.java
@@ -57,7 +57,7 @@ public final class CdaVersionHandler implements Handler {
     }
 
     @OpenApi(
-        description = "Determine the current active version of CWMS Data Access.",
+        description = "Determine the current active version of CWMS Data API.",
         responses = {
             @OpenApiResponse(status = STATUS_200, content = {
                 @OpenApiContent(type = Formats.JSON, from = CdaVersion.class)})
--- a/cwms-data-api/src/main/java/cwms/cda/api/CdaVersionHandler.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/CdaVersionHandler.java
@@ -28,10 +28,11 @@ package cwms.cda.api;
 
 import static cwms.cda.api.Controllers.GET_ONE;
 import static cwms.cda.api.Controllers.STATUS_200;
+import static cwms.cda.data.dao.JooqDao.getDslContext;
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
-import cwms.cda.ApiServlet;
+import cwms.cda.data.dao.CdaVersionDao;
 import cwms.cda.data.dto.CdaVersion;
 import cwms.cda.formatters.ContentType;
 import cwms.cda.formatters.Formats;
@@ -43,6 +44,7 @@ import io.javalin.plugin.openapi.annotations.OpenApiContent;
 import io.javalin.plugin.openapi.annotations.OpenApiResponse;
 import javax.servlet.http.HttpServletResponse;
 import org.jetbrains.annotations.NotNull;
+import org.jooq.DSLContext;
 
 public final class CdaVersionHandler implements Handler {
     static final String TAG = "Version";
@@ -69,9 +71,9 @@ public final class CdaVersionHandler implements Handler {
     @Override
     public void handle(@NotNull Context ctx) throws Exception {
         try (Timer.Context ignored = markAndTime(GET_ONE)) {
-            CdaVersion cdaVersion = new CdaVersion.Builder()
-                .withVersion(ApiServlet.getApiVersion())
-                .build();
+            DSLContext dsl = getDslContext(ctx);
+            CdaVersionDao dao = new CdaVersionDao(dsl);
+            CdaVersion cdaVersion = dao.getCdaVersion();
             String serialized = Formats.format(new ContentType(Formats.JSON), cdaVersion);
             ctx.result(serialized);
             ctx.contentType(Formats.JSON);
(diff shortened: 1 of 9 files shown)


rma-rripken:
looks good so far.
I could see that CdaVersion object containing a Properties style <String, String> map so that we could add arbitrary attributes as needed.

rma-rripken:
I wonder if there should be somesort of singleton CdaVersion and some way for components to find it and set specific properties?  Like maybe the TimeseriesController could find the CdaVersion singleton thing and then set a feature flag like "versionedTS=true" or something like that.   Like it would be nice if the thing implementing the feature was the same object that added the feature-flag to CdaVersion.  Otherwise you are stuck trying to remember to change CdaVersion in addition to the Controller or Dao or whatever.

cwms-data-api/src/main/java/cwms/cda/api/CdaVersionHandler.java
```diff
@@ -0,0 +1,81 @@
...
+    public CdaVersionHandler(MetricRegistry metrics) {
+        this.metrics = metrics;
+    }
+
+    private Timer.Context markAndTime(String subject) {
+        return Controllers.markAndTime(metrics, getClass().getName(), subject);
+    }
+
+    @OpenApi(
+        description = "Determine the current active version of CWMS Data Access.",
```
krowvin: Should be CWMS Data API?

krowvin (commented):
Seems like a good start. 

Could be expanded later to include capabilities of a given endpoint 

I.e. 

/version/{endpoint}

Which might be 

/version/timeseries

And that might return the current supposed payload versions and what features each supports?

Not sure if we have an issue for this or if it was a discussion somewhere.

MikeNeilson:
> I wonder if there should be somesort of singleton CdaVersion and some way for components to find it and set specific properties? Like maybe the TimeseriesController could find the CdaVersion singleton thing and then set a feature flag like "versionedTS=true" or something like that. Like it would be nice if the thing implementing the feature was the same object that added the feature-flag to CdaVersion. Otherwise you are stuck trying to remember to change CdaVersion in addition to the Controller or Dao or whatever.

As much as I usually dislike any singleton, that actually sounds fairly good. I definitely agree with inverting that control of information to point of usage.

MikeNeilson (commented):
As @krowvin said, good start.

Only issue is what he pointed out already.

The other thoughts are definitely interesting, but I think should be done in follow up as we better sort out what we need. Good news is the DTO allows for that expansion rather easily.

zack-rma:
I quite like the idea of a property map, as that would make future inclusion of capabilities or other information simple in the future. 

I also like the idea of a singleton for this DTO to store version information. Replacing the `VERSION` string in ApiServlet with a CdaVersion object would allow us to keep the existing functionality and store more information about capabilities/feature support in one location. This would be especially useful for indicating specialized feature support or keeping track of what custom endpoints an office might have implemented/in use.

MikeNeilson:
> I quite like the idea of a property map, as that would make future inclusion of capabilities or other information simple in the future.
> 
> I also like the idea of a singleton for this DTO to store version information. Replacing the `VERSION` string in ApiServlet with a CdaVersion object would allow us to keep the existing functionality and store more information about capabilities/feature support in one location. This would be especially useful for indicating specialized feature support or keeping track of what custom endpoints an office might have implemented/in use.

We could also add an interface with a "getFeatures" method that each handler can implement that we can search for on start/usage of the endpoint to have each respond back.

MikeNeilson (dismissed):

zack-rma:
> > I quite like the idea of a property map, as that would make future inclusion of capabilities or other information simple in the future.
> > I also like the idea of a singleton for this DTO to store version information. Replacing the `VERSION` string in ApiServlet with a CdaVersion object would allow us to keep the existing functionality and store more information about capabilities/feature support in one location. This would be especially useful for indicating specialized feature support or keeping track of what custom endpoints an office might have implemented/in use.
> 
> We could also add an interface with a "getFeatures" method that each handler can implement that we can search for on start/usage of the endpoint to have each respond back.

That would certainly resolve some of the issues we've had in the past related to verifying available functionality before usage. 

There is room within this task estimate to add a property map and add the DTO to the ApiServlet class. That would better position this endpoint for expanded future usages. It can always be put off into a follow-up issue if you'd rather this gets merged as-is.

MikeNeilson:
Yeah, add the property map, but don't go overboard on what gets put in it. Maybe just the database schema version to have something extra? and maybe the data-entry-date support?

zack-rma:
I've added the schema version and data-entry-date support to the CdaVersion DTO

cwms-data-api/src/main/java/cwms/cda/data/dao/CdaVersionDao.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    private String hasTsDataEntryDateSupport() {
+        TimeSeriesDaoImpl tsDao = new TimeSeriesDaoImpl(super.dsl);
+        boolean supported = false;
+        try {
+            tsDao.validateEntryDateSupport(true);
+            supported = true;
+        } catch (DataAccessException e) {
+            logger.log(Level.FINE, e.getMessage(), e);
```
MikeNeilson: set to finest.

MikeNeilson (changes requested):
Definitely a reasonable start.

I suspect though that we should nest the features to allow for a by endpoint listing.

zack-rma:
Added endpoint nesting support

MikeNeilson (approved):
close enough to what I was thinking, at least for the moment.

PR #1295, opened 2025-09-05
CWMSVUE-600 Updating effective data retrieval to not include aliases

rma-bryson:


rma-rripken:
I think you should add some comments so its clear how this works.  1 query to build the non-alias list and then skipping results from the ResultSet if its an alias.   Its easy to implement but not ideal.  Might be an acceptable trade-off.  Performance will depend on how many aliased items there typically are.

rma-rripken (approved):

PR #1296, opened 2025-09-05
Changing TimeseriesControllerTestIT test_big_create to test_medium_cr…

rma-rripken:
…eate with fewer points.   With 200k points the connection gets closed before the query completes


rma-rripken:
Related: https://github.com/USACE/cwms-data-api/issues/1297

PR #1309, opened 2025-09-10
CWMSVUE-600 Updated to account for "*" officeid mask when filtering aliases for eff dates.

rma-bryson:


cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecDao.java
```diff
@@ -526,18 +533,23 @@ private Set<String> getRatingIds(String office, String templateIdMask, boolean i
...
+                    .orderBy(officeField.asc(), idField.asc())
+                    .fetch();
+
+            // Group results into a Map of office to specs
+            return result.stream()
+                    .collect(Collectors.groupingBy(
+                            r -> r.get(officeField),
+                            LinkedHashMap::new, // preserves insertion order
+                            Collectors.mapping(r -> r.get(idField), Collectors.toCollection(LinkedHashSet::new))
+                    ));
```
rma-rripken: I think all this part can be done by jOOQ fetchGroups.  I don't think we've used that anywhere else in CDA but we should start.

rma-bryson: updated

rma-rripken (approved):

PR #1313, opened 2025-09-16
CDA-41: Add Location Alias Inclusion Flag to getOne

### LINKED ISSUE #1036: Locations Endpoint - Return Aliases
Requesting we add the location aliases to the location endpoint, something like:

```bash
curl -X 'GET' \
  'https://wm.swt.ds.usace.army.mil:8243/swt-data/locations/CATO?office=SWT' \
```
Returns
```json
{
  "office-id": "SWT",
  "name": "CATO",
  "latitude": 36.20388889,
  "longitude": -95.7066667,
  "active": true,
  "public-name": "Verdigris River near Catoosa, OK",
  "long-name": "Verdigris River near Catoosa, OK",
  "description": "Verdigris River near Catoosa, OK",
  "timezone-name": "US/Central",
  "location-kind": "STREAM_LOCATION",
  "nation": "US",
  "state-initial": "OK",
  "county-name": "Rogers",
  "nearest-city": "Catoosa, OK",
  "horizontal-datum": "NAD83",
  "published-longitude": 0,
  "published-latitude": 0,
  "vertical-datum": "NGVD29",
  "elevation": 0,
  "bounding-office-id": "SWT",
  "elevation-units": "ft"
}
```

The catalog DOES return these aliases, but the catalog is very slow at the moment compared to the above endpoint:
```bash
curl -X 'GET' \
  'https://cwms-data.usace.army.mil/cwms-data/catalog/LOCATIONS?office=SWT&like=CATO' \
```

```json
{
  "entries": [
    {
      "office": "SWT",
      "name": "CATO",
      "nearest-city": "Catoosa, OK",
      "public-name": "Verdigris River near Catoosa, OK",
      "long-name": "Verdigris River near Catoosa, OK",
      "description": "Verdigris River near Catoosa, OK",
      "kind": "STREAM_LOCATION",
      "time-zone": "US/Central",
      "latitude": 36.20388889,
      "longitude": -95.7066667,
      "published-latitude": 0,
      "published-longitude": 0,
      "horizontal-datum": "NAD83",
      "elevation": 0,
      "unit": "m",
      "vertical-datum": "NGVD29",
      "nation": "United States",
      "state": "OK",
      "county": "Rogers",
      "bounding-office": "SWT",
      "active": true,
      "aliases": [
        {
          "name": "Agency Aliases-CWMS Standard Naming",
          "value": "CATO"
        },
        {
          "name": "Agency Aliases-NWS Handbook 5 ID",
          "value": "CTOO2"
        },
        {
          "name": "Agency Aliases-DCP Platform ID",
          "value": "CE677A96"
        },
        {
          "name": "Agency Aliases-CWMS Legacy Naming",
          "value": "CATO"
        },
        {
          "name": "Agency Aliases-USGS Station Number",
          "value": "07178452"
        }
      ]
    }
  ],
  "page-size": 500,
  "total": 1
}
```

zack-rma:
Fixes #1036 

Adds support for including aliases in location getOne results.

Includes alias formatting fixes for location level DAO and tests.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
@@ -10,6 +10,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import cwms.cda.api.enums.Nation;
+import cwms.cda.data.dto.catalog.LocationAlias;
 import cwms.cda.formatters.Formats;
 import cwms.cda.formatters.annotations.FormattableWith;
 import cwms.cda.formatters.json.JsonV1;
@@ -17,7 +18,9 @@ import cwms.cda.formatters.json.JsonV2;
 import cwms.cda.formatters.xml.XMLv1;
 import cwms.cda.formatters.xml.XMLv2;
 import java.time.ZoneId;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.function.Consumer;
@@ -54,6 +57,8 @@ public final class Location extends CwmsDTO {
     private final String mapLabel;
     private final String boundingOfficeId;
     private final String elevationUnits;
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private final List<LocationAlias> aliases;
 
     private Location() {
         this(new Builder(null, null, null, null, null,null, null));
@@ -83,6 +88,7 @@ public final class Location extends CwmsDTO {
         this.mapLabel = builder.mapLabel;
         this.boundingOfficeId = builder.boundingOfficeId;
         this.elevationUnits = builder.elevationUnits;
+        this.aliases = builder.aliases;
     }
 
     public String getName() {
@@ -173,6 +179,10 @@ public final class Location extends CwmsDTO {
         return boundingOfficeId;
     }
 
+    public List<LocationAlias> getAliases() {
+        return aliases;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) {
@@ -182,6 +192,20 @@ public final class Location extends CwmsDTO {
             return false;
         }
         Location location = (Location) o;
+
+        for (LocationAlias alias : aliases) {
+            boolean found = false;
+            for (LocationAlias alias2 : location.getAliases()) {
+                if (alias.equals(alias2)) {
+                    found = true;
+                    break;
+                }
+            }
+            if (!found) {
+                return false;
+            }
+        }
+
         return Double.compare(location.getLatitude(), getLatitude()) == 0
                 && getActive() == location.getActive()
                 && getName().equals(location.getName())
@@ -212,7 +236,7 @@ public final class Location extends CwmsDTO {
                 getLocationType(), getLocationKind(), getNation(), getStateInitial(),
                 getCountyName(), getHorizontalDatum(), getPublishedLongitude(),
                 getPublishedLatitude(), getVerticalDatum(), getElevation(), getMapLabel(),
-                getBoundingOfficeId(), getOfficeId(), getElevationUnits());
+                getBoundingOfficeId(), getOfficeId(), getElevationUnits(), getAliases());
     }
 
     @Override
@@ -270,6 +294,7 @@ public final class Location extends CwmsDTO {
         private String mapLabel;
         private String boundingOfficeId;
         private String elevationUnits;
+        private List<LocationAlias> aliases = new ArrayList<>();
         private static final String MISSING_NAME_ERROR_MSG = "Location name is a required field";
         private final Map<String, Consumer<Object>> propertyFunctionMap = new HashMap<>();
 
@@ -328,6 +353,7 @@ public final class Location extends CwmsDTO {
             this.mapLabel = location.getMapLabel();
             this.boundingOfficeId = location.getBoundingOfficeId();
             this.elevationUnits = location.getElevationUnits();
+            this.aliases = location.getAliases();
             buildPropertyFunctions();
         }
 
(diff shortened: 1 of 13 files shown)


rma-rripken:
I'm curious about LocationAlias switching from the locationLevelId field to categoryId-groupId.  How and when are they different, what does that change do?

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
```diff
@@ -54,6 +57,8 @@ public final class Location extends CwmsDTO {
     private final String mapLabel;
     private final String boundingOfficeId;
     private final String elevationUnits;
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    private final List<LocationAlias> aliases;
```
rma-rripken: Most other places we used NON_NULL,  Is there a reason NON_EMPTY is better?  If you don't request the aliases I think it makes sense not to serialize the list (b/c its null).  But if you do request aliases and there aren't any I think its nice to see [].  I could be convinced that is wrong though

zack-rma:
The categoryId-groupId alias field is the correct format, as seen in the example input of the initial issue. It can also been seen here in code: https://github.com/USACE/cwms-data-api/blame/e56603d0de8c393326c1dc089250177c7c44163b/cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java#L576.

I mistakenly set this value to the locationId, so this was a fix. I also applied it to my recent alias change for location levels.

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
```diff
@@ -270,6 +294,7 @@ public static class Builder {
         private String mapLabel;
         private String boundingOfficeId;
         private String elevationUnits;
+        private List<LocationAlias> aliases = new ArrayList<>();
```
rma-rripken: whether this starts out null or empty is probably whether the serialization is  NON_NULL or NON_EMPTY

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
zack-rma: The alias list is created as an empty list as part of the builder to avoid NPE when using methods on the results of the getter, such as `isEmpty()`. This means that it is not considered null, so it will always be serialized unless the inclusion is set to NON_EMPTY. 

I could see having an empty list of aliases be serialized to `[]` as being handy to quickly recognize a lack of aliases. I could rework the builder to do so, but there would be a risk of throwing a NPE on certain operations expecting that list to be non-null.

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
zack-rma: Yes, though leaving it as null can cause NPE when using the value retrieved from the getter. Generally, I would avoid leaving a list as null and would use an empty list instead. If it would be preferred to have an empty list when including aliases in results, this can be changed.

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
MikeNeilson: I rather agree with @rma-rripken 's thoughts on this one. don't show if not requested, but empty list if it is.

can we return Optional and get the same behavior?

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
rma-rripken: I think your instincts about preferring an empty list over a null object are right but we've been using NOT_NULL across CDA and I'm not convinced that there is anything special about this case to be inconsistent and switch to NON_EMPTY here or to change everything across the board.  Returning an empty list can be reassuring even if its verbose.  Its really clear to the user that we looked and there weren't any aliases.  NOT_NULL also makes the DTO match the JSON - for better and worse.  It does mean that we have to handle null fields but we can do that and write tests for that.

MikeNeilson (changes requested):
I'm with Ryan. Field should be null and not render if aliases not requested and empty array if requested and no aliases.

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
zack-rma: I've updated the list handling to include aliases when the list is empty, but not null. The new annotation is `NOT_ABSENT` for the Optional field.

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
zack-rma: I've updated the Location DTO to return an Optional for aliases. It now includes aliases when they are present or an empty list, but leaves them off when null.

cwms-data-api/src/main/java/cwms/cda/formatters/json/JsonV2.java
```diff
@@ -67,6 +68,7 @@ public static ObjectMapper buildObjectMapper() {
         retVal.setPropertyNamingStrategy(PropertyNamingStrategies.KEBAB_CASE);
         retVal.setSerializationInclusion(JsonInclude.Include.NON_NULL);
         retVal.registerModule(new JavaTimeModule());
+        retVal.registerModule(new Jdk8Module());
```
MikeNeilson, line 71: excellent, saved by a simple plugin.

cwms-data-api/src/test/java/cwms/cda/helpers/DTOMatch.java
```diff
@@ -294,7 +295,25 @@ public static void assertMatch(Location first, Location second) {
                 () -> assertEquals(first.getCountyName(), second.getCountyName()),
                 () -> assertEquals(first.getTimezoneName(), second.getTimezoneName()),
                 () -> assertEquals(first.getOfficeId(), second.getOfficeId()),
-                () -> assertEquals(first.getLocationType(), second.getLocationType())
+                () -> assertEquals(first.getLocationType(), second.getLocationType()),
+                () -> {
+                        if (first.getAliases().isPresent() && second.getAliases().isPresent()) {
+                            assertEquals(first.getAliases().get().size(), second.getAliases().get().size(),
+                                "Alias list sizes do not match");
+                            for (LocationAlias alias : first.getAliases().get()) {
```
MikeNeilson, line 303: Not that it matters given these arrays are small but I can't help but think there's a cleaner way to do this.

Though perhaps the Aliases should be a Set, and not even an ordered one, might make the comparison easier, I'm pretty sure we can't have duplicate alias names.

But that doesn't need to happen this PR. and the duplication here make sense as you're providing more specific asserts for diagnostics.

MikeNeilson (approved):
Some thoughts, but nothing that rises to the level of "we absolutely must make the change."

PR #1321, opened 2025-09-18
Add support for `localDatumName` in `VerticalDatumInfo` DTO and its b…

rma-rripken:
…uilder, with tests and example XML

This doesn't implement all the requested features but should prevent the parsing error some were seeing


cwms-data-api/src/main/java/cwms/cda/data/dto/VerticalDatumInfo.java
```diff
@@ -20,6 +20,9 @@ public class VerticalDatumInfo extends CwmsDTOBase {
     String nativeDatum;
     Double elevation;
 
+    public String localDatumName;
```
MikeNeilson: public?

MikeNeilson (commented):
The jooq record change will definitely help with some Java 17 warning/error messages.

@ktarbet / @perrymanmd does the insert of local datum in this make sense to you.

I want to say it does to me, but I haven't really done much directly with how the vertical datum information is stored and shipped around.

cwms-data-api/src/main/java/cwms/cda/data/dto/VerticalDatumInfo.java
rma-rripken: yikes.  Good catch.

rma-rripken:
@perrymanmd  pointed out a stacktrace that came from CDA not properly parsing the xml when it included localDatumName.  This just fixes that exception.

rma-rripken:
I took the xml example from his post and made sure it could be parsed.   I also had a look at VerticalDatumContainer to see if there was anything else that was obviously missing.  I didn't see anything obvious.

MikeNeilson:
> @perrymanmd pointed out a stacktrace that came from CDA not properly parsing the xml when it included localDatumName. This just fixes that exception.

Now I remember that.

MikeNeilson (approved):

PR #1324, opened 2025-09-18
Updated total count query for location level retrieval

zack-rma:
Improved efficiency of total count retrieval query in location level retrieval.


rma-rripken (dismissed):

rma-rripken:
Interesting that the time_offset is one of the query columns but wasn't in the order columns.   Should we be ordering by time_offset too?

zack-rma:
We likely should be. In my testing, I kept running into the case where the number of results between the old and new query was just slightly off (4570 vs 4535) when the time_offset wasn't included.

rma-rripken:
If the time-offset needs to be in the selectDistinct part then it should be in the order-by clauses of the data fetching queries or else the pages might not return consistent contents.  If its not ordered by that field then two queries can return rows different orders, when we do the offset part to skip to the right page of data we could end up doubling some rows and skipping others.

rma-rripken (approved):

PR #1328, opened 2025-09-22
CDA-29: Fix TimeSeries group request slower than expected

### LINKED ISSUE #1022: TimeSeries group request slower than expected
```
curl -X 'GET' \
  'https://<host>/cwms-data/timeseries/group?office=NWDP&include-assigned=true&timeseries-category-like=NWP_A.%2A' \
  -H 'accept: application/json'
```

The returned result is rather large, but should be even then should be relatively fast given what the data is.

Tested on cwbi-test and future prod, equally slow on both.

zack-rma:
Fixes #1022

Update in progress to TimeSeries group retrieval query.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
@@ -128,7 +128,7 @@ public class TimeSeriesGroupDao extends JooqDao<TimeSeriesGroup> {
             joinCond = joinCond.and(grpAssgn.DB_OFFICE_ID.eq(tsOfficeId));
         }
 
-        SelectConditionStep<?> query = dsl.select(catGrp.CAT_DB_OFFICE_ID,
+        SelectSeekStep2<?, String, BigDecimal> query = dsl.select(catGrp.CAT_DB_OFFICE_ID,
                         catGrp.TS_CATEGORY_ID, catGrp.TS_CATEGORY_DESC, catGrp.GRP_DB_OFFICE_ID,
                         catGrp.TS_GROUP_ID, catGrp.TS_GROUP_DESC, catGrp.SHARED_TS_ALIAS_ID,
                         catGrp.SHARED_REF_TS_ID, grpAssgn.CATEGORY_ID, grpAssgn.DB_OFFICE_ID,
@@ -137,7 +137,8 @@ public class TimeSeriesGroupDao extends JooqDao<TimeSeriesGroup> {
                 .from(catGrp).leftJoin(grpAssgn)
                 .on(joinCond)
             .where(whereCond)
-            .and(whereCondGrpCat);
+            .and(whereCondGrpCat)
+            .orderBy(catGrp.TS_GROUP_ID, grpAssgn.ATTRIBUTE);
 
         logger.fine(() -> query.getSQL(ParamType.INLINED));
 
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
@@ -30,12 +30,8 @@ import cwms.cda.data.dto.AssignedTimeSeries;
 import cwms.cda.data.dto.TimeSeriesCategory;
 import cwms.cda.data.dto.TimeSeriesGroup;
 import java.math.BigDecimal;
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.logging.Logger;
-import kotlin.Pair;
 import org.jetbrains.annotations.NotNull;
 import org.jooq.*;
 import org.jooq.conf.ParamType;
@@ -106,14 +102,6 @@ public class TimeSeriesGroupDao extends JooqDao<TimeSeriesGroup> {
         AV_TS_CAT_GRP catGrp = AV_TS_CAT_GRP.AV_TS_CAT_GRP;
         AV_TS_GRP_ASSGN grpAssgn = AV_TS_GRP_ASSGN.AV_TS_GRP_ASSGN;
 
-        final RecordMapper<org.jooq.Record, Pair<TimeSeriesGroup, List<AssignedTimeSeries>>> mapper =
-                queryRecord -> {
-                    TimeSeriesGroup group = buildTimeSeriesGroup(queryRecord);
-                    List<AssignedTimeSeries> loc = (List<AssignedTimeSeries>) queryRecord.get("multiset");
-
-                    return new Pair<>(group, loc);
-                };
-
         Condition whereCondGrpCat = DSL.noCondition();
         if (categoryOfficeId != null) {
             whereCondGrpCat = whereCondGrpCat.and(catGrp.CAT_DB_OFFICE_ID.eq(categoryOfficeId.toUpperCase()));
@@ -128,8 +116,8 @@ public class TimeSeriesGroupDao extends JooqDao<TimeSeriesGroup> {
             joinCond = joinCond.and(grpAssgn.DB_OFFICE_ID.eq(tsOfficeId));
         }
 
-        SelectConditionStep<Record9<String, String, String, String, String, String, String, String,
-                List<AssignedTimeSeries>>> query = dsl
+        SelectSeekStep4<Record9<String, String, String, String, String, String, String, String,
+                List<AssignedTimeSeries>>, String, String, String, String> query = dsl
             .select(
                 catGrp.CAT_DB_OFFICE_ID,
                 catGrp.TS_CATEGORY_ID,
@@ -139,7 +127,6 @@ public class TimeSeriesGroupDao extends JooqDao<TimeSeriesGroup> {
                 catGrp.TS_GROUP_DESC,
                 catGrp.SHARED_TS_ALIAS_ID,
                 catGrp.SHARED_REF_TS_ID,
-
                 DSL.multiset(
                     dsl
                         .select(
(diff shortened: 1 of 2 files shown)


rma-rripken:
The order-by should probably be:  category_office_id, category_id, group_office_id, group_id, and then the attribute.   You might consider DSL.multiset.

rma-rripken:
or it could be multiple queries.   One to find the distinct keys (category_office_id, category_id, group_office_id, group_id)  Then another query to find assignments groupedBy keys ordered by attribute.

rma-rripken:
We might also consider adding paging.  I suggest paging by fixed number of groups per page, keep entire group within one page and not try to count assignments.  The LOE didn't include adding paging.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
```diff
@@ -128,30 +131,49 @@ private List<TimeSeriesGroup> getTimeSeriesGroupsWhere(Condition whereCond, Stri
...
+                catGrp.CAT_DB_OFFICE_ID,
+                catGrp.TS_CATEGORY_ID,
+                catGrp.TS_CATEGORY_DESC,
+                catGrp.GRP_DB_OFFICE_ID,
+                catGrp.TS_GROUP_ID,
+                catGrp.TS_GROUP_DESC,
+                catGrp.SHARED_TS_ALIAS_ID,
+                catGrp.SHARED_REF_TS_ID,
+
+                // Use multiset to efficiently fetch localized grpAssgn records
```
rma-rripken: remove this comment

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
```diff
@@ -128,30 +131,49 @@ private List<TimeSeriesGroup> getTimeSeriesGroupsWhere(Condition whereCond, Stri
...
+            .from(catGrp)
             .where(whereCond)
-            .and(whereCondGrpCat)
-            .orderBy(catGrp.TS_GROUP_ID, grpAssgn.ATTRIBUTE);
+            .and(whereCondGrpCat);
 
-        logger.fine(() -> query.getSQL(ParamType.INLINED));
 
-        List<Pair<TimeSeriesGroup, AssignedTimeSeries>> assignments =
+        logger.severe(() -> query.getSQL(ParamType.INLINED));
```
rma-rripken: shouldn't be severe

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
```diff
@@ -128,30 +131,49 @@ private List<TimeSeriesGroup> getTimeSeriesGroupsWhere(Condition whereCond, Stri
...
+
+        List<Pair<TimeSeriesGroup, List<AssignedTimeSeries>>> assignments =
```
rma-rripken: If I understand multiset there is probably a way to go right from the query to the Map<group, list> type that we want for return.   If you can't figure it out in a couple minutes just keep what you have

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
```diff
@@ -183,23 +205,33 @@ private List<TimeSeriesGroup> getTimeSeriesGroupsWithoutAssigned(Condition where
...
+    private List<AssignedTimeSeries> buildAssignedTimeSeries(org.jooq.Record queryRecord) {
+        List<AssignedTimeSeries> retval = new ArrayList<>();
+
+        String assignRecord = queryRecord.get("assignments", String.class);
+        if (assignRecord != null) {
+            try {
+                JSONArray jsonArr = (JSONArray) new JSONParser().parse(assignRecord);
+                for (Object obj : jsonArr) {
+
+                    JSONArray json = (JSONArray) obj;
```
rma-rripken: Curious.  We haven't gotten results as string and then parsed them as JSON anywhere else.  I respect the hustle but I prefer adding a mapper and pulling explicit fields out of the result and feeding them into a method reference to do the Result->DTO mapping.

rma-rripken: I'm going to remember this trick for more desperate situations

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
```diff
@@ -128,30 +131,49 @@ private List<TimeSeriesGroup> getTimeSeriesGroupsWhere(Condition whereCond, Stri
...
+                            grpAssgn.TS_ID,
+                            grpAssgn.DB_OFFICE_ID,
+                            grpAssgn.ATTRIBUTE,
+                            grpAssgn.ALIAS_ID,
+                            grpAssgn.REF_TS_ID
+                        )
+                        .from(grpAssgn)
+                        .where(joinCond)
+                        .orderBy(grpAssgn.ATTRIBUTE) // Localized ordering inside the group
+                ).as("assignments") // Alias for the resulting nested collection
```
rma-rripken: instead of .as("assignments")   I think you can do something like 
.convertFrom(rs -> rs.map(this::buildAssignedTimeSeries))

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
MikeNeilson: Arguably worth doing *if* the resulting speed justifies it. But really only then.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
MikeNeilson: Does anyone know if java.util.logging can do a "also include user configuration?"

If so we should really just implement that, we all do this when finding problems... and we all accidentally leave it in; I'm as guilty of that as anyone.

If not, perhaps we should consider a different logger for the tests... well, and actually using a logging facade... but that's a whole other question than this PR.

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesGroupControllerTestIT.java
```diff
@@ -1666,6 +1666,6 @@ void testRetrievalTiming() {
         .assertThat()
             .statusCode(is(HttpServletResponse.SC_OK))
             .body("assigned-time-series.size()", greaterThan(0))
-            .time(lessThan(250L)); // should be pretty quick, under 0.5 seconds. Old query was ~3 seconds
+            .time(lessThan(500L)); // should be pretty quick, under 0.5 seconds. Old query was ~3 seconds
```
MikeNeilson, line 1669: I certainly like 250ms better, but I'll concede that 500 is a large improvement over 3s.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
```diff
@@ -183,23 +205,33 @@ private List<TimeSeriesGroup> getTimeSeriesGroupsWithoutAssigned(Condition where
...
+                    Integer attr = null;
+                    if (attrBD != null) {
+                        attr = attrBD.intValue();
+                    }
+
+                    retval.add(new AssignedTimeSeries(officeId, timeseriesId, aliasId, refTsId, attr));
+                }
+            } catch (ParseException e) {
+                logger.severe(() -> "Error parsing assigned time series JSON: " + e.getMessage());
+                throw new RuntimeException(e);
```
MikeNeilson: Don't we have an exception system to handle this more cleanly?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
zack-rma: Updated mapping method to use Result type

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesGroupControllerTestIT.java
zack-rma: I was seeing closer to 250 regularly, but the occasional ~260ms was causing failure. I set it to 500 to be safe.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
zack-rma: Removed

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
zack-rma: Updated as suggested

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
zack-rma: Reverted to fine

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
zack-rma: Removed and transitioned to using Record type

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
zack-rma: Removed

rma-rripken:
I think it can be even cleaner.   I don't think the multiset needs to sort on all those fields. The advantage of the multiset is that the sorting can happen separately.
We shouldn't have to do the dance with the List of Pair<>, we should be able to just map to the final results.

Can you try something like this? 

```
        SelectSeekStep4<Record9<String, String, String, String, String, String, String, String, List<AssignedTimeSeries>>, 
                String, String, String, String> query = 
                dsl.select(
                        catGrp.CAT_DB_OFFICE_ID,
                        catGrp.TS_CATEGORY_ID,
                        catGrp.TS_CATEGORY_DESC,
                        catGrp.GRP_DB_OFFICE_ID,
                        catGrp.TS_GROUP_ID,
                        catGrp.TS_GROUP_DESC,
                        catGrp.SHARED_TS_ALIAS_ID,
                        catGrp.SHARED_REF_TS_ID,
                        DSL.multiset(
                                dsl.select(
                                                grpAssgn.DB_OFFICE_ID,
                                                grpAssgn.TS_ID,
                                                grpAssgn.ATTRIBUTE,
                                                grpAssgn.ALIAS_ID,
                                                grpAssgn.REF_TS_ID
                                        )
                                        .from(grpAssgn)
                                        .where(joinCond)
                                        .orderBy(grpAssgn.ATTRIBUTE)
                        ).convertFrom(rs -> rs.map(this::buildAssignedTimeSeries))
                )
                .from(catGrp)
                .where(whereCond)
                .and(whereCondGrpCat)
                .orderBy(
                        catGrp.CAT_DB_OFFICE_ID,
                        catGrp.TS_CATEGORY_ID,
                        catGrp.GRP_DB_OFFICE_ID,
                        catGrp.TS_GROUP_ID
                );

        logger.fine(() -> query.getSQL(ParamType.INLINED));

        RecordMapper<? super Record9<String, String, String, String, String, String, String, String, List<AssignedTimeSeries>>, TimeSeriesGroup> mapperToTimeSeriesGroup =
                queryRecord -> {
                    TimeSeriesGroup group = buildTimeSeriesGroup(queryRecord);
                    List<AssignedTimeSeries> assignedTS = (List<AssignedTimeSeries>) queryRecord.get("multiset");

                    return new TimeSeriesGroup(group, assignedTS);
                };

        return query.fetch(mapperToTimeSeriesGroup);
```

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
```diff
@@ -151,35 +138,29 @@ private List<TimeSeriesGroup> getTimeSeriesGroupsWhere(Condition whereCond, Stri
...
+                    return new TimeSeriesGroup(group, assignedTS);
+                };
 
-        List<TimeSeriesGroup> retval = new ArrayList<>();
-        for (final Map.Entry<TimeSeriesGroup, List<AssignedTimeSeries>> entry : map.entrySet()) {
-            List<AssignedTimeSeries> assigned = entry.getValue();
-            retval.add(new TimeSeriesGroup(entry.getKey(), assigned));
-        }
-        return retval;
+        return query.fetch(mapperToTimeSeriesGroup);
```
rma-rripken, line 163: That looks much cleaner.  Hopefully the performance is still there.

rma-rripken (approved):

PR #1345, opened 2025-10-01
Bugfix/remove auth from IT GET requests

rma-rripken:
I noticed that Timeseries IT were sending the Authentication header to some of the get requests.  Except for a couple places (which we should do a better job of documenting) all the GET shouldn't need auth.  I pulled those changes into this "clean" PR.


MikeNeilson (commented):
well, there's definitely nothing wrong with this PR. However, I wonder if we want this to actually be true. At least everywhere.

While we certainly have "guest" there is going to be additional read authorization requirements soon so we should keep at least a few.

But I'm not certain enough to even remotely dictate that so I'm going to leave this on comment so if someone else approves, or you feel strongly enough to bypass, go for it.

MikeNeilson (approved):

Issue #836, opened 2024-08-09
Browser Caching Swagger Page "Get" Requests at Standard Text Endpoint


msweier:
I noticed Chrome was caching my Swagger page "Get" request when I was playing around GET/POST/DELETE with the Standard Text end point. After a POST or DELETE the GET request on the Swagger page wouldn't reflect the change. Not sure if anything can be done, but something to be aware of.

krowvin:
> Not sure if anything can be done  

CDA sets the header for `Cache-Control` to 300 seconds or roughly 5 minutes. 

In Swagger if you want to do your testing there you can (on your end) click the box to "disable cache" in the dev tools. 
![image](https://github.com/user-attachments/assets/c02c9b1f-3126-43bd-92bd-6d3b1ed9bf04)

As for client applications that are either native, web based, or otherwise. You could manually decide if you want to cache or re-request. 

As just an example for web that would look like this (For VanillaJS) 

```javascript 
fetch("cda/ts/url", { cache: "no-store" }).then((response) => {
  /* consume the response */
});

```
Via
https://developer.mozilla.org/en-US/docs/Web/API/Request/cache#examples


But good point this is something to be aware of, and i'm not sure this is listed anywhere so thought i'd mention the above!

msweier:
It's odd that CLOB end point doesn't seem to have this issue.

krowvin:
Good chance the clob doesn't have the cache headers set 

I just assumed all were by default set to 5

@MikeNeilson any ideas on which endpoints might have this?

MikeNeilson:
Should be most of them. Which 99% of the time is the right thing to to do. This is clearly in that 1%. That said I think the full and proper solution is going to use etags and the database informing the API there's a new version.

MikeNeilson:
A simple solution can just be to add a random, but unused parameter to the parameters, but that only works outside of swagger.

rma-rripken:
@msweier  How are you using the Clob endpoint? Are you requesting as text/plain?  The Clob controller should have the same caching headers set but text/plain is being handled in a special way for clob so that large clob (gigabytes?) can be streamed directly from database and the clob end-point also supports range-requests.  It wouldn't surprise me if browser or libraries are making requests against clob in ways that aren't going to benefit from normal cache-control headers.

msweier:
> @msweier How are you using the Clob endpoint? Are you requesting as text/plain? 

[...]

PR #1350, opened 2025-10-06
Add a cache busting parameter "_cb" to requests sent from swagger-ui page.   Also request proxies not to cache.

rma-rripken:


MikeNeilson (approved):

krowvin (commented):
Wonder why the formatting was different. Might be worth adding husky in to ensure we don't toggle between formatter preferences. 

Edit:
- #1351

krowvin:
Resolves 
- #836

Issue #987, opened 2024-12-19
CDA can't handle Timezones that don't exist anymore (or at all)


DanielTOsborne:
The CWMS schema contains timezones that either never existed (placeholders), or no longer exist in the host system (java runtime).
As such, any attempt to retrieve locations fails with the following:

```
18-Dec-2024 23:55:55.954 SEVERE [https-openssl-nio-8243-exec-48] cwms.cda.api.LocationController.getAll 7896909191952785378: failed to process request
        java.time.zone.ZoneRulesException: Unknown time-zone ID: US/Pacific-New
                at java.time.zone.ZoneRulesProvider.getProvider(ZoneRulesProvider.java:272)
                at java.time.zone.ZoneRulesProvider.getRules(ZoneRulesProvider.java:227)
                at java.time.ZoneRegion.ofId(ZoneRegion.java:120)
                at java.time.ZoneId.of(ZoneId.java:411)
                at java.time.ZoneId.of(ZoneId.java:359)
                at cwms.cda.data.dao.LocationsDaoImpl.buildLocation(LocationsDaoImpl.java:147)
                at org.jooq.RecordMapper.apply(RecordMapper.java:87)
                at org.jooq.RecordMapper.apply(RecordMapper.java:72)
                at java.util.stream.Collectors.lambda$mapping$8(Collectors.java:355)
                at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
                at java.util.Iterator.forEachRemaining(Iterator.java:116)
                at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
                at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
                at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
                at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
                at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
                at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
                at org.jooq.impl.AbstractCursor.collect(AbstractCursor.java:78)
                at org.jooq.impl.ResultQueryTrait.collect(ResultQueryTrait.java:361)
                at org.jooq.impl.ResultQueryTrait.fetch(ResultQueryTrait.java:1465)
[...]

PR #1352, opened 2025-10-07
Additional timezone mappings.

rma-rripken:
I'm not positive these are needed in production but we do hit some of these timezones in our CDA test data.


rma-rripken:
#987

MikeNeilson (approved):

PR #1354, opened 2025-10-13
Bugfix/cda 33 forecast spec save permissions issue

rma-rripken:
https://github.com/USACE/cwms-data-api/issues/1108


cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
```diff
@@ -183,14 +183,17 @@ public static void register_users() throws Exception {
                 }
                 if (user == TestAccounts.KeyUser.SPK_OTHER_NORMAL_SAME_ROLES) {
                     String name = user.getName();
-                    String officeId = db.getOfficeId();
+//                    String officeId = db.getOfficeId();  // This comes back as HQ
```
MikeNeilson: don't we assign an office to the TestAccounts.KeyUser instance?

If not I'm good with this, but we should open an issue to track making this more generic.

MikeNeilson (dismissed):

rma-bryson (dismissed):

cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
rma-rripken: Oh, right.  OperatingOffice. I'll fix that.

rma-rripken:
closes #1108

rma-rripken:
I'm not sure how much this PR actually now proves.  We're creating a user account and giving them write access.  We're doing it a slightly different way than the rest of the harness but the end-result is the same. We have a user that has write access for an office and then we show that the user can create a forecast spec.  I guess it proves that more than one user can create.

rma-bryson:
I guess if someone tries something similar in the future and things fail, this is a reference point that shows a need to add users to the TS ID Creators group, or things don't work?

As for the original issue - I think we can close that out as Not Reproduceable (by us or by the original reporter).

rma-bryson (approved):

PR #1357, opened 2025-10-15
Feature/1356 openid configuration

### LINKED ISSUE #1356: Modify OpenID configuration settings to process and allow an `x-kc_idp_hint field`
CWMS Usage will allow direct CAC loggin as well as Login.gov, thus we need to support multiple OpenID Connect providers

Make use of https://swagger.io/docs/specification/v3_0/openapi-extensions/ to allow setting the parameter and use openid connect discover to retrieve the bulk of details.

None

**Additional context**
All access is still threw already used keycloak instance, this simply alters how we're doign it.

MikeNeilson:
Fixes #1356.

Definitely still needs work (should only try the initOAuth if an OAuth based security scheme is actually present, but it gets the idea across.

Unfortunately, at least using the default Swagger Authentication system, I don't think I can do what I wanted where one, in the auth button, could have a login that would automatically work and pass  the additional required query parameter. The SwaggerUI code simply doesn't do anything with extension values provided in the scheme.

However, that really only affects the SwaggerUI not usages of the application itself. So I'm inclined not to worry about it and will strip of some of the extra elements I added to the code.

Good news is, does pre-populate the client id, and the login flow behaves correctly; I was redirected to the local docker-compose keycloak and saw the kc_idp_hint field on the request so in environment should behave the same.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,9 @@ ENV cwms.dataapi.access.providers="KeyAccessManager,OpenID"
 ENV cwms.dataapi.access.openid.wellKnownUrl="https://<prefix>/.well-known/openid-configuration"
 ENV cwms.dataapi.access.openid.issuer="<issuer>"
 ENV cwms.dataapi.access.openid.timeout="604800"
+# Putting default values here to easy configuration
+ENV cwms.dataapi.access.openid.clientId=cwms
+ENV cwms.dataapi.access.openid.idpHint=federation-eams
 #ENV cwms.dataapi.access.openid.altAuthUrl="https://identityc-test.cwbi.us/auth/realms/cwbi"
 
 # used to simplify redeploy in certain contexts. Update to match -<marker> in image label
--- a/cda-gui/src/pages/swagger-ui/index.jsx
+++ b/cda-gui/src/pages/swagger-ui/index.jsx
@@ -5,32 +5,42 @@ import { useEffect } from "react";
 import { getBasePath } from "../../utils/base";
 
 export default function SwaggerUI() {
-  useEffect(() => {
-    // document.querySelector("#swagger-ui").prepend(Index)
-    // TODO: Add page index to top of page
-    // Alter the page title to match the swagger page
-    document.title = "CWMS Data API for Data Retrieval - Swagger UI";
-    // Begin Swagger UI call region
-    // TODO: add endpoint that dynamic returns swagger generated doc
-    SwaggerUIBundle({
-      url: getBasePath() + "/swagger-docs",
-      dom_id: "#swagger-ui",
-      deepLinking: false,
-      presets: [SwaggerUIBundle.presets.apis],
-      plugins: [SwaggerUIBundle.plugins.DownloadUrl],
-      requestInterceptor: (req) => {
-        // Add a cache-busting query param
-        const sep = req.url.includes("?") ? "&" : "?";
-        req.url = `${req.url}${sep}_cb=${Date.now()}`;
+    useEffect(() => {
+        // document.querySelector("#swagger-ui").prepend(Index)
+        // TODO: Add page index to top of page
+        // Alter the page title to match the swagger page
+        document.title = "CWMS Data API for Data Retrieval - Swagger UI";
+        // Begin Swagger UI call region
+        // TODO: add endpoint that dynamic returns swagger generated doc
+        const ui = SwaggerUIBundle({
+            url: getBasePath() + "/swagger-docs",
+            dom_id: "#swagger-ui",
+            deepLinking: false,
+            presets: [SwaggerUIBundle.presets.apis],
+            plugins: [SwaggerUIBundle.plugins.DownloadUrl],
+            requestInterceptor: (req) => {
+                // Add a cache-busting query param... but only if it's to our api. Some
+                // external systems, like keycloak, don't allow random unknown parameters.
+                const origin = window.location.origin;                
+                const re = new RegExp(`^${origin}.*`)
+                if (re.test(req.url))
+                {
+                    const sep = req.url.includes("?") ? "&" : "?";
+                    req.url = `${req.url}${sep}_cb=${Date.now()}`;
 
-        // Also ask intermediaries not to serve from cache
-        req.headers["Cache-Control"] = "no-cache, no-store, max-age=0";
-        req.headers["Pragma"] = "no-cache";
-
-        return req;
-      },
-    });
-  }, []);
+                    // Also ask intermediaries not to serve from cache
+                    req.headers["Cache-Control"] = "no-cache, no-store, max-age=0";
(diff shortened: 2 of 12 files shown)


MikeNeilson:
@krowvin Perhaps it would make more sense to just integrate the groundwater login dialog instead of relying on the Swagger UI system? I'm fairly certain it's not difficult to do; should be able to hook into the internal auth bits that fill in those details.

MikeNeilson:
Now includes client_id value within an extension field. Considering 
1. In the Swagger-UI we have to tell users what it is
2. Anyone with developer tools, or possibly even just eyeballs on an address bar, can see it
Thus this is not secret information and we should just provide it.

cda-gui/src/pages/swagger-ui/index.jsx
```diff
@@ -34,13 +36,17 @@ export default function SwaggerUI() {
                 }
                 return req;
             },
+            onComplete: () => {
+                const spec = JSON.parse(ui.spec().get("spec"));
+                console.log(JSON.stringify(spec.components.securitySchemes));
```
krowvin: Remove the logs if you are ready for prime time!

cda-gui/vite.config.js
```diff
@@ -3,7 +3,7 @@ import react from "@vitejs/plugin-react";
 
 // https://vitejs.dev/config/
 export default defineConfig(({ mode }) => {
-  // const env = loadEnv(mode, process.cwd(), "");
+  const env = loadEnv(mode, process.cwd(), "");
```
krowvin, line 6: This works with Node, but you should also see

https://vite.dev/guide/env-and-mode

i.e.

`const CUSTOM_VAR = import.meta.env.VITE_CUSTOM_VAR`

Where you prefix with VITE_ to avoid accidental env var injection.. from your `.env` files

krowvin (dismissed):
Up to you if you want to change these things. Tossing info out there!

(There's CI/CD you can setup to remove console logs... before prod)

cda-gui/src/pages/swagger-ui/index.jsx
MikeNeilson: Still draft, just push that change up before lunch in case the computer crashed.

cda-gui/vite.config.js
MikeNeilson: Thanks, was wondering why that that was comment out.

MikeNeilson: well, that doesn't actually work within the vite.config.js

MikeNeilson:
Okay, now it's ready. Once it deploys to -dev you should be able to use the very first OIDC option ("Code") and not have to know the client id and get forwarded through the various identity provides and otherwise login as usual.

krowvin:
> Should be able to use the very first OIDC option

Confirming the user still clicks the green `Authorize` button, then selected the option in there? 

Then later we might add a Login button somewhere that could also tie into that? 

I imagine SwaggerUI (constructor) has a callback we can use to setup a Login button. Let me know if you want me to look into that first before trying to setup the GWW login methods. (It only supports direct-login right now afaik)

MikeNeilson:
> > Should be able to use the very first OIDC option
> 
> Confirming the user still clicks the green `Authorize` button, then selected the option in there?
> 
> Then later we might add a Login button somewhere that could also tie into that?
> 
> I imagine SwaggerUI (constructor) has a callback we can use to setup a Login button. Let me know if you want me to look into that first before trying to setup the GWW login methods. (It only supports direct-login right now afaik)

Correct, other than not needing to already know the client id, and being able to choose the first option, the user side flow is the same. (They will get redirected to the auth server, but people should be used to that by now in general.)

MikeNeilson:
@jbkolze @oskarhurst @krowvin  FYI, keycloak login in -dev is currently broken due to the cache buster code sending unknown parameters to the keycloak instance which then intentionally fails since it doesn't like unknown query parameters.

Somebody give this a look when you have time. There will likely be improvements to the SecurityScheme extension in a follow up, like providing a link to a graphic. Work on https://github.com/opendcs/rest_api/pull/574 is informing that. (Plan to share the extension design since... why not, same need.)

cda-gui/src/pages/swagger-ui/index.jsx
MikeNeilson: The log statements should be gone now.

Dockerfile
```diff
@@ -56,6 +56,9 @@ ENV cwms.dataapi.access.providers="KeyAccessManager,OpenID"
 ENV cwms.dataapi.access.openid.wellKnownUrl="https://<prefix>/.well-known/openid-configuration"
 ENV cwms.dataapi.access.openid.issuer="<issuer>"
 ENV cwms.dataapi.access.openid.timeout="604800"
```
rma-bryson, line 58: Not related to these changes, but is that timeout value of "604800" in minutes or seconds? I believe it gets parsed as minutes which is like 420 days

cda-gui/src/pages/swagger-ui/index.jsx
```diff
@@ -5,32 +5,63 @@ import { useEffect } from "react";
...
+                }
+                return req;
+            },
+            onComplete: () => {
+                const spec = JSON.parse(ui.spec().get("spec"));
+                for (const schemeName in spec.components.securitySchemes) {
+                    const scheme = spec.components.securitySchemes[schemeName];
+                    if (scheme.type === "openIdConnect") {
+                        let additionalParams = null;
+                        let hints = scheme["x-kc_idp_hint"];
```
rma-bryson, line 45: Assumption here is that Keycloak is the only IdP that will be supported?

Dockerfile
MikeNeilson: should be seconds, so about a week.

cda-gui/src/pages/swagger-ui/index.jsx
MikeNeilson: At the moment, yes, I'm working on another project that will handle this more generically but for our current purpose keycloak is the target. That said it does verify if the value is there before it tries to do anything so it can still work with any generic OIDC provider.

Dockerfile
rma-bryson: UrlResolver thinks its minutes. `public UrlResolver(URL jwksUrl, int keyTimeoutMinutes)`

MikeNeilson: ... well then, I can math. 🤦‍♂️ .

MikeNeilson: but I'm not going to fix it this pr, will adjust separate it's been there for ever and the places it's used I can override easily and things are restarted regularly anyways.

rma-bryson: yeah, doesn't need to be addressed here, just something i noticed.

rma-bryson (dismissed):

PR #1358, opened 2025-10-22
CDA-40: Exception Handling Implementation Updates

### LINKED ISSUE #1106: Improve Exception/Error handling logic, heirarchy, etc
> There is no parent class that RateException can extend?  Otherwise what keeps ApiServlet from filling up with implementation specific exception types?  I would think that if we really can't think of anything more generic at the very least we should be able to make a class for Controller exceptions and another for DaoExceptions depending on where the issue is.  I don't see ApiServlet knowing how to handle RateException in any helpful or meaningfully different way than other exceptions.

_Originally posted by @rma-rripken in https://github.com/USACE/cwms-data-api/pull/1082#discussion_r2096272543_

Based on the PR discussion, we need to revisit the error handling. I think we have the crux of the right solution and some recent experience may provide us the next improvement.

zack-rma:
Fixes #1106 

Improves exception handling hierarchy and messages returned to users.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
+++ b/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
@@ -348,8 +348,7 @@ public class ApiServlet extends HttpServlet {
                     ctx.header("X-XSS-Protection", "1; mode=block");
                 })
                 .exception(ApplicationException.class, (e, ctx) -> {
-                    CdaError re = new CdaError(e.getCdaErrorMessage(), e.getCdaHttpErrorCode(),
-                        e.getSource(), e.getDetails());
+                    CdaError re = new CdaError(e.getCdaErrorMessage(), e.getSource(), e.getDetails());
                     logger.atInfo().withCause(e).log(re.toString());
                     ctx.status(e.getCdaHttpErrorCode()).json(re);
                 })
@@ -360,7 +359,7 @@ public class ApiServlet extends HttpServlet {
                     ctx.status(HttpServletResponse.SC_NOT_IMPLEMENTED).json(re);
                 })
                 .exception(BadRequestResponse.class, (e, ctx) -> {
-                    CdaError re = new CdaError("Bad Request", e.getStatus(),
+                    CdaError re = new CdaError("Bad Request",
                         "User Input", new HashMap<>(e.getDetails()));
                     logger.atInfo().withCause(e).log(re.toString());
                     ctx.status(e.getStatus()).json(re);
--- a/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
+++ b/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
@@ -349,7 +349,9 @@ public class ApiServlet extends HttpServlet {
                 })
                 .exception(ApplicationException.class, (e, ctx) -> {
                     CdaError re = new CdaError(e.getCdaErrorMessage(), e.getSource(), e.getDetails());
-                    logger.atInfo().withCause(e).log(re.toString());
+                    if (e.getLoggerLevel().isPresent()) {
+                        logger.at(e.getLoggerLevel().get()).withCause(e).log(re.toString());
+                    }
                     ctx.status(e.getCdaHttpErrorCode()).json(re);
                 })
                 .exception(UnsupportedOperationException.class, (e, ctx) -> {
--- a/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
+++ b/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
@@ -31,9 +31,6 @@ import static cwms.cda.api.Controllers.OFFICE;
 import static cwms.cda.api.Controllers.PROJECT_ID;
 import static cwms.cda.api.Controllers.RATING_ID;
 import static cwms.cda.api.Controllers.WATER_USER;
-
-import cwms.cda.api.CdaVersionHandler;
-import cwms.cda.api.rating.RatingEffectiveDatesController;
 import static io.javalin.apibuilder.ApiBuilder.crud;
 import static io.javalin.apibuilder.ApiBuilder.delete;
 import static io.javalin.apibuilder.ApiBuilder.get;
@@ -55,6 +52,7 @@ import cwms.cda.api.BinaryTimeSeriesController;
 import cwms.cda.api.BinaryTimeSeriesValueController;
 import cwms.cda.api.BlobController;
 import cwms.cda.api.CatalogController;
+import cwms.cda.api.CdaVersionHandler;
 import cwms.cda.api.ClobController;
 import cwms.cda.api.Controllers;
 import cwms.cda.api.CountyController;
@@ -68,6 +66,7 @@ import cwms.cda.api.LevelsController;
 import cwms.cda.api.LocationCategoryController;
 import cwms.cda.api.LocationController;
 import cwms.cda.api.LocationGroupController;
+import cwms.cda.api.LocationKindController;
 import cwms.cda.api.LookupTypeController;
 import cwms.cda.api.MeasurementTimeExtentsGetController;
 import cwms.cda.api.OfficeController;
@@ -75,14 +74,6 @@ import cwms.cda.api.ParametersController;
 import cwms.cda.api.PoolController;
 import cwms.cda.api.ProjectController;
 import cwms.cda.api.PropertyController;
-import cwms.cda.api.errors.RateException;
-import cwms.cda.api.rating.RateTimeSeriesController;
-import cwms.cda.api.rating.RateValuesController;
-import cwms.cda.api.rating.RatingController;
-import cwms.cda.api.rating.RatingLatestController;
-import cwms.cda.api.rating.RatingMetadataController;
-import cwms.cda.api.rating.RatingSpecController;
-import cwms.cda.api.rating.RatingTemplateController;
 import cwms.cda.api.SpecifiedLevelController;
 import cwms.cda.api.StandardTextController;
 import cwms.cda.api.StateController;
@@ -93,10 +84,10 @@ import cwms.cda.api.TextTimeSeriesController;
 import cwms.cda.api.TextTimeSeriesValueController;
 import cwms.cda.api.TimeSeriesCategoryController;
 import cwms.cda.api.TimeSeriesController;
+import cwms.cda.api.TimeSeriesFilteredController;
 import cwms.cda.api.TimeSeriesGroupController;
 import cwms.cda.api.TimeSeriesIdentifierDescriptorController;
 import cwms.cda.api.TimeSeriesRecentController;
-import cwms.cda.api.TimeSeriesFilteredController;
 import cwms.cda.api.TimeZoneController;
 import cwms.cda.api.TurbineChangesDeleteController;
 import cwms.cda.api.TurbineChangesGetController;
@@ -111,15 +102,8 @@ import cwms.cda.api.auth.users.roles.AddRoleController;
 import cwms.cda.api.auth.users.roles.DeleteRolesController;
 import cwms.cda.api.auth.users.roles.GetRolesController;
 import cwms.cda.api.enums.UnitSystem;
-import cwms.cda.api.errors.AlreadyExists;
+import cwms.cda.api.errors.ApplicationException;
 import cwms.cda.api.errors.CdaError;
-import cwms.cda.api.errors.DeleteConflictException;
-import cwms.cda.api.errors.FieldException;
-import cwms.cda.api.errors.InvalidItemException;
-import cwms.cda.api.errors.JsonFieldsException;
-import cwms.cda.api.errors.NotFoundException;
-import cwms.cda.api.errors.RequiredQueryParameterException;
-import cwms.cda.api.errors.ValueTooLongException;
 import cwms.cda.api.location.kind.GateChangeCreateController;
 import cwms.cda.api.location.kind.GateChangeDeleteController;
 import cwms.cda.api.location.kind.GateChangeGetAllController;
@@ -138,11 +122,16 @@ import cwms.cda.api.project.ProjectLockRevokeDeny;
 import cwms.cda.api.project.ProjectPublishStatusUpdate;
 import cwms.cda.api.project.RemoveAllLockRevokerRights;
 import cwms.cda.api.project.UpdateLockRevokerRights;
+import cwms.cda.api.rating.RateTimeSeriesController;
+import cwms.cda.api.rating.RateValuesController;
+import cwms.cda.api.rating.RatingController;
+import cwms.cda.api.rating.RatingEffectiveDatesController;
+import cwms.cda.api.rating.RatingLatestController;
+import cwms.cda.api.rating.RatingMetadataController;
+import cwms.cda.api.rating.RatingSpecController;
+import cwms.cda.api.rating.RatingTemplateController;
 import cwms.cda.api.rating.ReverseRateTimeSeriesController;
 import cwms.cda.api.rating.ReverseRateValuesController;
-import cwms.cda.api.LocationKindController;
-import cwms.cda.api.watersupply.AccountingCatalogController;
-import cwms.cda.api.watersupply.AccountingCreateController;
 import cwms.cda.api.timeseriesprofile.TimeSeriesProfileCatalogController;
 import cwms.cda.api.timeseriesprofile.TimeSeriesProfileController;
 import cwms.cda.api.timeseriesprofile.TimeSeriesProfileCreateController;
@@ -155,6 +144,8 @@ import cwms.cda.api.timeseriesprofile.TimeSeriesProfileParserCatalogController;
 import cwms.cda.api.timeseriesprofile.TimeSeriesProfileParserController;
 import cwms.cda.api.timeseriesprofile.TimeSeriesProfileParserCreateController;
 import cwms.cda.api.timeseriesprofile.TimeSeriesProfileParserDeleteController;
+import cwms.cda.api.watersupply.AccountingCatalogController;
+import cwms.cda.api.watersupply.AccountingCreateController;
 import cwms.cda.api.watersupply.WaterContractCatalogController;
 import cwms.cda.api.watersupply.WaterContractController;
 import cwms.cda.api.watersupply.WaterContractCreateController;
@@ -171,12 +162,8 @@ import cwms.cda.api.watersupply.WaterUserDeleteController;
 import cwms.cda.api.watersupply.WaterUserUpdateController;
 import cwms.cda.data.dao.JooqDao;
 import cwms.cda.formatters.Formats;
-import cwms.cda.formatters.FormattingException;
-import cwms.cda.formatters.UnsupportedFormatException;
 import cwms.cda.security.Authenticator;
 import cwms.cda.security.CdaAccessManager;
-import cwms.cda.security.CwmsAuthException;
-import cwms.cda.security.MissingRolesException;
 import cwms.cda.security.Role;
 import io.javalin.Javalin;
 import io.javalin.apibuilder.CrudFunction;
@@ -198,7 +185,6 @@ import io.swagger.v3.oas.models.PathItem;
 import io.swagger.v3.oas.models.info.Info;
 import io.swagger.v3.oas.models.security.SecurityRequirement;
 import io.swagger.v3.oas.models.servers.Server;
-
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
@@ -209,7 +195,6 @@ import java.time.DateTimeException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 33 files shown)


MikeNeilson (commented):
Looks reasonable to me. Not sure why that test is failing as you didn't change any configuration related to it (other than some formatting).

zack-rma:
Looks like the error message wasn't being passed to the super constructor correctly. I've fixed that and resolved some other similar message pass-through issues.

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesIdentifierDescriptorControllerTestIT.java
```diff
@@ -197,6 +199,38 @@ void test_create_delete_new_LRTS_identifier(String format) throws JsonProcessing
...
+            .redirects().max(3)
+            .post("/timeseries/identifier-descriptor/")
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_BAD_REQUEST))
+            .body("message", equalTo("Bad Request."))
+            .body("source", equalTo("User Input"))
+            .body("details.message",
+                is("Invalid Time Series Description: 12HoursLocal is not a valid duration is not a valid interval"));
```
MikeNeilson: This sentence is... weird. Is the error about a duration or an interval?

MikeNeilson (changes requested):
Looks good other than one error message I simply don't understand the context of.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -362,22 +347,10 @@ public void init() {
                     ctx.header("X-Frame-Options", "SAMEORIGIN");
                     ctx.header("X-XSS-Protection", "1; mode=block");
                 })
-                .exception(UnsupportedFormatException.class, (e, ctx) -> {
-                    CdaError re = new CdaError(e.getMessage());
+                .exception(ApplicationException.class, (e, ctx) -> {
+                    CdaError re = new CdaError(e.getCdaErrorMessage(), e.getSource(), e.getDetails());
                     logger.atInfo().withCause(e).log(re.toString());
```
rma-psmorris: This new exception catch handles many previous caught exceptions. The loggers varied with the majority at info, I see one at severe, one at fine, and some weren't logged. Given these are exceptions, do you want to change the logged level to be higher than info? Warning for example?

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesIdentifierDescriptorControllerTestIT.java
zack-rma: The error message builder was incorrectly appending `is not a valid interval` to the end of the error, regardless of the actual error cause coming from the database. I've removed it and updated the integration tests accordingly.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/errors/ApplicationException.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public String getCdaErrorMessage() {
+        return cdaErrorMessage;
+    }
+
+    public int getCdaHttpErrorCode() {
+        return cdaHttpErrorCode;
+    }
+
+    static Map<String, Serializable> buildDetailsMap(String message) {
```
rma-psmorris: This is scoped for the usage that you have but there are sub-classes that will not be able to use this method. If you intend for the other classes to be able to use this method, then it should be scoped protected.

cwms-data-api/src/main/java/cwms/cda/api/errors/CdaError.java
```diff
@@ -1,94 +1,123 @@
...
 import java.util.Objects;
-import kotlin.random.Random;
+import java.util.UUID;
 
 /**
  * Class for reporting error to users, primary used for default error handlers for exceptions,
  * however you can initialize and return more detail to the user if it makes sense for the given
  * endpoint.
  */
 public class CdaError {
```
rma-psmorris: This class should be declared as final?

cwms-data-api/src/main/java/cwms/cda/api/errors/CdaError.java
```diff
@@ -1,94 +1,123 @@
...
-    private String message;
-    @Schema(description = "A randomly generated number to help identify your request in the logs "
+    private final String message;
+    @Schema(description = "A randomly generated UUID to help identify your request in the logs "
             + "for analysis..")
-    private String incidentIdentifier;
-    private Map<String, ? extends Object> details;
+    private final String incidentIdentifier;
+    private String source;
```
rma-psmorris: Why is source not final?

cwms-data-api/src/main/java/cwms/cda/api/errors/CdaError.java
```diff
@@ -1,94 +1,123 @@
...
     }
 
-    public CdaError(String message, Map<String, ? extends Object> details,
+    /**
+     * Constructor with message and detail map, with option to suppress incident ID generation.
+     * @param message the error message
+     * @param details additional details about the error
+     * @param suppressIncidentId if true, suppresses the incident ID generation
+     */
+    public CdaError(String message, Map<String, Serializable> details,
```
rma-psmorris, line 79: In getDetails you return an unmodifiable details, but here you still allow for mutability given you assign the incoming ref to details. You can push the unmodifiable over to here in the constructor rather than do the unmodifiable during the get.

cwms-data-api/src/main/java/cwms/cda/api/errors/CdaError.java
```diff
@@ -1,94 +1,123 @@
...
+    }
+
     @Override
     public String toString() {
-        return String.format("%s: %s", incidentIdentifier, message);
+        String result;
+        if (source != null) {
+            result = String.format("%s: %s. Originates from %s", incidentIdentifier, message, source);
+        } else {
+            result = String.format("%s: %s", incidentIdentifier, message);
```
rma-psmorris: "%s: %s. Originates from %s"
"%s: %s"

I would include the "." delimiting the two, so "%s: %s."

rma-psmorris: Note that you have other messages that follow this pattern and should be corrected, "%s. %s". DeleteConflictException for example.

cwms-data-api/src/main/java/cwms/cda/api/errors/DeleteConflictException.java
```diff
@@ -24,25 +24,28 @@
...
 import java.util.Map;
+import javax.servlet.http.HttpServletResponse;
 
-public final class DeleteConflictException extends RuntimeException {
+public final class DeleteConflictException extends ApplicationException {
 
     public DeleteConflictException(String message, SQLException cause) {
-        super(message, cause);
+        super(message, "Database", "Cannot perform requested delete. "
+            + "Data is referenced elsewhere in CWMS.", HttpServletResponse.SC_CONFLICT, new HashMap<>(), cause);
```
rma-psmorris: Recommend revision of message, "Cannot delete this record because it is linked to other data in CWMS.”

cwms-data-api/src/main/java/cwms/cda/api/errors/ExclusiveFieldsException.java
```diff
@@ -1,27 +1,35 @@
...
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 public class ExclusiveFieldsException extends FieldException {
     private static final String DETAIL_KEY = "Use only one of";
-    private Map<String, List<String>> details = new HashMap<>();
+    private static final String MESSAGE = "Mutually exclusive fields used.";
```
rma-psmorris: If this is generated from a request, then the error message should be: "Mutually exclusive fields were provided in the request."

cwms-data-api/src/main/java/cwms/cda/api/errors/ExclusiveFieldsException.java
```diff
@@ -1,27 +1,35 @@
 package cwms.cda.api.errors;
 
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 public class ExclusiveFieldsException extends FieldException {
```
rma-psmorris: Exceptions that you do not intend to be sub-classed should be marked final

cwms-data-api/src/main/java/cwms/cda/api/errors/ExclusiveFieldsException.java
```diff
@@ -1,27 +1,35 @@
...
+        details.put(DETAIL_KEY, String.join(", ", fields));
     }
 
     @Override
-    public Map<String, ? extends List<String>> getDetails() {
+    public Map<String, Serializable> getDetails() {
+        return details;
+    }
+
+    private static Map<String, Serializable> createDetails(Set<String> fields) {
```
rma-psmorris: This is repeated code and could be refactored into FieldException.

cwms-data-api/src/main/java/cwms/cda/api/errors/NotFoundException.java
```diff
@@ -1,21 +1,25 @@
 package cwms.cda.api.errors;
 
-public class NotFoundException extends RuntimeException {
+import java.util.HashMap;
+import javax.servlet.http.HttpServletResponse;
+
+public class NotFoundException extends ApplicationException {
+    private static final String DATABASE = "Database";
```
rma-psmorris: Use of constant here, where as in other classes it is inline "Database". Could be refactored to all use one constant.

cwms-data-api/src/main/java/cwms/cda/api/errors/RateException.java
```diff
@@ -25,10 +25,13 @@
...
+import java.util.HashMap;
+import javax.servlet.http.HttpServletResponse;
 
-public final class RateException extends RuntimeException {
+public final class RateException extends ApplicationException {
 
     public RateException(String message, SQLException cause) {
-        super(message, cause);
+        super(message, "Database", "Error performing rate function: " + message,
+            HttpServletResponse.SC_BAD_REQUEST, new HashMap<>(), cause);
```
rma-psmorris: Is bad request appropriate here? This a returned error from the database rate function. Is 500, SC_INTERNAL_SERVER_ERROR, more correct here?

cwms-data-api/src/main/java/cwms/cda/api/errors/RequiredFieldException.java
```diff
@@ -9,30 +11,39 @@
...
+        details.put(MISSING_FIELDS, fieldString);
     }
 
     @Override
-    public Map<String, ? extends List<String>> getDetails() {
+    public Map<String, Serializable> getDetails() {
+        return details;
+    }
+
+    private static Map<String, Serializable> createDetails(Set<String> fields) {
```
rma-psmorris: see note above on refactor

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
```diff
@@ -26,45 +26,58 @@
...
-public final class ValueTooLongException extends RuntimeException {
-	private String parameter;
-	private int length;
-	private int maxLength;
-	private boolean suppressIncidentId = true;
+import java.util.HashMap;
+import javax.servlet.http.HttpServletResponse;
 
-	private static final String DEFAULT_ERROR = "One or more provided values exceeds the maximum length for the parameter.";
+public final class ValueTooLongException extends ApplicationException {
```
rma-psmorris: Suggest renaming this class to: FieldLengthExceededException and carry these terms into this class where appropriate.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -393,7 +393,7 @@ static InvalidItemException buildInvalidTSIDIntervalException(RuntimeException i
         if (localizedMessage != null) {
             String[] parts = localizedMessage.split("\n");
             if (parts.length > 2) {
-                return new InvalidItemException(String.format("Invalid Time Series Description: %s is not a valid interval",
+                return new InvalidItemException(String.format("Invalid Time Series Description: %s",
```
rma-psmorris: I would like to have this be written as: "Invalid time series description" but the error was already coded with the caps...

cwms-data-api/src/main/java/cwms/cda/formatters/FormattingException.java
```diff
@@ -1,12 +1,21 @@
 package cwms.cda.formatters;
 
-public class FormattingException extends RuntimeException{
+import cwms.cda.api.errors.ApplicationException;
+import io.jsonwebtoken.io.IOException;
```
rma-psmorris: Is this the correct import?

cwms-data-api/src/main/java/cwms/cda/security/CwmsAuthException.java
```diff
@@ -1,41 +1,54 @@
...
-    public boolean suppressMessage() {
-        return this.suppress;
+    private static String getAuthErrorMessageFromCodeAndMessage(String msg, int code) {
+        switch (code) {
+            case HttpServletResponse.SC_UNAUTHORIZED:
+                return msg.isBlank() ? INVALID_USER : msg;
+            case HttpServletResponse.SC_FORBIDDEN:
+                return NOT_AUTHORIZED;
+            default:
+                return "Unknown auth error";
```
rma-psmorris: Auth should be fully written out as authorization to be clear that we are talking about authorization and not authentication.

rma-psmorris (changes requested):
Reviewed and changes requested. There are some questions too.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
zack-rma: I've added a log level field to the ApplicationException class to allow individual exceptions to retain their own logging levels. I've added it as an Optional value, so logging will only occur when it is present.

cwms-data-api/src/main/java/cwms/cda/api/errors/ApplicationException.java
zack-rma: Updated scope to be protected

cwms-data-api/src/main/java/cwms/cda/api/errors/CdaError.java
zack-rma: I've added a default value of "Unknown" and made source final.

cwms-data-api/src/main/java/cwms/cda/api/errors/CdaError.java
zack-rma: Moved unmodifiable to constructor

cwms-data-api/src/main/java/cwms/cda/api/errors/CdaError.java
zack-rma: There was an issue with the cause message also containing a ".", resulting in double periods at the end of the message. I've updated other errors to match and added a check for messages already ending with "."

cwms-data-api/src/main/java/cwms/cda/api/errors/DeleteConflictException.java
zack-rma: Revised as suggested.

cwms-data-api/src/main/java/cwms/cda/api/errors/ExclusiveFieldsException.java
zack-rma: Revised message as suggested.

cwms-data-api/src/main/java/cwms/cda/api/errors/ExclusiveFieldsException.java
zack-rma: Marked final

cwms-data-api/src/main/java/cwms/cda/api/errors/ExclusiveFieldsException.java
zack-rma: Refactored as suggested

cwms-data-api/src/main/java/cwms/cda/api/errors/NotFoundException.java
zack-rma: Moved constant into ApplicationException class for reuse

cwms-data-api/src/main/java/cwms/cda/api/errors/RateException.java
zack-rma: I had reused the existing error code, but internal server error would be more accurate. Updated as suggested.

cwms-data-api/src/main/java/cwms/cda/api/errors/RequiredFieldException.java
zack-rma: Refactored

cwms-data-api/src/main/java/cwms/cda/api/errors/ValueTooLongException.java
zack-rma: Renamed as suggested

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
zack-rma: I've updated the error message and the appropriate tests

cwms-data-api/src/main/java/cwms/cda/formatters/FormattingException.java
zack-rma: No, replaced with `java.io.IOException`

cwms-data-api/src/main/java/cwms/cda/security/CwmsAuthException.java
zack-rma: Fixed

cwms-data-api/src/main/java/cwms/cda/api/errors/CdaError.java
zack-rma: Updated

zack-rma (commented):
Updated classes as suggested and added improved log level handling

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
MikeNeilson: @krowvin @jbkolze thoughts on the capitalization there?

I'm with Peter on this one in that I think he's technically correct, grammar wise, but curious what other thinks.

cwms-data-api/src/main/java/cwms/cda/api/errors/AlreadyExists.java
```diff
@@ -1,16 +1,20 @@
 package cwms.cda.api.errors;
 
 import java.util.HashMap;
+import java.util.logging.Level;
 import javax.servlet.http.HttpServletResponse;
 
 public class AlreadyExists extends ApplicationException {
     private static final String ALREADY_EXISTS = "Already exists";
+    private static final Level LOG_LEVEL = Level.INFO;
```
MikeNeilson: Lower log level. This is a common enough issue and we've already told the user, we likely *don't* need more details in the log itself.

cwms-data-api/src/main/java/cwms/cda/api/errors/DeleteConflictException.java
```diff
@@ -28,13 +28,15 @@
 import java.sql.SQLException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.logging.Level;
 import javax.servlet.http.HttpServletResponse;
 
 public final class DeleteConflictException extends ApplicationException {
+    private static final Level LOG_LEVEL = Level.INFO;
```
MikeNeilson, line 35: arguably same comment as AlreadyExists, but this is likely more reasonable.

cwms-data-api/src/main/java/cwms/cda/api/errors/NotFoundException.java
```diff
@@ -1,25 +1,30 @@
 package cwms.cda.api.errors;
 
 import java.util.HashMap;
+import java.util.logging.Level;
 import javax.servlet.http.HttpServletResponse;
 
 public class NotFoundException extends ApplicationException {
-    private static final String DATABASE = "Database";
+    private static final Level LOG_LEVEL = Level.INFO;
```
MikeNeilson: Same comment as AlreadyExists.

cwms-data-api/src/main/java/cwms/cda/api/errors/FieldException.java
```diff
@@ -3,20 +3,30 @@
 import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
+import java.util.logging.Level;
 import javax.servlet.http.HttpServletResponse;
 
 public class FieldException extends ApplicationException {
+    private static final Level LOG_LEVEL = null;
```
MikeNeilson: why null? maybe put at lowest log level?

MikeNeilson (changes requested):
Some nitpicky details, that to be fair I could be wrong about and should have discussion with others.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
jbkolze: Agreed that in a vacuum I'd go with the lower-case version. If there are existing similar cases in the API I would just use whatever is consistent.  Realistically I wouldn't bat an eye at it either way.

Big fan of `time series` either way -- trying to move toward two separate words in new development, though i know it's not the case in a lot of legacy products.

cwms-data-api/src/main/java/cwms/cda/api/errors/AlreadyExists.java
zack-rma: That's fair. I've changed it to `Level.FINE`

cwms-data-api/src/main/java/cwms/cda/api/errors/FieldException.java
zack-rma: I had it as null since this error was not previously logged. I've updated it to `Level.FINEST`

MikeNeilson (approved):

zack-rma:
Note: In a divergence from the initial design, the HTTP error codes were moved from the CdaError class to the individual exception classes, as each exception has its own thrown exception type. Since the exceptions are already being provided with the error message presented to the user, adding this data seems appropriate.

rma-psmorris (approved):
No further feedback

PR #1362, opened 2025-10-29
Add static analysis unit test for Controller classes

### LINKED ISSUE #1349: Create unit test that does static analysis of ALL Controllers and finds parameter mismatches
On the bi-weekly documentation call it was discovered that the swagger documentation for a Text-timeseries method was describing one query parameter but the code was looking for another.  That has happened on several other occasions.   While looking into that problem I manually found a couple of other places with similar bug.  Think of the parameters described in Swagger as 'Set<String> documented'.   The set 'documented' can be found at test-time by inspecting the method annotations.

The controllers pull query parameters out of the Context argument using one of several methods or helper methods in our Controllers class.   A static analysis tool like JavaParser can be used to find all the parameters our implementations pull out of the Context.  The parameters actually used is 'Set<String> used'.

The text-timeseries bug from the bi-weekly call would be found if:
documented.removeAll(used) is not empty.

A different but similar bug is when the controller uses parameters that aren't documented.   I found at least one class where that currently happens.  Those would be found if:
used.removeAll(documented) is not empty

All our Controller classes could be found by either looking at path registration in ApiServlet or by finding all implementations of CrudHandler or Handler.

Create static test that checks every Controller and verifies that the query parameters used in Swagger match the query parameters used in the implementation.

If JavaParser cannot be used it seems like we could make a Context spy and record what each Controller pulls out of Context.   This would be annoying because it would require running each Controller in a mock environment.

**Additional context**
I have tried the annotation inspection and can find 'documented".  I believe finding "used" is similarly possible but I haven't tried it yet.

RyanM-RMA:
Related to #1349, I'll be adding a unit test for static analysis of of the Controller implementations.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,10 +73,9 @@ jobs:
       - name: build and test
         id: thebuild
         run: ./gradlew :cwms-data-api:test --tests "cwms.cda.api.OpenApiDocTest" --info --init-script init.gradle
-        continue-on-error: true
       - name: Publish Test Report
         uses: mikepenz/action-junit-report@v5
-        if: success() || failure() # always run even if the previous step fails
+        if: always()
         with:
           annotate_only: true
           include_passed: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,6 +56,31 @@ jobs:
           annotate_only: true
           include_passed: true
           report_paths: '**/TEST-*.xml'
+  open-api-static-analysis:
+    name: OpenApi Static Analysis Tests
+    runs-on: ubuntu-latest
+    env:
+      OPEN_API_TEST: "true"
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v5.0.0
+      - name: setup java
+        uses: actions/setup-java@v5.0.0
+        with:
+          distribution: 'temurin'
+          java-version: 11
+          cache: 'gradle'
+      - name: build and test
+        id: thebuild
+        run: ./gradlew :cwms-data-api:test --tests "cwms.cda.api.OpenApiDocTest" --info --init-script init.gradle
+        continue-on-error: true
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v5
+        if: success() || failure() # always run even if the previous step fails
+        with:
+          annotate_only: true
+          include_passed: true
+          report_paths: '**/TEST-*.xml'
   build-docker-image:
     runs-on: ubuntu-latest
     steps:
--- a/cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
+++ b/cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
@@ -1,0 +1,57 @@
+/*
+ * MIT License
+ * Copyright (c) 2025 Hydrologic Engineering Center
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package cwms.cda.api;
+
+import helpers.OpenApiDocTestInfo;
+import helpers.OpenApiTestHelper;
+import io.javalin.apibuilder.CrudHandler;
+import io.javalin.http.Handler;
+import java.util.List;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+class OpenApiDocTest {
+
+    @MethodSource(value = "getHandlerDocInfo")
+    @ParameterizedTest
+    void test_handler_documentation(OpenApiDocTestInfo testInfo) {
+
+    }
+
+    @MethodSource(value = "getCrudHandlerDocInfo")
+    @ParameterizedTest
+    void test_crud_handler_documentation(OpenApiDocTestInfo testInfo) {
+
+    }
+
+    static Stream<OpenApiDocTestInfo> getHandlerDocInfo() {
+        List<Class<Handler>> handlers = OpenApiTestHelper.findClassesOfType(Handler.class);
+        return handlers.stream()
+                       .map(clazz -> OpenApiTestHelper.readOpenApiDocs(Handler.class, clazz));
+    }
+
+    static Stream<OpenApiDocTestInfo> getCrudHandlerDocInfo() {
+        List<Class<CrudHandler>> handlers = OpenApiTestHelper.findClassesOfType(CrudHandler.class);
+        return handlers.stream()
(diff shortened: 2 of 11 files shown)


cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
```diff
@@ -0,0 +1,486 @@
...
+            }
+        }
+
+        String extraInfo = receivedItems.stream()
+                                        .map(p -> p.getParamInfo().getName())
+                                        .collect(Collectors.joining(", "));
+        String missingInfo = missingItems.stream()
+                                         .map(OpenApiParamInfo::getName)
+                                         .collect(Collectors.joining(", "));
+        assertAll(() -> assertTrue(receivedItems.isEmpty(), "Found extra path parameters: " + extraInfo),
```
rma-rripken: I feel like this should explain what "extra" means.

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
```diff
@@ -0,0 +1,486 @@
...
+                  () -> assertTrue(missingItems.isEmpty(), "Found missing path parameters: " + missingInfo),
```
rma-rripken: Also here, what does "missing" mean?

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
```diff
@@ -0,0 +1,486 @@
...
+                  () -> assertAll(expectedParams.stream().map(expectedParam -> testParamInfo(expectedParam, verifiedUsages))));
+    }
+
+    private void testQueryParameters(List<OpenApiParamInfo> expectedQueryParameters,
+                                     Set<OpenApiParamUsageInfo> receivedQueryParameters) {
+        List<OpenApiParamUsageInfo> verifiedUsages = new ArrayList<>();
+        List<OpenApiParamInfo> expectedParams = new ArrayList<>();
+        List<OpenApiParamInfo> missingItems = new ArrayList<>();
+        List<OpenApiParamUsageInfo> receivedItems = new ArrayList<>(receivedQueryParameters);
```
rma-rripken: These List seem like they should be Set

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
```diff
@@ -0,0 +1,486 @@
...
+            // It's using a field accessor - this means it's calling Controllers.NAME for instance.
+            // This doesn't apply to when the field is statically imported though.
+            FieldAccessExpr fieldExp = arg.asFieldAccessExpr();
+            value = parseParameterName(fieldExp.getNameAsExpression());
+        } else if (arg instanceof NameExpr) {
+            // Assume it's coming from the Controllers class, as that's a normal location for constants
+            // This is effectively like...we have a name, but we're not sure where it's coming from and we're
+            // Not a string literal.
+            try {
+                Field field = Controllers.class.getField(arg.asNameExpr().getNameAsString());
```
rma-rripken: This is using Reflection to resolve the reference not JavaParser.  Thats interesting.   And its assuming the declaring class is always Controllers.   I think you could do something like this and know what the declaring class is.    Its usually going to be Controllers or the current class but it would be better to know than to guess.  

} else if (arg instanceof NameExpr) {
            NameExpr ne = (NameExpr) arg;
            ResolvedValueDeclaration resolved = ne.resolve();
            String declaringClass = resolved.asField().declaringType().getQualifiedName();
            String fieldName = arg.asNameExpr().getNameAsString();

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
```diff
@@ -0,0 +1,486 @@
...
+                                        .stream()
+                                        .anyMatch(throwStmt -> {
+                                            if (throwStmt.getExpression() instanceof ObjectCreationExpr) {
+                                                ObjectCreationExpr creation = (ObjectCreationExpr) throwStmt.getExpression();
+                                                return creation.getType().getNameAsString().equals("UnsupportedOperationException");
+                                            }
+                                            return false;
+                                        });
+        return () -> assertTrue(throwsException, clazz.getSimpleName() + "::" + testInfo.getMethod().getName() + " is marked as ignored, but does not throw an UnsupportedOperationException.");
+    }
```
rma-rripken, line 142: I'm not convinced that we need to look for UnsupportedOperationException.  Or, I'm not convinced that is the only "right" way for something to be unimplemented.    There's at least one controller that does: 
```
 ctx.status(HttpCode.NOT_IMPLEMENTED).json(CdaError.notImplemented());
```
And that seems pretty clear and clean to me.  I'd be curious what other patterns are in use besides those two.

rma-rripken:
This ended up needing more code from JavaParser than I anticipated from the beginning.  I don't know JavaParser well enough to know if there are shortcuts we are missing but from what I've seen in the results it is catching the errors I was hoping we'd find.

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
MikeNeilson: One could argue, in light of that, that we should determine a single right way for us.

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
RyanM-RMA: Updated to Sets, let me know if I missed any

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
RyanM-RMA: Updated to say that it's a documented path parameter that's not used

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
RyanM-RMA: Updated it to say it's a used undocumented path parameter.

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
RyanM-RMA: I updated this to look for `ctx.status(HttpServletResponse.SC_NOT_IMPLEMENTED).json(CdaError.notImplemented())`

cwms-data-api/src/test/java/cwms/cda/api/OpenApiDocTest.java
RyanM-RMA: This was a bit of a wild ride to get right, but I got the symbol resolver stuff figured out.  It does make the tests slower, but not by a significant amount.

rma-rripken (dismissed):
We should check with Mike on whether we submit this enabled or not?  Maybe a @Tag on the test class so that its submitted but not run during the automated build?

MikeNeilson:
> We should check with Mike on whether we submit this enabled or not? Maybe a @tag on the test class so that its submitted but not run during the automated build?

Gentleman, I shall scream it again... `EnableIfProperty`! and then add another job to the build so we see the results but it doesn't block the builds yet.

RyanM-RMA:
> > We should check with Mike on whether we submit this enabled or not? Maybe a @tag on the test class so that its submitted but not run during the automated build?
> 
> Gentleman, I shall scream it again... `EnableIfProperty`! and then add another job to the build so we see the results but it doesn't block the builds yet.

Oooh, I don't believe I have used this, thank you for the quick feedback @MikeNeilson.  Give me a few minutes, and I'll get that going.

MikeNeilson:
> > > We should check with Mike on whether we submit this enabled or not? Maybe a @tag on the test class so that its submitted but not run during the automated build?
> > 
> > 
> > Gentleman, I shall scream it again... `EnableIfProperty`! and then add another job to the build so we see the results but it doesn't block the builds yet.
> 
> Oooh, I don't believe I have used this, thank you for the quick feedback @MikeNeilson. Give me a few minutes, and I'll get that going.

FYI did the name for memory, it might be slightly different than that exact text.

.github/workflows/build.yml
```diff
@@ -56,6 +56,31 @@ jobs:
...
+      - name: setup java
+        uses: actions/setup-java@v5.0.0
+        with:
+          distribution: 'temurin'
+          java-version: 11
+          cache: 'gradle'
+      - name: build and test
+        id: thebuild
+        run: ./gradlew :cwms-data-api:test --tests "cwms.cda.api.OpenApiDocTest" --info --init-script init.gradle
+        continue-on-error: true
```
MikeNeilson: remove this and set the below to "always"

We've tweaked the repo to have specific "required builds" this one for now is informational, but we still want it to show up as "failed" to draw eyes.

MikeNeilson (changes requested):

.github/workflows/build.yml
RyanM-RMA: I think I updated this right, let me know.

MikeNeilson: Yep, though looks like you need to update the branch, below is still showing the update button.

MikeNeilson (approved):

PR #1474, opened 2025-11-17
Test updates for latest schema and correct release schema image.

MikeNeilson:
NOTE: bit hot mess still. Additional cleanup to happen with #1480 

Update and error messaging to handle to output from https://github.com/HydrologicEngineeringCenter/cwms-database/pull/72 for LRTS time series validation.

Additionally had to update testcontainers as I had updated docker and only the latest version of test containers properly accounted for some altered behavior. And then discovered some incorrectly imported dependencies and corrected that.

Modifications were made to how the Schema version was acquired. There is some minor duplication of schema versions. Will update in further PR. This has waited long enough and is causing confusion on other work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
           - env: latest
             image: "ghcr.io/hydrologicengineeringcenter/cwms-database/cwms/database-ready-ora-23.5:latest-dev"
           - env: release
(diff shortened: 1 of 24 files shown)


MikeNeilson:
It's almost there. Unfortunately the ready database image doesn't have the version in the database table set correctly... which I thought I had fixed.

adamkorynta (approved):

Issue #991, opened 2024-12-20
CDA doesn't run against a readonly database anymore


DanielTOsborne:
Older versions of CDA worked even against an Oracle database set to readonly (COOP standby).
Now it doesn't work, and just fills up catalina log with tons of stacktraces (8 GB/day with current schedule).

If I'm not writing to the database, I would expect it to work even against a readonly database, since the request doesn't need to login.
I have this configured to always perform report generation, even on a standby COOP server, so the reports are continuous even in the event of a failover situation (also ensures COOP process is working and data matches with primary).

```
20-Dec-2024 00:00:00.007 WARNING [https-openssl-nio-8243-exec-7] cwms.cda.ApiServlet.lambda$init$17 error on request[6108007947412956016]: /spk-data/timeseries
        org.jooq.exception.DataAccessException: Unable to set user session.  user empty = false
                at cwms.cda.datasource.DirectUserPreparer.prepare(DirectUserPreparer.java:24)
                at cwms.cda.datasource.DelegatingConnectionPreparer.prepare(DelegatingConnectionPreparer.java:42)
                at cwms.cda.datasource.ConnectionPreparingDataSource.getConnection(ConnectionPreparingDataSource.java:25)
                at cwms.cda.datasource.ConnectionPreparingDataSource.getConnection(ConnectionPreparingDataSource.java:22)
                at org.jooq.impl.DataSourceConnectionProvider.acquire(DataSourceConnectionProvider.java:87)
                at org.jooq.impl.DefaultDSLContext.connectionResult(DefaultDSLContext.java:652)
                at cwms.cda.data.dao.JooqDao.connectionResult(JooqDao.java:577)
                at cwms.cda.data.dao.TimeSeriesDaoImpl.isVersioned(TimeSeriesDaoImpl.java:449)
                at cwms.cda.data.dao.TimeSeriesDaoImpl.getVersionType(TimeSeriesDaoImpl.java:429)
                at cwms.cda.data.dao.TimeSeriesDaoImpl.getTimeseries(TimeSeriesDaoImpl.java:353)
                at cwms.cda.api.TimeSeriesController.getAll(TimeSeriesController.java:465)
                at io.javalin.apibuilder.CrudFunction$1.invoke$lambda-0(CrudHandler.kt:30)
                at io.javalin.apibuilder.CrudFunctionHandler.handle(CrudHandler.kt)
                at cwms.cda.security.GuestAccessManager.manage(GuestAccessManager.java:28)
[...]

PR #1477, opened 2025-11-19
Feature/1392 use flogger

### LINKED ISSUE #1392: Choose logging framework
I think we can all agree the the usage of "java.util.logging"... exists.

We should standardize on slf4j or flogger. For CDA I (@MikeNeilson have no preference, though I'm more experienced with slf4j)

J.U.L is sufficient for the backend for now, but using either of those two facades does give us more options in the future.

Personally I also like the limit of debug/trace vs fine, finer, finest. How often have we actually wanted something in between the basic or everything?

rma-rripken:
closes #1392


cwms-data-api/src/main/java/cwms/cda/datasource/DirectUserPreparer.java
```diff
@@ -14,15 +18,23 @@ public DirectUserPreparer(String user) {
...
     public Connection prepare(Connection conn) {
-        if (user != null) {
-            String sql = "begin cwms_env.set_session_user_direct(upper(?)); end;";
-            try (PreparedStatement setApiUser = conn.prepareStatement(sql)) {
-                setApiUser.setString(1,user);
-                setApiUser.execute();                
-            } catch (Exception e) {
-                throw new DataAccessException("Unable to set user session.  "
-                        + "user empty = " + user.isEmpty(), e);
+        try {
```
MikeNeilson, line 21: well, it might work. presumably if the database is in read only mode the oracle jdbc driver would actually respond as such.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/datasource/DirectUserPreparer.java
rma-rripken: I was hoping this would fix the issue reported in https://github.com/USACE/cwms-data-api/issues/991 
I didn't explicitly test that case but it seemed like in #991 the connection was read-only and the user was set.    Should I take this out and just get the flogger changes merged first?  The only reason I include it is b/c flogger gives us the option to easily do .atMostEvery(...) and (hopefully) not fill up their logs.   8Gb/day is kinda impressive.

MikeNeilson: It's certainly worth trying, and definitely worth the atMost logging filter. Though I suspect most of the issue is the call within the JdbcInterceptor which will call set_session_direct on every connection retrieve so that it always starts in a valid state. That database call has a logging call to a table that *should* be independent and writable even in the standby database, but I'm not sure we can actually configure that.

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
```diff
@@ -72,7 +78,7 @@ public Optional<User> getByUniqueName(String uniqueName, String cac_role) {
                             logger.atInfo().log("Building user object.");
                            return new User(userName, principalName, email, cac_role != null,  roles);
                         } else {
-                            return (User)null;
+                            return null;
```
MikeNeilson, line 81: Thank you, not sure why that didn't click in my head.

MikeNeilson (approved):

PR #1481, opened 2025-11-20
CDA-66: Updated TS identifier descriptor paging

### LINKED ISSUE #1472: timeseries-descriptor paging info missing total and not handling empty office
**Describe the bug**
In the cda doc meeting it was discovered that the ts-descriptor end-point is not returning the total in the paging information but total field is shown as part of swagger response.  Also its not handling empty office parameter.  Also the documentation is inconsistent, its sometimes called timeseries identifier but the url is timeseries-identifier-descriptor 

**Expected behavior**
expect total in the paging results and expect office-id to either accept null or be marked as required.  probably prefer null

2025-09-02

zack-rma:
Now includes total number of results in response.

Contributes to resolve #1472 

More work to come for addition of alias inclusion query parameter to filter out aliased items.


cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
```diff
@@ -116,11 +110,34 @@ public Collection<TimeSeriesIdentifierDescriptor> getTimeSeriesIdentifiers(Strin
...
+        if (!retval.isEmpty() && total == null) {
+            var totalCount = dsl.selectCount().from(dsl
+                .selectDistinct(AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.DB_OFFICE_ID,
+                    AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.CWMS_TS_ID,
+                    AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.INTERVAL_UTC_OFFSET,
+                    AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.TS_ACTIVE_FLAG,
+                    AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.TIME_ZONE_ID)
+                .from(AV_CWMS_TS_ID2.AV_CWMS_TS_ID2)
+                .where(whereCondition)
+                .orderBy(AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.DB_OFFICE_ID, AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.CWMS_TS_ID,
```
rma-rripken: If we're just getting the total then there shouldn't be an offset and we shouldn't have to do the orderBy bit -right?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
```diff
@@ -116,11 +110,34 @@ public Collection<TimeSeriesIdentifierDescriptor> getTimeSeriesIdentifiers(Strin
...
+                    AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.INTERVAL_UTC_OFFSET,
+                    AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.TS_ACTIVE_FLAG,
+                    AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.TIME_ZONE_ID)
+                .offset(offset)).fetch();
+            total = totalCount.get(0).value1();
```
rma-rripken: Instead of doing a fetch() and then get(0).value1() you could just use fetchOne(0, Integer.class)

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
```diff
@@ -116,11 +110,34 @@ public Collection<TimeSeriesIdentifierDescriptor> getTimeSeriesIdentifiers(Strin
                         AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.TS_ACTIVE_FLAG,
                         AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.TIME_ZONE_ID)
                 .limit(pageSize)
-                .offset(firstRow)
+                .offset(offset)
                 .stream()
                 .map(this::toDescriptor)
                 .filter(Objects::nonNull)
```
rma-rripken: Its not in your changes but this filter(Objects::nonNull) seems weird - how can rows be mapped thru ::toDescriptor and come out null?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
```diff
@@ -24,6 +24,8 @@
 
 package cwms.cda.data.dao;
 
+import static org.jooq.impl.DSL.noCondition;
```
rma-rripken: I really love noCondition.  Is such a cool concept.

rma-rripken (changes requested):
Couple changes to the total query are needed

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
zack-rma: Correct, removed.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
zack-rma: Updated

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
zack-rma: Shouldn't be the case, removed.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
```diff
@@ -86,52 +91,96 @@ public TimeSeriesIdentifierDescriptors getTimeSeriesIdentifiers(String cursor, i
...
-                .offset(firstRow)
+                .offset(offset)
                 .stream()
-                .map(this::toDescriptor)
-                .filter(Objects::nonNull)
+                .map(this::toDescriptorWithAliases)
                 .collect(Collectors.toList());
+        }
+
+        if (!retval.isEmpty() && total == null) {
```
rma-rripken, line 155: The query for the total doesn't depend on whether or not we are returning aliases?  That doesn't feel right.  I saw that you had tests with and without aliases.  Can you explain how the total calc works?

rma-rripken: oh, its in the where condition

rma-rripken (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
zack-rma: The current total implementation will report the same number of results regardless of aliases, since the aliases are now included in the structure of the individual records returned. The where clause is doing that filtering on the `ALIASED ITEM` field.

For paging purposes, the total refers to the number of structured records returned, not the number of total db records, which would include the number of aliases. This allows for consistent paging and identifies the unique records, whether they have aliases or not.

If the aliases were still included as unique returned identifier records (as they were before this change), including them in the total would be the logical choice.

PR #1483, opened 2025-11-24
Enhancements/blob clob query

DanielTOsborne:
Add blob-id or clob-id query parameter, respectively, to all required methods, so ones with invalid path characters can be referenced.

This is required for https://github.com/HydrologicEngineeringCenter/cwms-cli/issues/72


cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
```diff
@@ -130,21 +130,33 @@ public void getAll(@NotNull Context ctx) {
             description = "Returns the binary value of the requested blob as a seekable stream with the "
                     + "appropriate media type.",
             queryParams = {
-                    @OpenApiParam(name = OFFICE, description = "Specifies the owning office."),
+                @OpenApiParam(name = OFFICE, description = "Specifies the owning office."),
+                @OpenApiParam(name = BLOB_ID, description = "If this _query_ parameter is provided the id _path_ parameter "
+                    + "is ignored and the value of the query parameter is used.   "
+                    + "Note: this query parameter is necessary for id's that contain '/' or other special "
+                    + "characters.  Because of abuse even properly escaped '/' in url paths are blocked.  "
```
MikeNeilson: It's not because of abuse. It's because of how Javalin handles the path parameter. In theory it's *supposed* to handle it, but it doesn't so do correctly.

Another approach I was thinking of taking (which I think would be complimentary to this one) was to base64 encode the id for the path parameter. Not easy for the user to do, but not difficult for cwms-python/cwms-cli.

MikeNeilson (changes requested):
The change itself is good. We'd already done similar for a portion of the clob end point.

However, the explanation text is incorrect. 

The inability to use `/` in the path parameter is due to limitations in pattern matching.

cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
DanielTOsborne: So, I didn't originally write that, I just copied what was already there: https://github.com/USACE/cwms-data-api/blob/develop/cwms-data-api/src/main/java/cwms/cda/api/ClobController.java#L159

DanielTOsborne:
> The change itself is good. We'd already done similar for a portion of the clob end point.
> 
> However, the explanation text is incorrect.
> 
> The inability to use `/` in the path parameter is due to limitations in pattern matching.

So, I didn't originally write that, I just copied what was already there: https://github.com/USACE/cwms-data-api/blob/develop/cwms-data-api/src/main/java/cwms/cda/api/ClobController.java#L159

MikeNeilson:
> > The change itself is good. We'd already done similar for a portion of the clob end point.
> > However, the explanation text is incorrect.
> > The inability to use `/` in the path parameter is due to limitations in pattern matching.
> 
> So, I didn't originally write that, I just copied what was already there: https://github.com/USACE/cwms-data-api/blob/develop/cwms-data-api/src/main/java/cwms/cda/api/ClobController.java#L159

well, now, that's quite a reasonable thing to do then isn't it. I'll go suggest updated text inline to accept.

cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
```diff
@@ -130,21 +130,33 @@ public void getAll(@NotNull Context ctx) {
...
+                    + "When using this query parameter a valid path parameter must still be provided for the request"
+                    + " to be properly routed.  If your blob id contains '/' you can't specify the blob-id query "
+                    + "parameter and also specify the id path parameter because firewall and/or server rules will "
+                    + "deny the request even though you are specifying this override. \"ignored\" is suggested."),
```
MikeNeilson:
```suggestion
                    + "Note: this query parameter is necessary for id's that contain '/' or other special "
                    + "characters. This is due to limitations in path pattern matching. "
                    + "We will likely add support for encoding the ID in the path in the future. For now use the id field for those IDs. "
                    + "Client libraries should detect slashes and choose the appropriate field." ),
```

MikeNeilson (changes requested):
Suggested text provided.

MikeNeilson (approved):

PR #1495, opened 2025-12-03
Issue/forecast spec update failure

AndreasChristmann:
added check to show that updating a ForecastSpec fails


MikeNeilson:
Your title says "add test for X" and you have changes to a dozen files that have nothing to do with "X".

It appears you need to update your local branch. Please cleanup the PR to have the correct focus.

AndreasChristmann:
> Your title says "add test for X" and you have changes to a dozen files that have nothing to do with "X".
> 
> It appears you need to update your local branch. Please cleanup the PR to have the correct focus.

Sorry, I abandoned this PR. It was meant to show that saving a ForecastSpec does not update the time series associated with it. But in an earlier conversation with you, you pointed out that it is the expected behavior that saving a ForecastSpec does not remove time series from its list but only adds to it.
For this reason, I'll be closing this pull request.

PR #1497, opened 2025-12-04
Entity endpoint Controller and Integration test

rma-kayla:
Implemented new Entity controller that exposes all CRUD methods. 
Created accompanying integration tests to ensure the endpoints work correctly and return the appropriate error responses.


rma-kayla:
@rma-bryson this PR is ready for your review

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
```diff
@@ -0,0 +1,253 @@
...
+            String formatHeader = ctx.req.getContentType();
+            ContentType contentType = Formats.parseHeader(formatHeader, Entity.class);
+            Entity entity = Formats.parseContent(contentType, ctx.bodyAsInputStream(), Entity.class);
+            if (entity.getId() == null || entity.getId().getOfficeId() == null || entity.getId().getName() == null) {
+                ctx.status(HttpServletResponse.SC_BAD_REQUEST);
+                ctx.result("Entity ID and Office ID must be provided in the request body.");
+                return;
+            }
+            EntityDao dao = new EntityDao(dsl);
+            // verify entity exists before updating
```
rma-bryson: I should have done this check in the dao. Can you move the check into the `updateEntity` method? You shouldn't need to catch any exception, it will throw a NotFoundException automatically and should then return the appropriate status - should verify this with integration test (see other comment).

cwms-data-api/src/test/java/cwms/cda/api/EntityControllerTestIT.java
```diff
@@ -0,0 +1,294 @@
...
+        .when()
+            .redirects().follow(true)
+            .redirects().max(3)
+            .patch("/entity/" + ENTITY_ID)
+        .then()
+            .log().ifValidationFails(LogDetail.ALL,true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_OK));
+
+        // GET to confirm updated field persisted
```
rma-bryson, line 115: Add an update-with-non-existing-entity-id check here with a bogus entity id to make sure its returning the not-found status.

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
```diff
@@ -0,0 +1,253 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.util.List;
+import java.util.logging.Logger;
+
+import static com.codahale.metrics.MetricRegistry.name;
+import static cwms.cda.api.Controllers.*;
+import static cwms.cda.data.dao.JooqDao.getDslContext;
+
+public class EntityController implements CrudHandler {
+    public static final String TAG = "Entity";
```
rma-bryson: the TAG and LOGGER constants can be private

cwms-data-api/src/test/java/cwms/cda/api/EntityControllerTestIT.java
```diff
@@ -0,0 +1,294 @@
...
+            .log().ifValidationFails(LogDetail.ALL, true)
+        .assertThat()
+            .statusCode(is(HttpServletResponse.SC_OK))
+            .body("entities", notNullValue())
+            .body("entities.size()", greaterThanOrEqualTo(0));
+    }
+
+    // Show simple filtering works with getAll and parent entity id only
+    @Test
+    void getAll_with_parent_filter_returns_200_empty_list_ok() {
```
rma-bryson, line 286: Add a test for the match null parents flag. 
1. Create a entity with a null parent
2. GET request with flag set to true -> should get it back
3. GET request with flag set to false -> shouldn't get it back. 
4. delete entity

rma-bryson (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
rma-kayla: made Tag private, and removed unused logger

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
rma-kayla: Moved the check if entity exists before updating into the dao

cwms-data-api/src/test/java/cwms/cda/api/EntityControllerTestIT.java
rma-kayla: I have a test for this outside of this test.  I'll rename it to make it more explicit.
// Entity ID in the URL must match the id.name in the request body
    @Test
    void update_mismatched_entity_id_bad_request()

cwms-data-api/src/test/java/cwms/cda/api/EntityControllerTestIT.java
rma-kayla: created test to verify match-null-parents functionality

rma-bryson (approved):

PR #1498, opened 2025-12-04
Fix wrong query parameter name for blob.

DanielTOsborne:
Part of #1483.
Forgot to fix the query name. I think I considered making the parameter the same (just 'id') across both blob and clob, but changed my mind and forgot to adjust blob.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,12 @@ jobs:
     outputs:
       thewar: ${{steps.thebuild.outputs.WARFILE}}
     steps:
+      - name: Clean up disk space, so we don't run out.
+        if: matrix.platform == 'ubuntu-latest'
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
       - name: checkout code
         uses: actions/checkout@v5.0.0
       - name: setup java
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
         jdk: [11]
         schema:
           - env: latest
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -16,14 +16,20 @@ jobs:
...
+            image: "ghcr.io/hydrologicengineeringcenter/cwms-database/cwms/database-ready-ora-23.5:main"
           - env: release
             image: "ghcr.io/hydrologicengineeringcenter/cwms-database/cwms/database-ready-ora-23.5:25.07.01"
     name: build and test (jdk ${{matrix.jdk}}, schema ${{matrix.schema.env}})
     runs-on: ubuntu-latest
     outputs:
       thewar: ${{steps.thebuild.outputs.WARFILE}}
     steps:
+      - name: Clean up disk space, so we don't run out.
+        if: matrix.platform == 'ubuntu-latest'
```
MikeNeilson:
```suggestion
```

I just realized this block probably isn.t even running since the platform isn't in the matrix. On my fork I got "disk out of space" and after removing that tests are actually running.

MikeNeilson:
@DanielTOsborne the correct fix to the build got merged in, you'll need to rebase this.

.github/workflows/build.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         jdk: [11]
         schema:
           - env: latest
-            image: "ghcr.io/hydrologicengineeringcenter/cwms-database/cwms/database-ready-ora-23.5:latest-dev"
+            image: "ghcr.io/hydrologicengineeringcenter/cwms-database/cwms/database-ready-ora-23.5:main"
```
MikeNeilson: revert this back to latest-dev.

MikeNeilson:
test failure is due to database change and with an endpoint outside the scope of this PR.

MikeNeilson (approved):

PR #1507, opened 2025-12-12
Clean runner before any steps

MikeNeilson:

PR #1509, opened 2025-12-13
Fix for Entity Controller setup causing build fails.

rma-kayla:
Refactored and cleaned Entity Controller to address clean runner build fail PR #1507.

- Fixed entity controller integration tests to incorporate unique data per test 
- Added test utility helpers for cleanup and unique entityJson test values
- Cleaned up global variables to remove hard coded param values
- Fixed teardown method to ensure all test entities cleaned up after testing.
- Added check in Entity Controller to fix unused documented query param errors and check for required params

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
@@ -55,7 +55,7 @@ public class EntityController implements CrudHandler {
             },
             responses = {
                     @OpenApiResponse(status = STATUS_200, content = {
-                            @OpenApiContent(isArray = true, from = Entity.class, type = Formats.JSONV2)
+                            @OpenApiContent(isArray = true, from = Entity.class, type = Formats.JSONV1)
                     })
             },
             tags = {TAG}
@@ -106,7 +106,7 @@ public class EntityController implements CrudHandler {
             },
             responses = {
                     @OpenApiResponse(status = STATUS_200, content = {
-                                    @OpenApiContent(from = Entity.class, type = Formats.JSONV2)
+                                    @OpenApiContent(from = Entity.class, type = Formats.JSONV1)
                     })
             },
             tags = {TAG}
@@ -139,7 +139,7 @@ public class EntityController implements CrudHandler {
             description = "Create CWMS Entity",
             requestBody = @OpenApiRequestBody(
                     content = {
-                            @OpenApiContent(from = Entity.class, type = Formats.JSONV2)
+                            @OpenApiContent(from = Entity.class, type = Formats.JSONV1)
                     },
                     required = true),
             responses = {
@@ -167,7 +167,7 @@ public class EntityController implements CrudHandler {
     @OpenApi(
             description = "Update an existing Entity.",
             requestBody = @OpenApiRequestBody(
-                    content = {@OpenApiContent(from = Entity.class, type = Formats.JSONV2)},
+                    content = {@OpenApiContent(from = Entity.class, type = Formats.JSONV1)},
                     required = true),
             pathParams = {
                     @OpenApiParam(name = ENTITY_ID, required = true, description = "Specifies the entity ID of the " +
--- a/cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
@@ -167,10 +167,6 @@ public class EntityController implements CrudHandler {
                     @OpenApiParam(name = ENTITY_ID, required = true, description = "Specifies the entity ID of the " +
                             " Entity to be updated. (e.g., NWS)")
             },
-            queryParams = {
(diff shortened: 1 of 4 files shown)


rma-kayla:
@rma-bryson and/or @adamkorynta this PR is ready for a review
@MikeNeilson this should fix the build fails from PR #1507

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
```diff
@@ -190,13 +191,27 @@ public void update(@NotNull Context ctx, @NotNull String entityId) {
...
+
+            if (!entityId.equalsIgnoreCase(entity.getId().getName())) {
+                ctx.status(HttpServletResponse.SC_NOT_FOUND);
+                ctx.result("Entity not found for the given entity-id.");
+                return;
+            }
+
+            if (!officeId.equalsIgnoreCase(entity.getId().getOfficeId())) {
+                ctx.status(HttpServletResponse.SC_BAD_REQUEST);
+                ctx.result("Office ID in query parameter must match the Office ID in the request body.");
```
MikeNeilson: don't we have "global" entities under the cwms office?

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
```diff
@@ -190,13 +191,27 @@ public void update(@NotNull Context ctx, @NotNull String entityId) {
                 ctx.result("Entity ID and Office ID must be provided in the request body.");
                 return;
             }
+
+            if (!entityId.equalsIgnoreCase(entity.getId().getName())) {
+                ctx.status(HttpServletResponse.SC_NOT_FOUND);
+                ctx.result("Entity not found for the given entity-id.");
```
rma-bryson: shouldn't this be a similar message and error response code to the mismatched office-id? something like:

""Entity ID in path parameter must match the Entity ID in the request body.""

rma-bryson (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
rma-kayla: updated message to be more consistent.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
MikeNeilson: It doesn't look like the change Bryson asked for made it to github.

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
rma-kayla: After discussion with Bryson, we decided the office query parameter is not needed in update method and will be validated in the body.  I removed unnecessary office id checks in the update method.

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
rma-kayla: @MikeNeilson, it is in next commit.

rma-kayla:
Refactored entity update controller and tests.
- Removed office-id query param and redundant checks, cleaned up error message.
- Added missing entity-id validation test and invalid office authorization test
- Improved variable naming and fixed failing .replace() logic.

cwms-data-api/src/test/resources/cwms/cda/data/dto/entity_test.json
MikeNeilson, line 1: I like the naming scheme you've chosen here. It's very explicit as to what's going on.

MikeNeilson (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/EntityController.java
rma-bryson, line 1: Thinking about the match-null-parents flag in the getAll. It doesn't seem right that we would ask for entities with a specific-parent and also get back entities with null-parents. 

I'm thinking we should add some logic here that will switch match-null-parents to false IF the parent-id parameter is provided by the user

rma-bryson: @MikeNeilson - What are your thoughts on that?

MikeNeilson: That make sense to me.

rma-kayla: Added logic to getAll when parent id param is provided, match-null-parents = false.
Added IT to verify logic

rma-bryson (approved):

rma-bryson (changes requested):
The version type should be V1

rma-kayla:
Version in DTO, Controller and Integration tests are updated to JsonV2

rma-bryson (approved):

PR #1512, opened 2025-12-16
add RSS feed endpoint for querying AQ message views

adamkorynta:
Alternate considerations:

- Looked into using a third-party RSS library for building up the XML. Seemed like the only somewhat maintained library is Rome, but that hasn't been updated since 2023 and contains the jdom2 dependency which has a known vulnerability (wouldn't really matter to our use-case of producing XML only, but would matter on the client). didn't seem too much a lift to just use our own DTO's and jackson serialization. If this ends up being a burden, it also wouldn't be too hard to update to something else in the future.
- opted to not set the atom links for self or previous as those would be moving targets given the week lookback limit. We do need a next cursor so the fact that the URL would not be reproducible after a time period is unavoidable. We could also throw out the cursor and only use the `since` field.
- don't search for messages older than a week. This could be shortened as needed.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/rss/MessageDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/rss/MessageDao.java
@@ -97,7 +97,7 @@ public final class MessageDao extends JooqDao<RssFeed> {
             default:
                 description = null;
         }
-        RssChannel channel = new RssChannel(office + "_" + name, nextLink, description, items);
+        RssChannel channel = new RssChannel(name, nextLink, description, items);
         return new RssFeed(channel);
     }
 
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/rss/MessageDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/rss/MessageDao.java
@@ -1,0 +1,145 @@
+/*
+ * MIT License
+ *
+ * Copyright (c) 2025 Hydrologic Engineering Center
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package cwms.cda.data.dao.rss;
+
+import static java.util.stream.Collectors.toList;
+import static org.jooq.impl.DSL.currentTimestamp;
+import static org.jooq.impl.DSL.field;
+import static org.jooq.impl.DSL.name;
+import static org.jooq.impl.DSL.partitionBy;
+import static org.jooq.impl.DSL.rowNumber;
+import static org.jooq.impl.DSL.select;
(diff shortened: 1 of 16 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/rss/MessageDao.java
```diff
@@ -0,0 +1,145 @@
...
+    private static RssItem rssItem(Record record, String p) {
+        ZonedDateTime enqTimestamp = record.get("ENQ_TIMESTAMP", Timestamp.class)
+            .toInstant().atZone(ZoneOffset.UTC);
+        String msgId = record.get("MSG_ID", String.class);
+        return new RssItem(p, enqTimestamp, msgId);
+    }
+
+    private Result<?> retrieveMessages(int offset, int pageSize, Instant since, String office, AqTable name) {
+        Timestamp sinceTimestamp = since == null ? null : Timestamp.from(since);
+        Table<?> t = table(name("CWMS_20", "AQ$" + office + "_" + name.name() + "_TABLE")).as("t");
```
MikeNeilson, line 121: interesting trick, taking advantage of the queue persistence.

cwms-data-api/src/test/resources/cwms/cda/data/dto/rss/ts_stored.xml
```diff
@@ -0,0 +1,13 @@
+<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
+    <channel>
+        <title>SWT_TS_STORED</title>
```
MikeNeilson: Would it make more sense for SWT to be not shown here and this show up and "TS_STORED"?

I think there is a use case for "everything" and long term these wont be from AQJMS queues anyways and it would be one global queue/topic where we filter by parameter.

Then again it is under "channel". Can RSS present more than one "channel" at a time?

MikeNeilson (commented):
Overall I like it. I do want to make sure that if at all practical we're exposing more standard "queue/topic" semantics.

While this will always be read only, WM staff and core developers will be interacting with the same sources of data and we should make sure things map, preferably into the future and not the past.

Will very likely be making good use of SNS *and* SQS for queues and topics (of which this is technically the later regardless of Oracle naming).

I'm also currently inclined towards a "queue/topic per concept" vs per-office per-concept. The latter may be an impratical amount of things to deal with when the applications can easily choose to filter and let whatever broker deal.

cwms-data-api/src/main/java/cwms/cda/data/dao/rss/MessageDao.java
adamkorynta: yeah makes it so that we don't have to pull in all the aqapi and jms dependencies and keeps us from being stateful

cwms-data-api/src/test/resources/cwms/cda/data/dto/rss/ts_stored.xml
adamkorynta: Unfortunately, RSS doesn't allow for multiple channels. I was originally thinking we could do a channel per office. If we want to join AQ$SWT_STATUS, AQ$SWT_TS_STORED, and AQ$SWT_REALTIME_OPS into a single channel we could utilize the <category> elements on the items itself to distinguish types. I didn't go that route because there's no description field on categories, where channels have both title and description. 

Even with topics externalized, you'd do a single global? I'd think we would still want to separate by at least office and setup authorization per topic, rather than implementing authorization per message.

MikeNeilson: For sqs/sns the authorization is different, there isn't going to be user level authorization on the queues as we don't have a practical way to expose them with the correct semantics... well, not without unneccassary difficulty. If an user or application needs to allow a message to put entered into the queue the backend service would have the authorization, and the front end... say CDA, would take care of any authorization details.

at a minimum though a hierarchy, at least visually of say "concept/office" may be appropriate, i know for mqtt you get "and everything underneath", need to look at sqs and rabbitmq/activemq to get a better sense.

MikeNeilson (dismissed):
Your new tests seem to be breaking, but seems like a reasonable start.

cwms-data-api/src/test/java/cwms/cda/api/rss/RssHandlerIT.java
```diff
@@ -0,0 +1,205 @@
...
+                    "    DBMS_AQADM.ADD_SUBSCRIBER(" +
+                    "      queue_name => ?, " +
+                    "      subscriber => sys.aq$_agent(?, NULL, NULL)" +
+                    "    ); " +
+                    "  EXCEPTION " +
+                    "    WHEN OTHERS THEN " +
+                    "      IF SQLCODE != -24034 THEN RAISE; END IF; " + // Ignore "Already a subscriber"
+                    "  END; " +
+                    "END;",
+                "CWMS_20.SPK_STATUS",
```
MikeNeilson, line 85: Don't need to do this in this PR, but this seems like something that should perhaps just be part of the CDA startup sequence.

MikeNeilson (approved):

cwms-data-api/src/test/java/cwms/cda/api/rss/RssHandlerIT.java
adamkorynta: that's a good point. For national we'd need to subscribe to every queue, local deployments would only be for the specific office

PR #1518, opened 2025-12-22
CDA-70 - Fixes getAll not returning results with office or name provided

### LINKED ISSUE #1489: ForecastInstance getAll does not returned expected results.  Needed by data-exchange
**Describe the bug**
The ForecastInstance getAll end-point is not returning expected results. 

**Expected behavior**
I've modified the existing ForecastInstance integration tests to demonstrate that the newly created ForecastInstances are not included in the getAll output.  I expect newly created instances to be found.
 https://github.com/USACE/cwms-data-api/tree/bugfix/forecast-instance-getall 

**To Reproduce**
Run test_get_create_get.  I believe the test should pass. 

I see this in current develop but it is also happening in the data-exchange development that uses 
- Version 2025.07.21-023344

rma-bryson:
fixes https://github.com/USACE/cwms-data-api/issues/1489

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/ForecastInstanceController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/ForecastInstanceController.java
@@ -128,9 +128,9 @@ public final class ForecastInstanceController implements CrudHandler {
                 @OpenApiParam(name = NAME, description = "Specifies the "
                         + "spec id of the forecast spec whose forecast instance data is to be "
                         + "included in the response. Default will be all names."),
-                @OpenApiParam(name = DESIGNATOR, description = "Specifies the "
-                        + "designator of the forecast spec whose forecast instance data to be included "
-                        + "in the response. "),
+                @OpenApiParam(name = DESIGNATOR, description = "Specifies the designator of the forecast spec to filter the "
+                        + "forecast instance data in the response. "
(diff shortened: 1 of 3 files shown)


cwms-data-api/src/main/java/cwms/cda/api/ForecastInstanceController.java
rma-rripken, line 1: Looks like this fixes a parameter mismatch between what the openapi documented and what the code was actually using.   Awesome.   CDAC looks for "designator" so this change seems right. 

Can you update the designator description to include the sort of standard sentence about the field being a posix regexp and include the link?

rma-rripken: Wait, Is it a regular expression?

rma-bryson: No, `designator` in this query is not treated as a regular expression. It is matched exactly using = (or checked for NULL)

rma-bryson: Updated the description for designator to add more details

rma-rripken (approved):

MikeNeilson (approved):

PR #1519, opened 2025-12-23
cda-37 Store blobs in object-store

rma-rripken:


rma-rripken:
still lots to cleanup and reorganize

MikeNeilson (commented):
Seems reasonable so far.

cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
```diff
@@ -62,33 +64,58 @@ protected DSLContext getDslContext(Context ctx) {
         return JooqDao.getDslContext(ctx);
     }
 
+    private BlobAccess chooseBlobAccess(DSLContext dsl) {
+        boolean useObjectStore = isObjectStorageEnabled();
+        try {
+            // Prefer Togglz if available
```
MikeNeilson: why wouldn't it be available?

rma-rripken:
I think I've implemented all the blob parts.   Would need to follow up with a similar clob dao that handles text encoding.

cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
```diff
@@ -62,33 +64,58 @@ protected DSLContext getDslContext(Context ctx) {
...
+            // fall back to system/env property check
+        }
+        if (useObjectStore) {
+            ObjectStorageConfig cfg = ObjectStorageConfig.fromSystem();
+            return new ObjectStorageBlobDao(cfg);
+        }
+        return new BlobDao(dsl);
+    }
+
+    private boolean isObjectStorageEnabled() {
```
MikeNeilson: Shouldn't this just use the CdaFeatures thing?

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
```diff
@@ -790,7 +790,7 @@ void test_large_data_url(String format) throws Exception {
                                 .log().ifValidationFails(LogDetail.ALL, true)
                             .assertThat()
                                 .statusCode(is(HttpServletResponse.SC_OK))
-                                .header("Transfer-Encoding", equalTo("chunked"))
+//                                .header("Transfer-Encoding", equalTo("chunked"))
```
MikeNeilson: remove or uncomment

cwms-data-api/src/test/java/cwms/cda/api/BlobControllerObjectStorageTestIT.java
```diff
@@ -0,0 +1,131 @@
...
+            if (!MINIO_CONTAINER.isRunning()) {
+                MINIO_CONTAINER.start();
+                setupMinioResources();
+            }
+
+            setObjectStoreProperties();
+        }
+    }
+
+    private static void setupMinioResources() {
```
MikeNeilson: 1. This should just be in the Cda Setup extension and can just be started whenever, eventually it'll be the only option and it doesn't really get in the way just being on.
2. https://testcontainers.com/modules/minio/ (favor the testcontainer helper modules unless you have reason not to, we do our own for CWMS to make the CWMS setup easier to handle.)

MikeNeilson (changes requested):
Overall looks good. A few nitpicks and probably at least one just misunderstanding of togglz on my part I need clarified.

cwms-data-api/src/test/java/cwms/cda/api/BinaryTimeSeriesControllerTestIT.java
rma-rripken: removed

cwms-data-api/src/main/java/cwms/cda/api/BlobController.java
rma-rripken: should always be available.  I removed

cwms-data-api/src/test/java/cwms/cda/api/BlobControllerObjectStorageTestIT.java
rma-rripken: As suggested in #2 - switching to the testcontainer module makes it simpler.   

I'm not sure I understand the suggestion in #1 - What does "Cda Setup" reference?  I think you are saying to move the minio-container out of this test and put it somewhere that runs all the time. 

Would the minio-container be its own @ExtendWith callback that gets added to DataApiTestIT sort of like KeyCloakExtension ?

MikeNeilson: I was thinking here: https://github.com/USACE/cwms-data-api/blob/develop/cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java

But making it its own like the KeyCloakExtension is likely the better idea. Proper encapsulation of concerns and all.

MikeNeilson (commented):
Primary logic looks good. As silly as it may seem for one test set, the minio setup should be setup the same as the other extensions.

MikeNeilson (approved):

PR #1524, opened 2025-12-30
add persistent queue subscriber on Javalin startup

adamkorynta:
this will ensure that queued messages remain in the queue until timeout. otherwise if there are no subscribers, messages won't be enqueued.


MikeNeilson (commented):
Need to rebase/merge develop

MikeNeilson (approved):

PR #1527, opened 2026-01-06
bug fixes found testing REGI against SWT db export

adamkorynta:
- fix openapi for location kind endpoint return type
- fix issues where parameterized connection was getting the office id session set, but then a separate configuration was used for the jOOQ call (I tried to do an audit and catch as many errors as I could)
- add null checks for outlet/turbine child location retrieval objects
- fix null pointers on water contract effective/expiration dates
- null check and ignore "CWMS" office ids for session variables
- fix time series vert datum retrieval trying to convert units to "SI" or "EN"

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
@@ -109,7 +109,9 @@ public abstract class Dao<T> {
     }
 
     protected void setOffice(Connection c, String office) throws SQLException {
-        CWMS_ENV_PACKAGE.call_SET_SESSION_OFFICE_ID(DSL.using(c).configuration(), office);
+        if(office != null && !"CWMS".equals(office)) {
+            CWMS_ENV_PACKAGE.call_SET_SESSION_OFFICE_ID(DSL.using(c).configuration(), office);
+        }
     }
 
     public abstract Optional<T> getByUniqueName(String uniqueName, String office);
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
@@ -102,13 +102,15 @@ public abstract class Dao<T> {
      * Sets session office on specific connection.
      * @param c opened connection
      * @param object Data containing a valid CWMS office
-     * @throws SQLException if the underlying database throws an exception
      */
-    protected void setOffice(Connection c, CwmsDTO object) throws SQLException {
-        this.setOffice(c,object.getOfficeId());
+    protected void setOffice(Connection c, CwmsDTO object) {
+        setOffice(c,object.getOfficeId());
     }
 
-    protected void setOffice(Connection c, String office) throws SQLException {
+    protected static void setOffice(Connection c, String office) {
+        //null office id is invalid for the session office, the client may send null if looking for data across all offices
+        //CWMS is not a valid session office id and maybe be parameterized from the client when
+        //searching for CWMS-owned data.
         if(office != null && !"CWMS".equals(office)) {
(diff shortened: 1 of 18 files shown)


MikeNeilson:
Please add a brief description. It helps, at least me, with reviewing the code.

cwms-data-api/src/test/java/cwms/cda/data/dto/LocationLevelTest.java
```diff
@@ -154,9 +154,9 @@ void test_serialization_om_Virtual() throws JsonProcessingException {
 
 	@Test
 	void test_mutual_exclusivity_seasonal() {
-		assertThrows(RequiredFieldException.class, () -> new SeasonalLocationLevel.Builder("Test", ZonedDateTime.now()).build());
+		assertThrows(RequiredFieldException.class, () -> new SeasonalLocationLevel.Builder("Test", ZonedDateTime.now()).build().validate());
 		assertThrows(RequiredFieldException.class, () -> new SeasonalLocationLevel.Builder("Test", ZonedDateTime.now())
-						.withIntervalMinutes(25).withIntervalMonths(12).build());
+						.withIntervalMinutes(25).withIntervalMonths(12).build().validate());
```
MikeNeilson: We're asserting an exception is thrown and then calling build on the exception?

or is this just formatting.

if so the .withIntervalMinutes should really start right under .Builder.

or the entirely supplier brought down to the next line. 

or just ingore the line length limit here.

As-is this was confusing on initial look.

cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
```diff
@@ -109,7 +109,9 @@ protected void setOffice(Connection c, CwmsDTO object) throws SQLException {
     }
 
     protected void setOffice(Connection c, String office) throws SQLException {
-        CWMS_ENV_PACKAGE.call_SET_SESSION_OFFICE_ID(DSL.using(c).configuration(), office);
+        if(office != null && !"CWMS".equals(office)) {
```
MikeNeilson, line 114: I feel like this check is going to have a lot of random implications. Please provide a description at each if, or perhaps just add in the javadoc for it that explain why the null/CWMS check is present and what it causes. Otherwise someone is likely to just revert this again should there be an issue.

Additionally, is it even valid that null would be passed in to this method?

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/Dao.java
adamkorynta: I'm not sure what kind of implications this could have. Null is likely not passed, but I figured if I was adding the CWMS check I might as well also check null. I saw several calls where the JooqDao.getDslContext was used on endpoints that can retrieve "CWMS" owned data which would cause issues.

I didn't see this check in the history, but will update with a comment

cwms-data-api/src/test/java/cwms/cda/data/dto/LocationLevelTest.java
adamkorynta: I moved the build() out of the lambda, reads easier (at least to me) now for what's being tested

MikeNeilson (approved):

adamkorynta:
The integration test that is failing is referencing the new changes in: https://github.com/HydrologicEngineeringCenter/cwms-database/pull/93/changes#diff-11704aa41440be32c8a54f8cd2b717f5096a928f608c1242528fabf9e11ed73aR2014

I'm going to merge this PR (since it didn't change anything there) and we'll have to resolve the failure.

MikeNeilson:
Yeah, I just did the same for Daniel's, it's the change Mike P made to the database to better report the seasonable location level error. So that test *should* be failing. Just make sure it's that one that's failing before we force anything.

PR #1538, opened 2026-01-13
Update test branch

MikeNeilson:
Pull in latest changes to develop since the last time we did this.

NOTE: I'm pretty sure this is everything, but it's a lot to keep track of so is possible I've missed something.


rma-rripken (approved):
I'm not sure how I'd even verify that this was everything but it looks like a plausible list.

MikeNeilson:
> I'm not sure how I'd even verify that this was everything but it looks like a plausible list.

yeah. that is a challenge in several projects. Something I'm hoping to solve with https://github.com/opendcs/branch-issue-tracker/ ... If I can ever find time to work on it.

PR #1539, opened 2026-01-16
CDA-75 - Adds flow-unit support to CDA for pump accounting.

### LINKED ISSUE #1180: Pump Account Endpoint POST Does Not Specify Units
When storing flow to the pump accounting endpoint in CDA you must specify a flow. However, there is no key to provide the units you are storing as. 

It would appear you must provide SI (cms) units as these are what are stored in the database by default. 

Please provide the ability to specify units/unit in the API when storing flow. 

*Another option considered was to document that the required units for flow are metric in the annotation?* This is arguably the fastest way to pre-warn someone on what to store. 

Looking further into @AndySWT and I saw that the Object for the accounting record CDA is most likely using in the DTO does not offer the ability to specify units. Which lends to, where possible, being sure to annotate expected units.  

Seen here:

Image

rma-bryson:
fixes https://github.com/USACE/cwms-data-api/issues/1180

Note this allows things to work with CDA converting to SI prior to store - a more permanent fix would be in the db to include native-units to the object holding the flow

### DIFF
--- a/cda-gui/package-lock.json
+++ b/cda-gui/package-lock.json
@@ -2269,6 +2269,24 @@
         "node": ">=6"
       }
     },
+    "node_modules/color-convert": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/color-convert/-/color-convert-2.0.1.tgz",
+      "integrity": "sha512-RRECPsj7iu/xb5oKYcsFHSppFNnsj/52OVTRKb4zP5onXwVF3zVmmToNcOfGC+CRDpfK/U584fMg38ZHCaElKQ==",
+      "license": "MIT",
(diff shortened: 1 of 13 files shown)


rma-rripken (approved):

rma-rripken:
@rma-bryson  I think this change or the corresponding change in cwms-database broke the WaterSupplyAccountingControllerIT on the latest schema.  I got the test running in the debugger.  Its trying to parse "Added water to the system" as a Timestamp.   CDA doesn't seem to know that wat_usr_contract_acct_obj_t now has an additional field and so I think its off-by-one when jooq is trying to construct the oracle udt from an array of values.   Maybe something needs to happen to get latest schema re-jooq'd ?

PR #1548, opened 2026-01-26
Update tomcat.

MikeNeilson:
Update Tomcat version in test and docker. dev/test already using this version.


rma-rripken (approved):

MikeNeilson:
Update the test code to handle the required docker environment with test-containers issue we had. While most of the update was just "correct import", it exposed that one test was using the wrong Junit assertion and thus behavior was not covered. Since the intent of this PR is *just* to update tomcat and then allow a settings change I've altered the behavior of the test itself so as not to risk a cascade of changes at this time.

I may also decide that instead of a "prod" branch that whatever is correctly functional in -test is what it then prod - if that decision is made I will adjust the various branches and github actions accordingly.

rma-rripken (approved):

krowvin:
@MikeNeilson this is approved if you are ready for it! If not suggest we mark it to draft and update why we are holding off

MikeNeilson:
already updated. and we've gone with a different branching strategy.

PR #1551, opened 2026-01-27
CSV Design Documentaiton

### LINKED ISSUE #1525: Format Design Document
# Statement
Continue to allow CSV and JSON as formats going forward. CDA will officially support JSON and CSV. Additional formats may be introduced via versioned changes or explicit content negotiation if a documented use case arises.

### CSV, without a viewing software, is also highly viewable by humans compared to JSON.

CSV isolates tabular datasets and minimizes structural overhead, making it easier to inspect in spreadsheet-oriented workflows. It is easily imported into various end-user client software (MS Excel/Sheets), and does not require extra code on the client in order to convert from one format to another when JSON is the only option.

If a student, stakeholder, news outlet, other agency, and others wanted to easily manipulate data without first needing to parse it in JSON they would be able to do so with CSV. CSV, without a viewing software, is also highly viewable by humans compared to JSON. 

Example: Sharing a link to a specific dataset. This can be handled via a direct CDA link that when clicked, downloads the CSV file directly.

*Relies on*: `Content-Type: text/csv` being set by default.

### JSON
JSON is the default response format. Some legacy endpoints currently return XML due to upstream constraints; these are considered transitional and should be normalized where feasible.

It works well with downstream programming needs and offers additional metadata beyond what CSV provides. 

## Expectations
1. Query parameter (?format=) overrides Accept header to support browser downloading via URI control.
  - If neither is present, default to JSON.
2. CSV must only include the headers and data fields. No additional metadata at the top. 
  - UTF-8 Encoded
  - Empty values MUST return empty fields, NOT `null`. I.e. `date,4.0,,` NOT `date,4.0,null,null`
3. If data response types must change, such as headers or keyvalue pairs in a response, the version MUST be updated to reflect this in the `accept` header to match the practices put in place already for versioning outlined here:
  - #942 
4. Proper errors must be returned when a format is failed to parse. I.e. invalid json or CSV file that does not contain any data/fails to render CSV. 
  - For generic errors, caused by the user, make sure to return `400 Bad Request`
  - For client-side/invalid format errors, must return `406 Not Acceptable` or `415 Unsupported Media type` where `406` are media types we deem will never be implemented. 

## Related Issues

- #1257

krowvin:
Resolves 
- #1525 

See the issue above for cross issue references, those were removed for RST publication.

### DIFF
--- a/docs/source/decisions/0005-comma-separated-values.rst
+++ b/docs/source/decisions/0005-comma-separated-values.rst
@@ -1,0 +1,70 @@
+Statement
+=========
+
+Continue to allow Comma Separated Values (CSV) and JavaScript Object Notation (JSON) as formats going forward. CWMS Data API (CDA) will officially
+support JSON and CSV. Additional formats may be introduced via versioned
+changes or explicit content negotiation if a documented use case arises.
+
+Purpose / Need
+==============
+
+CSV
+---
+
+CSV, without a viewing software, is also highly viewable by humans compared to
+JSON.
+
+CSV isolates tabular datasets and minimizes structural overhead, making it
+easier to inspect in spreadsheet-oriented workflows. It is easily imported
+into various end-user client software (MS Excel/Sheets), and does not require
+extra code on the client in order to convert from one format to another when
+JSON is the only option.
+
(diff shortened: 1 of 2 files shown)


MikeNeilson:
superseded by #1634

PR #1555, opened 2026-01-30
Bugfix/1531 rating spec

### LINKED ISSUE #1531: RatingSpec formats inconsistent
During the documentation meeting Charles attempted to read/post/request RatingSpec.     For some reason some of the RatingSpec end-points deal in JSON and others in XML.  

We should support XML and JSON

Or - pick one and only support XML or only JSON.

rma-rripken:
Fixes #1531


MikeNeilson (commented):
marking place for future review. Seems reasonable so far.

cwms-data-api/src/main/java/cwms/cda/api/rating/RatingSpecController.java
```diff
@@ -201,58 +205,50 @@ protected RatingSpecDao getRatingSpecDao(DSLContext dsl) {
...
-            retval = translateJsonToXml(body);
-        } else {
-            throw new IllegalArgumentException("Unexpected contentType format:" + contentType);
+            try {
+                RatingSpec spec = Formats.parseContent(contentType, body, RatingSpec.class);
+                // If we can parse it into our CDA RatingSpec object have the DAO use it.
+                dao.create(spec, failIfExists);
+                ctx.status(HttpServletResponse.SC_CREATED);
+            } catch (FormattingException e) {
+                if (contentType.getType().contains(Formats.XML)) {
```
rma-bryson, line 240: should this be XMLV2?

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecXmlUtils.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @JsonIgnore
+    abstract List<ZonedDateTime> getEffectiveDates();
+}
+
+
+class RatingSpecXmlUtils {
+    private static final XmlMapper mapper = buildMapper();
+
+    private static XmlMapper buildMapper() {
```
rma-bryson, line 64: Alternatively you could extend XMLv2 into a custom RatingSpecXml class and override the buildXmlMapper() method if you wanted to keep the exception handling logic by calling format(ratingspec). Is the `.withRootName("rating-spec")` necessary in the `toPlSqlXml` method? I would have thought the root name annotation on the spec itself would have done that.

cwms-data-api/src/main/java/cwms/cda/api/rating/RatingSpecController.java
rma-rripken: Looked to me like ContentType returned just the type in getType() - any extra parameters are returned via getParameters() (e.g. "version=2" ).   Formats.XML is just "application/xml"   So checking if getType contains "application/xml" seemed like the most basic "is it supposed to be xml" type check I could think of.

rma-bryson: oh just noticed its calling `contains` not `equals`

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecXmlUtils.java
rma-bryson: There is also a @JacksonXmlRootElement annotation that allows you to specify the root name `localName` and a `namespace` - if you create a RatingSpecWrapper class.

rma-rripken: I actually started with an extension of XMLv2 but I hated that OutputFormatter wasn't made generic.  I ended up having a bunch of instanceof and it was really ugly and I gave up on it.  But maybe I'm only partially understanding your suggestion.   XMLv2 has a format() but it also has a parse().  Would you suggest parsing the input into a CDA RatingSpec in either case and only having one Dao create() method?

rma-rripken: wait, are you saying extend XMLv2 and override _only_ the buildXmlMapper() method?  I don't think that will work b/c I'm also adding an extra root element and inserting that namespace crud.   Maybe it could work.  Looking at XMLv2 does have me thinking that maybe I should have my method just throw InternalServerErrorResponse.

rma-bryson: Wrapper class looks good to me. 

No need to extend XMLv2 with your current changes - though looking at the exception handling in `format` vs `parseContent` in XMLv2 makes me wonder why one throws a FormattingException and the other an InternalServerErrorResponse. My initial suggestion was to extend XMLv2 so you could use the existing format/parseContent methods just to preserve consistency in exception handling.

rma-bryson (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecXmlUtils.java
rma-rripken: It bugged me that I'd have to have the Dao thrown an exception from the javalin package.  The dao really shouldn't be concerned with the web.

PR #1566, opened 2026-02-04
CDA-92: Fix to resolve TimeSeriesGroup DAO serializing AssignedTimeSeries nulls

### LINKED ISSUE #1564: Timeseries Groups return "null" if attribute field is blank
**Describe the bug**
In  2026.01.21-testa, the timeseries group will return "null" if attribute field is blank.
```
     {
       "office-id": "LRL",
       "timeseries-id": "DanvilleKY.Stage.Inst.5Minutes.0.USGS-raw",
       "alias-id": "null",
       "ref-ts-id": "null"
     },
```
**Expected behavior**
Prior versions (September) did not return the alias.
```
      {
       "office-id": "LRL",
       "timeseries-id": "DanvilleKY.Stage.Inst.5Minutes.0.USGS-raw",
       "attribute": 0
     },
```

I would expect it to be omitted like prior versions or just `null`. 

**CURL Commands**
```
curl -X 'GET' \
  'https://DEV/cwms-data/timeseries/group?office=LRL&group-office-id=CWMS&timeseries-group-like=USGS&_cb=1770236258817' \
  -H 'accept: application/json' \
  -H 'Cache-Control: no-cache, no-store, max-age=0' \
  -H 'Pragma: no-cache'
```

zack-rma:
Fix to resolve TimeSeriesGroup DAO serializing AssignedTimeSeries null attributes as strings. 

Includes integration test and serialization unit test. 

Resolves #1564

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
@@ -25,17 +25,24 @@
 package cwms.cda.data.dao;
 
 import static com.google.common.flogger.LazyArgs.lazy;
-
 import static java.util.stream.Collectors.toList;
 
+import com.google.common.flogger.FluentLogger;
 import cwms.cda.data.dto.AssignedTimeSeries;
 import cwms.cda.data.dto.TimeSeriesCategory;
 import cwms.cda.data.dto.TimeSeriesGroup;
 import java.math.BigDecimal;
 import java.util.List;
-import com.google.common.flogger.FluentLogger;
 import org.jetbrains.annotations.NotNull;
-import org.jooq.*;
+import org.jooq.Condition;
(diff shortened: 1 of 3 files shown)


rma-rripken (approved):

rma-rripken:
I did a search across the code base and couldn't find any other places were the same pattern was used so we shouldn't hit this error on any other DTO

rma-rripken:
LocationGroups were built slightly differently so they shouldn't exhibit this either

MikeNeilson:
So why did this bypass the "latest schema" check? at least without any explanation. Fast resolution is got, not if it risks breaking something else though.

rma-rripken:
I bypassed the latest schema check b/c it has been broken.  There are two failing IT tests, one in WaterSupplyAccounting and another in TimeSeriesFiltered that only fail in latest.  They failed for me and Bryson both in our recent PR's.

Issue #1567, opened 2026-02-05
Several Dao make use of jOOQ stream() without closing


rma-rripken:
Several Dao make use of jOOQ's stream() but don't explicitly close the stream.   This does not seem to be causing any immediate issues but in trying to debug WaterSupplyAccounting in my IDE I noticed several tests that were triggering Tomcat's abandoned connection warning and noticed that some of the tests in LevelsControllerTestIT were taking extreme amounts of time (4-5 minutes per test).  When the streams were wrapped in try-with-resources those same tests complete in seconds.
I looked in the github environment and I didn't see any of the tests there taking crazy amounts of time so maybe this is only a problem when the tests are run in the debugger?  
If we're supposed to be closing these streams seems like we should be doing it correctly whether they are currently showing problems or not.

This link suggests that users must close jooq streams: https://github.com/jOOQ/jOOQ/issues/4932  I'm sure there are other jOOQ documentation pages that explain this better. 

I'd suggest just adding try-with-resource around the existing stream() calls and not trying to rewrite anything.

These are the Dao that are affected:
LocationGroupDao
LocationLevelsDaoImpl
LocationsDaoImpl
PoolDao
TimeSeriesIdentifierDescriptorDao
TimeZoneDao
StandardTextDao

PR #1570, opened 2026-02-06
Wrap uses of jOOQ stream() with try-with-resources.

rma-rripken:


rma-rripken:
closes #1567

MikeNeilson (approved):
Sorry about the delay, wanted to sanity check that those test errors weren't from this PR. Just the usual water supply update and LRTS/PRTS per session behavior.

rma-rripken:
No rush.  Would be nice to get those failing tests fixed.

PR #1572, opened 2026-02-07
Fixing time series cascade delete issues.

### LINKED ISSUE #1184: Cacade Delete for Timeseries Group?
Is it possible to get a cascade delete option for the timeseries group similar to the location group? Otherwise it's not possible to delete a timeseries category with CDA.

rma-rripken:
Fixes #1184


rma-rripken:
The only thing that is left to do (IMO) is to have the create and update either throw an error if the user-provided body includes groups and assignments or to cascade the create/update.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesGroupDao.java
```diff
@@ -56,6 +57,16 @@ public class TimeSeriesGroupDao extends JooqDao<TimeSeriesGroup> {
...
     public static final String CWMS = "CWMS";
 
+    private enum DeleteTsGroupCascadeMode {
+        UNKNOWN,
+        USE_CASCADE_ROUTINE,
+        USE_UNASSIGN
+    }
+
+    private static volatile DeleteTsGroupCascadeMode deleteTsGroupCascadeMode = DeleteTsGroupCascadeMode.UNKNOWN;
+    private static final Object deleteTsGroupCascadeModeLock = new Object();
```
MikeNeilson: I think you can and should just synchronize on the deleteTsGroupCascadeMode object. Though maybe only if it isn't volatile, alternatively, maybe use AtomicReference?

rma-rripken: I can't synchronize on deleteTsGroupCascadeMode b/c the reference changes.   I could switch to holding the mode in an AtomicReference but I think I'd still want another object to synchronize on (could maybe use TimeSeriesGroupDao.class).  That block keeps a compound action (test + db call + set) atomic.  Although keeping it atomic is probably overkill. What would it matter if two requests both try to call a non-existing procedure as long as they both failover to a method that works?  The mode flag was an optimization - once we know which method works we don't need to keep probing and failing-over.

rma-rripken: I went ahead and removed the synchronized block.  Volatile is good enough given the way we're using the flag.

rma-rripken:
@MikeNeilson   I can apparently assign time series to a CWMS owned group by I can't unassign them.  The error I see is:  

```
Caused by: Error : 20048, Position : 0, Sql = BEGIN "CWMS_20"."CWMS_TS"."UNASSIGN_TS_GROUP" (:1 , :2 , :3 , :4 , :5 ) ; END;, OriginalSql = { call "CWMS_20"."CWMS_TS"."UNASSIGN_TS_GROUP" (?, ?, ?, ?, ?) }, Error Msg = ORA-20048: NO_WRITE_PRIVILEGE: User doesnt have write privileges
ORA-06512: at "CWMS_20.CWMS_ERR", line 80
ORA-06512: at "CWMS_20.ST_TS_GROUP_ASSIGNMENT", line 9
ORA-04088: error during execution of trigger 'CWMS_20.ST_TS_GROUP_ASSIGNMENT'
ORA-06512: at "CWMS_20.CWMS_TS", line 10803
ORA-06512: at line 1
```

If the category/group are owned by CWMS but there are assignments for SPK time series should I (as normal SPK user) be able to unassign those SPK timeseries?   Should I be able to unassign non-SPK time series from a CWMS group?   Should I be able to create/delete CWMS groups?

MikeNeilson:
Create/Delete CWMS group: no, those are permanent
Remove offices to which no permissions: no
Remove own timeseries en mass: That does seem reasonable. Just wouldn't be deleting the group, just unassigning for that office.

rma-rripken:
I've disabled the two tests that deal with CWMS owned groups and created a new issue to resolve that bug.  

https://github.com/USACE/cwms-data-api/issues/1631

MikeNeilson (approved):

PR #1546, opened 2026-01-22
CDA-56 - Enhance exception handling to include detailed error message

RyanM-RMA:
Closes #1287 
This provides additional details to the error message so there's less guessing about the issue.

PR #1575, opened 2026-02-10
Bugfix/cda 56 add error

rma-rripken:
Supersedes #1546

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/errors/NoDataRateException.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/errors/NoDataRateException.java
@@ -1,0 +1,21 @@
+package cwms.cda.api.errors;
+
+import java.sql.SQLException;
+import java.util.LinkedHashMap;
+import java.util.logging.Level;
+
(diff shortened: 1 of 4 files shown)


cwms-data-api/src/main/java/cwms/cda/api/errors/NoDataRateException.java
```diff
@@ -0,0 +1,21 @@
...
+import java.sql.SQLException;
+import java.util.LinkedHashMap;
+import java.util.logging.Level;
+
+public class NoDataRateException extends RateException{
+    // 404 is Not Found - this isn't quite right b/c the rate() or reverse-rate() isn't failing b/c
+    // the ts isn't found its failing b/c a necessary rating table isn't found.
+    // 424 is Failed Dependency - this is closer to what we want, but it's not quite right either.
+
+    // 422 is Unprocessable Entity.  I think this is the closest.
```
adamkorynta, line 12: I do think this is the best/closest option. 

I'm going to throw out a wild idea though...... what about 204 No Content?

adamkorynta (approved):

PR #1578, opened 2026-02-11
fix paging for rating spec getAll endpoint

adamkorynta:
- the total size of rating specs was not queries, so next-page was never generated
- optimize the query by paging and filtering just on the rating specs, not the rating specs + effective dates
- optimize `caseInsensitiveLikeRegex` condition for `"*"` and `".*"` by setting it to `noCondition()`. This is especially helpful when columns are concatenations like TSIDs, Rating Ids, etc

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecDao.java
@@ -55,8 +55,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
 import java.util.NavigableSet;
+import java.util.Objects;
 import java.util.Optional;
-import java.util.Set;
 import java.util.TimeZone;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -130,8 +130,8 @@ public class RatingSpecDao extends JooqDao<RatingSpec> {
                     } catch (NumberFormatException e) {
                         logger.atInfo().log("Could not parse %s", parts[1]);
                     }
-                    pageSize = Integer.parseInt(parts[2]);
                 }
+                pageSize = Integer.parseInt(parts[2]);
             }
         }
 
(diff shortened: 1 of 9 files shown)


rma-rripken:
* isn' t really a regex.  Oracle treats it as .*

adamkorynta:
do you think I should capture both `"*"` and `".*"` as no condition?

rma-rripken:
things that use * might work but imo they are broken and the default should be .*

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecDao.java
```diff
@@ -191,150 +130,109 @@ public RatingSpecs retrieveRatingSpecs(String cursor, int pageSize, String offic
...
-        Map<RatingSpec, List<ZonedDateTime>> map = new LinkedHashMap<>();
-        try (Stream<? extends Record> stream = query.fetchStream()) {
-            stream.forEach(rec -> {
-                RatingSpec template = buildRatingSpec(rec);
-
-                Timestamp effectiveDate = rec.get(ratView.EFFECTIVE_DATE);
-                ZonedDateTime effective = toZdt(effectiveDate);
-
-                List<ZonedDateTime> list = map.computeIfAbsent(template, k -> new ArrayList<>());
+        // Preserve stable ordering
```
rma-rripken: "Preserve stable ordering" isn't whats happening.  You are putting them in HashMap which doesn't preserve stable ordering then you are re-sorting in Java.

adamkorynta: oh right, forgot to remove that comment after I switched from LinkedHashMap

rma-rripken: Have you looked at multiset?  I feel like this would be much clearer as a multiset query and you could have the database do all the sorting

rma-rripken:
Maybe an integration test that inserts a bunch and then pages the results and verifies they are all there?  That is really what should have happened before you found this error.

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecDao.java
```diff
@@ -191,150 +130,108 @@ public RatingSpecs retrieveRatingSpecs(String cursor, int pageSize, String offic
                     } catch (NumberFormatException e) {
                         logger.atInfo().log("Could not parse %s", parts[1]);
                     }
+                    pageSize = Integer.parseInt(parts[2]);
```
rma-rripken: why does this line move?

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecDao.java
adamkorynta: I hadn't looked at multiset before, I implemented it and it seems to behave nicely.

cwms-data-api/src/main/java/cwms/cda/data/dao/RatingSpecDao.java
adamkorynta: that was a mistake, reverted

rma-rripken (dismissed):

rma-rripken (approved):

PR #1581, opened 2026-02-12
Add flag to allow encoded slashes.

rma-rripken:


MikeNeilson (approved):
... are we actually waiting on anything for this?

Seems ready.

PR #1587, opened 2026-02-17
CDA-72: OpenAPI Documentation: Added verification of requirement status for query parameters to tests

### LINKED ISSUE #1494: Expand static code analysis testing to verify implementations respect OpenApi annotations with "required = true"
Static code analysis testing is not ensuring that tests on endpoint implementations with "required = true" parameters enforce that the parameter is supplied.

Update static code analysis testing to fail when the above case occurs.

**Additional context**
Endpoint example that would need to be tested, getOne from TimeSeriesGroupController.
https://github.com/USACE/cwms-data-api/blob/develop/cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesGroupController.java#L184

zack-rma:
Resolves #1494

Adds verification of required status for query parameters. Also checks parameter type.


rma-rripken (dismissed):
These all look good but there are still some failing static analysis tests related to required parameters for timezone and effective-date.  I assume you'll look into those in a follow-up?

rma-rripken (approved):
The annotation isn't my favorite solution but its very effective from a cost/value standpoint

MikeNeilson:
> The annotation isn't my favorite solution but its very effective from a cost/value standpoint

Agreed, certainly can't think of anything better and it *is* clear to the reader (developer) what might be happening.

MikeNeilson (approved):

PR #1586, opened 2026-02-17
CDA-44 - Adds location vertical-datum endpoint. Implements vertical-datum conversion for location retreival.

rma-bryson:
Fixes: https://github.com/USACE/cwms-data-api/issues/1577
Part of fix for: https://github.com/USACE/cwms-data-api/issues/1101

CDA-44 - Adds test for both xml and json content-types for vertical datum info

CDA-44 - Improved error handling of create, update, delete. Added workarounds to some db behavior. Adds tests for vertical datum endpoint.

Implements vertical datum endpoint. Failing test needs to be resolved.

Fixed failing tests. Leaving in this state until decision on Location DTO is made to handle VerticalDatumInfo, as we need a means of storing to the datum offset table.

Defaulted units for storing vertical datum location

Reverts much of the DTO with internal json ignored Vertical Datum Info. Updated tests, Updated dao and controller to handle vertical datum info conversions with DTO changes. Assumption that we can only convert if we have a vdi entry in table for location.

adds back store with fail-if-exists param

adds updated json that uses vertical datum info

Fixes broken build. Achieves passing retrieval test.

adds vertical datum info to retrieved location data. Adds conversion logic. Adds integration tests.

CDA-44 - DTO updates and converter added on update and update to table on create

PR #1591, opened 2026-02-18
Devops/update keycloak

MikeNeilson:
Updates keycloak to a version that continuous to run in gh actions (weird, but not gonna worry about it as we *should* keep that up to date.

Also partially reverts a change made in #1586 that incorrectly sets the user session to the provided office id, which is intended purely as a search constraint.

### DIFF
--- a/compose_files/keycloak/healthcheck.sh
+++ b/compose_files/keycloak/healthcheck.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+{ printf 'HEAD /auth/health/ready HTTP/1.0\r\n\r\n' >&0; grep 'HTTP/1.0 200'; } 0<>/dev/tcp/localhost/9000
--- a/cwms-data-api/logging.properties
+++ b/cwms-data-api/logging.properties
@@ -67,7 +67,8 @@ org.apache.catalina.core.ContainerBase.[Catalina].[localhost].[/cwms-data].handl
 
 fixtures.level=FINE
 cwms.cda.datasource.level=FINE
-org.apache.level=FINE
+org.apache.level=ERROR
+org.apache.tomcat.jdbc.pool.level=INFO
 org.apache.catalina.realm.level=INFO
 org.apache.catalina.realm.useParentHandlers=true
 org.apache.catalina.authenticator.level=INFO
--- a/cwms-data-api/src/test/resources/tomcat/conf/context.xml
+++ b/cwms-data-api/src/test/resources/tomcat/conf/context.xml
@@ -16,7 +16,7 @@
               testWhileIdle="true"
               timeBetweenEvictionRunsMillis="5000"
               logValidationErrors="true"
-              suspectTimeout="15"
+              suspectTimeout="8"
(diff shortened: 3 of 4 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
rma-bryson: This gets called from getLocations(). Does this need to have the session office set?

MikeNeilson: If it does we have a design issue. getLocations is quite often called by "guests", even a non-guest should be able to search the entire database regardless of what their user office permissions are. The officeId parameter to getLocation is not a session related office, like it would be for say POST, but is instead just a limiter on the locations query that is run.

rma-bryson: Perhaps it isn't necessary here then. Makes me wonder where else the session office is being set but not needed.

rma-rripken (approved):

MikeNeilson:
well, it appears @rma-bryson is correct, the vertical datum system requires that office id; however that is rather problematic as requiring the office id at this level breaks the basic functionality of you know, letting a user search though the data of the databases whether they're logged in or not.

rma-rripken:
So its needed for vertical-datum conversion? Okay, but that only happens if the user adds the datum flag - right?  Seems like we could shift the call to set the office in the session to inside just that branch and maybe add a comment on the datum flag that users need to be logged-in to use that feature or something like that.

rma-bryson:
Wondering if we can replace the loc package call with a view query of the AV_VERT_DATUM_OFFSET?

MikeNeilson:
> Wondering if we can replace the loc package call with a view query of the AV_VERT_DATUM_OFFSET?

that might work.

> So its needed for vertical-datum conversion? Okay, but that only happens if the user adds the datum flag - right? Seems like we could shift the call to set the office in the session to inside just that branch and maybe add a comment on the datum flag that users need to be logged-in to use that feature or something like that.


Doesn't really change the situation, if the user doesn't have permissions for that specific office for the datum, the set session office will still fail. Thus breaking the search in which we allow anyone to ask for the datum.

rma-bryson:
https://github.com/USACE/cwms-data-api/pull/1594 - here's a view query update that works @MikeNeilson

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
rma-rripken: I did a quick search for places that set the session office id in GET type methods: 
getLocation getLocations  (this pr is about these)
retreiveLocationLevel
getRatingSet
retrieveLatestXML
getSpecifiedLevels
retrieveVerticalDatumInfo
getAllBasins getBasin
getStream getStreamLocations
getReachesOnStream
retrieveLockCatalog
retrieveLock
retrieveProject
catProject 
retrieveStandardText

PR #1594, opened 2026-02-19
CDA-44 - Updates getLocations to query the view for datum conversions.

rma-bryson:


rma-bryson:
There's a failing test - that passes when i run from my IDE. Will look more into that, but this at leaves gives an idea towards using a view query for performing datum conversions when retrieving locations.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
```diff
@@ -142,14 +153,62 @@ public List<Location> getLocations(String nameRegex, String unitSystem, String d
...
+        }
+
+        // Determine the location's native vertical datum (stored on the Location)
+        String nativeDatum = loc.getVerticalDatum();
+        if (nativeDatum == null || nativeDatum.isBlank()) {
+            return loc; // nothing to convert from
+        }
+
+        // Query offsets from the materialized view for this location/office
+        DSLContext ctx = DSL.using(configuration);
```
MikeNeilson: seems like it might be better/more clean if this method takes the context instead of configuration. Seems rare this would every need to run in a different JDBC session that whatever the primary query is.

rma-rripken: yes.  I think Bryson was just trying to keep the signature the same but I agree, if we don't specifically need Configuration (which is usually needed by pl/sql calls) we shouldn't pass it just to convert it.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
```diff
@@ -142,14 +153,62 @@ public List<Location> getLocations(String nameRegex, String unitSystem, String d
...
+            }
+            String id1Normalized = id1.replace("-", "");
+            String nativeDatumNormalized = nativeDatum.replace("-", "");
+            if (id1Normalized.equalsIgnoreCase(nativeDatumNormalized)) {
+                String toDatum = r.get(AV_VERT_DATUM_OFFSET.AV_VERT_DATUM_OFFSET.VERTICAL_DATUM_ID_2);
+                Double value = r.get(AV_VERT_DATUM_OFFSET.AV_VERT_DATUM_OFFSET.OFFSET);
+                String desc = r.get(AV_VERT_DATUM_OFFSET.AV_VERT_DATUM_OFFSET.DESCRIPTION);
+                boolean estimate = desc != null && desc.toLowerCase().contains("estimate");
+                offsets.add(new VerticalDatumInfo.Offset(estimate, toDatum, value));
             }
```
rma-rripken: put this part in a method and add it to jooq as a mapper.  It will be a little bit cleaner.  Jooq will do the for(...) part for you and you'll get your offset right out of the fetch()

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
rma-rripken: I think Mike is right.  The DSL.using(configuration) is probably going to get a new connection from the pool -which we don't want to have happen

MikeNeilson: I was seeing "no connections available" during the tests, I thought it was from something failing but it likely was this. The pool size is 1 intentionally in the integration tests.

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
rma-bryson: updated per your suggestion

cwms-data-api/src/test/java/cwms/cda/api/LocationControllerTestIT.java
```diff
@@ -256,83 +255,20 @@ void test_location_create_get_delete() throws Exception {
...
+        createLocationWithVerticalDatum(locNgvd, true, officeId, VerticalDatum.NGVD29);
+        addVerticalDatumOffsetForExistingLocation(locNgvd, officeId, VerticalDatum.NGVD29, VerticalDatum.NAVD88, -0.5, true);
 
         // Request NAVD88 for NGVD29 location
         given()
             .log().ifValidationFails(LogDetail.ALL,true)
             .accept(Formats.JSONV2)
             .queryParam(OFFICE, officeId)
-            .queryParam(DATUM, VerticalDatum.NAVD88.toString())
+            .queryParam(DATUM, VerticalDatum.NGVD29.toString())
```
rma-rripken: Is the comment wrong on this?

cwms-data-api/src/main/java/cwms/cda/data/dao/LocationsDaoImpl.java
rma-bryson: I updated this to parameterize the context.

cwms-data-api/src/test/java/cwms/cda/api/LocationControllerTestIT.java
```diff
@@ -256,83 +255,20 @@ void test_location_create_get_delete() throws Exception {
...
-            .queryParam(OFFICE, officeId)
-        .when()
-            .redirects().follow(true)
-            .redirects().max(3)
-            .post("/location/" + locNavd + "/vertical-datum")
-        .then()
-            .log().ifValidationFails(LogDetail.ALL,true)
-            .assertThat()
-            .statusCode(is(HttpServletResponse.SC_CREATED));
+        // Create two locations with explicit native datums
```
rma-rripken: Is this still creating two locations?  Seems like we are creating one?

cwms-data-api/src/test/java/cwms/cda/api/LocationControllerTestIT.java
rma-bryson: yeah it is. Updated

cwms-data-api/src/test/java/cwms/cda/api/LocationControllerTestIT.java
rma-bryson: Just one, forgot to update the comments. Fixed now

rma-bryson:
@MikeNeilson  - tests are now passing for the release schema

cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
```diff
@@ -285,6 +288,45 @@ protected static void addNewUser(String username) throws SQLException {
         }, "cwms_20");
     }
 
+    protected static void addVerticalDatumOffsetForExistingLocation(String location, String officeId, VerticalDatum from, VerticalDatum to, double offset, boolean isEstimate) throws SQLException {
+        CwmsDatabaseContainer<?> db = CwmsDataApiSetupCallback.getDatabaseLink();
+        String desc = isEstimate ? "ESTIMATE" : "";
+        final String insertSql =
+                "INSERT INTO AT_VERT_DATUM_OFFSET " +
+                        " (LOCATION_CODE, VERTICAL_DATUM_ID_1, VERTICAL_DATUM_ID_2, EFFECTIVE_DATE, OFFSET, DESCRIPTION) " +
```
MikeNeilson, line 296: seems like it would be better use an API call, but this is test code.

MikeNeilson (approved):

MikeNeilson:
verified only the expected tests targeting CWMS Database latest are failing.

PR #1596, opened 2026-02-19
Moving LRTS handling into a Connection Preparer

rma-rripken:
This handles the LRTS flag in a way that unsets the flag in the session when the connection is closed.


cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -135,34 +134,25 @@ protected JooqDao(DSLContext dsl) {
      */
     public static DSLContext getDslContext(Context ctx) {
         DSLContext retVal;
-        final String officeId = ctx.attribute(ApiServlet.OFFICE_ID);
+
         final DataSource dataSource = ctx.attribute(ApiServlet.DATA_SOURCE);
         final Boolean isNewLRTS = ctx.header(ApiServlet.IS_NEW_LRTS) == null
```
adamkorynta: null should really be false for now

cwms-data-api/src/main/java/cwms/cda/datasource/LrtsSessionPreparer.java
```diff
@@ -0,0 +1,103 @@
...
+
+        // Return a proxy that will unset the flag on close()
+        return (Connection) Proxy.newProxyInstance(
+                connection.getClass().getClassLoader(),
+                new Class<?>[]{Connection.class},
+                new CloseUnsettingHandler(connection));
+    }
+
+
+    private static class CloseUnsettingHandler implements InvocationHandler {
```
adamkorynta, line 69: I don't think we should be unsetting. it should always be explicitly true or false. That way we aren't letting the database decide what the default behavior of CDA is.

cwms-data-api/src/main/java/cwms/cda/datasource/LrtsSessionPreparer.java
```diff
@@ -0,0 +1,103 @@
...
+                    : JooqDao.REQUIRE_OLD_LRTS_ID_FORMAT;
+            CWMS_UTIL_PACKAGE.call_SET_SESSION_INFO(dsl.configuration(),
+                    JooqDao.SESSION_USE_LRTS_ID_FORMAT, requireBool, requireIntValue);
+            logger.atFine().log("Set LRTS session flag to %s (%d) for connection %s",
+                    requireBool, requireIntValue, connection);
+        } catch (RuntimeException ex) {
+            // Convert to SQLException contract if the call fails
+            SQLException sqlEx = new SQLException("Failed setting LRTS session flag", ex);
+            try {
+                connection.close();
```
adamkorynta: We shouldn't close the connection, the calling API should handle that.

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
rma-rripken: ok. I made null be false.

cwms-data-api/src/main/java/cwms/cda/datasource/LrtsSessionPreparer.java
rma-rripken: Ok. I put the unsetting bit behind a flag and default to false.

cwms-data-api/src/main/java/cwms/cda/datasource/LrtsSessionPreparer.java
rma-rripken: Took out the close class.

cwms-data-api/src/test/java/cwms/cda/api/DataApiTestIT.java
```diff
@@ -677,7 +676,7 @@ public void cleanupLocationGroups() throws Exception {
     @AfterEach
     public void cleanupLocationCategories() throws Exception {
         if (this.categoriesCreated.isEmpty()) {
-            logger.atInfo().log("No location categories to cleanup.");
+            logger.atFine().log("No location categories to cleanup.");
```
MikeNeilson, line 679: good change.

adamkorynta:
I believe the office id context and this LRTS context is designed with the understanding that the session state would be cleared when the connection is returned to the pool. Since that seem to not be the case, I think the real solution would be to add something like: 

```
BEGIN
  DBMS_SESSION.CLEAR_ALL_CONTEXT;
  DBMS_SESSION.RESET_PACKAGE;
END;
```

either as the first connection preparer or on connection return.

rma-rripken:
I made a new issue for any work to add session clearing code.

adamkorynta (approved):

PR #1597, opened 2026-02-20
If the paging condition uses upper-case on the cursorId then the orde…

rma-rripken:
…rBy should also use upper-case on the column.

The office-id columns in the view will already be upper-cased and we don't need the database to try and do extra work.


rma-rripken:
I think this fixes a test that sometimes fails in latest schema.  Not sure why it only sometimes fails though

adamkorynta (approved):

MikeNeilson (changes requested):
While the failed tests are clearly unrelated (but do we know why those started failing?), there should be a test setup for this particular condition to prevent regressions in the future.

PR #1598, opened 2026-02-20
Changing sql resource file to use named parameters

rma-rripken:
I think this will fix a Catalog test that started failing in latest schema


zack-rma:
Catalog test appears to still be failing from 

```
PLS-00307: too many declarations of 'DELETE_TS_GROUP' match this call
```

rma-rripken:
yep.  Unfortunately this isn't going to fix it.  Related to: https://github.com/HydrologicEngineeringCenter/cwms-database/pull/120  and https://github.com/HydrologicEngineeringCenter/cwms-database/pull/101#issuecomment-3937838360

PR #1603, opened 2026-02-24
update jooq codegen to 26.02.17-RC01

adamkorynta:


MikeNeilson:
```
at app//cwms.cda.data.dao.watersupply.WaterSupplyAccountingDaoIT.testStoreAndRetrieveWaterSupplyPumpAccounting(WaterSupplyAccountingDaoIT.java:157)

        Caused by:
        java.sql.SQLException: Error while writing value at bind index: 1
            at org.jooq.impl.DefaultBindContext.bindValue0(DefaultBindContext.java:80)
            at org.jooq.impl.AbstractBindContext.bindValue(AbstractBindContext.java:74)
            ... 26 more

```

It would appear we may need find some way to allow for multiple gone gens. Maybe we should start namespacing the packages of the code gen output by version... or like shadowing it or something. e.g. have something that works with out the version but have everything truly exist under that namespace to allow these transitions.... definitely not getting fixed right for this though.

rma-rripken:
I was looking into that WaterSupplyAcct thing.  We could switch based on schema version and call the pl/sql directly (not thru codegen).    Another option would be to introspect the arguments by calling something like:
```
SELECT OBJECT_NAME, OVERLOAD, ARGUMENT_NAME, POSITION, DATA_TYPE, IN_OUT
FROM ALL_ARGUMENTS
WHERE OWNER = 'CWMS_20'
  AND PACKAGE_NAME = 'CWMS_WATER_SUPPLY'
  and OBJECT_NAME = 'RETRIEVE_ACCOUNTING_SET'
ORDER BY OVERLOAD, POSITION;
``` 

Or we just call one signature and catch the BindException and then try the other signature and maybe cache/remember which one worked.

MikeNeilson:
> I was looking into that WaterSupplyAcct thing. We could switch based on schema version and call the pl/sql directly (not thru codegen). Another option would be to introspect the arguments by calling something like:
> 
> ```
> SELECT OBJECT_NAME, OVERLOAD, ARGUMENT_NAME, POSITION, DATA_TYPE, IN_OUT
> FROM ALL_ARGUMENTS
> WHERE OWNER = 'CWMS_20'
>   AND PACKAGE_NAME = 'CWMS_WATER_SUPPLY'
>   and OBJECT_NAME = 'RETRIEVE_ACCOUNTING_SET'
> ORDER BY OVERLOAD, POSITION;
> ```
> 
> Or we just call one signature and catch the BindException and then try the other signature and maybe cache/remember which one worked.

Yeah, either of those options makes sense and would be easier to handle.

rma-rripken (approved):
Did the codegen issues related to plsqlBoolean get resolved?

adamkorynta:
> Did the codegen issues related to plsqlBoolean get resolved?

No - TeamCity is not happy with the 19c database for some reason. I've requested more time on the associated Jira ticket to proceed further. Marking this PR as draft until that is sorted out.

Issue #1568, opened 2026-02-05
WaterSupplyAccounting IT failing in latest


rma-rripken:
The Integration Tests for WaterSupplyAccounting are failing with the latest schema.   

I believe cwms-database changes affected the shape of some Oracle user-defined-types and there is a mismatch between what jOOQ thinks the types need vs what the database is expecting.  This mismatch is leading to jOOQ sending a description field into what the database is expecting to be a timestamp.  The string is not parsing as a timestamp and eventually triggering a bind exception.

CDA might be able to work around the issue by doing its own type mapping or perhaps codegen can be run against the latest schema and the CDA dependency updated depending on the schema ?

PR #1609, opened 2026-02-27
Bugfix/1568 ws plsql mismatch

rma-rripken:


rma-rripken:
This is ugly. I'm open for suggestions on how else to resolve the errors or how to better tag what database version these manually created types/procedures go with.

rma-rripken:
Fixes #1568

rma-rripken:
It might be possible to do something like: select * from ALL_TYPE_ATTRS where OWNER = 'CWMS_20' and TYPE_NAME = 'WAT_USR_CONTRACT_ACCT_OBJ_T'   and see if the results contain PUMP_FLOW_UNIT.

rma-bryson:
This seems reasonable for now. If this is becomes more common going forward, maybe there is a solution involving 2 codegens and an abstraction layer atop of that.

rma-bryson (approved):

PR #1614, opened 2026-03-02
CDA-88 - Implements username regex filter for /users endpoint

rma-bryson:


cwms-data-api/src/main/java/cwms/cda/api/auth/users/UsersController.java
```diff
@@ -66,9 +66,11 @@ public void delete(Context ctx, String username) {
 
     @OpenApi(
         queryParams = {
-            @OpenApiParam(allowEmptyValue = true, name = OFFICE, type = String.class,
+            @OpenApiParam(allowEmptyValue = true, name = OFFICE,
                     description = "Show only users with active privileges in a given office." 
                                 + Controllers.OFFICE_DESCRIPTION ),
+            @OpenApiParam(name = USERNAME,
```
rma-rripken: If its a regex the parameter name should be username-regex or username-like or username-mask.  username-regex seems the most explicit to me.

Also, look at any other regex parameter - the swagger desc should link to the regex doc page.

rma-rripken (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UsersController.java
MikeNeilson: don't we consistently use "<root name>-like" for the regex fields? Probably makes sense here to do both. username and username-like, backend can easily turn the username into a regex that means "I mean this exactly."

rma-rripken: +1 username-like.  I'd rather not have two parameters that do the same thing.  If it takes a regex you can always pass in an exact match.  Unless there is/was an existing parameter that we need to maintain support for.

rma-rripken (dismissed):

cwms-data-api/src/main/java/cwms/cda/data/dao/UserDao.java
```diff
@@ -189,6 +194,11 @@ public Users getAll(String cursor, int pageSize, String office, boolean includeR
                         .from(vUserGroups)
                         .where(upper(vUserGroups.DB_OFFICE_ID).eq(upper(limitOffice)))
                         .and(vUserGroups.IS_MEMBER.eq("T")).asField().gt(1);
+
+                    // Note: usernameRegex is not included in the cursor; apply it if present in current request
```
MikeNeilson: huh, if it's not in the cursor, how are we paginating through it? the value of next-page is supposed to be self contained.

rma-rripken: Many end-points, maybe even more than half, of the end-points do not include all the query parameters in the next-page cursor.  Not saying that its right not to include them all but that is where we are at right now.

rma-rripken: Maybe we make a list of which end-points could be fixed with regard to the next-page cursor?

rma-bryson: upon closer inspection, Users does include the username query param in the next-page cursor. Removed those lines of code. Also added test coverage.

rma-rripken: Look again.  If it was in the cursor somewhere you'd be pulling it out of parts[] and your Users object would be passing it into encodeCursor.  If its not in the cursor and you removed it from the query params in the paging test and your IT is still passing that tells me the IT isn't right or there aren't enough IT.

rma-bryson: Okay, so it includes the last username, but that is not the regex query param. 

If we want it self-contained, then we should include the regex in the cursor, otherwise the regex query param needs to be included in each subsequent paging web requests.

rma-bryson: Seems like it belongs in the cursor rather than a query param. Updated to include it.

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UsersController.java
```diff
@@ -105,6 +99,7 @@ public void getAll(Context ctx) {
         try (final Timer.Context ignored = markAndTime(GET_ALL)) {
             DSLContext dsl = getDslContext(ctx);
             String office = ctx.queryParam(OFFICE);
+            String username = ctx.queryParam(USERNAME_LIKE);
```
rma-rripken: call the variable usernameRegex or something similar to help track what it holds

cwms-data-api/src/main/java/cwms/cda/data/dto/PageCursor.java
```diff
@@ -0,0 +1,14 @@
+package cwms.cda.data.dto;
+
+import java.util.Base64;
+
+public interface PageCursor {
+    void decodeCursor(String cursor, String delimiter);
+    String encode(Base64.Encoder encoder, String delimiter);
```
MikeNeilson: Add a new line between things -- all kinda blurred into one method on first read. Maybe a short javadoc explaining expectations for each method. Might be obvious to us but not the new reader.

MikeNeilson (changes requested):
Good change with the new interface. Have a little nitpick on the formatting and documentation for future devs.

cwms-data-api/src/main/java/cwms/cda/api/auth/users/UsersController.java
rma-bryson: updated to usernameRegex

cwms-data-api/src/main/java/cwms/cda/data/dto/PageCursor.java
rma-bryson: added line breaks and javadocs

MikeNeilson (approved):

Issue #1618, opened 2026-03-03
Cannot Store Forecast After Initial Store


rma-bryson:
**Describe the bug**
After initial creation of a forecast instance, any subsequent patch will fail. Issue appears to be related to storing a null spec-id. 

**Expected behavior**
This should update the existing forecast instance, rather than failing with an exception. 

**To Reproduce**

1. Create a forecast instance and store
2. Attempt to store the same forecast instance
3. See error

**Logs/Incident Identifier**
3-Mar-2026 10:10:15.876 WARNING [http-nio-auto-1-exec-10] cwms.cda.ApiServlet.lambda$init$8 error on request[2db143e2-a925-488b-a62e-34716aa1ab9b]: /cwms-data/forecast-instance/
	org.jooq.exception.IntegrityConstraintViolationException: SQL [{ call "CWMS_20"."CWMS_FCST"."STORE_FCST" (?, ?, cast(? as date), cast(? as date), ?, ?, ?, ?, ?, ?, ?, ?) }]; ORA-01400: cannot insert NULL into ("CWMS_20"."AT_FCST_INST"."FCST_SPEC_CODE")

**Additional context**
This is blocking work for Extract/Post enhancements, which requires storing a forecast instance more than once. There is an existing integration test for create/update, but it is disabled with a comment describing this issue.

rma-bryson:
@MikeNeilson - This issue is now causing failures with the Extract/Post enhancements in RTS, and as such is blocking that work. This is likely a db side issue, but could perhaps have a workaround on the Java side of things.

PR #1621, opened 2026-03-04
Java-side stop-gap solution to storing/updating forecast instance 

rma-bryson:
...after initial creation. Real fix is database-side.

stop-gap solution to https://github.com/USACE/cwms-data-api/issues/1618

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
@@ -307,7 +307,7 @@ public final class ForecastInstanceDao extends JooqDao<ForecastInstance> {
         String office = officeArg;
 
         String query = INSTANCE_QUERY + GET_ONE_CONDITIONS;
-        return ctx.connectionResult(c -> {
+        return connectionResult(ctx, c -> {
             try (PreparedStatement preparedStatement = c.prepareStatement(query)) {
                 preparedStatement.setString(1, office);
                 preparedStatement.setString(2, name);
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
@@ -150,7 +150,7 @@ public final class ForecastInstanceDao extends JooqDao<ForecastInstance> {
                 .withOperatorKey("{noop}");
         try {
             // If the instance doesn't exist this will throw a NotFoundException which we can ignore, if it does exist we want to delete it before storing the new one
-            getForecastInstance(0, noopUrlBuilder, officeId, forecastInst.getSpec().getSpecId(), forecastInst.getSpec().getDesignator(), forecastDate.toInstant(), issueDate.toInstant());
+            getForecastInstance(ctx, 0, noopUrlBuilder, officeId, forecastInst.getSpec().getSpecId(), forecastInst.getSpec().getDesignator(), forecastDate.toInstant(), issueDate.toInstant());
             CWMS_FCST_PACKAGE.call_DELETE_FCST(ctx.configuration(), forecastInst.getSpec().getSpecId(), forecastInst.getSpec().getDesignator(),
                     forecastDate, issueDate, "UTC", officeId);
         } catch (NotFoundException e) {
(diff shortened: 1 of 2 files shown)


rma-rripken:
Could probably use a transaction - so if the store fails then the delete would get rollled back.

Its not in your changes but I'm curious about that thread local utc calendar.  I wonder how things break if that isn't there , maybe Zack needs that for his timezone issue.

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
```diff
@@ -127,15 +127,28 @@ public void create(ForecastInstance forecastInst) {
        blob.setQUALITY_CODE(0L);
        blob.setTHE_BLOB(fileData);
        connection(dsl, conn -> {
-           setOffice(conn, officeId);
+           DSLContext ctx = getDslContext(conn, officeId);
            DefaultBinding.THREAD_LOCAL.set(UTC_CALENDAR);
-           CWMS_FCST_PACKAGE.call_STORE_FCST(DSL.using(conn).configuration(), forecastInst.getSpec().getSpecId(),
+           clearExistingForecastInstance(forecastInst, officeId, forecastDate, issueDate, ctx);
```
rma-rripken: I noticed that Jooq codegen wants the config first and I've been trying to follow suit when I pass around a Configuration or DslContext.

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
```diff
@@ -127,15 +127,28 @@ public void create(ForecastInstance forecastInst) {
...
                    "UTC", forecastInst.getMaxAge(), forecastInst.getNotes(), forecastInfo,
                    blob, "F", "T", officeId);
        });
     }
 
+    private void clearExistingForecastInstance(ForecastInstance forecastInst, String officeId, Timestamp forecastDate, Timestamp issueDate, DSLContext ctx) {
+        ReplaceUtils.OperatorBuilder noopUrlBuilder = new ReplaceUtils.OperatorBuilder().withTemplate("")
+                .withOperatorKey("{noop}");
+        try {
+            getForecastInstance(0, noopUrlBuilder, officeId, forecastInst.getSpec().getSpecId(), forecastInst.getSpec().getDesignator(), forecastDate.toInstant(), issueDate.toInstant());
```
rma-rripken: I don't see the dsl/connection/configuration passed into this.  Doesn't this cause another connection to be pull from the pool?

rma-rripken: Also, if you aren't doing anything with the result why call getForecastInstance?

rma-rripken: So that it triggers a not found exception and skips the delete?

rma-rripken (changes requested):
Can you wrap with a transaction and add a comment about why you are calling getForecastInstance but ignoring the result.

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
rma-bryson: I grabbed that from what the update method was doing to check for existing. But i'll add a comment explaining.

rma-bryson: I also updated to pass in the connection

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
rma-bryson: updated to move that as the first param

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
rma-rripken: I think the getForecastInstance call is going to go to its dsl member var in order to access the db which is what would try and open a new connection.

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
```diff
@@ -271,13 +293,21 @@ private static ForecastInstance map(int byteLimit, ReplaceUtils.OperatorBuilder
...
+    private static ForecastInstance getForecastInstance(DSLContext ctx, int byteLimit, ReplaceUtils.OperatorBuilder urlBuilder, String officeArg,
+                                                        String name, String designator, Instant forecastDate, Instant issueDate) {
         if(officeArg != null){
             officeArg = officeArg.toUpperCase();
         }
         String office = officeArg;
 
         String query = INSTANCE_QUERY + GET_ONE_CONDITIONS;
-        return connectionResult(dsl, c -> {
+        return ctx.connectionResult(c -> {
```
rma-rripken: Is there a reason this uses the jooq connectionResult and not the one in JooqDao that wraps the exceptions?  ctx.connectionResult(...) vs connectionResult(ctx, ...)

rma-rripken (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastInstanceDao.java
rma-bryson: Updated to use the JooqDao's connectionResult(ctx, ...)

rma-rripken (approved):

rma-rripken:
Why did the build checks not have to run after your last force push?

rma-bryson:
> Why did the build checks not have to run after your last force push?

Not sure, but looks like they ran now

PR #1627, opened 2026-03-09
Explicitly set max body size.

MikeNeilson:
may not exactly fix the problem. but I want a known value for the text test on cwbi-dev.


msweier (commented):
Looks good. Although we weren't seeing anything in the DEBUG logs for these failures. I still think it's on the WAF.

msweier (approved):

MikeNeilson:
> Looks good. Although we weren't seeing anything in the DEBUG logs for these failures. I still think it's on the WAF.

certainly possible. just want to make sure we cover all the basis. Plus the Default for Tomcat is 2MB so we'd need to increase this anyways.

PR #1630, opened 2026-03-10
Correct Api Spec locations and behavior.

MikeNeilson:
Additional set the RSS handler to use an Enum for the message queue instead of a raw string. Queue and Topic names are most definitely a controlled resource and not something willy nilly.

Graphic of RSS change:
image

Location vertical datum fix:
image


cwms-data-api/src/main/java/cwms/cda/api/enums/MessageQueue.java
```diff
@@ -0,0 +1,36 @@
+package cwms.cda.api.enums;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+@Schema(
+    name = "Message Queue",
+    description = "Desired set of status messages. Must be one of the named options."
+                + " Letter casing is ignored."
+)
+public enum MessageQueue {
```
adamkorynta, line 10: Replace `AqTable` and usage of it within `MessageDao` with this

MikeNeilson: Thanks, I didn't even think to look for an existing enum. Should've pointed out when you first did it that it should've been used that way.

adamkorynta (dismissed):

adamkorynta (approved):

PR #1634, opened 2026-03-10
CDA-74 Created ADR for timeseries csv formatting

### LINKED ISSUE #1525: Format Design Document

rma-bryson:


MikeNeilson:
Don't forget the document number as seen in @krowvin 's initial csv ADR: https://github.com/USACE/cwms-data-api/pull/1551

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,85 @@
...
+Opinions
+========
+
+Opinion 1
+---------
+
+@brysonspilman
+
+Summary:
+Since the intended use of the CSV format is for retrieval only, a customized format that follows standardize csv practices is appropriate.
```
MikeNeilson:
```suggestion
Since the intended use of the CSV format is for retrieval only, a customized format that follows standardized csv practices is appropriate.
```

MikeNeilson (dismissed):

krowvin (commented):
See if you can improve the formatting of the numbering and add more sub section headers to break things up

i.e.

`1)` vs `1.` 

Might help in reading this visually [when it is rendered](https://github.com/USACE/cwms-data-api/blob/8ad927b7bb35fa8c7f0def100b756c36545fc70e/docs/source/decisions/0008-timeseries-csv-format.rst)!

Optional fix, not a blocker for me.

rma-bryson:
Updated formatting to be more readable.

MikeNeilson (commented):
Looks reasonable to me.

Will let @krowvin give the final approval.

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+@brysonspilman
+
+Summary
+~~~~~~~
+Since the intended use of the CSV format is for retrieval only, a customized format that follows standardized csv practices is appropriate.
+
+Key points
+~~~~~~~~~~
+
+- CSV will be serialized via the jackson api
```
adamkorynta:
```suggestion
- CSV will be serialized via the jackson api for consistency with JSON and XML serialization.
```

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+- Timeseries CSV will include the date-time and value fields as well as optionally the time-series id, office, quality, units, and version-date.
+- By default, optional fields will not be included.
+- A flag to include optional fields as metadata comments at the top of the file may be added. This would be a commented line indicating the amount of metadata followed by commented metadata rows using key:value pairs. See example below.
+- Column names use kebab-case for consistency with JSON and XML.
+- Units will either be included in metadata comments or as a column. Recommend not including this in the value column-header, as column names should match DTO fields. A custom serializer could be created if units need to be included in the value-column-header.
```
adamkorynta: I can see a reasoning for not including the units at all (as in example 1) only if the user has requested explicit units. If they only request a unit system (or use the default unit system), units should be required.

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+- date-time values will be serialized as ISO-8601 strings. NOTE this differs from JSON and XML for date-time which are serialized as epoch-millis.
+- Null values are empty fields. Missing values use quality-code = 5, for consistency with JSON and XML.
+- UTF-8 encoding, comma delimiter, LF line endings, header always included.
+- One row is produced per Record.
+- Multi-retrieve never includes multiple time-series IDs.
```
adamkorynta: What is multi-retrieve?

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+- CSV will be serialized via the jackson api
+- Timeseries CSV will include the date-time and value fields as well as optionally the time-series id, office, quality, units, and version-date.
+- By default, optional fields will not be included.
+- A flag to include optional fields as metadata comments at the top of the file may be added. This would be a commented line indicating the amount of metadata followed by commented metadata rows using key:value pairs. See example below.
+- Column names use kebab-case for consistency with JSON and XML.
+- Units will either be included in metadata comments or as a column. Recommend not including this in the value column-header, as column names should match DTO fields. A custom serializer could be created if units need to be included in the value-column-header.
+- date-time values will be serialized as ISO-8601 strings. NOTE this differs from JSON and XML for date-time which are serialized as epoch-millis.
+- Null values are empty fields. Missing values use quality-code = 5, for consistency with JSON and XML.
+- UTF-8 encoding, comma delimiter, LF line endings, header always included.
+- One row is produced per Record.
```
adamkorynta: Record should be defined.

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+Key points
+~~~~~~~~~~
+
+- CSV will be serialized via the jackson api
+- Timeseries CSV will include the date-time and value fields as well as optionally the time-series id, office, quality, units, and version-date.
+- By default, optional fields will not be included.
+- A flag to include optional fields as metadata comments at the top of the file may be added. This would be a commented line indicating the amount of metadata followed by commented metadata rows using key:value pairs. See example below.
+- Column names use kebab-case for consistency with JSON and XML.
+- Units will either be included in metadata comments or as a column. Recommend not including this in the value column-header, as column names should match DTO fields. A custom serializer could be created if units need to be included in the value-column-header.
+- date-time values will be serialized as ISO-8601 strings. NOTE this differs from JSON and XML for date-time which are serialized as epoch-millis.
```
adamkorynta: should this be part of the accept header and be applied consistently to all formats?

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+      2021-06-22T00:00:00Z, 1.0
+      2021-06-23T00:00:00Z, 2.0
+      2021-06-24T00:00:00Z, 3.0
+
+3. All optionals turned on, with metadata-as-comments not turned on:
+
+   .. code-block:: text
+
+      time-series-id, office-id, date-time, value, units, version-date, quality-code
+      ALAT2.Flow-Out.Inst.1Hour.0.Rev-SWF-REGI, SWT, 2021-06-21T00:00:00Z, 0.0, ft, 2021-06-21T00:00:00Z, 1
```
adamkorynta: What is the purpose of repeating the time series id and units on every line?

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+Opinion 1
+---------
+
+@brysonspilman
+
+Summary
+~~~~~~~
+Since the intended use of the CSV format is for retrieval only, a customized format that follows standardized csv practices is appropriate.
+
+Key points
```
adamkorynta, line 24: Missing a statement that column headers will always be present, matching RFC 4180 stating they are allowed (but optional).

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+
+- CSV will be serialized via the jackson api
+- Timeseries CSV will include the date-time and value fields as well as optionally the time-series id, office, quality, units, and version-date.
+- By default, optional fields will not be included.
+- A flag to include optional fields as metadata comments at the top of the file may be added. This would be a commented line indicating the amount of metadata followed by commented metadata rows using key:value pairs. See example below.
+- Column names use kebab-case for consistency with JSON and XML.
+- Units will either be included in metadata comments or as a column. Recommend not including this in the value column-header, as column names should match DTO fields. A custom serializer could be created if units need to be included in the value-column-header.
+- date-time values will be serialized as ISO-8601 strings. NOTE this differs from JSON and XML for date-time which are serialized as epoch-millis.
+- Null values are empty fields. Missing values use quality-code = 5, for consistency with JSON and XML.
+- UTF-8 encoding, comma delimiter, LF line endings, header always included.
```
adamkorynta: worth calling out that comma-only CSV follows RFC 4180 compliance specifically as sometimes tab/pipe/semicolon are sometimes referred to as CSV.

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+~~~~~~~
+Since the intended use of the CSV format is for retrieval only, a customized format that follows standardized csv practices is appropriate.
+
+Key points
+~~~~~~~~~~
+
+- CSV will be serialized via the jackson api
+- Timeseries CSV will include the date-time and value fields as well as optionally the time-series id, office, quality, units, and version-date.
+- By default, optional fields will not be included.
+- A flag to include optional fields as metadata comments at the top of the file may be added. This would be a commented line indicating the amount of metadata followed by commented metadata rows using key:value pairs. See example below.
```
adamkorynta: Are optional fields variable? i.e. if they changed column ordering or if a new optional column were to be added, does that break the format in terms of backwards compatibility? calling API should always check the index based on the column header position (which is why column headers are always present).

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+Summary
+~~~~~~~
+Since the intended use of the CSV format is for retrieval only, a customized format that follows standardized csv practices is appropriate.
+
+Key points
+~~~~~~~~~~
+
+- CSV will be serialized via the jackson api
+- Timeseries CSV will include the date-time and value fields as well as optionally the time-series id, office, quality, units, and version-date.
+- By default, optional fields will not be included.
```
adamkorynta: optional fields should be enumerated; it is unclear if example 3 is an exhaustive list or if this is left ambiguous on purpose (meaning the schema is purposefully variable to allow for more fields at runtime).

docs/source/decisions/0008-timeseries-csv-format.rst
adamkorynta: additionally, comments are not support by RFC 4180. This needs to call out what characters are being used for comments and where we expect these comments to be supported in common CSV tooling (pandas, excel, etc)

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+
+2. All optionals turned on, with metadata-as-comments turned on:
+
+   .. code-block:: text
+
+      # metadata-count: 5
+      # time-series-id: ALAT2.Flow-Out.Inst.1Hour.0.Rev-SWF-REGI
+      # office-id: SWT
+      # version-date: 2021-06-21T00:00:00Z
+      # quality-code: 1
```
adamkorynta: Why is there a quality code in the metadata, quality is defined per value

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+      2021-06-24T00:00:00Z, 3.0
+
+2. All optionals turned on, with metadata-as-comments turned on:
+
+   .. code-block:: text
+
+      # metadata-count: 5
+      # time-series-id: ALAT2.Flow-Out.Inst.1Hour.0.Rev-SWF-REGI
+      # office-id: SWT
+      # version-date: 2021-06-21T00:00:00Z
```
adamkorynta: If this is an aggregate time series, how is version-date specified? I believe in CWMS-Vue we use `base` for the `1111-11-11 11:11` unversioned date and `aggregate` when we are grabbing the latest for every timestamp.

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,88 @@
...
+      2021-06-21T00:00:00Z, 0.0
+      2021-06-22T00:00:00Z, 1.0
+      2021-06-23T00:00:00Z, 2.0
+      2021-06-24T00:00:00Z, 3.0
+
+2. All optionals turned on, with metadata-as-comments turned on:
+
+   .. code-block:: text
+
+      # metadata-count: 5
```
adamkorynta: what is the metadata-count used for?

docs/source/decisions/0008-timeseries-csv-format.rst
adamkorynta: It should be noted that a custom serializer means that all clients would then possibly need a custom deserializer probably utilizing a consistent and document regex pattern

adamkorynta (changes requested):

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Its intended to be a way to make it easier for parsers to know how many meta-data "comment" lines there are

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: good point, maybe this becomes non-optional then and gets returned with the date-time and values. 

Do we have timeseries in practice with changing quality codes for its values ?

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: That question is the reason behind the meta-data "comments". Since those comments are not a standardized csv format, this is offered as a way to get the time-series-id without using meta-data comments. The trade-off being it repeats on every line.

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Added definition of Record

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: getAll - which is the only thing CDA supports for timeseries, but its a getAll records for this one time-series id. 

I update the terminology here to say a payload never includes multiple time-series IDs.

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Added comment

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: It was at the end of line 35, but i moved it to its own line

adamkorynta: from RFC 4180

```
3.  MIME Type Registration of text/csv
....
      The "header" parameter indicates the presence or absence of the
      header line.  Valid values are "present" or "absent".
      Implementors choosing not to use this parameter must make their
      own decisions as to whether the header line is present or absent.
```
this ADR should explicitly state whether CDA will include that in the header (`text/csv; header=present`)

docs/source/decisions/0008-timeseries-csv-format.rst
adamkorynta: isn't every comment line a metadata line?

docs/source/decisions/0008-timeseries-csv-format.rst
adamkorynta: that's not stated above

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,90 @@
...
+@brysonspilman
+
+Summary
+~~~~~~~
+Since the intended use of the CSV format is for retrieval only, a customized format that follows standardized csv practices is appropriate.
+
+Key points
+~~~~~~~~~~
+
+- CSV will be serialized via the jackson api for consistency with JSON and XML serialization.
```
MikeNeilson: Counter point, though maybe this is possible with the jackson api, we should allow streaming of the CSV data, at least eventually.

I don't know if we can do that with the retrieve_ts_f fucntion, though I believe so as it returns a cursor like, but that would go a long way into usability for bulk data sets. Not faster, but not an idle connection so doesn't time out so long as data is flowing.

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Updated to include units in value header field after discussion. Added justifications

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: I think so - that seems like that right sport for date format. Updated

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Added note and table row on why headers are always included

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: listed which fields are optional

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Updated to have quality code and data entry date be optional columns, not metadata

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Updated decision to use human-readable terms

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Right now it is. I guess I was thinking of it more as meta-data lines being a subset of commented lines. At least, this format allows it to be a subset.

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,127 @@
...
+
+   * - Topic
+     - Decision
+     - Justification
+   * - Serialization
+     - We will utilize the Jackson API
+     - Maintain consistency with JSON and XML serialization infrastructure and object-mapper settings already used across the API.
+   * - Required columns
+     - Always include ``date-time`` and ``value``; include units in the value column header as parentheses (e.g., ``value (ft)``)
+     - Units must exist in exactly one canonical location in all modes.
```
adamkorynta:
```suggestion
     - Units should exist in exactly one canonical location in all modes. Conditionally adding them as metadata comments will cause confusion over the inconsistency
```

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,127 @@
...
+   * - Optional columns
+     - Optional (off by default): ``time-series-id``, ``office-id``, ``version-date``, ``data-entry-date``, ``quality``
```
adamkorynta: I don't think `time-series-id` or `office-id` should ever be columns. If this information is needed, metadata comments should be turned on.

We had decided for JSON/XML timeseries, that there has not been a use-case for `version-date` at the row level. So far we've only needed base, a specific version, or aggregate.

Should be added somewhere that a future update could include an optional `text-annotation` column.


```suggestion
     - Optional (off by default): ``data-entry-date``, ``quality``. 
```

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,127 @@
...
+     - Everything except ``date-time`` and ``value`` (with units in the header) is optional. Because headers are always included, optional columns can be toggled without breaking parsing. Clients should rely on column names, not indices.
```
adamkorynta:
```suggestion
     - Everything except ``date-time`` and ``value`` (with units in the header) is optional. Because headers are always included, optional columns can be toggled without breaking parsing. Clients should rely on column names, not indices. Given units are in the `value` header, clients will need to handle this appropriately to determine the correct column index.
```

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,127 @@
...
+   * - Metadata fields
+     - May be emitted as top-of-payload comments (``metadata-format=comment``) or as actual columns (``metadata-format=column``)
```
adamkorynta:
```suggestion
     - The payload may include prefix comments using the mime-type property: (``metadata-format=comment``)
```

If there is a case where meta-data columns would be preferred, please reference here as that redundant data is unnecessary given any parsing API (python, javascript, hand-edited excel, java, etc) can remove comments.

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,127 @@
...
+     - The following fields can be treated as metadata comments at top-of-payload rather than columns: ``time-series-id``, ``office-id``, ``version-date``. These are optional (off by default). When included as comments, the payload starts with a line indicating count (e.g., ``# metadata-count: 3``) to aid parsing.
```
adamkorynta: `payload starts with a line indicating count (e.g., ``# metadata-count: 3``)` - there is nothing else in the comments other than metadata, how does a line count aid with parsing?

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,127 @@
...
+     - A bitmask (integer) compactly represents multiple boolean flags with fast native bitwise operations; a ``byte[]`` adds overhead without improving expressiveness for fixed flag sets.
+   * - Nulls and missing values
+     - Render nulls as empty fields; use ``quality-code = 5`` for missing values
+     - Keeps behavior consistent with JSON and XML.
+   * - Encoding and delimiters
+     - UTF-8, comma delimiter, LF line endings
+     - Comma-only CSV follows RFC 4180 compliance.
+   * - Record structure
+     - One row per record
+     - A record is a single date-time and value pair; ``quality-code`` and ``data-entry-date`` may be included as optional columns.
```
adamkorynta: The definition of record includes the version-date even if that column won't be shown.

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,127 @@
...
+     - Kebab-case names
+     - Keeps naming consistent with JSON and XML.
+   * - Accept header for format and columns
+     - Use HTTP Accept header parameters to select date format and optional columns
+     - Default CSV serialization uses ISO-8601 strings. Examples: ``text/csv;date-format=ISO8601-Instant`` (default), ``text/csv;date-format=epoch-millis``. Use Accept header parameters to enable optional columns (e.g., ``quality=present``, ``data-entry-date=present``). If these were query params instead, toggling would be easier in a browser, but Accept keeps content negotiation consistent.
+   * - Quality representation
+     - ``quality`` (aka quality-code) is an optional integer bitmask
+     - A bitmask (integer) compactly represents multiple boolean flags with fast native bitwise operations; a ``byte[]`` adds overhead without improving expressiveness for fixed flag sets.
+   * - Nulls and missing values
+     - Render nulls as empty fields; use ``quality-code = 5`` for missing values
```
adamkorynta: `quality-code = 5` description should be under `Quality representation` as it is separate from nulls

adamkorynta (changes requested):

docs/source/decisions/0008-timeseries-csv-format.rst
MikeNeilson: null is the representation of missing. For regular it's not as important but for irregular interval it can be interpreted along the lines of "at this time a measurement was attempted but things happen and we don't know, but we tried." Granted not the most common context.

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: I was thinking about other potential information being included in comments in the future - that may not necessarily be metadata fields. General comments vs metadata field comments. But, since we control the serialization, I think you are correct and we wouldn't ever be including general comments here. This would probably be more useful if we were round tripping data.

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Is the TimeSeries DTO incorrect then? the version-date field isn't in the Record DTO

MikeNeilson: Only if we allow "all versions" to be retrieved from that endpoint. Which I don't think we do, it's either latest,earliest, exact, or non-versioned.

docs/source/decisions/0008-timeseries-csv-format.rst
adamkorynta: Ah I read this section as how nulls are handling more generally, i.e. if a future optional `text-annotation` column were added, it would have an empty column rather than `null` or `-` or some other annoying representation.

```suggestion
     - Missing values will be represented with an empty value field (null) and will have ``quality-code = 5``. Constants will not be used to represent missing values.
```

docs/source/decisions/0008-timeseries-csv-format.rst
adamkorynta: A record in this case is date-time, value, quality, data entry date, and version-date. Uniqueness for this endpoint records is just the date-time.

This is worth calling out because like Mike said, we don't allow multi-date-version retrieval.

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: updated to remove metadata-count line

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: updated to include version-date in record definition

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: This makes sense to me so as to avoid repeated data. The one benefit of using columns is that follows standards (using '#' comments isn't standard). The idea here was to allow for either comments or columns. Will let @MikeNeilson weigh in

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,126 @@
...
+     - Do not include units as a separate column or in metadata comments. This avoids the anti-pattern of dual representation; units live in exactly one canonical location. Custom deserialization may be required to extract units from the header, which is preferable to duplicate representations.
+   * - Version-date encoding
+     - Use ``base`` for 1111-11-11T11:11, ``aggregate`` for aggregate versions, ISO-8601 timestamp for actual version dates, and omit the field if unversioned
+     - Matches CWMS-VUE behavior. A separate CSV column per case was rejected due to lack of use-cases and schema bloat. Note this requires custom serialization handling.
+   * - Column headers
+     - Always include headers
+     - RFC 4180 allows headers; including them keeps the format scalable if optional columns are introduced later and prevents reliance on fixed column indices. We will include a header param of ``headers=present`` in the Accept header to explicitly indicate that headers are included, even though they will always be present. This allows for future flexibility if we ever need to emit headerless CSV for some reason.
```
adamkorynta:
```suggestion
     - RFC 4180 allows headers; including them keeps the format scalable if optional columns are introduced later and prevents reliance on fixed column indices. We will include a header param of ``header=present`` in the Accept header to explicitly indicate that headers are included, even though they will always be present. This allows for future flexibility if we ever need to emit headerless CSV for some reason.
```

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,126 @@
...
+     - Default CSV serialization uses ISO-8601 strings. Examples: ``text/csv;date-format=ISO8601-Instant`` (default), ``text/csv;date-format=epoch-millis``. Use Accept header parameters to enable optional columns (e.g., ``quality=present``, ``data-entry-date=present``). If these were query params instead, toggling would be easier in a browser, but Accept keeps content negotiation consistent.
+   * - Quality representation
+     - ``quality`` (aka quality-code) is an optional integer bitmask
+     - A bitmask (integer) compactly represents multiple boolean flags with fast native bitwise operations; a ``byte[]`` adds overhead without improving expressiveness for fixed flag sets.
+   * - Nulls and missing values
+     - Missing values will be represented with an empty value field (null) and will have ``quality-code = 5``. Constants will not be used to represent missing values.
+     - Keeps behavior consistent with JSON and XML.
+   * - Encoding and delimiters
+     - UTF-8, comma delimiter, LF line endings
+     - Comma-only CSV follows RFC 4180 compliance.
```
adamkorynta:
```suggestion
     - Comma-only CSV follows RFC 4180 compliance. Tab/Pipe/semicolon delimiters will not be supported.
```

docs/source/decisions/0008-timeseries-csv-format.rst
adamkorynta: do we even need to call out the serialization library being used? it has no bearing on client consumption, whereas the rest of the document is focused on API usage.

rma-bryson: Yeah i'm going to just remove this, don't think this decision has to be made at this level

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,126 @@
...
+   * - Record structure
+     - One row per record
+     - A record is a single date-time and value pair; ``quality-code`` and ``data-entry-date`` may be included as optional columns. ``version-date`` is also an attribute of the record, but is not expected to vary within a payload and can be treated as metadata.
```
adamkorynta:
```suggestion
     - A record is a single date-time and value pair; ``quality-code`` and ``data-entry-date`` may be included as optional columns. ``version-date`` is also an attribute of the record but is covered under the optional metadata comments.
```

docs/source/decisions/0008-timeseries-csv-format.rst
```diff
@@ -0,0 +1,126 @@
...
+      2021-06-22T00:00:00Z, 1.0
+      2021-06-23T00:00:00Z, 2.0
+      2021-06-24T00:00:00Z, 3.0
+
+3. All optionals turned on, with metadata-as-comments not turned on:
+
+   .. code-block:: text
+
+      time-series-id, office-id, date-time, value (cfs), version-date, data-entry-date, quality-code
+      ALAT2.Flow-Out.Inst.1Hour.0.Rev-SWF-REGI, SWT, 2021-06-21T00:00:00Z, 0.0, aggregate, 2021-06-21T00:05:00Z, 5
```
adamkorynta: If you're proposing to add version-date to a column, the representation is then no longer `aggregate` as the term `aggregate` refers to using the latest version-date across all time steps. I don't think we should add `version-date` here unless that feature is requested.

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: removed

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: good point - this makes me more inclined to not have metadata columns at all. comments only.

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: No need to include those as columns given they can be included in metadata rows.

MikeNeilson: might be useful to set the content-disposition to download and set the filename to the timeseries name.

MikeNeilson: I just though of a valid reason to include a row level version date, data transfer (e.g. test to dev) and inspection. That said it's a rather uncommon use-case so we don't need to worry about it yet in this PR... plus it doesn't really need to be "easy" anyways given it would be a rare use-case.

krowvin: Office in column seems more useful to me if you were to to return multiple timeseries, one TS per row and the data be in the columns? (Not that we are, just trying to think when Office might go in a column)

MikeNeilson (commented):
I think this is a good start. (NOTE: I mean that in since it's complete and we can go with it, but I'm sure we'll have to change something later... e.g. normal software development.)


I did just share this again with the more focus portion of the community of practice, so will give them a day or two to review, otherwise I would've marked approved.

MikeNeilson:
@jeffsuperglide @rgoss @adamscarberry As y'all have or might work on A2W and likely don't get arbitrary notifications from CDA developement you might be interested and want to comment.

We also have https://github.com/USACE/cwms-data-api/pull/1692 getting worked on so this should eventually replace the existing A2W CSV download.

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: Decision is to not include metadata as columns

docs/source/decisions/0008-timeseries-csv-format.rst
rma-bryson: There is no case for this - will not be supporting metadata columns.

adamkorynta (approved):

PR #1635, opened 2026-03-12
1622 data query tool should have a refresh button

### LINKED ISSUE #1622: Data Query tool should have a refresh button
When you query data you select a date and a timeseries for either mode. A refresh button would allow the user to refresh a dataset for the same time window and TSID(s) without having to click a new date time or select another/the same TSIDs

Create a refresh button next to the calendar items that when clicked sends a `no-cache` header to fetch the absolute latest data of the timeseries ids selected

Setup auto refresh everytime you come back to the browser window (on focus), but there is no explicit control for this and could cause extra API calls / confusion with data reloading if you were coming back to explicitly check _that_ one data point.

**Additional context**
Someone working in CWBI Dev is using it to check their data and wants to be able to hit refresh after they load data to the DB via CDA

krowvin:
- Fixes env var naming
- Adds a refresh button
- adds new menu icon
  - Enable / Disable Cache (state / browser)
  - Set asc/desc order
  - make settings menu icon red when something is disabled (cache)
- Change table to desc by default
- Disable the auto complete default input box features to ensure they do not get in the way of the smart input items (combobox items)

Various snippets:
image
image

### DIFF
--- a/cda-gui/src/pages/data-query/index.jsx
+++ b/cda-gui/src/pages/data-query/index.jsx
@@ -23,7 +23,9 @@ const v2_config = new Configuration({
 const ts_api = new TimeSeriesApi(v2_config);
 const offices_api = new OfficesApi();
 const DATA_QUERY_CACHE_KEY = "data-query-cache-enabled";
+const DATA_QUERY_SORT_ASC_KEY = "data-query-sort-ascending";
 const DEFAULT_CACHE_ENABLED = true;
+const DEFAULT_SORT_ASCENDING = true;
 
 // const config = cwmsConfigs["SWF"];
 // async function fetchConfig(configUrl) {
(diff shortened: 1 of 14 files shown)


cda-gui/src/pages/data-query/index.jsx
```diff
@@ -21,6 +22,10 @@ const v2_config = new Configuration({
 });
 const ts_api = new TimeSeriesApi(v2_config);
 const offices_api = new OfficesApi();
+const DATA_QUERY_CACHE_KEY = "data-query-cache-enabled";
+const DATA_QUERY_SORT_ASC_KEY = "data-query-sort-ascending";
```
krowvin: Might be worth getting more creative with these key names if we ever plan to have other apps and a local store key is too generic?

MikeNeilson (approved):

MikeNeilson:
Though it looks like the Dockerfile needs a javascript/node update somewhere.

PR #1638, opened 2026-03-12
1177 absorb javascript client

### LINKED ISSUE #1177: Absorb javascript client
At this time the current javascript client is maintained independantly, this is made it difficult for the JavaScript client to keep up with changes. Additionally it is the client that finds the majority of issues with the OpenAPI Spec Output.

krowvin:
This is a rough copy from 

https://github.com/hydrologicengineeringcenter/cwms-data-api-client-javascript

Will need

- [x] Plumbed into gradle / made sure JSON swagger spec is updated to pull during the build and not from the URL
~- [ ] CI/CD Updated~
~- [ ] Access added to NPM for this repo~


cda-client/scripts/package-updates/modPackage.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+# Set date variable for use in version
+printf -v date '%(%Y.%m.%d)T'
```
MikeNeilson: We use tags for versioning, Look at the various release.yamls. So the version should come through an environment variable or argument.

MikeNeilson: then again I now see the note in the readme so maybe that's already taken care of.

cda-client/scripts/docker/buildSpec.cmd
MikeNeilson, line 1: batch scripts?

krowvin: Will remove these scripts for this repo, helped with local setup in windows

MikeNeilson (changes requested):
Change this all to use java/gradle to generate the API itself, there's an existing subproject, adjust options as required to match existing output.

Additionally, for now, remove the publish steps we'll deal with that in a follow up after we verify everything works.

The existing tests should be valid, you might have to experiment with exactly where they go and how they get referenced.

You might be able to reference https://github.com/opendcs/opendcs/tree/main/java/api-clients for some additional examples.

Also make sure that there is some form of build task, should be in the gradle, that does the equivalent of what `npm run build` would do (the OpenDCS Typescript client does this). The generators can sometimes output incorrect code and that catches that before it's used anywhere.

krowvin:
Note to self from Mike

Make sure i'm using gradle commands to do things like build the docs/examples/run the generator

Check how OpenDCS does this - build.gradle file that knows how to integrate into the gradle build. Gradle plugin should do generation

Move tests to the javascript dir with the other tests


Make sure to target current gradle version for CDA

Node plugin for gradle

Make a `typescript` directory under clients

clients/typescript/scripts/spec-updates/servers.json
```diff
@@ -0,0 +1,12 @@
+{
+  "servers": [
+    {
+      "url": "https://cwms-data.usace.army.mil/cwms-data",
+      "description": "PRIMARY production server"
+    },
+    {
+      "url": "https://water.usace.army.mil/cwms-data",
```
krowvin, line 8: This no longer works and needs to be removed

krowvin:
- Moved cwmsjs to clients/typescript from https://github.com/HydrologicEngineeringCenter/cwms-data-api-client-javascript
- Replaced old cda-client Gradle project
- Added Gradle OpenAPI client generation
- Added generated TypeScript compile task
- Added TypeDoc and examples generation
- Removed npm publish steps for follow-up, will make PR for this later
- Updated repo metadata to USACE/cwms-data-api repo
- Added npm bundler entrypoints
- Added jsDelivr/unpkg browser bundle, good for our static build of users that don't want to use react/vite/bundlers/etc (Old repo was doing this)
- Brought in develop changes

MikeNeilson (dismissed):

MikeNeilson (dismissed):
Does need the conflicts fixed though.

MikeNeilson (approved):

PR #1647, opened 2026-03-19
CDA-98 Workaround to updating forecast location for a spec

rma-bryson:
…ation id, does not overwrite the original spec, but creates a new one for the new location.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/ForecastSpecController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/ForecastSpecController.java
@@ -57,6 +57,8 @@ public final class ForecastSpecController extends BaseCrudHandler {
                 ForecastSpec existing = dao.getForecastSpec(forecastSpec.getOfficeId(), forecastSpec.getSpecId(), forecastSpec.getDesignator());
                 if(locationsAreDifferent(forecastSpec, existing)) {
                     dao.updateSpecWithLocationIdChange(forecastSpec);
+                } else {
+                    dao.create(forecastSpec);
                 }
             } catch (NotFoundException e) {
                 dao.create(forecastSpec);
--- a/cwms-data-api/src/main/java/cwms/cda/api/ForecastSpecController.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/ForecastSpecController.java
@@ -2,6 +2,7 @@ package cwms.cda.api;
 
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import cwms.cda.api.errors.NotFoundException;
 import cwms.cda.data.dao.DeleteRule;
 import cwms.cda.data.dao.ForecastSpecDao;
 import cwms.cda.data.dao.JooqDao;
@@ -52,10 +53,24 @@ public final class ForecastSpecController extends BaseCrudHandler {
             ForecastSpecDao dao = new ForecastSpecDao(dsl);
             ForecastSpec forecastSpec = deserializeForecastSpec(ctx);
 
-            dao.create(forecastSpec);
+            try {
+                ForecastSpec existing = dao.getForecastSpec(forecastSpec.getOfficeId(), forecastSpec.getSpecId(), forecastSpec.getDesignator());
+                if(locationsAreDifferent(forecastSpec, existing)) {
+                    dao.updateSpecWithLocationIdChange(forecastSpec);
+                }
+            } catch (NotFoundException e) {
+                dao.create(forecastSpec);
+            }
 
             ctx.status(HttpServletResponse.SC_CREATED);
         }
+    }
+
+    private static boolean locationsAreDifferent(ForecastSpec forecastSpec, ForecastSpec existing) {
+        String newLocationId = forecastSpec.getLocationId();
+        String existingLocationId = existing.getLocationId();
+        return (newLocationId == null && existingLocationId != null)
+                || (newLocationId != null && !newLocationId.equalsIgnoreCase(existingLocationId));
     }
 
     @OpenApi(
@@ -240,7 +255,12 @@ public final class ForecastSpecController extends BaseCrudHandler {
             ForecastSpec forecastSpec = deserializeForecastSpec(ctx);
             DSLContext dsl = getDslContext(ctx);
             ForecastSpecDao dao = new ForecastSpecDao(dsl);
-            dao.update(forecastSpec);
+            ForecastSpec existing = dao.getForecastSpec(forecastSpec.getOfficeId(), forecastSpec.getSpecId(), forecastSpec.getDesignator());
+            if(locationsAreDifferent(forecastSpec, existing)) {
+                dao.updateSpecWithLocationIdChange(forecastSpec);
+            } else {
+                dao.create(forecastSpec);
+            }
             ctx.status(HttpServletResponse.SC_OK);
         }
     }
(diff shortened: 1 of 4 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastSpecDao.java
```diff
@@ -36,19 +43,71 @@ public ForecastSpecDao(DSLContext dsl) {
...
-           CWMS_FCST_PACKAGE.call_STORE_FCST_SPEC(DSL.using(conn).configuration(), forecastSpec.getSpecId(),
-                   forecastSpec.getDesignator(), forecastSpec.getSourceEntityId(),
-                   forecastSpec.getDescription(), forecastSpec.getLocationId(),
-                   timeSeriesIds, "F", "F", forecastSpec.getOfficeId());
+           final String tsIdsFinal = timeSeriesIds;
+
+           // Use office-scoped DSL context
+           DSLContext ctx = getDslContext(conn, officeId);
+
+           ctx.transaction(configuration -> {
```
adamkorynta: update should be an explicit operation (even in a workaround). The controller should be checking if the spec already exists and then tell the dao to perform an update. The code route shouldn't be querying/deleting instances if the spec doesn't exist.

cwms-data-api/src/test/java/cwms/cda/api/ForecastSpecControllerTestIT.java
```diff
@@ -88,17 +90,64 @@ static void truncateFcstTimeSeries() throws SQLException {
...
 
     static void deleteSpec() throws SQLException {
-       try {
            CwmsDataApiSetupCallback.getDatabaseLink()
                    .connection(c -> {
-                       CWMS_FCST_PACKAGE.call_DELETE_FCST_SPEC(OracleDSL.using(c).configuration(), SPEC_ID, "designator",
-                               DeleteRule.DELETE_ALL.getRule(), OFFICE);
-                       CWMS_FCST_PACKAGE.call_DELETE_FCST_SPEC(OracleDSL.using(c).configuration(), SPEC_ID + "-NULL-DESIGNATOR", null,
-                               DeleteRule.DELETE_ALL.getRule(), OFFICE);
+                       try {
```
adamkorynta: `Map.of(<spec-id1>, <designator1>,
<spec-id2>, <designator2>,
<spec-id3>, <designator3>
...)` and then iterating reads better.

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastSpecDao.java
```diff
@@ -36,19 +43,71 @@ public ForecastSpecDao(DSLContext dsl) {
...
+               // 3) Re-store the previously removed forecast instances
+               for (ForecastInstance fi : existingInstances) {
+                   instDao.create(fi);
+               }
+           });
        });
     }
 
+    private void removeExistingForecastSpec(DSLContext ctx, String officeId, String specId, String designator) {
+        // Find any existing designators (including NULL) for this spec in this office. Probably is only one, but just to be safe.
```
adamkorynta: Your comment describes the already self-describing code, but I don't understand why you're deleting all the other specs with different designators.

rma-bryson: Removed this whole method as its not needed anymore since there can't be 2 specs with the same specId-designator.

cwms-data-api/src/test/java/cwms/cda/api/ForecastSpecControllerTestIT.java
rma-bryson: updated

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastSpecDao.java
rma-bryson: Updated to make this an explicit operation

cwms-data-api/src/test/java/cwms/cda/api/ForecastSpecControllerTestIT.java
```diff
@@ -88,19 +92,31 @@ static void truncateFcstTimeSeries() throws SQLException {
...
 
     static void deleteSpec() throws SQLException {
-       try {
            CwmsDataApiSetupCallback.getDatabaseLink()
                    .connection(c -> {
-                       CWMS_FCST_PACKAGE.call_DELETE_FCST_SPEC(OracleDSL.using(c).configuration(), SPEC_ID, "designator",
-                               DeleteRule.DELETE_ALL.getRule(), OFFICE);
-                       CWMS_FCST_PACKAGE.call_DELETE_FCST_SPEC(OracleDSL.using(c).configuration(), SPEC_ID + "-NULL-DESIGNATOR", null,
-                               DeleteRule.DELETE_ALL.getRule(), OFFICE);
+                       Map<String, Optional<String>> specs = Map.of(
```
adamkorynta: Usage of Optional is intended primarily for public API. Nulls are fine in local scope as you control where and how they are handled.

cwms-data-api/src/main/java/cwms/cda/data/dao/ForecastSpecDao.java
```diff
@@ -46,9 +47,54 @@ public void create(ForecastSpec forecastSpec) {
                    forecastSpec.getDesignator(), forecastSpec.getSourceEntityId(),
                    forecastSpec.getDescription(), forecastSpec.getLocationId(),
                    timeSeriesIds, "F", "F", forecastSpec.getOfficeId());
+
        });
     }
 
+    public void updateSpecWithLocationIdChange(ForecastSpec forecastSpec) {
```
adamkorynta, line 54: This does not only update the location id, it updates all fields given your deletes and re-store call. 

Additionally, if the location id is the only the thing that needs to be updated, and this is a workaround, direct table edits are much better so you can avoid such destructive calls.

cwms-data-api/src/main/java/cwms/cda/api/ForecastSpecController.java
```diff
@@ -52,12 +53,26 @@ public void create(@NotNull Context ctx) {
             ForecastSpecDao dao = new ForecastSpecDao(dsl);
             ForecastSpec forecastSpec = deserializeForecastSpec(ctx);
 
-            dao.create(forecastSpec);
+            try {
+                ForecastSpec existing = dao.getForecastSpec(forecastSpec.getOfficeId(), forecastSpec.getSpecId(), forecastSpec.getDesignator());
+                if(locationsAreDifferent(forecastSpec, existing)) {
```
adamkorynta, line 58: This reverts previous behavior of allowing POST to provide full updates to the spec

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/ForecastSpecController.java
rma-bryson: good catch. Fixed

cwms-data-api/src/test/java/cwms/cda/api/ForecastSpecControllerTestIT.java
rma-bryson: Can't use null values with Map.of(...)

But I made it work without using Optional

PR #1656, opened 2026-03-25
update jooq codegen to 26.02.17-RC01

adamkorynta:
- this time using Oracle 19C to generate the codegen
- This contains some bug fixes for fragile exception handling that has changed between ojdbc8 and ojdbc11
- replaced the hand-copied codegen jars with a shaded version of the previous codegen jar used explicitly for the water supply accounting UDT's


adamkorynta:
Supersedes: https://github.com/USACE/cwms-data-api/pull/1603

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -597,7 +597,7 @@ public static boolean isInvalidUnits(RuntimeException input) {
             SQLException sqlException = optional.get();
             String message = sqlException.getLocalizedMessage();
 
-            if (INVALID_UNIT.matcher(message).matches()) {
```
rma-rripken, line 600: If this is changing this from find() to matches() does the pattern still need to have the 6 line prefix and 10line suffix?

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -608,7 +608,7 @@ public static boolean isInvalidUnits(RuntimeException input) {
                 }
             }
 
-            if (CONVERSION_ERROR.matcher(message).matches()) {
```
rma-rripken, line 611: what does this change accomplish?

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
```diff
@@ -677,7 +677,7 @@ private static FieldLengthExceededException buildFieldLengthExceededException(Ru
 
         if (localizedMessage != null) {
             Matcher matcher = FIELD_LENGTH_EXCEEDED.matcher(localizedMessage);
-            if (matcher.matches()) {
+            if (matcher.find()) {
```
rma-rripken, line 680: What does this accomplish?   Is this related to the shaded jooq jar change

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
```diff
@@ -492,8 +494,15 @@ public ApiKey apiKeyForUser(DataApiPrincipal p, String keyName) {
...
-        ZonedDateTime expires = rs.getObject("expires",ZonedDateTime.class);
+
+        ZonedDateTime created = Optional.ofNullable(rs.getObject("created", Timestamp.class))
+            .map(Timestamp::toInstant)
+            .map(i -> i.atZone(ZoneOffset.UTC))
+            .orElse(null);
+        ZonedDateTime expires = Optional.ofNullable(rs.getObject("expires", Timestamp.class))
+            .map(Timestamp::toInstant)
+            .map(i -> i.atZone(ZoneOffset.UTC))
+            .orElse(null);
```
rma-rripken, line 505: Is this related to the jooq jar changes?

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyAccountingDao.java
```diff
@@ -130,23 +129,25 @@ public List<WaterSupplyAccounting> retrieveAccounting(String contractName, Water
...
             setOffice(c, projectLocation.getOfficeId());
             try {
-                return retrieveFromCodegen(units, c, contractRefT, startTimestamp, endTimestamp, timeZoneId, startInclusiveFlag, endInclusiveFlag, ascendingFlagStr, rowLimitBigInt, transferType);
+                var contractRefT = WaterSupplyUtilsShadow.toContractRef(waterUser, contractName);
+                return retrieveViaManual(units, c, contractRefT, startTimestamp, endTimestamp, timeZoneId, startInclusiveFlag, endInclusiveFlag, ascendingFlagStr, rowLimitBigInt, transferType);
             } catch (DataAccessException e){
                 if(isInvalidColumn(e)){
-                   return retrieveViaManual(units, c, contractRefT, startTimestamp, endTimestamp, timeZoneId, startInclusiveFlag, endInclusiveFlag, ascendingFlagStr, rowLimitBigInt, transferType);
+                    var contractRefT = WaterSupplyUtils.toContractRef(waterUser, contractName);
+                    return retrieveFromCodegen(units, c, contractRefT, startTimestamp, endTimestamp, timeZoneId, startInclusiveFlag, endInclusiveFlag, ascendingFlagStr, rowLimitBigInt, transferType);
```
rma-rripken: Why did you flip these?

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyUtils.java
```diff
@@ -233,32 +233,38 @@ static WAT_USR_CONTRACT_ACCT_TAB_T toWaterUserContractAcctTs(WaterSupplyAccounti
...
+    private static @NonNull WAT_USR_CONTRACT_ACCT_OBJ_T getWatUsrContractAcctObjT(WaterSupplyAccounting accounting,
+        Map.Entry<Instant, List<PumpTransfer>> entry, PumpTransfer transfer,
+        LOCATION_REF_T pumpIn, LOCATION_REF_T pumpOut, LOCATION_REF_T pumpBelow) {
         WAT_USR_CONTRACT_ACCT_OBJ_T watUsrContractAcctObjT = new WAT_USR_CONTRACT_ACCT_OBJ_T();
         WATER_USER_CONTRACT_REF_T contractRef = toContractRef(accounting.getWaterUser(),
                 accounting.getContractName());
         watUsrContractAcctObjT.setWATER_USER_CONTRACT_REF(contractRef);
         watUsrContractAcctObjT.setACCOUNTING_REMARKS(transfer.getComment());
         watUsrContractAcctObjT.setPUMP_FLOW(transfer.getFlow());
+        watUsrContractAcctObjT.setPUMP_FLOW_UNIT(transfer.getFlowUnit());
```
rma-rripken, line 253: The object had a pump_flow_unit and we weren't previously setting it?  That seems like a big oversight.  @rma-bryson

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyUtils.java
```diff
@@ -305,7 +311,7 @@ static LOC_REF_TIME_WINDOW_TAB_T toTimeWindowTabT(WaterSupplyAccounting accounti
     }
 
     static List<WaterSupplyAccounting> toWaterSupplyAccountingList(Connection c, WAT_USR_CONTRACT_ACCT_TAB_T
-            watUsrContractAcctTabT, String flowUnits) {
```
rma-rripken, line 308: How come flowUnits goes away?  I'm confused.

rma-rripken (commented):
@rma-bryson should probably weigh in.

adamkorynta:
This PR is essentially flip-flopping the work that was done in: https://github.com/USACE/cwms-data-api/pull/1539 and https://github.com/USACE/cwms-data-api/pull/1609

The hand-gen was the forwards compatible that supported flow units.
The codegen was the backwards compatible that didn't.

With the new codegen we have:
- codegen matches the previous hand-gen so it replaces it
- need a new backwards compatible so instead of writing a new handgen, use a shaded jar of the old codegen

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
adamkorynta: related to updating to ojdbc11

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
adamkorynta: related to updating to ojdbc11, updated to find to be less fragile if more changes occur in newer drivers.

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyUtils.java
rma-bryson: iirc this update was done in 2 stages. One was a cda side fix to just convert to SI, then a db update was made to add in the pump flow units. It didn't exist before. 

Things still worked because of the cda bandaid, but yeah this is the correct update now that the db supports pump flow units for this object.

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
rma-rripken: how is this less fragile?

adamkorynta: if the driver further pads the message, the check won't break. All we should care about is that the error pattern is somewhere in the exception message right?

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
rma-rripken: With ojdbc11 we can't pull ZonedDateTime out of ResultSet ?  Or else it gets the timezone wrong?

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
adamkorynta: good point, removed

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
adamkorynta: oh I missed this was handling the full line. reverted

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyUtils.java
adamkorynta: This route is now the new schema way where the units are part of the UDT as opposed to the old schema where the units were just set to what was parameterized in. The parameter was moved to the shaded flow.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: I suspect it's because we use Date instead of the appropriate "Timestamp with Time zone" column.

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyUtils.java
MikeNeilson: you're forgetting n-1 support. We still need to support the version of the database that doesn't include it. Example nat'l prod hasn't been updated, nor districts, and yes some are starting to want the newer builds and I'm trying to get better about releases.

MikeNeilson: NOTE: unless you've handled that in a different way, I've currently only looked at these discussions.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: oh wait, that's the ApiKey that should be using Timestamp with Time zone, or at least timestamp.

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyUtils.java
adamkorynta: That basically sums up this whole PR. What's in the current codebase is hand-copied codegen `n` support and codegenned `n-1` support. Now that the codegen has been updated to `n`, instead of hand-copying the `n-1` codegen, I shaded the jar. So supporting both `n` and `n-1`. 

We could also shade a codegen for `n+1` for features we want to code against that haven't made it to a release branch yet (ex. when the new level views get merged in).

For my sanity in this context:

`n` = current rc schema release
`n-1`= production schema release
`n+1` = schema:latest

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyAccountingDao.java
adamkorynta: We've reversed the support now, it first attempts the new schema, then falls back on the old.

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
adamkorynta:
```
Caused by: java.sql.SQLException: ORA-18716: {0} not in any time zone.DATE
    https://docs.oracle.com/error-help/db/ora-18716/
    		at oracle.jdbc.driver.DateAccessor.getZonedDateTime(DateAccessor.java:181)
    		at oracle.jdbc.driver.Redirector$36.redirect(Redirector.java:887)
    		at oracle.jdbc.driver.Redirector$36.redirect(Redirector.java:879)
    		at oracle.jdbc.driver.Representation.getObject(Representation.java:606)
    		at oracle.jdbc.driver.Accessor.getObject(Accessor.java:1077)
    		at oracle.jdbc.driver.OracleStatement.getObject(OracleStatement.java:7462)
```

This is the error ojdbc11 gave me: https://github.com/USACE/cwms-data-api/actions/runs/23608697782/job/68758390276#step:6:7652

adamkorynta: They are dates in the schema: https://github.com/HydrologicEngineeringCenter/cwms-database/blob/7b4bbafbe613252235f85470fd3b679162777944/schema/src/cwms/tables/at_api_keys.sql#L5

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyUtils.java
rma-rripken: Maybe the shaded packages should name the schema they were created from?  Trying to think how this might work.

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyAccountingDao.java
rma-rripken: Isn't shadow the old schema?  looks to me like your having it try the old schema first and then the current.   I'd think you'd first want to try whatever was in codegen first and then fallback to anything else.

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyUtils.java
adamkorynta: I didn't want to force us to name the schema as that is already defined in the versions catalog, updating would not only require updating the version catalog, but also the gradle shade plugin and the java source.

Thoughts on `codegen_legacy` instead of `codegen_shadow`? The other one I considered was `codegend_deprecated`

cwms-data-api/src/main/java/cwms/cda/data/dao/JooqDao.java
rma-rripken: For find() to work that way you need to update the regex pattern and remove ^ and $

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyUtils.java
rma-rripken: oh this makes more sense to me now.  The handgen classes that added pump_flow_units are now the "regular" classes.  We previously didn't set the pump_flow_units from the regular classes b/c they didn't support it before.  Now that the codegen has been updated the "regular" classes support the new field.  got it

cwms-data-api/src/main/java/cwms/cda/data/dao/watersupply/WaterSupplyAccountingDao.java
adamkorynta: oh I see, yes. I flip-flopped it back

cwms-data-api/src/main/java/cwms/cda/data/dao/AuthDao.java
MikeNeilson: ... well, that was foolish of me. But explains why you have to handle the time field that way.

rma-rripken (approved):

PR #1668, opened 2026-04-01
2026 04 01 test updates

MikeNeilson:
Mostly opened now to make the all scheme version tests faster... or at least not manual.


MikeNeilson:
... and wrong branch target. oi.

PR #1669, opened 2026-04-01
2026 04 01 test updates

MikeNeilson:
Mostly opened now to make the all scheme version tests faster... or at least not manual.


MikeNeilson:
As I'm not concerned about the latest database branch on the test and prod environments, I'm going to accept this as-is. Should the acceptance testing by the users fail for anything that wasn't already known and not solved yet... I'm just going to abandon the test and prod branch and start doing proper trunk based. Since we have the build always running for the databases we can do that a lot more safely now.

PR #1672, opened 2026-04-03
Correct query to ignore aliases.

### LINKED ISSUE #1671: TimeSeriesIdentifier end point returns error if aliases are present
**Describe the bug**
A clear and concise description of what the bug is. This should include the instance of CDA tried, the endpoints, and what you were intending to do.

**Expected behavior**
Appropriate data returned

**To Reproduce**

1. curl <url>/cwms-data/timeseries/identifier-descriptor/FTRA.Flow-In.Ave.~1Day.1Day.Best-MRBWM
2. get 500 error
3. Lookup incident id

**Logs/Incident Identifier**

75f78a21-1277-4cc6-aada-0782a1d16ed7

**If you do not have access to this, ignore this section!**

**CURL Commands**
If applicable, add the curl command from the SWAGGER UI to help reproduce your problem.  
**REDACT** any internal domains/IP addresses.

At the top of any instance of CDA you will see a version. Please provide that version.

- Version [e.g. 2026.04.03-testa, develop]

**Additional context**
The view used included aliases, query needs to include "aliased_item is null" or deal with the aliases.

MikeNeilson:
Fixes #1671 

NOTE: I'm aware I still need the test. End of day, can't figure out how to create a aliases for one of those items in 3 minutes.

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
@@ -224,22 +224,32 @@ public class TimeSeriesIdentifierDescriptorDao extends JooqDao<TimeSeriesIdentif
 
     public Optional<TimeSeriesIdentifierDescriptor> getTimeSeriesIdentifier(String office, String timeseriesId) {
         AV_CWMS_TS_ID2 view = AV_CWMS_TS_ID2.AV_CWMS_TS_ID2;
-        return connectionResult(dsl, connection -> {
-            Record5<String, String, Long, String, String>
-                result = dsl.select(view.CWMS_TS_ID, view.DB_OFFICE_ID, view.INTERVAL,
-                                    view.TIME_ZONE_ID, view.TS_ACTIVE_FLAG)
+        Table<?> innerTable = AV_CWMS_TS_ID2.AV_CWMS_TS_ID2.as("alias_table");
+            Field<String> tsId = innerTable.field("CWMS_TS_ID", String.class);
+            Field<BigDecimal> innerTsCode = innerTable.field("TS_CODE", BigDecimal.class);
+            Field<String> aliasedItem = innerTable.field("ALIASED_ITEM", String.class);
+        return dsl
+                .select(view.DB_OFFICE_ID,
+                    view.CWMS_TS_ID,
+                    view.INTERVAL_UTC_OFFSET,
+                    view.TS_ACTIVE_FLAG,
+                    view.TIME_ZONE_ID,
+                    DSL.multiset(
+                        dsl.selectDistinct(
+                            tsId
+                        ).from(innerTable)
+                            .where(view.TS_CODE.eq(innerTsCode))
+                            .and(aliasedItem.isNotNull())
+                    ).convertFrom(rs -> rs.map(r -> r.get(tsId, String.class)))
+                )
                 .from(view)
                 .where(view.CWMS_TS_ID.eq(timeseriesId)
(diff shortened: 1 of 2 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesIdentifierDescriptorDao.java
```diff
@@ -224,19 +224,32 @@ private TimeSeriesIdentifierDescriptor toDescriptorWithAliases(Record6<String, S
...
-
-            return retval;
-        });
+                .where(view.CWMS_TS_ID.eq(timeseriesId)
+                .and(view.DB_OFFICE_ID.eq(office.toUpperCase())))
+                .orderBy(view.DB_OFFICE_ID, view.CWMS_TS_ID,
+                    view.INTERVAL_UTC_OFFSET,
+                    view.TS_ACTIVE_FLAG,
+                    view.TIME_ZONE_ID)     
+                .fetchOptional(this::toDescriptorWithAliases);
```
MikeNeilson, line 252: So I copied this from the getAll handler above, and initially left it at .stream().findFirst(). Now Idea why but in that usage the underlying connection never closed so the fetchOptional was used and that error went away.

Not a huge deal, but figured I'd mention it for others to be aware of.

adamkorynta: I believe Ryan has run into this before and did an audit previously to check for that kind of usage. The jOOQ documentation is misleading: https://www.jooq.org/doc/latest/manual/sql-execution/fetching/lazy-fetching-with-streams/ - in that second example, it makes it look like you don't have to handle the resources unlike the first example.

Lukas has commented on: https://stackoverflow.com/questions/35558364/do-i-risk-a-jdbc-connection-leak-when-streaming-jooq-results-outside-a-try-with which is the reference we had looked at before to determine where the leak was coming from.

MikeNeilson: That would seem to imply the queries I copied are incorrect. And yet they seem to pass the test just fine. Checking that was part of why I set it to a repeated test.

rma-rripken: I have previously double-checked all our usages of stream wrt to jooq.  I believe the query you copied from was modified in Dec.  It may have been missed in review.  Stream() works and its super natural to use and it looks clean but to work correctly with jooq I believe it must be inside a try-with-resources block.   My understanding is that jooq's stream() call also pulls the entire result set before streaming it.  Since you need to wrap the block in the try-with-resources anyway seems likr you might as well use fetchLazy and get whatever minor performance improvements you can. 
I skimmed the jooq implementation of fetch and it looked to me like it was doing try-with-resources around fetchLazy with an internal call to collect so I'd suggest replacing any usages of the pattern of calling stream().map(...).collect(...) with .fetch(Mapper). 

I'm not sure why your findFirst wouldn't close the connection but if you wanted to find out I'd set a breakpoint right before that jooq call and then when you are at that break-point I'd add/enable a new generic breakpoint on any thrown exceptions and see what pops up.   It sounds like fetchOptional is what you want anyways so if you believe my previous arg against using stream() then fetchOptional is the right thing anyways and its not worth investigating.

MikeNeilson: Thanks. I'll just leave it for now, as all the tests pass. I'm I'm curious enough that I'll investigate later in the week so we can know for sure.

adamkorynta (approved):

PR #1676, opened 2026-04-07
Update various packages.

MikeNeilson:
Am manually testing the CDA GUI now, wanted to get the rest of the build started.

Manually handles several dependabot/security reports that weren't getting updated by the automation.


MikeNeilson:
UI still works, also corrected the behavior for the swagger-docs link, previous fix did not quite do the trick.

cda-client/package.json
```diff
@@ -35,7 +35,7 @@
     "es6-promise": "^4.2.4"
   },
   "devDependencies": {
-    "@openapitools/openapi-generator-cli": "^2.15.3",
+    "@openapitools/openapi-generator-cli": "^0.0.6",
```
krowvin: Why are we downgrading the generator? I think we might have picked the wrong version because this could introduce some unforeseen security issues going this far back?

I assumed it might be because of the old javalin/java version but wanted to make sure.

cda-gui/src/main.jsx
```diff
@@ -36,6 +37,16 @@ const router = createBrowserRouter(
...
         },
+        {
+          path: "swagger-docs",
+          // yes, this is a bit odd, but the swagger-docs are "rendered" by the api backend, not client side.
+          loader: () => {
+            const h = href(getBasePath() + "/swagger-docs");
+            globalThis.location.replace(h);
+            return null;
+          },
+          component: <div />,
```
krowvin: I think this should be element: <div /> or Component: SomeComponent. Lowercase component doesn’t look like a valid route object field here. 

Assuming we are doing this as a 'don't know don't care' because the location.replace will exit the current frontend anyways?

krowvin (changes requested):
Wondering about the version downgrade on the generator before you merge

cda-client/package.json
MikeNeilson: Huh, thanks, didn't catch that. I'm not sure what happened there, I ran `npm audit fix --force` might have had something to do with the force, but definitely going backwards there.

cda-gui/src/main.jsx
MikeNeilson: One of those doesn't really matter, just needs to exist and would get ignored.

cda-gui/src/route-paths.js
```diff
@@ -10,6 +10,17 @@ export const routePaths = [
     path: "swagger-ui",
     sitemapPath: "swagger-ui",
   },
+  {
+    id: "swagger-docs",
+    path: "swagger-docs",
```
MikeNeilson, line 15: @krowvin I had put some logging in here, and it just never gets called. Just goes right to the ErrorFallback.jsx page that's defined, and the error itself is undefined.

So far as I can tell this should be the correct way to specify this but it's not working. Perhaps the location replace defined in a handler for one of the "routeComponenents" but that seems odd. This was working before the recent inclusion of the route-paths.js and just... really isn't different so I'm a bit confused.

PR #1681, opened 2026-04-08
Remove settings of CORS heads in the options block.

MikeNeilson:


jbkolze (dismissed):
Looks reasonable to me -- my understanding is that the WAF is expected to handle any required CORS settings.

PR #1682, opened 2026-04-08
X-Trace-ID and structured logging output.

MikeNeilson:
Okay, somewhat got away from me, but it all works now.

The combination of slf4j and logback, while not required, does simplify the creation of structured logs. As well as having the same log output as OpenDCS uses in cloud environment which will simplify log searches (we aren't learning two formats.)

There is a gRPC context mechanism provided by flogger but I'm still reading up. I think I'm likely to create a custom ContextDataProvider (what Flogger needs implemented) to handle mapping to the SLF4j MDC. so still using that system, just cleaner.

Additionally, since we know we have async things, we needed to sort out how the nested contexts work on a thread pool to make sure the Trace ID passing along correctly.

For those that do not like the log formatting, one can always setup your local environment to use something you find more friendly. However, the cda-plain.log was thrown together at the last moment so suggestions welcome.

The structured output uses jsonl as the extension. At least in VS Code this validates and renders correctly as a defined JSON file... e.g. nice element coloring.


MikeNeilson:
NOTE: I'm not hard set on the whole logback thing, it was a rather obnoxious pain to get working; However, I believe there is going to be custom work no matter what as most items that use JUL don't really provide much support for things like the `MappedDiagnosticContext` concept.

cwms-data-api/src/main/java/cwms/cda/logging/TraceIdFilter.java
```diff
@@ -0,0 +1,62 @@
...
+            chain.doFilter(request, response);
+        }
+    }
+    
+    private static String validate(String id) throws IOException
+    {
+        if (UUID_MATCHER.matcher(id).matches()) {
+            return id;
+        } else {
+            throw new IOException("Trace id '" + id + "' is not a valid UUIDish value.");
```
krowvin, line 72: Throwing here will likely bubble up as a 500. This makes it look like a server issue but really it's for bad client input. 

Should we fallback to a generated ID or return a 400 instead?

cwms-data-api/src/main/java/cwms/cda/logging/TraceIdFilter.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.common.flogger.MetadataKey;
+import com.google.common.flogger.context.ScopedLoggingContexts;
+
+@WebFilter("/*")
+public class TraceIdFilter implements Filter
+{
+    public static final String CONTEXT_TRACE_ID = "traceID";
+    public static final String HEADER_TRACE_ID = "X-Trace-ID";
+
+    public static final Pattern UUID_MATCHER = Pattern.compile("[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}");
```
krowvin: This looks like it will only cover lowercase UUID?

Dockerfile
```diff
@@ -30,10 +30,25 @@ RUN mkdir /download && \
     tar xzf apache-tomcat-*tar.gz && \
     mv apache-tomcat-9.0.112 /usr/local/tomcat/ && \
     cd / && \
-    rm -rf /download && \
     rm -rf /usr/local/tomcat/webapps/* && \
     mkdir /usr/local/tomcat/webapps/ROOT && \
-    echo "<html><body>Nothing to see here</body></html>" > /usr/local/tomcat/webapps/ROOT/index.html
+    echo "<html><body>Nothing to see here</body></html>" > /usr/local/tomcat/webapps/ROOT/index.html && \
+    mkdir -p /usr/local/tomecat/conf/Catalina/localhost
```
krowvin:
```suggestion
    mkdir -p /usr/local/tomcat/conf/Catalina/localhost
```

flogger-cda-context/src/main/java/cwms/cda/logging/CdaLoggingContext.java
```diff
@@ -0,0 +1,87 @@
...
+                    MDC.put(key, String.join(",", entry.stream()
+                                                       .map(Object::toString)
+                                                       .collect(Collectors.toList())))
+                );
+            }
+
+            // Just reset back the MDC map to whatever it was before we decided to muck with it.
+            return () -> {
+                MDC.clear();
+                MDC.setContextMap(currentMdc);
```
krowvin, line 83: if currentMdc is null will it throw an unhandled error? Might need a guard clause for null

flogger-cda-context/src/main/java/cwms/cda/logging/CdaLoggingContext.java
```diff
@@ -0,0 +1,87 @@
...
+         * MDC data is ThreadLocal, so the above behavior is required to allow the context to
+         * properly propagate to threads, which are currently used in the {@see RatingMetaDataDao} (though no logging is done there)
+         *
+         */
+        @Override
+        public LoggingContextCloseable install() {
+            final var meta = this.getMetadata();
+            final var currentMdc = MDC.getCopyOfContextMap();
+
+            atCreationMdc.putAll(atCreationMdc);
```
krowvin, line 60: Is this putting the current value back into itself? 

Is that the same as doing this

myVar = 1

myVar = myVar

print(myVar)
> 1

krowvin (changes requested):
Tomecat is def a fix, the rest is up to you!

krowvin:
This would end up needing merged with #1428 - i'm happy to merge this into 1428 after you get this into main though!

cwms-data-api/src/main/java/cwms/cda/logging/TraceIdFilter.java
MikeNeilson: Good catch, spec says *should* be lower case, but to accept both.

cwms-data-api/src/main/java/cwms/cda/logging/TraceIdFilter.java
MikeNeilson: Fair point, but would have to sort out how to do that. Probably need to hijack the response... which is fine since why process any more if the request is invalid.

flogger-cda-context/src/main/java/cwms/cda/logging/CdaLoggingContext.java
MikeNeilson: ... kinda, but not.

It does look super weird, but we have to copy the current context as it is in the thread where install was called, then add all of the values this LoggingContext already has, and then put them onto the current thread.

MDC values are "ThreadLocal" hence all the magic to shift the contexts around.

flogger-cda-context/src/main/java/cwms/cda/logging/CdaLoggingContext.java
MikeNeilson: `MDC::setContextMap` handles null inputs since several version earlier than we're using.

Dockerfile
MikeNeilson: ...Smooth. Thanks.

MikeNeilson (commented):
Thanks, that Dockerfile change was probably why a specific form of override wasn't working.

flogger-cda-context/src/main/java/cwms/cda/logging/CdaLoggingContext.java
MikeNeilson: Bit more, The CdaLoggingContext instance was (possibly) created on a different thread than where install() is called. If they're on the same thread, yeah it's basically a no-op, but it's done this way to cover the case it was made for, separate threads.

MikeNeilson:
> This would end up needing merged with #1428 - i'm happy to merge this into 1428 after you get this into main though!

That is separate, that's more about the transparent proxy, not something within CDA itself.

cda-gui/package-lock.json
adamkorynta, line 1: out of curiosity, why did this get modified for this PR?

flogger-cda-context/src/main/java/cwms/cda/logging/CdaLoggingContext.java
```diff
@@ -0,0 +1,87 @@
...
+                    var key = meta.getKey(i);
+                    var val = meta.getValue(i);
+                    MDC.put(key.getLabel(), val.toString());
+                }
+            }
+
+            final var tags = this.getTags();
+            if (tags != null) {
+                tags.asMap().forEach((key, entry) ->
+                    MDC.put(key, String.join(",", entry.stream()
```
adamkorynta, line 74: Since you're already streaming, use `Collectors.joining(",")` in the collect instead of joining the end product.

flogger-cda-context/src/main/java/cwms/cda/logging/CdaLoggingContext.java
adamkorynta: this seems a little too clever. I think rewriting to:
```
            final Map<String, String> mergedMdc = new HashMap<>();
            if (currentMdc != null) {
                mergedMdc.putAll(currentMdc);
            }
            mergedMdc.putAll(atCreationMdc);
```
and then populating a `mergedMdc` in the two loops followed by 
```
  MDC.setContextMap(mergedMdc);
```
would accomplish the same thing, be less confusing to look at, and not muck with `atCreationMdc` (which by its name, shouldn't be modified).

cwms-data-api/src/main/java/cwms/cda/logging/TraceIdFilter.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.common.flogger.MetadataKey;
+import com.google.common.flogger.context.ScopedLoggingContexts;
+
+@WebFilter("/*")
+public class TraceIdFilter implements Filter
+{
+    public static final String CONTEXT_TRACE_ID = "traceID";
+    public static final String HEADER_TRACE_ID = "X-Trace-ID";
```
adamkorynta, line 23: Why "X-Trace-ID" instead of (or maybe in addition to?) `traceparent`: https://www.w3.org/TR/trace-context/

adamkorynta: would be a 16-byte hex value as standard, not a UUID

MikeNeilson: To be honest, I wasn't aware of that, and from my brief search of what people have done it never came up. Just about everyone just does a UUID with X-Trace-Id or X-Client-Trace-Id.

But I'll go do some more research, all for standards. Especially if used by default in systems like Granafa/etc for searching.

cda-gui/package-lock.json
MikeNeilson: There's a post commit hook for the javascript, so it just always gets updated. I usually do a git restore or something to exclude it and just forgot on this one.

flogger-cda-context/src/main/java/cwms/cda/logging/CdaLoggingContext.java
MikeNeilson: Thanks, wasn't trying to be clever, just dealing with the which thread has what situtation. I'll look at it again. that said MDC.getCopy... can never be null; according to the SLF4j docs.

adamkorynta: I had checked the BasicMDCAdapter implementation and it read like this:
```
    public Map<String, String> getCopyOfContextMap() {
        Map<String, String> oldMap = (Map)this.inheritableThreadLocalMap.get();
        return oldMap != null ? new HashMap(oldMap) : null;
    }
```

adamkorynta: and actually docs say it can be null: https://www.slf4j.org/api/org/slf4j/MDC.html#getCopyOfContextMap() gross

MikeNeilson: 🤦‍♂️ right, it's the setContext that can *take* null now.

cwms-data-api/src/main/java/cwms/cda/logging/TraceIdFilter.java
adamkorynta: yeah, if you bring in the opentelemetry sdk, it can generate the traceparent for you so you can get the current trace + parent trace id, so if you roundtrip the header back to the client, or in this case, the spawned thread would be the current trace and the parent trace would be built-in.

MikeNeilson: Follow up so everyone is aware. We're going to do that in addition to this. This is simple at gets us something now; however, following the traceparent standard will definitely get us further in the future. It's just going to take a bit more to get there. The standard isn't super complex but there are a lot of steps involved. 

The change to logging so we could even either is the larger portion of this PR so we'll stick with what it is for now and follow up later.

cwms-data-api/src/main/java/cwms/cda/logging/TraceIdFilter.java
MikeNeilson: though arguably we could also just silently ignore the value.

krowvin (approved):
Issue I had was resolved

MikeNeilson:
I'm going to redo this.

1. Just going to do straight to the w3c trace context
2. will split PRs,
  a. structured logging
  b. then the trace context

Will likely also use the opentelemetry helpers. Since we want to start using otel anyways, doesn't make sense to create our own context/thread wrappers if they exist already.

Issue #1625, opened 2026-03-06
Improve exception handling


rma-rripken:
**Files:** `LocationController.java:200`, `ParametersController.java:124`, `SpecifiedLevelController.java:113`, `TimeZoneController.java:124`, `UnitsController.java:118`, `ConnectionPreparingDataSource.java:26,44`

Multiple controllers are catching `Exception` broadly. This can mask unexpected errors and make debugging harder. It also risks swallowing security-relevant exceptions.

In all five cases I believe the catch blocks can be removed b/c ApiServlet is already handling the expected exceptions. 

The controller-level catches are worse than ApiServlet handlers because:
1.
They returned generic 500 for all errors (even client errors like bad format)
2.
SpecifiedLevelController leaked ex.getLocalizedMessage() to the client
3.
TimeZoneController and UnitsController returned plain text instead of JSON
4.
They logged everything at SEVERE even for client errors
5.
They didn't include the incidentIdentifier that the ApiServlet DataAccessException handler provides for tracing

The request is to remove the broad "catch (Exception e)" blocks and verify that the tests pass and that there isn't any Controller-specific handling needed.

Issue #1342, opened 2025-09-30
Cannot retrieve Elev Timeseries


rma-rripken:
**Describe the bug**
I'm working on #1102 and I was testing timeseries retrieval with Elev parameters.   I started getting an error I hadn't seen before.   I created some tests and a draft pr ( here https://github.com/USACE/cwms-data-api/pull/1341) to see if it duplicates in this environment.   The message is "ORA-14551: cannot perform a DML operation inside a query". 

The test that triggers this is here: https://github.com/USACE/cwms-data-api/blob/7c849aa6048eb851922ba44fcc3308126cae9070/cwms-data-api/src/test/java/cwms/cda/api/TimeseriesControllerTestIT.java#L1818

 I tried refactoring the timeseries query so that the vertical-datum retrieval is done in its own query and I got a different error - I don't have the exact wording but it was basically something like User doesn't have write-access. Seems weird.  Does cwms-loc.GET_VERTICAL_DATUM_INFO_F cause a write?  I wouldn't think any of this code had changed recently.

**Expected behavior**
I'd expect timeseries to be returned including the vertical-datum

**To Reproduce**

1. Query for an Elev timeseries

**Logs/Incident Identifier**
```
 TimeseriesControllerTestIT (schema: latest-dev) > TimeseriesControllerTestIT.test_get_for_elev_has_datum (schema: latest-dev) STANDARD_ERROR
2025-09-30T18:32:58.6011938Z     30-Sep-2025 18:32:58.573 WARNING [http-nio-auto-1-exec-18] cwms.cda.ApiServlet.lambda$init$20 error on request[-1402968679892546527]: /cwms-data/timeseries/
[...]

Issue #350, opened 2023-04-27
JooqDao changes removed exception wrapping


rma-rripken:
It looks like changes to JooqDao.getDslContext method removed the single exit point and added an early return that does not add the ExecuteListener.  That listener must be present if we want sql exceptions automatically wrapped as NotFound, NullArgument, AlreadyExists etc. 

ApiServlet has handlers for specific exceptions that makes sure NotFound exception returns code 404. etc.   I'm not sure if Javalin exception handlers can throw additional exceptions.  If so we might be able to move the exception wrapping code out of the jooq listener and into a javalin exception handler.  Or do both - belt and suspenders.

Failing to get those exceptions wrapped means that certain end-points started returning 500 instead of 404.  
cwms-data/timeseries?name=... for example.   The radar client code was coded to handle 404 differently than 500.  So things that internally use the client code stopped working the way they previously were working.

rma-rripken:
This line: 
https://github.com/USACE/cwms-radar-api/blob/05e6ddc2a03bc924a036d5ba941fc2e328cd76e5/cwms_radar_api/src/main/java/cwms/radar/data/dao/JooqDao.java#L45

MikeNeilson:
yeah, looks like those should always assign to retval.

PR #1683, opened 2026-04-09
1677 return stack trace for authenticated users with specific roles

### LINKED ISSUE #1677: Return stack trace for authenticated users with specific role(s)
Currently we get an identifier, click through multiple logs, and hope that the log has made it to the UI/place we can see the logs. On our shared DEV environment there can be large delays and many logs which require searching. 

This makes logging a pain-point for down stream developers that are not running on localhost/on-prem with more direct log access.

When a user is authenticated, has the CWMS Admins role, and the environment is "development" then return the traceback in the error JSON response. 

**Default to identifier ONLY response if no env set.**

BONUS: Also return SQL error # if applicable, link to the issue in grafana/cloudwatch, and the traceback. 

Only require auth, no role, and only on dev environment. 

Will this slow requests down? Possibly only the error response to do a sec table lookup of the role/auth state. Consider a bootup check for the env and remove this fetch/feature altogether if not development.

krowvin:
Add conditional stack traces for local and development debugging and keeps the existing `incidentIdentifier` behavior used for log discovery.

Stack traces are now included in `CdaError.details.stackTrace` when the request is
explicitly configured for trace output:
- request is on localhost (docker compose)
OR
- request is authenticated as a `CWMS User Admins` AND in `dev` environment (env.lower().includes('dev'))

The implementation keeps `CdaError` as a response DTO and places the request-aware policy
in `ErrorTraceSupport`.

To make that behavior apply consistently, this PR also removes several redundant
controller-level `catch (Exception)` wrappers that were bypassing `ApiServlet`, and
updates remaining endpoint `500` handlers to make their responses through
`ErrorTraceSupport`.

Partly fixes some of 
- #1625 
but not the ConnectionPreparingDataSource

Other potentially related issues include:
- #1106
- #299
- #350
- #1342 (helps with diagnosing/better traceback visibility)  

Trying to centralize general errors where possible and reduce fragmentation

Not trying to solve a new base exception direction, or a better inheritance model mainly because @MikeNeilson said to discuss more before doing that in another issue.


krowvin:
Something to note is that it returns JSON and a stack trace in JSON is not the prettiest of things with all the \n in the JSON string. 

The last commit shows an alternative where we could keep both, or pick one, of the options. Putting the traces in a list split by the newline (Could probably do \n\t too) does render nicely in the browser at least and possible in a fetch/output of a downstream client. 

List
image

Vs

String
image

MikeNeilson (changes requested):
Instead of using System.getProperty or System.getenv directly, integrate into the Togglz feature system.

But otherwise I don't hate it. Do want to chew on it some more, but overall seems reasonable.

krowvin:
Spoke with Mike. Lets go with stackTraceLines and remove the string. 

Need to fix the OpenAPI spec. Latest Schema can break if Oracle out of sync (but check). But search for FAILED all caps and first instance all caps FAILED should have your error for the other failed tests.

cwms-data-api/src/main/java/cwms/cda/api/errors/ErrorTraceSupport.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Map;
+import java.util.Optional;
+import org.togglz.core.context.FeatureContext;
+import org.togglz.core.manager.FeatureManager;
+
+public final class ErrorTraceSupport {
+    public static final String STACK_TRACE_KEY = "stackTrace";
+    public static final String STACK_TRACE_LINES_KEY = "stackTraceLines";
+    private static final String DEV_MARKER = "dev";
+    private static final Role CWMS_USER_ADMINS_ROLE = new Role("CWMS User Admins");
```
MikeNeilson: Let's assume a new role `SHOW STACK TRACE` or something.

cwms-data-api/src/main/java/cwms/cda/api/errors/ErrorTraceSupport.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    static boolean localhostRequestOverrideEnabled(Context ctx) {
+        if (ctx == null || ctx.req == null) {
+            return false;
+        }
+        return localhostRequestOverrideEnabled(ctx.req.getServerName());
+    }
+
+    static boolean localhostRequestOverrideEnabled(String serverName) {
```
MikeNeilson: ditch this and just only render if the required role is set.

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/api/errors/ErrorTraceSupport.java
```diff
@@ -0,0 +1,124 @@
...
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.togglz.core.context.FeatureContext;
+import org.togglz.core.manager.FeatureManager;
+
+public final class ErrorTraceSupport {
```
adamkorynta: I'm already not fond of the `CdaError` name because in Java `Error` is a special subclass of Throwable. I'd recommend renaming this class at least to `ExceptionTraceSupport`

cwms-data-api/src/main/java/cwms/cda/api/errors/ErrorTraceSupport.java
```diff
@@ -0,0 +1,124 @@
...
+    static boolean hasAdminRole(DataApiPrincipal principal) {
+        return principal != null
+                && principal.getRoles().contains(CWMS_USER_ADMINS_ROLE);
+    }
+
+    static boolean stackTraceFeatureEnabled() {
+        try {
+            FeatureManager featureManager = FeatureContext.getFeatureManager();
+            return featureManager.isActive(CdaFeatures.INCLUDE_ERROR_STACK_TRACES);
+        } catch (Throwable ignore) {
```
adamkorynta: Never catch `Throwable`, is there a need to catch anything?

cwms-data-api/src/main/java/cwms/cda/api/errors/ExceptionTraceSupport.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+        if (cause != null && includeStackTrace) {
+            String stackTrace = stackTraceOf(cause);
+            merged.put(STACK_TRACE_KEY, stackTrace);
+            merged.put(STACK_TRACE_LINES_KEY, stackTraceLinesOf(stackTrace));
+        }
+        return Collections.unmodifiableMap(merged);
+    }
+
+    static boolean shouldIncludeStackTrace(Context ctx) {
```
adamkorynta, line 58: I believe you package scope a lot of these methods to make them testable (love tests!), but it is making the logic hard to discern when I have to go through multiple layers to figure out exactly what set of toggles get me a stack trace. Recommend reducing the logic down to a single method if possible.

e.g. from this method it looks like if the stack trace feature is enabled, principal can be null.

Based on our earlier discussion I think we can end up just simplifying this check to:
`stackTraceFeatureEnabled() && resolvePrincipal(ctx).map(p::hasStackTraceRole).orElse(false)`

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesIdentifierDescriptorController.java
```diff
@@ -343,7 +344,7 @@ public void delete(@NotNull Context ctx, @NotNull String timeseriesId) {
             ctx.status(HttpServletResponse.SC_OK);
 
         } catch (DataAccessException ex) {
```
adamkorynta: This should get handled in ApiServlet

docker-compose.README.md
```diff
@@ -45,3 +45,20 @@ The following users and permissions are available:
...
 Traefik uses port 8081 by default, if this conflicts with existing services on your machine it can
 be changed by setting the APP_PORT variable.
+
+## Local error trace debugging
+
+The API supports conditional stack traces in JSON error responses through
+`cwms.cda.api.errors.ErrorTraceSupport`.
+
+Stack-trace exposure is controlled through the Togglz feature flag
+`INCLUDE_ERROR_STACK_TRACES`.
```
adamkorynta, line 55: recommend adding to the docker-compose.yml to turn on by default

docker-compose.README.md
```diff
@@ -45,3 +45,20 @@ The following users and permissions are available:
...
+
+For shared environments, enable that feature in the active `features.properties` and the API will
```
adamkorynta: Wouldn't this be an environment variable? not editing the features.properties which just declares the defaults?

adamkorynta (changes requested):

docker-compose.README.md
krowvin: Updated here
https://github.com/USACE/cwms-data-api/pull/1683/changes#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R75

and

https://github.com/USACE/cwms-data-api/pull/1683/changes#diff-0ca46033e95da2cb909f1242eeb7c7808d92c40b6b34bedba371f26fcb03d228R2

docker-compose.README.md
krowvin: I wondered the same but to my knowledge we do not have the vars available yet? 

@MikeNeilson [suggested we use togglz](https://github.com/USACE/cwms-data-api/pull/1683#pullrequestreview-4083630972)

Attempted it here

https://github.com/USACE/cwms-data-api/pull/1683/changes/b3593701ce16ef92afdb067c933169ab1734ebb0#diff-17e15a1833942c6fa1e1473e638b835b8b3b98c9a5534d2e591a047c3979e343R76

But has been removed. Unless togglz could somehow target the env and that's what Mike meant?

cwms-data-api/src/test/java/cwms/cda/api/errors/ExceptionTraceSupportTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+class ExceptionTraceSupportTest {
+
+    @Test
+    void includesStackTraceLinesWhenFeatureEnabledForTraceRole() {
+        DataApiPrincipal principal = new DataApiPrincipal("trace-user",
+                Set.of(new Role(ExceptionTraceSupport.SHOW_STACK_TRACE_ROLE_NAME)));
+
+        Map<String, Serializable> details = ExceptionTraceSupport.buildDetails(Map.of(),
+                new IllegalStateException("boom"),
```
MikeNeilson: "... wouldn't it be if it was: boom, boom, and *then* kaboom?"

docker-compose.README.md
MikeNeilson: The idea behind putting all of the "features" behind togglz is that it gives the primary code a single place to ask "should/can I do this" will punting off the actual acquisition of that option to something else.

For example this is an easy one to put into at at_properties, or something.

MikeNeilson: the features.properties file is just a default location.

MikeNeilson (changes requested):
Looks good. However, there is one challenge you need to do.

An integration test that actually renders a stacktrace. NOTE: you may need to be a bit contrived here, except that reality.

cwms-data-api/src/test/java/cwms/cda/api/errors/ExceptionTraceSupportTest.java
krowvin: let me hear you say wayo

docker-compose.README.md
adamkorynta: Ah I thought that worked like Spring Boot where you can have either config file, env, system property, or database flags. I mean I guess you can, but it doesn't sound like by default with togglz

cwms-data-api/src/test/resources/cwms/cda/data/sql/users.sql
```diff
@@ -82,13 +82,11 @@ begin
     cwms_sec.add_user_to_group('l2hectest_vt','CWMS PD Users', 'HQ');
     cwms_sec.add_user_to_group('l2hectest_vt','CWMS User Admins', 'HQ');
     cwms_sec.add_user_to_group('l2hectest_vt','CWMS DBA Users', 'HQ');
-    cwms_sec.add_user_to_group('l2hectest_vt','SHOW STACK TRACE', 'HQ');
```
krowvin: Removed these so the tests would pass

docker-compose.README.md
MikeNeilson: I believe that's how it works, and that right now we're just focused on file and env/properties.

cwms-data-api/src/test/java/cwms/cda/api/errors/ExceptionTraceRenderingTestIT.java
```diff
@@ -0,0 +1,138 @@
...
+import javax.servlet.http.HttpServletRequest;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.togglz.core.context.FeatureContext;
+import org.togglz.core.manager.FeatureManager;
+
+@Tag("integration")
+class ExceptionTraceRenderingTestIT {
```
MikeNeilson: This test should be of the actual API returning the stack trace appropriate.

There should be no need to Proxy HTTP response or Request. This should be derived from `DataApiTestIT` so that it runs after the CDA instance is setup.

However, I will grant that it may be a bit difficult to force something to actually output a stack trace. You should be able insert a new handler, within the test code, that does that.

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesFilteredControllerTestIT.java
```diff
@@ -24,14 +29,42 @@
...
+import org.junit.jupiter.api.Test;
+import org.togglz.core.context.FeatureContext;
+import org.togglz.core.manager.FeatureManager;
 
 @Tag("integration")
 class TimeSeriesFilteredControllerTestIT extends DataApiTestIT {
     static FluentLogger logger = FluentLogger.forEnclosingClass();
     public static final String JSON_FILE = "/cwms/cda/api/lrl/1hour.json";
 
+    static class StackTraceFeatureExtension implements BeforeEachCallback, AfterEachCallback {
```
MikeNeilson: Move this to it's own class under fixtures.

MikeNeilson (changes requested):
Generally looks good. A few nitpicks about some of the tests.

krowvin:
@MikeNeilson bump

Wasn't sure how to request re-review

MikeNeilson:
> @MikeNeilson bump
> 
> Wasn't sure how to request re-review

This works, but if you look up at the "reviewers" block, you'll notice a little spinner icon. That's the request review from button.

Needs a rebase though, sorry, probably because I had merged in some other exception work.

MikeNeilson (approved):

adamkorynta (approved):

Issue #1288, opened 2025-09-03
POR Record Timeseries Query Triggers 500 Error


msweier:
**Describe the bug**
When querying a large data set like:
```
curl -X 'GET' \  'https://CWBITEST/cwms-data/timeseries?name=Baldhill_Dam.Flow-Out.Inst.~15Minutes.0.best&office=MVP&begin=1945-09-03T16%3A11%3A00.307Z&end=2025-09-03T16%3A11%3A00.307Z' \  -H 'accept: application/json;version=2'
```
a 500 error is typically returned. Sometimes it will complete though. Similar behavior with local production CDA instance. @MikeNeilson suggested it related to the volume of data and the initial indexing query may be timing out.

**CDA Version (please complete the following information):**
 2025.09.03-test

rma-rripken:
Part of the ts query gets a total so that you know where you are in the full set of results as you page thru the data.  I wonder if that part of the query is just too slow to run across the full POR.

msweier:
I'm still getting an error on test w/ this query (some other POR timeseries will work though).

```
Failed to fetch.
Possible Reasons:

CORS
Network Failure
URL scheme must be "http" or "https" for CORS request.
```

On the local server it worked fine.

rma-psmorris:
From 2026-01-27 CDA Meeting: Still an issue, cache or job queue. Ryan R suggests 
1. adding a timer or metric around item count part of query to see if that is the culprit. If count is > 100k or some limit, then short circuit the count. Run in a thread and if ready by time needed, great otherwise don't return a count. 
2. Mike N adds: Adding in a time in the API handler to avoid returning 500 and issue a better message on the error occurring (taking too long).
3. Another alternative is to drop total from the return, this would be an API change, so limited to v2.
4. Total could also be a representative value from extents.

Pursue steps 1 and 2.

krowvin:
Possibly resolved with #1692 

Will want to confirm 500 does not exist for versioned/other TS that do not short circuit

Followup to see if anything has been discussed since Peter's comment above in January

rma-rripken:
My pr #1700 added metrics to the ts retrieve so we should get some idea of whether it times out on the count or on the data retrieve.  I also added a timer in the Controller that should give a better error message.  

When the changes make it into test perhaps someone could try and duplicate Mitch's original request and see how the changes are working.

Its very possible #1692 will resolve the issue as well.

PR #1700, opened 2026-04-20
Feature/1288 ts timer

rma-rripken:
I've added metrics to dao to try and detect where the request is timing out (in total calc or in data retrieve).   I've also added to the Controller so that it should cause a timeout if the dao is taking longer than 45 seconds.


rma-rripken:
Having the Controller call the dao method asynchronously caused Formatter classes to get loaded on a different thread which led to a bug caused by class loading.  Multiple ways to fix it but ObjectMappers are meant to be reused so one fix is just to have the Formatter classes construct and hold a static ObjectMapper.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
```diff
@@ -503,7 +466,25 @@ public void getAll(@NotNull Context ctx) {
...
+                // Execute DAO call with a timeout so we can return a clearer message instead of a generic 500
+                int apiTimeoutMs = Integer.getInteger("cwms.cda.api.apiTimeoutMs", 45000);
+                CompletableFuture<TimeSeries> daoFuture = CompletableFuture.supplyAsync(() -> dao.getTimeseries(cursor, pageSize, requestParameters));
+                TimeSeries ts;
+                try {
+                    ts = daoFuture.get(apiTimeoutMs, TimeUnit.MILLISECONDS);
+                } catch (java.util.concurrent.TimeoutException ex) {
+                    daoFuture.cancel(true);
+                    cwms.cda.api.errors.CdaError re = new cwms.cda.api.errors.CdaError("Request is taking too long; try narrowing the date range.");
+                    ctx.status(HttpServletResponse.SC_GATEWAY_TIMEOUT);
```
MikeNeilson: This should be `408 Request Timeout` or `413 Content To Large`, I'm inclined to 408 as it is an arbitrary time out we're forcing the call to deal with.

504 implies some form of proxy/gateway in the middle.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -170,21 +181,45 @@ public TimeSeriesDaoImpl(DSLContext dsl) {
     public TimeSeriesDaoImpl(DSLContext dsl, @Nullable MetricRegistry metrics) {
         super(dsl);
 
+        this.metrics = metrics;
+
         if (metrics != null) {
+            String className = this.getClass().getName();
```
MikeNeilson: why do we let this be null?

MikeNeilson (changes requested):
Overall looks good, couple of nitpick and questions.

MikeNeilson:
> but ObjectMappers are meant to be reused so one fix is just to have the Formatter classes construct and hold a static ObjectMapper

I recently merged a PR in that I think is related to that problem, there were two hashmap used for some of the format mapping and I was seeing a lot of concurrentmodificationexceptions in the cloud logs; definitely a lot more actual parallel operations in the cloud environment than the local tests.

cwms-data-api/src/main/java/cwms/cda/api/TimeSeriesController.java
rma-rripken: agreed

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-rripken: You are right, better to just pass in a valid registry

rma-rripken:
Do we have any tests that test out concurrent access?  With only one connection in the IT pool even if we jumped through hoops to simulate multiple clients execution would still end up serialized by the db connection.   The frameworks (javalin, jooq) do a pretty good job protecting us from lots of threading troubles but yeah we can still make bad choices.

MikeNeilson:
> Do we have any tests that test out concurrent access? With only one connection in the IT pool even if we jumped through hoops to simulate multiple clients execution would still end up serialized by the db connection. The frameworks (javalin, jooq) do a pretty good job protecting us from lots of threading troubles but yeah we can still make bad choices.

No, the thread pool size is very intentionally one to catch resource issues.

It's in the back of my mind though, perhaps we can just create a 2nd CDA instance that has the increased connection pool size.

Though for the current issues we've discussed I wonder if that would even be required - if we broke the stateful Dao design. the formatter problems is before we even need a connection.

But that still only covers a few cases.

MikeNeilson (approved):

rma-rripken:
#1288

Issue #1613, opened 2026-03-02
OTHER and NATIVE vertical datum support.


rma-rripken:
As discussed here:  https://github.com/USACE/cwms-data-api/issues/1611  

Vertical datum doesn't work for OTHER because it didn't seem like the database was supporting it when the code was written but it might be supported now.  NATIVE might also work but it doesn't appear to be tested.

The request is to test OTHER and NATIVE support and see if they can be easily made to work for Location, Rating and Time Series.

rma-bryson:
Draft PR that has "NATIVE" vertical-datum storage test passing. I was able to get OTHER to store by converting the string to "LOCAL" - but fails on retrieval because the test is expecting offsets to be present, but there are none. The offset table doesn't appear to have any rows for converting between the "other" datum and NAVD88/NGVD29. How is that intended to work? Does that offset conversion need to be manually added as a row from CDA on store? Should the vertical datum endpoint handle that? Seems the timeseries endpoint supports getting back the offsets, so it should be able to support returning the offsets from the OTHER to NAVD88/NGVD29 (assuming that conversion factor was provided on storage). 

Any feedback on expected behavior would be appreciated - @msweier @MikeNeilson

msweier:
Thanks for digging into this @rma-bryson. I used a chatbot to look into this, and it had some insights, but I would also check with someone familiar with the vertical datum DB design too. 

In short, yes you should have to manually add an offset from OTHER to NAVD88. So something like this:
`  
{
    "native-datum": "OTHER",
    "local-datum-name": "MSL1912",
    "unit": "ft",
    "offsets": [
  {
    "native-datum": "OTHER",
    "local-datum-name": "MSL1912",
    "unit": "ft",
    "offsets": [
      { "to-datum": "NAVD-88", "value": -0.11, "estimate": false }
    ]
  }
`
Now the DB wants to label a custom datum generically as `LOCAL` inside the database and `OTHER` outward facing. The DB should handle this (no need for CDA to do this ), so the user could store `LOCAL` or `OTHER` under native-datum (the actual name of the native datum would be stored in at_vert_datum_local):

- [`normalize_vertical_datum_in`](https://github.com/HydrologicEngineeringCenter/cwms-database/blob/main/schema/src/cwms/cwms_loc_pkg_body.sql#L258-L266): `OTHER → LOCAL`
[...]

PR #1710, opened 2026-04-24
Bugfix/cda 101 native and other vert datum support

rma-bryson:
related to https://github.com/USACE/cwms-data-api/issues/1613


rma-bryson:
IT test fails due to db not recognizing OTHER vertical datum.

rma-bryson:
> IT test fails due to db not recognizing OTHER vertical datum.

I was able to get OTHER to store, by converting it to 'LOCAL'. This works, however, when we retrieve the ts we don't get back the offsets. I'm not sure exactly how conversions between local "other" datum and NAVD-88 and NGVD-29 is intended to work.

perrymanmd (approved):

rma-bryson:
Leaving this PR open to showcase the failure in retrieving vertical datum offsets for OTHER/LOCAL (which are synonyms)

PR #1716, opened 2026-05-01
CWMSVUE-765 add support for level-refs endpoint

adamkorynta:
this removes added query time for each level's values which can get dense (especially with seasonal levels)

## Summary

Add /level-ref endpoint that retrieves only level ids and the list of effective dates. This endpoint does not retrieve constant, seasonal, timeseries, or virtual level data. This will allow for on-demand loading via CWMS-Vue row expansion

## Related Issue

https://jira.hecdev.net/browse/CWMSVUE-765

## Validation

- [x] integration testing
- [X] manual HTTP client calls

## Checklist

- [ ] AI tools used


cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevelRefs.java
```diff
@@ -0,0 +1,83 @@
...
+                this.workingLevels.nextPage = encodeCursor(cursor,
+                        this.workingLevels.pageSize,
+                        this.workingLevels.total);
+            } else {
+                this.workingLevels.nextPage = null;
+            }
+            return workingLevels;
+        }
+
+        public LocationLevelRefs.Builder add(LocationLevelRef level) {
```
rma-bryson, line 78: What is the benefit of this method instead of just building up the list and passing it into the constructor?

rma-bryson (approved):
Had one question, but approved

cwms-data-api/src/main/resources/psw4j.properties
MikeNeilson, line 1: I see I'm not the only one who thought the logging for that was... excessive.

MikeNeilson (commented):
Looks reasonable to me.

Thought I thought we had fixed this one:
```
TimeSeriesGroupControllerTestIT (schema: latest-dev) > TimeSeriesGroupControllerTestIT.test_patch_district_permission (schema: latest-dev) > [2] */* FAILED
    java.lang.AssertionError: 1 expectation failed.
    JSON path assigned-time-series.size() doesn't match.
    Expected: <0>
      Actual: <1>
```

Though I'll grant I think some work has gone into the cwms-database go that may have affected it recently.

cwms-data-api/src/main/resources/psw4j.properties
adamkorynta: Yeah.... I get why they issue the warnings, but I wish there was a better way to turn them off without adjusting the log levels.

cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevelRefs.java
adamkorynta: Yeah I could have ditched the builder, but was trying to stay consistent with all our other newer DTO's and calculate next page inside the builder. I also could have just passed into the builder constructor, but the builder pattern is meant for field mutation.

Issue #1004, opened 2024-12-31
Move to latest Javalin


MikeNeilson:
Need to coordinate with district first and discuss about timelines but javalin 4.x is not maintain and 6.x contains some features we'd like, like case insensitive path matching. Would also drastically simplify our writing of queries by being able to use the `var` data type.

MikeNeilson:
Note this more requires at least Java 17.

PR #1727, opened 2026-05-09
add compile and unit test checks for java 21

adamkorynta:
## Summary

Add Java 21 compile/unit test targets to build matrix. currently only running Java 11 integration tests due to target production environments and not wanting to overly abuse GitHub Actions.

Cannot currently add Java 25 support because there is no gradle version that supports both Java 11 and Java 25 unless setting the toolchain version to Java 11 (which defeats the whole purpose).

## Related Issue

Related to  #1004 

## Validation

Tested through CI/CD pipeline.

## Checklist

- [ ] AI tools used

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        jdk: [11]
+        jdk: [11, 21, 25]
         # place holder for now and leaving commented but in place. General scheme target tests again the current release, the next release, the previous release, and latest development.
         # at this time every possible instance of concern is on 25.07.01, once 26.02.17 is finalized it becomes release and 25.07.01 becomes previous.
         # next-release will be comment out until such time as a RC gets made
@@ -48,6 +48,7 @@ jobs:
         id: thebuild
         run: ./gradlew clean build --info --init-script init.gradle
       - name: integration tests
+        if: matrix.jdk == 11
         run: ./gradlew integrationtest --info --init-script init.gradle -PCDA.oracle.database.image=${{matrix.schema.image}}
       - name: Create matrix job badge
         if: success() || failure() # always run even if the previous step fails
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        jdk: [11, 21, 25]
+        jdk: [11, 21]
         # place holder for now and leaving commented but in place. General scheme target tests again the current release, the next release, the previous release, and latest development.
         # at this time every possible instance of concern is on 25.07.01, once 26.02.17 is finalized it becomes release and 25.07.01 becomes previous.
         # next-release will be comment out until such time as a RC gets made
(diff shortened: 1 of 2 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -342,7 +364,7 @@ protected TimeSeries getRequestedTimeSeries(String page, int pageSize, @NotNull
 
         // put all those columns together as "valid"
         CommonTableExpression<Record7<BigDecimal, String, String, String, String, BigDecimal,
-                String>> valid =
+                        String>> valid =
```
MikeNeilson, line 367: should probably just use `var` at this point.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -74,7 +74,29 @@
 import java.util.stream.Collectors;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
-import org.jooq.*;
+import org.jooq.CommonTableExpression;
+import org.jooq.Condition;
+import org.jooq.Cursor;
+import org.jooq.DSLContext;
+import org.jooq.Field;
+import org.jooq.Record;
```
krowvin, line 82: Ah I assume this is also what was keeping java 25 from building?

.github/workflows/build.yml
```diff
@@ -48,6 +48,7 @@ jobs:
         id: thebuild
         run: ./gradlew clean build --info --init-script init.gradle
       - name: integration tests
+        if: matrix.jdk == 11
```
krowvin, line 51: If i'm reading correct this is saying only do an integration test for 11? Just making sure!

If we have other toggles like this it might be worth using an env in the github repo to have a "JAVA_VERSION" and we can toggle that going forward? Might be extra! Would be good to make an issue for this if someone agrees with the deb hard coding 11 in places gives?

krowvin (commented):
Looks good, few followup thoughts. Thanks for the PR!

krowvin (approved):
Approving this, 

Was not sure if @MikeNeilson had a blocker but realized he would request changes if it was.

And it's listed as draft

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
adamkorynta: yeah pretty simple fix, didn't test integration tests though so there might be more at runtime.

.github/workflows/build.yml
adamkorynta: right now there's some sort of caching issue that's causing the builds to fail so I didn't even try to get 21 or even 25 to run integration tests.

For 25 we would need to update gradle to v9 which drops support for Java 11..... So we could use the gradle toolchain mechanism, which would require downloading a whole separate JRE in the pipeline and referencing that. That could work with a repo level env variable.

MikeNeilson:
going to close, not for lack of any value, but the upgrade issue isn't just Java 11+ it's which Jakarta EE version we can use... which at this point is none due to existing limitations.

PR #1736, opened 2026-05-18
Update tomcat to 9.0.118

ktarbet:


gradle/libs.versions.toml
```diff
@@ -11,7 +11,7 @@ google-findbugs = "3.0.2"
 error_prone_annotations = "2.15.0"
 cwms-ratings = "4.2.3"
 javalin = "4.6.8"
-tomcat = "9.0.115"
+tomcat = "9.0.118"
```
MikeNeilson, line 14: This should also be changed in the docker file. Unfortunately I haven't found a good way to keep those in sync automatically yet.

MikeNeilson (changes requested):

MikeNeilson (approved):

PR #1744, opened 2026-05-22
Correct role specification.

MikeNeilson:
(cherry picked from commit b03ee09e6b4735753b912a38873d1cc020d45bc0)

## Summary

Provide requireRoles list to gate change POST/DELETE to prevent unauthenticated requests from getting any further than that check.

## Related Issue

Closes #

## Validation

Manually with Curl.

before: error was could not deserialize (provided `{}` as the data)
after: error is "credentials not provided"

## Checklist

- [ ] AI tools used

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
+++ b/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
@@ -616,11 +616,11 @@ public class ApiServlet extends HttpServlet {
                                           Controllers.PROJECT_ID, NAME);
         String virtualOutletCreatePath = "/projects/virtual-outlets";
         cdaCrudCache(outletPath, new OutletController(metrics), requiredRoles, 1, TimeUnit.DAYS);
-        post(gateChangeCreatePath, new GateChangeCreateController(metrics));
+        post(gateChangeCreatePath, new GateChangeCreateController(metrics), requiredRoles);
         get(gateChangePath, new GateChangeGetAllController(metrics));
-        delete(gateChangePath, new GateChangeDeleteController(metrics));
+        delete(gateChangePath, new GateChangeDeleteController(metrics), requiredRoles);
         cdaCrudCache(virtualOutletPath, new VirtualOutletController(metrics), requiredRoles, 1, TimeUnit.DAYS);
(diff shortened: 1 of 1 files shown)


krowvin (approved):
No concerns!

rma-rripken:
Some possible avenues to prevent this from being possible in the future.   
1) Wrap the object returned from the createStandalone() call with a proxy that doesn't allow delete/put/patch/post from being registered without roles or with empty roles and then call ApiBuilder.setStaticInstance() with the wrapped object. 
2) Change AuthDao isAuthorized to check if a request is put/patch/delete/post and fail unless roles are provided.  

We will have to implement workarounds or allow-lists and someone could always incorrectly add a controller to the workaround or allow but I feel like that is going to be easier to catch (someone doing the wrong thing) vs something being forgotten (roles accidentally not provided).  

In fact, maybe the way to implement the Rate end-point exception is to add a new marker role, something long and obvious like POST_IS_ALLOWED_FOR_ANYONE_AT_THIS_URL and AuthDao could allow POST requests when that was the only role present.   I feel like reviewers are more likely to notice something obnoxious being used in a PR. 

It isn't as obvious to me how to go about verifying that every controller delete fails unless auth is provided.   There might be a way to do something like the OpenApi introspection tests but for some reason it feels harder.

MikeNeilson:
> In fact, maybe the way to implement the Rate end-point exception is to add a new marker role, something long and obvious like POST_IS_ALLOWED_FOR_ANYONE_AT_THIS_URL and AuthDao could allow POST requests when that was the only role present. I feel like reviewers are more likely to notice something obnoxious being used in a PR.

>> ... POST_IS_ALLOWED_FOR_ANYONE_AT_THIS_URL

As much as I'm a fan of verbosity in names, I would think `GUEST` or `Anonymous` would suffice. 

The main issue I see is that there would be *some* confusion. We once accidentally had everything have the lock icon and there were several group that though they needed to get permissions.

There was a suggestion of a simple wrapper to the calls of the javalin API, similar to how we have the cdaCrudCache such that it would expect roles by default and need to explicitly have a lack of roles defined.

Perhaps something as similar as a constant array of just the "GUEST" role that could be passed to those few methods. Anything else just seems like it's asking for trouble.

PR #1692, opened 2026-04-14
Perf: 1Million Challenge for CDA TS GET

krowvin:
Work on #1685

# Improve unfiltered /timeseries read performance + perf coverage

- move the main unfiltered `/timeseries` read path off the Oracle `retrieve_ts_out_tab` hot path and onto direct `AV_TSV_DQU`/jOOQ reads while preserving `retrieve_ts` response semantics in Java
- keep the direct path aligned with Oracle behavior using integration parity coverage, including a DST-window regression case for regular series
- Java/Gradle benchmark harness
- add explicit fixture teardown for the benchmark runner so local perf runs exit cleanly and leave the workspace in a predictable state

## Benchmark

Local single-request benchmarks against the Docker/Testcontainers Oracle dev environment, using `page-size` equal to the requested row count:

- legacy path, `250000` points: `40.679s`
- new direct path, `250000` points: `4.652501s`
- legacy path, `500000` points: `HTTP 500`

The current Java benchmark harness also confirms the direct path now returns the expected count for the large seeded series:

- direct path, `250000` points: `250000` reported rows, correct final timestamp, `4.652501s`

The benchmark task writes JSON artifacts under `load_data/performance/results/`.

## Validation

- `./gradlew :cwms-data-api:integrationTests --tests cwms.cda.api.TimeSeriesDirectReadParityIT`
- `./gradlew :cwms-data-api:timeseriesReadBenchmark -Pbenchmark.pointCount=250000 -Pbenchmark.runs=1 -Pbenchmark.forceReseed=true`

## Notes

- scope here is the main unfiltered `/timeseries` read path; `/timeseries/filtered` is still on the legacy path
- could move to larger streaming/lower-materialization for more improvements

Seeding 250,000 TS values for SPK and reading them back via CDA returns in `4.6s`

Before changes
```
Current 500k+ Fails
Current 250k in 40.679s
```

Changes
```
250k in 4.793s
500k in 7.498s
1,000,000 in about 14.9s
```

# Tool Usage

- [X] AI Tools used

PR #1757, opened 2026-06-01
Bugfix/session office not set

### LINKED ISSUE #1753: Intermitant 500 error reading timeseries when threading returns ORA-20047 / "Database Error" or "System Error"
**Describe the bug**
When using the `GET /timeseries` with cwms-python (which spawns 20 threads by default), I get intermittent 500 errors on dev server (running  2026.05.11-develop ) returning `"Database Error"` or `"System Error"`. When looking at the CDA logs, I see `ORA-20047: SESSION_OFFICE_ID_NOT_SET: Session office id is not set by the application`, raised from `CWMS_UTIL.GET_DB_OFFICE_ID`. The store will succeed on retry, so it has something to do with the frequent requests.

**Expected behavior**
I would think the office would always be assigned to the request.

**To Reproduce**
1. Use cwms-python to read period of record for `Baldhill_Dam.Flow-Out.Inst.~15Minutes.0.best` (e.g. `cwms.get_timeseries(ts_id = 'Baldhill_Dam.Flow-Out.Inst.~15Minutes.0.best', office_id=MVP, begin= datetime.datetime(1950,1,1))`) on dev server or issue concurrent requests to any office-scoped endpoint (e.g. 20 parallel `GET /timeseries?...&office=<OFFICE>` requests, chunked by time window).
2. Observe that a fraction of responses come back as HTTP 500 with body `"Database Error"` or `"System Error"`, while the rest succeed.

**Logs/Incident Identifier**
### The Error Chain
Server log shows the following sequence:

1. **ORA-01422**: `exact fetch returns more than requested number of rows`
   - Occurs inside `CWMS_UTIL.GET_DB_OFFICE_ID` when trying to resolve the office code
   - This is **detection, not the root cause** — without session office context, the office lookup query returns multiple rows instead of exactly one
   
2. **ORA-20047**: `SESSION_OFFICE_ID_NOT_SET: Session office id is not set by the application`
   - The PL/SQL catches ORA-01422, recognizes it means session context is broken, and re-raises with a clear error message
   - Stack trace shows: `CWMS_UTIL.GET_DB_OFFICE_ID` → `CWMS_TS.*` or other package calls

3. **Client sees**: HTTP 500 with body `"Database Error"` (for main query failures) or `"System Error"` (for async count query failures)
   - The ORA code is wrapped by jOOQ `DataAccessException` and caught by `ApiServlet`'s generic handler, which hard-codes the response body
   - Clients cannot distinguish this from other database errors
### Example Log Sequence
```
ORA-01422: exact fetch returns more than requested number of rows
ORA-06512: at "CWMS_20.CWMS_UTIL", line 599
ORA-20047: SESSION_OFFICE_ID_NOT_SET: Session office id is not set by the application
ORA-06512: at "CWMS_20.CWMS_ERR", line 80
ORA-06512: at "CWMS_20.CWMS_UTIL", line 605
```
`2026.05.11-develop`

**Additional context**
I didn't see these errors in test or prod so I'm wondering if the new timeseries query may expose this a bit more?

AI suggests the reason is the request-time DSL builder never installs `SessionOfficePreparer` on pooled connections, but I'm wondering why we haven't seen this before. It's been that way for a long time.

msweier:
## Summary

This is an AI assisted PR to help get the ball rolling on this bug fix.

Adds `SessionOfficePreparer` to the connection preparer chain in `JooqDao.getDslContext(Context)` so that every pooled connection gets its Oracle session office set on checkout.

I've seen intermittant 500 errors on develop CDA that traces to (`ORA-20047: SESSION_OFFICE_ID_NOT_SET`) on authenticated requests to the timeseries endpoint when multithreading a POR timeseries pull, but AI suggests this applies to all office scoped endpoints.

This may have been latent since 2022 but exposed with PR #1692 made it visible by switching timeseries reads from `cwms_ts.retrieve_ts_out_tab` (which handles office context internally) to a direct SQL path that calls `CWMS_UTIL.CONVERT_UNITS` inline. Non-authenticated requests ppear unaffected because `AuthDao.prepareGuestContext()` already wires in a `SessionOfficePreparer`; the authenticated path via `prepareContextWithUser()` doesn't.

## Related Issue

Closes #1753

## Validation

This is difficult to test.  The failure depends on connection pool state, which connections are warm, what office they were last used for, and whether the pool has recycled them. AI couldn't reliably reproduce it in a single-threaded integration test, and even under concurrent load it only fails a fraction of the time. I am able to reporduce using a cwms-python POR query in develop, but I couldn't validate on a local instance. 

The new `JooqDaoTestIT` does what it can: it verifies `SessionOfficePreparer` against the Oracle instance by checking that the session variable gets set on a cold connection and correctly overridden when switching offices, but it can't replicate the race conditions that causes the error. The tests also pass on the current `develop` branch since I couldn't replicate the error condition in the test enviorment.

If anyone has ideas for a better way to exercise the pool-level behavior, I'm all ears.

## Checklist

- [x] AI tools used - Claude Opus 4.6


msweier:
I updated the ClobControllerTest because it is the only mock setup that is building the request by hand and it needed to match the new context. The integration tests didn't needed because it is testing for real.

rma-rripken:
Seems worth trying.  If we're always setting the office here should SessionOfficePreparer get removed from the AuthDao?

adamkorynta (dismissed):

msweier:
I can still replicate this in dev. @MikeNeilson if you review and merge, I can test in dev to see if this works in the real world.

krowvin:
- https://github.com/USACE/cwms-data-api/pull/1757/commits/3314407a19a9e11186cde8ef2a075812b349e988

Should fix the tests that were failing before. Remaining failed tests look to be unrelated.

MikeNeilson (approved):

MikeNeilson:
Failing tests are existing and will be addresses in a separate PR>

PR #1772, opened 2026-06-09
Add signed batch machine run context / ADR

krowvin:
## Summary

This is the parent PR for the CWMS Batch Events M2M authentication work. Initial runtime execution and district scheduling now have separate PRs that can ship before this M2M rollout.

CDA now supports the preferred production shape when Keycloak can mint the batch machine context directly into the normal access token. CDA consumes validated OIDC claims such as `machine_auth` and `run_as_office`, while still rejecting unregistered machine principals instead of auto-creating them.

The signed `X-CWMS-Job-Context` path remains in the design as a fallback for cases where Keycloak cannot provide the needed machine-run context without a custom extension. The current rollout avoids a Keycloak SPI by using office-scoped scheduler and runner service accounts.

## Features extracted ahead of M2M

- Runtime: USACE-WaterManagement/cwms-batch-events#203, supported by USACE/cwbi-wm-images#61.
- Scheduling/timezones: USACE-WaterManagement/cwms-batch-events#204, supported by USACE/airflow-config#438.

Runtime uses the existing combined image and office job definitions and does not depend on Airflow or new Keycloak clients. The initial scheduler uses an authorized CDA API key stored in Airflow; Keycloak clients remain a prerequisite for the later M2M paths below. These extractions have their own local test evidence. The existing diagrams describe the later M2M topology.

## Related Draft PRs

| Area | PR | Notes |
| --- | --- | --- |
| Batch Events registry/API/UI/dispatcher | USACE-WaterManagement/cwms-batch-events#184 | Remaining M2M broker, shared dispatch, resource profiles/timeouts, repository browser, and authorization work; initial runtime/schedules extracted separately. |
| AWS Batch CDK infra | cwbi-dev-infrastructure/cwms-batch#20 | Shared runtime job definitions, Batch Events app secret wiring, runtime token key exposure, and dispatcher/API permissions. |
| Shared runner image | USACE/cwbi-wm-images#51 | Brokered env and office-runner CDA bearer-token bootstrap; installed-command startup extracted to USACE/cwbi-wm-images#61. |
| cwms-python client | HydrologicEngineeringCenter/cwms-python#295 | Supports `CDA_BEARER_TOKEN` and `BATCH_JOB_CONTEXT_TOKEN`. |
| cwms-cli client | HydrologicEngineeringCenter/cwms-cli#226 | Routes CLI CDA sessions through batch-aware auth helper. |
| Airflow DAG config | USACE/airflow-config#429 | Later Keycloak scheduler-client authentication; initial registry/timezone scheduler extracted to USACE/airflow-config#438. |
| Airflow infra secrets | cwbi-dev-infrastructure/airflow#6 | Exposes Batch Events scheduler auth variables, optional per-office scheduler client map, and scheduled office list through Airflow app secrets. |
| SWT proof jobs | USACE-WaterManagement/swt-wm-cwbi-jobs#23 | Office proof branch for bearer-token batch runtime. |

### System Overview

CWMS Batch M2M overview

Editable source: [`batch-m2m-overview.drawio`](https://github.com/USACE/cwms-data-api/blob/batch-dynamic-runtime/cda-job-context/docs/source/decisions/diagrams/batch-m2m-overview.drawio)

### End User UI Flow

End user Batch Events UI flow

Editable source: [`batch-ui-job-flow.drawio`](https://github.com/USACE/cwms-data-api/blob/batch-dynamic-runtime/cda-job-context/docs/source/decisions/diagrams/batch-ui-job-flow.drawio)

### Airflow Scheduled Flow

Airflow scheduled Batch Events flow

Editable source: [`batch-airflow-scheduler-flow.drawio`](https://github.com/USACE/cwms-data-api/blob/batch-dynamic-runtime/cda-job-context/docs/source/decisions/diagrams/batch-airflow-scheduler-flow.drawio)

## Validation

CDA:
- `JAVA_HOME=C:\Program Files\Java\jdk-21` `./gradlew.bat :cwms-data-api:test --tests cwms.cda.security.BatchJobContextTest --no-daemon --stacktrace`
- `JAVA_HOME=C:\Program Files\Java\jdk-21` `./gradlew.bat :cwms-data-api:integrationTests --tests cwms.cda.api.auth.OpenIdConnectTestIT --no-daemon --stacktrace`
- The integration test proves Keycloak service-account `client_credentials` tokens can carry `machine_auth`/`run_as_office`, CDA rejects unregistered machine principals, and CDA accepts registered office-scoped machine principals.

Cross-repo/local E2E evidence:
- Batch Events local UI created and ran jobs using the new registry model.
- A command-mode timeout job with `timeoutMinutes=1` and a 90-second sleep failed locally with `Local executor timeout after 60 seconds`.
- Airflow helper tests verify registry timezone evaluation, DST gap skip, and repeated local occurrence run-once behavior.
- Batch Events UI build passed after Scripts Manager source/command/path browser/timezone changes.

## Checklist

- [x] AI tools used


cwms-data-api/src/main/java/cwms/cda/security/OpenIdConnectIdentitityProvider.java
```diff
@@ -105,12 +105,16 @@ private DataApiPrincipal getUserFromToken(Context ctx) throws CwmsAuthException
...
+                DataApiPrincipal dataApiPrincipal = principal.get();
+                BatchJobContext.prepareContext(ctx, dataApiPrincipal);
+                return dataApiPrincipal;
             } else if (CREATE_USERS) {
                 final String preferredUserName = claims.get(PREFERRED_USERNAME_CLAIM, String.class);
                 final String givenName = claims.get(GIVEN_NAME_CLAIM, String.class);
                 final String email = claims.get(EMAIL_CLAIM, String.class);
-                return dao.createUser(preferredUserName, oidcPrincipal, givenName, email);
+                DataApiPrincipal dataApiPrincipal = dao.createUser(preferredUserName, oidcPrincipal, givenName, email);
+                BatchJobContext.prepareContext(ctx, dataApiPrincipal);
```
MikeNeilson: If the "batch" user is getting created randomly here, we have an issue, In the context of a batch process this should be a failure.

cwms-data-api/src/main/java/cwms/cda/security/BatchJobContext.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final FluentLogger logger = FluentLogger.forEnclosingClass();
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    public static final String HEADER = "X-CWMS-Job-Context";
+    public static final String RUN_AS_OFFICE_ATTR = "BatchRunAsOffice";
+    public static final String JOB_ID_ATTR = "BatchJobId";
+    public static final String REQUESTED_BY_ATTR = "BatchRequestedBy";
+    public static final String DISPATCH_SOURCE_ATTR = "BatchDispatchSource";
+
+    public static final String SECRET_PROPERTY = "cwms.dataapi.batch.jobContext.secret";
```
MikeNeilson, line 42: why is Batch becoming an issuer of secrets?

My understanding is that Keycloak would provide the JWT and CDA would just consume it.

cwms-data-api/src/main/java/cwms/cda/security/BatchJobContext.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final String DEFAULT_MACHINE_USERS = "";
+
+    private BatchJobContext() {
+    }
+
+    public static boolean isBatchMachineUser(String username) {
+        if (username == null) {
+            return false;
+        }
+        String machineUsers = readSetting(MACHINE_USERS_PROPERTY, DEFAULT_MACHINE_USERS);
```
MikeNeilson, line 62: CDA should not need to know anything about this, the JWT provided by Keycloak can embed a claim of "machine-auth" or something and decisions made from that.

cwms-data-api/src/main/java/cwms/cda/security/BatchJobContext.java
```diff
@@ -0,0 +1,165 @@
...
+            Claims claims = parse(token);
+            String office = claims.get("run_as_office", String.class);
+            if (office == null || office.isBlank()) {
+                office = claims.get("office", String.class);
+            }
+            if (office == null || office.isBlank()) {
+                throw new CwmsAuthException("Batch job context missing run_as_office",
+                    HttpServletResponse.SC_UNAUTHORIZED);
+            }
+            ctx.attribute(RUN_AS_OFFICE_ATTR, office.toUpperCase(Locale.ROOT));
```
MikeNeilson: While appropriate to put in to a future logging context... these shouldn't need to be part of the Request/Response attributes. Downstream components needing to know that is a definite code smell.

docs/source/decisions/0009-batch-machine-run-context.md
```diff
@@ -0,0 +1,198 @@
...
+| **Author(s)**  | CWBI Batch Runtime Team |
+| **Sponsor**    | HEC/USACE               |
+| **Date**       | 6/8/2026                |
+| **Supersedes** | N/A                     |
+
+## Objective
+
+Provide CWMS Data API with a trusted batch run context for jobs that execute through a shared machine identity.
+
+Batch runtimes will authenticate to CDA with a service account (via Keycloak). Each job will also provide a signed context token that identifies the authorized job launch context, including the office for which the scheduler or API approved the run.
```
MikeNeilson: Setting up additional signing is going to be difficult to get right, and will involve yet-more CDK changes. Granted they won't be difficult.

I would like us to first determine if we can, in some way, setup keycloak to be able to receive the required information (such as the office, and specific job identification) and return it back in the already signed JWT access token.

Some of the other information can, and should, still be provided but it doesn't really required singing it's just informational. The office would be used to set the session context (or by the future authorization system) to appropriately limit operations.

e.g. office + job identification can be readily tied to a policy to appropriately limit operations..

MikeNeilson (commented):
I'm being a bit pushy here. *AND* I may in fact be wrong about what we can accomplish with Keycloak and thus may end up having to go this route (there's nothing fundmentally wrong with the design... other that too much reliance on properties, though that's an initial design problem not a new problem) but I think if we can get Keycloak to handle the load the transition is a lot simpler.

cwms-data-api/src/main/java/cwms/cda/security/OpenIdConnectIdentitityProvider.java
krowvin: Corrected in bc40f8b7

krowvin:
CDA now takes a validated OIDC claims  `machine_auth` and `run_as_office`

Stoped auto creating machine principle 

After checking the Keycloak paths, I do not think we can get dynamic per-job values like `run_as_office` and `job_id` into the access token using normal realm/client configuration alone. Built in protocol mappers work for static claims, service account/user attributes, roles, audiences, etc but they do not give us a clean supported way for Batch to send trusted per job context during `client_credentials` and have Keycloak validate and mint it into the access token.


Keycloak has a way to do this via an SPI/provider, such as a custom protocol mapper or script mapper deployed into Keycloak. Keycloak SPI docs  https://www.keycloak.org/docs/latest/server_development/index.html#_providers


I do not think CWBI is likely to accept that operationally. If they did, we would need an owner and process for packaging, deploying, versioning, and maintaining that custom provider inside their Keycloak infrastructure. 


Using a signed `X-CWMS-Job-Context` looks to be the way to go when Keycloak cannot create dynamic job context for us. CDA validates the normal Keycloak token for the machine principal, then separately validates the dispatcher signed job context for the run office.

I'm working right now to verify all this works completely on localhost and make sure it works as expected

cwms-data-api/src/main/java/cwms/cda/security/BatchJobContext.java
krowvin: Initially when I did this I was trying to figure out how to make this work with only one service account in keycloak, I now have it where we can have a service account per office 

Now Batch is not the issuer of that context in the preferred path. Keycloak mints the normal access token for a per-office service account.

cwms-data-api/src/main/java/cwms/cda/security/BatchJobContext.java
krowvin: CDA now consumes this from the validated Keycloak/OIDC access token via claims like `machine_auth=true` and `run_as_office=<office>`, rather than relying on Batch to issue a separate signed context.

The remaining CDA-side knowledge is just the claim contract and normal user/principal lookup. It still rejects unregistered machine principals instead of auto-creating them. Locally I verified this with per-office Keycloak service accounts that mint the claims directly into the access token.

docs/source/decisions/0009-batch-machine-run-context.md
krowvin: Addressed this some here

https://github.com/USACE/cwms-data-api/pull/1772#issuecomment-4805595804

docs/source/decisions/0010-batch-machine-run-context.md
```diff
@@ -0,0 +1,276 @@
...
+completion; Batch Events owns dispatch, status, log lookup, and runtime broker
+behavior after the job is accepted.
+
+### Keycloak-Minted Run Context
+
+When available, Keycloak mints the batch run context into the normal access token. CDA validates that token through the existing OIDC flow and reads these claims:
+
+| Claim           | Description                                  |
+| --------------- | -------------------------------------------- |
+| `machine_auth`  | Marks the access token as a batch machine run |
```
MikeNeilson: Based on conversations today it was determined this was not pratical.

The machine_auth claim is fixed, so that one shouldn't be difficult to have provided.

docs/source/decisions/0010-batch-machine-run-context.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Alternatives Considered
+
+### Per-Office Keycloak Service Accounts
+
+Create one scheduler service account and one runner service account per office or trust boundary.
+
+- **Pros**: Office context is represented directly by the service account.
+- **Cons**: Recreates service-account and secret sprawl as offices and runtimes grow.
+- **Selected for current rollout**: This avoids Keycloak custom extensions while still allowing AWS Batch job definitions and runner images to be shared across offices.
```
MikeNeilson: Statement needs updating.

docs/source/decisions/0010-batch-machine-run-context.md
```diff
@@ -0,0 +1,276 @@
...
+
+Normal CDA route authorization and CWMS database permissions determine whether the machine user can act on the requested resource office.
+
+### Configuration
+
+The Java API is configured with system properties or environment variables:
+
+| Setting                                  | Description                                                    |
+| ---------------------------------------- | -------------------------------------------------------------- |
+| `cwms.dataapi.batch.jobContext.secret`   | Signing secret for validating HS256 job context tokens         |
```
MikeNeilson: would be better to do async and use a public key here, with the secret key over in cwms-batch secrets.... or even better some sort of AWS provided thing (might cost more though.)

cwms-data-api/src/main/java/cwms/cda/security/BatchJobContext.java
MikeNeilson: For the record, it was decided that the context is still required.

MikeNeilson (changes requested):
Based on recent decussions, Needs to be updated.

Will leave it to you if you want to leave stale information place as reference, but ADR should reflect the current state of understanding.

PR #1774, opened 2026-06-09
Add ADR doc for JMS Message Structure

perrymanmd:
## Summary

Add ADR doc for JMS Message Structure

## Related Issue

Closes # <None>

## Validation

No code changed

## Checklist

- [ ] AI tools used


krowvin:
Is it possible to add a nested structure to the queues? 

I'm thinking in terms of topics vs queues

For example

`timeseries/stored`
or
`timeseries/renamed`
or - subscribe to all with
`timeseries`

This would let you subscribe to a topic containing multiple queues. Bringing this up now if we are deciding on a standard for naming of the queues.

MikeNeilson:
> Is it possible to add a nested structure to the queues?
> 
> I'm thinking in terms of topics vs queues
> 
> For example
> 
> `timeseries/stored` or `timeseries/renamed` or - subscribe to all with `timeseries`
> 
> This would let you subscribe to a topic containing multiple queues. Bringing this up now if we are deciding on a standard for naming of the queues.

This ADR is documenting current state only. Follow up ADRs will address future semantics.

MikeNeilson (changes requested):
Formatting looks fine to me, so does the content, seeing only a few responses were made we can call this sufficient.

However, 9 is the wrong number, please use 11 and add it to the index.rst in that folder.

(NOTE: I took the liberty of rebasing your PR, remember to pull changes before updating those.

MikeNeilson (changes requested):
Change number to 10. (well, or whatever the next digit is by the time you get to updating it.) We already have a 9, and there's two other ADRs in the pipeline.

MikeNeilson (changes requested):
You may need to update your clone's develop branch. There's a bunch of commits that you didn't create and I'm concerned even squashing may cause issues.

Please clean up the commit history to something that looks reasonable.

MikeNeilson:
still  showing 281 files changed.

MikeNeilson:
This was merged in as part of https://github.com/USACE/cwms-data-api/pull/1830

Issue #1118, opened 2025-06-04
Swagger Lock Icons


krowvin:
All of the swagger endpoints in the UI are showing a lock icon. 

Is it possible to update to only show locks for the *required* endpoints that need authentication? 

This was present in the past and will give third party users requesting read only data some ideas as to what endpoints they will have access to.

MikeNeilson:
Odd, that's supposed to be happening that way already.

MikeNeilson:
It's already fixed, just not on the -prod version.
Will be updating as soon as a few other changes get merged in and moved through -dev and -test.

PR #1779, opened 2026-06-11
Add spec-driven Swagger auth controls

krowvin:
﻿## Summary
- Update the GUI to use Groundwork Water auth helpers for Swagger UI login
- Select the custom login method from the generated OpenAPI security schemes
- Add a login mode dropdown so users can switch between the custom login flow and Swagger's built-in Authorize controls
- Support localhost development by falling back to local Keycloak direct-grant auth when CWMSLogin is advertised but not reachable (It was showing in the security controls response)
- Fix localhost Keycloak dev redirect and healthcheck behavior

## Validation
- `npx.cmd eslint src/pages/swagger-ui/index.jsx --ext js,jsx --report-unused-disable-directives --max-warnings 0`
- `npm.cmd run build`
- Tested manually on localhost against local CDA instance

## Screenshots

With "CWBI Login" selected the locks on the swagger UI are not shown. The hope is to not add login confusion by default. While still allowing the option to authenticate via other means (API key/etc)  

image

image

- [x] AI Tools Used


cda-gui/src/css/swagger.css
```diff
@@ -14,3 +14,130 @@ body {
   margin: 0;
   background: #fafafa;
 }
+
+.swagger-auth-bar {
```
krowvin: I was wondering if this should be tailwind css? Preference?

cda-gui/package.json
```diff
@@ -21,7 +21,7 @@
   },
   "dependencies": {
     "@tanstack/react-query": "^5.64.1",
-    "@usace-watermanagement/groundwork-water": "^3.3.1",
+    "@usace-watermanagement/groundwork-water": "^3.9.0",
```
krowvin, line 24: older GWW only supported direct grant

cda-gui/vite.config.js
```diff
@@ -4,29 +4,40 @@ import react from "@vitejs/plugin-react";
...
+  const cdaApiRoot = new URL(env.CDA_API_ROOT || "http://localhost:8081").origin;
   // const BASE_PATH = env?.BASE_PATH ?? "/cwms-data";
   return {
     base: "/cwms-data",
     plugins: [react()],
     server: {
       proxy: {
-        "^/cwms-data/timeseries/.*": {
-          target: env.CDA_API_ROOT,
+        "/cwms-data/timeseries": {
```
krowvin: I debated doing one single wildcard for the localhost proxy here

compose_files/keycloak/realm.json
```diff
@@ -664,7 +664,8 @@
       "redirectUris": [
         "https://cwms-data.test:8444/*",
         "https://localhost:5010/*",
-        "http://localhost:*"
+        "http://localhost:*",
+        "http://127.0.0.1:*"
```
krowvin, line 668: Sometimes it'll launch in loopback, othertimes it's localhost. Figured we'd have both.

krowvin (commented):
Few thoughts of my own with these changes

cda-gui/src/css/swagger.css
MikeNeilson: What ever is inline with GWW and its usages by USACE?

cda-gui/vite.config.js
MikeNeilson: May as well so everything can be tested correctly locally by default.

compose_files/keycloak/realm.json
MikeNeilson: Yep, 127.0.0.1 is the ostensibly "correct" way, but in local development mode having both will make life easier.

MikeNeilson (commented):
Sorry for the delay, thought the review had actually submitted thursday and it had not.

cda-gui/src/css/swagger.css
```diff
@@ -14,3 +14,130 @@ body {
...
+  background: #b42318;
+  border-color: #912018;
+  color: #ffffff;
+}
+
+.swagger-auth-button.danger:not(:disabled):hover {
+  background: #912018;
+}
+
+.swagger-ui-host:not(.swagger-login-mode) .swagger-ui .auth-wrapper {
```
rma-psmorris: I don't think this should be done as the lock icon is informative as to what authentication is required by the endpoint. Swagger also doesn't have the best UI for their auth state (see https://github.com/swagger-api/swagger-ui/issues/4402).

- **No lock icon**: the endpoint is public. Authentication is not involved with the operation.
- **Gray unlocked lock icon**: authentication is optional. The endpoint allows anonymous access, but also accepts authenticated requests if credentials are provided.
- **Black locked lock icon**: authentication is required. The endpoint cannot be called successfully without valid credentials.

The change you are implementing here by not displaying the lock icon suggests that all endpoints are public.
Also, CDA does not correctly define Swagger icon state for black locked lock icons -- but that is a separate issue (that I'm checking on to see if it needs an issue to be created).

krowvin: I'm good to remove the drop-down and leave a simple login button that logs you in via AAA or KeyCloak depending on the swagger spec. 

@mikeneilson 

I'll do this unless you have thoughts on why we should leave locks out. 

My thinking was to have users opt into the swagger locks if they want this route of auth. Intenteded for internal use. But I realize that won't always be the case.

MikeNeilson: The locks should stay. Definitely interested in the no/gray/black though. Long term there will be items that won't show up from GETs unless authenticated with the new authorization work.

rma-psmorris: Here is the existing lock issue: https://github.com/USACE/cwms-data-api/issues/1118

MikeNeilson (approved):
This good, the failing tests are clearly something else, but given you have a PR that should fix at least one of those I would wait to merge until that one is and this one is merged/rebased with it.

PR #1780, opened 2026-06-11
CDA-113 Adding csv support for timeseries

### LINKED ISSUE #1257: Changing format= Does Not Work with Request Headers
# Problem
When attempting to query for `csv` data with `format=csv` on timeseries (and other endpoints with `format` as an option) you get an error. 

## CDA Versions
This is a problem on
- `2025.08.18` - TEST
- `2025.08.26-cwbi-prod-as-of-2025-08-26` - PROD

## Query
```bash
curl -X 'GET' \
  'https://cwms-data.usace.army.mil/cwms-data/timeseries?name=KEYS.Elev.Inst.1Hour.0.Ccp-Rev&office=SWT&format=csv' \
```
However if you curl with */* as the header (or no request header) it works
```bash
curl 'https://cwms-data.usace.army.mil/cwms-data/timeseries?name=KEYS.Elev.Inst.1Hour.0.Ccp-Rev&office=SWT&format=csv'
```

## Details
Part of the problem is that it's hard to control what request header the browser will decide on if a user, for example, wants to set an `<a>` tag to trigger a download for this data. 

<details>
<summary>
For example this (click me) fails due to the inability to force a header with the a tag. 
</summary>

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>Download Time Series Data</title>
</head>
<body>
  <h1>Download KEYS Elevation Data</h1>
  <p>
    <a href="https://cwms-data.usace.army.mil/cwms-data/timeseries?name=KEYS.Elev.Inst.1Hour.0.Ccp-Rev&office=SWT&format=csv" 
       download="KEYS_Elev_Inst_1Hour.csv">
      Click here to download CSV
    </a>
  </p>
</body>
</html>
```
</details>

Whereas,

<details>
<summary>
This works, but is involved and subject to CORS issues.
</summary> 

```html
<!DOCTYPE html>
<html lang="en">

    <head>
        <meta charset="UTF-8">
        <title>Download CSV with Fetch</title>
    </head>

    <body>
        <h1>Download KEYS Elevation Data</h1>
        <button id="downloadBtn">Download CSV</button>

        <script>
            document.getElementById("downloadBtn").addEventListener("click", async () => {
                const url = "https://cwms-data.usace.army.mil/cwms-data/timeseries?name=KEYS.Elev.Inst.1Hour.0.Ccp-Rev&office=SWT&format=csv";

                try {
                    // Fetch with no extra headers
                    const response = await fetch(url, { headers: { "accept": "*/*" } });
                    if (!response.ok) throw new Error("Network response was not ok");

                    // Convert to Blob
                    const blob = await response.blob();

                    // Create a temporary download link
                    const downloadUrl = window.URL.createObjectURL(blob);
                    const a = document.createElement("a");
                    a.href = downloadUrl;
                    a.download = "KEYS_Elev_Inst_1Hour.csv"; // Suggested filename
                    document.body.appendChild(a);
                    a.click();
                    a.remove();

                    // Release object URL
                    window.URL.revokeObjectURL(downloadUrl);
                } catch (err) {
                    console.error("Download failed:", err);
                }
            });
        </script>
    </body>

</html>
```

</details>

You can also work around the error in swagger by selecting the blank request header in the dropdown for 200; where application/json;version=2 is selected by default.

Image

## Suggested Solution

Consider ignoring the mime-type of the request and taking priority of the URI `format=` when provided. Still make use of the version provided in the accept header. 

## Other users

@SWF-WM originally reported this for stakeholders that were looking to use CSV. 

## Alternatives Considered

There are other WM data retrieval options that offer CSV:
Namely, 

https://water.usace.army.mil/cda/reporting/providers/swf/timeseries?name=ALAT2.Flow-Out.Inst.1Hour.0.Rev-SWF-REGI&begin=2025-08-20T16:34:46.267Z&end=2025-08-27T16:34:46.267Z&format=csv

That work out of the box with A tag, as you can see on the webpage that the link was pulled from:
[Data Sources - Aquilla Lake - A2W](https://water.usace.army.mil/overview/swf/locations/alat2)

rma-bryson:
## Summary

Implements csv support for timeseries. This should also streamline other data types supporting our custom csv formatting in the future - including metadata comment headers at top of payload, unit in column header, and optional-column inclusion via DTO annotations in the spirit of Jackson. 

This also adds support for specifying date-format. Currently this only works for csv, but could be extended to other formats in the future.

Also noting decision during cda dev meeting to use query parameters for optional column inclusion and date-format. Updated ADR to reflect that. 
## Related Issue

Closes https://github.com/USACE/cwms-data-api/issues/1257

## Validation

Added unit tests and integration tests

## Checklist

- [ ] Junie assisted in getting annotations working correctly.


cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
```diff
@@ -172,9 +174,8 @@ public final class Controllers {
     public static final String TS_IDS = "ts-ids";
 
     public static final String EXAMPLE_DATE = "2021-06-10T13:00:00-07:00";
-    public static final String DATE_FORMAT = "YYYY-MM-dd'T'hh:mm:ss[Z'['VV']']";
     public static final String TIME_FORMAT_DESC = "The <a href=\"times.html\">format for this field</a> is ISO 8601 extended" +
-            ", with optional offset and timezone, i.e., '" + DATE_FORMAT + "', e.g., '" + EXAMPLE_DATE + "'."            ;
+            ", with optional offset and timezone, i.e., '" + "YYYY-MM-dd'T'hh:mm:ss[Z'['VV']']" + "', e.g., '" + EXAMPLE_DATE + "'."            ;
```
MikeNeilson: All timestamps should be UTC as was discussed and is stated in the index page of CDA>

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamConsumer.java
```diff
@@ -4,7 +4,20 @@
...
 import java.sql.SQLException;
 
+/**
+ * A consumer for streaming binary data out to callers (e.g., HTTP layer).
+ *
+ * The primary abstract method accepts an InputStream with optional position and total length.
+ * For generated content where the total length is not known up front (e.g., CSV rendered on-demand),
+ * use the default two-argument overload which delegates with null position/length values, allowing
+ * callers to choose chunked transfer without setting Content-Length.
+ */
```
MikeNeilson, line 14: Don't we have this for clob/clob already? Are they that different?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -220,6 +232,270 @@ public String getTimeseries(String format, String names, String office, String u
...
+        Field<BigDecimal> qualityForNormalization = DSL.nvl(
+                view.QUALITY_CODE.cast(BigDecimal.class),
+                DSL.val(BigDecimal.valueOf(5))
+        );
+
+        Field<BigDecimal> normalizedQuality = CWMS_TS_PACKAGE.call_NORMALIZE_QUALITY(
+                qualityForNormalization
+        ).as("quality_norm");
+
+        Field<Double> convertedValue = CWMS_UTIL_PACKAGE.call_CONVERT_UNITS(
```
MikeNeilson: `AV_TSV_DQU where unit_id = 'X'` will already provide only units in `X`, a call to `cwms_util.convert_units` is redundant.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -220,6 +232,270 @@ public String getTimeseries(String format, String names, String office, String u
...
+                .and(view.TS_CODE.eq(tsCode))
+                .and(view.OFFICE_ID.eq(officeResolved))
+                .and(view.UNIT_ID.equalIgnoreCase(nativeUnits))
+                .and(view.DATE_TIME.ge(beginTimestamp))
+                .and(view.DATE_TIME.le(endTimestamp))
+                .and(view.START_DATE.le(endTimestamp))
+                .and(view.END_DATE.gt(beginTimestamp));
+
+        ResultQuery<? extends Record4<Timestamp, Double, BigDecimal, Timestamp>> query;
+        if (versionDate != null) {
```
MikeNeilson, line 267: It would seem this logic is the same in the updated tsv retrieval that @krowvin had created.

Perhaps that should be extended to take some sort of callback to allow the streaming or gathering instead of duplicating the logic. (especially considering how critical that logic is.)

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -220,6 +232,270 @@ public String getTimeseries(String format, String names, String office, String u
...
+            try (stream) {
+                consumer.accept(stream, Formats.CSV);
+            } catch (IOException | SQLException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    // InputStream that renders CSV rows on-demand from a jOOQ Cursor
+    private static final class CsvOnDemandInputStream extends InputStream {
```
MikeNeilson: This is large enough it should be in its own file.

MikeNeilson (changes requested):
Generally good. Definitely concerned about some critical logic duplication.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-bryson: refactored into own file

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-bryson: extracted query logic into a method for re-use

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-bryson: removed redundant unit conversion

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamConsumer.java
rma-bryson: Its the same StreamConsumer used for clob. Just added a default overload method for cases where total length isn't known instead of passing in -1L for length. But functionally its the same.

cwms-data-api/src/main/java/cwms/cda/api/Controllers.java
rma-bryson: Updated description to reflect timestamps being in UTC

rma-bryson:
image
Was able to get the example csv working

cwms-data-api/src/main/java/cwms/cda/data/dao/StreamConsumer.java
MikeNeilson: ... duh. This is what happens when you review code before a 3-day weekend. It appears my brain saw the files as all new, when you've simply added some useful comments to it.

cwms-data-api/src/main/java/cwms/cda/ApiServlet.java
```diff
@@ -932,8 +938,41 @@ private void getOpenApiOptions(JavalinConfig config) {
                                    .addSecurityItem(new SecurityRequirement().addList(provider))
         );
         ops.path("/swagger-docs")
-            .responseModifier((ctx,api) -> {
-                api.getPaths().forEach((key,path) -> setSecurityRequirements(key,path,secReqs));
+            .responseModifier((ctx, api) -> {
```
MikeNeilson: well, this block clearly needs to be cleaned up into multiple methods and likely a different class... but that is for a future PR.

MikeNeilson (dismissed):
Nice trick with the getExample method to fill out that detail.

Seems within that we could also setup our own annotation to retrieve a sample JSON.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-bryson: Turns out this was needed to handle bad user-requested units. The test_wrong_units() was failing in the TimeSeriesControllerTestIT. Adding this back in fixes it because the conversion procedure handles validation.

MikeNeilson: Right, but why are will killing performance or a possible, but not common, failure?

I can see the value in the check, but a single called to the unit conversion itself will inform us if the unit it valid, without an additional context switch for every row.

... yes yes, it might be "common" but not in the various places of automation that will call this endpoint a lot, or from GUIs that may artificially limit the choices in the UI.

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-bryson: Updated validation to a single indexed existence check before the data query, eliminating any per-row conversion overhead while still failing fast on invalid units.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -957,6 +962,23 @@ private DirectReadMetadata fetchRequestedTimeSeriesMetadataRecord(
...
+            boolean unitRequestedExists = dsl.fetchExists(
+                    selectOne()
+                            .from(AV_TSV_DQU.AV_TSV_DQU)
+                            .where(AV_TSV_DQU.AV_TSV_DQU.TS_CODE.eq(tsCode.cast(Long.class)))
+                            .and(AV_TSV_DQU.AV_TSV_DQU.OFFICE_ID.eq(officeId))
+                            .and(AV_TSV_DQU.AV_TSV_DQU.UNIT_ID.equalIgnoreCase(units))
+            );
+
+            if (!unitRequestedExists) {
+                String msg = units + " is not a valid unit for time series " + name;
```
MikeNeilson: User input, needs to go through the sanitizer.

I would've just called the procedure, but wasn't thinking about the unit system level.

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
rma-bryson: Updated to sanitize message. Also added a check that a row actually exists to guarantee failure is actually due to bad units

MikeNeilson (dismissed):
Sorry, apparently I merged something in that caused a conflict in this PR.

rma-bryson:
> Sorry, apparently I merged something in that caused a conflict in this PR.

Rebased, ready to merge once approved

MikeNeilson (approved):

PR #1781, opened 2026-06-12
Fix pseudo-irregular time series direct reads

### LINKED ISSUE #1745: TimeSeries Endpoint - Pseudo Irregular Fixes
## Problem

For pure irregular TS we should not be doing a gap fill for pseudo irregular. We must actually be checking if it truly local regular and do a check for this before just assuming. Typically 15min is not "local", Mike says it's a 2hour TS interval minimum for that. 

The direct SQL timeseries read path appears to classify any interval part beginning with `~` as regular/local-regular and therefore generates expected timestamps plus synthetic missing rows.

This is too broad. Some `~` series, especially short-interval series like:

`Baldhill_Dam.Flow-Out.Inst.~15Minutes.0.best` / office `MVP`

appear to be pseudo-irregular/observed series and should not be gap-filled unless CDA can prove the series is truly local-regular.

Current logic:

- `isRegularSeries(intervalMinutes, intervalPart)` returns true when `intervalMinutes != 0 || isLocalRegularInterval(intervalPart)`
- `isLocalRegularInterval(intervalPart)` returns true when normalized interval starts with `~`

For `Baldhill_Dam.Flow-Out.Inst.~15Minutes.0.best`, catalog metadata currently reports:

- interval: `~15Minutes`
- interval-offset: `-2147483648`
- time-zone: `US/Central`

The direct read path then builds expected timestamps and merges synthetic null rows. In a 2020-present comparison, CDA returned additional null/gap-fill rows that A2W did not return. For all-history reads this also appears to contribute to poor performance/failure.

## Expected Behavior

CDA should only perform expected-time generation/gap-fill when the timeseries is truly regular/local-regular under CWMS rules.

Pure irregular or pseudo-irregular `~` series should return observed rows only unless the underlying CWMS/retrieve logic explicitly defines gap-fill behavior for that series.

## Acceptance Criteria

- Add tests for a short-interval `~15Minutes` pseudo-irregular series proving it does not get synthetic gap rows.
- Keep tests proving true LRTS/local-regular series still get expected-time handling where required.
- Do not rely only on the TSID interval string prefix `~` to decide regularity.
- Use CWMS metadata/package behavior or another authoritative database signal to distinguish:
  - true regular
  - true local-regular/LRTS
  - pseudo-irregular/observed-only
- Preserve data correctness for versioned reads, quality codes, units, trim behavior, and entry dates.
- Add a regression/performance case for a long-window `~15Minutes` pseudo-irregular series.
- Must use method to check the timeseries specification to check if it's local regular. Should be a proceedure for this.

krowvin:
﻿## Summary

Fixes the time-series direct-read path for pseudo-irregular old-style local-regular IDs such as `*.~15Minutes.*` when `X-CWMS-LRTS-Formatting: true` is present.

## Root cause

The direct-read path could resolve pseudo-irregular old-style IDs through new LRTS formatting, producing local-regular metadata and returning no values. It also treated `~` intervals as regular without first honoring `interval_utc_offset = UTC_OFFSET_IRREGULAR`, which could incorrectly invoke gap-fill behavior for pseudo-irregular data.

## Changes

- Detect old-style `~` IDs whose old-format metadata has irregular interval offset.
- Resolve/read those IDs under old LRTS formatting for direct reads.
- Suppress regular gap-fill and report zero interval duration for truly irregular offset metadata.
- Add an integration regression test that seeds a pseudo-irregular `~15Minutes` series and verifies the CDA response matches `retrieve_ts` under the LRTS formatting header.

## Validation

- `./gradlew.bat :cwms-data-api:compileJava :cwms-data-api:compileTestJava --no-daemon`
- `./gradlew.bat :cwms-data-api:integrationTests --tests "*TimeSeriesDirectReadParityIT.pseudoIrregularReadWithLrtsHeaderMatchesRetrieveTs" "-PCDA_POOL_INIT_SIZE=5" "-PCDA_POOL_MAX_ACTIVE=10" "-PCDA_POOL_MAX_IDLE=5" "-PCDA_POOL_MIN_IDLE=2" --no-daemon`
- `./gradlew.bat :cwms-data-api:checkstyleMain :cwms-data-api:checkstyleTest --no-daemon`

Note: broader `TimeSeriesDirectReadParityIT` runs were not clean in this local container due fixture/setup issues before the relevant endpoint path was exercised, including Oracle seed partition check failures and a default-data county error.

Closes #1745


cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -679,6 +736,10 @@ private TimeSeries getRequestedTimeSeriesDirect(String page, int pageSize,
         String cursor = null;
         Timestamp tsCursor = null;
 
+        if (forceOldLrtsFormatting) {
```
MikeNeilson: I'm not sure why this cares about the LRTS formatting. There is a function that just tells you:

https://github.com/HydrologicEngineeringCenter/cwms-database/blob/648b849254db226bedd9d9a38e94628df28ec9d5/schema/src/cwms/cwms_ts_pkg.sql#L1064

use JOOQ to call (may already be a helper somewhere in CDA that caches the response.

Basic logic:
1. Is LRTS  -> build expected times and map it out
2. Is not LRTS -> just return the data.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -890,6 +959,103 @@ private List<TimeSeries.Record> fetchRequestedTimeSeriesRows(long tsCode, String
...
+                    return rows;
+                }
+            } catch (SQLException ex) {
+                throw new DataAccessException("Unable to fetch direct time series rows", ex);
+            }
+        });
+    }
+
+    private static String buildVersionedRowsSql(boolean includeEntryDate) {
+        return "select date_time,"
```
MikeNeilson: you can build and return a JOOQ query. keeps the type safety.

NOTE: some of those elements will still need to be text.

NOTE2: since we have to do things in a loop anyways to build the output, it will likely be faster to do the normalization of the quality_code in Java. Moving between plsql and sql does cause a context switch which has some performance concerns.

NOTE3: av_tsv_dqu already handles the unit conversions (that's what the U stands for in dqu), you just put `units_id = ...` in the where clause. Additional conversions are no-ops, but force the context switch.

MikeNeilson (changes requested):

krowvin:
pseudo-irregular `~15Minutes` series were being handled like true local-regular series in the direct-read path, which caused incorrect LRTS/session behavior and large window reads

# changes
- This fix uses database authoritative LRTS detection via `CWMS_TS_PACKAGE.call_IS_LRTS__2(...)`.
- Pseudo-irregular series with `interval_utc_offset == UTC_OFFSET_IRREGULAR` are treated as irregular and return observed rows only.
- True local-regular/LRTS series still use expected-time generation and gap-fill behavior.
- Removed the old workaround that forced old LRTS formatting through recursive DAO construction and raw JDBC row reads.
- Kept the direct row read in jOOQ against `AV_TSV_DQU`.
- Row reads now filter by requested `UNIT_ID` and normalize quality in Java.
- Added/adapted regression coverage for `ITPARPIRR.Flow.Inst.~15Minutes.0.BENCH` with `X-CWMS-LRTS-Formatting: true`.
- Verified the pseudo-irregular regression test passes.
- Verified full `TimeSeriesDirectReadParityIT` passes locally.
- Verified compile and checkstyle tasks pass locally with JDK 21.
- Tested MVP timeseries `Baldhill_Dam.Flow-Out.Inst.~15Minutes.0.best`.

## tests
- For a 70,177-row production sized window (`2024-01-01` to `2026-01-01`), local CDA returned the same row count, timestamps, qualities, and pseudo-irregular metadata.
- Local timing for that 70k-row window was about 2 seconds, compared with production  runs around 29-38 seconds.
- pulled data from each instance, local and prod, to confirm no drift in datasets. 

## timings

For this command 

localhost
```bash
curl -sS -w '\nstatus=%{http_code} time=%{time_total}s bytes=%{size_download}\n' \
  -H 'Accept: application/json;version=2' \
  -H 'X-CWMS-LRTS-Formatting: true' \
  'http://localhost:8081/cwms-data/timeseries?name=Baldhill_Dam.Flow-Out.Inst.~15Minutes.0.best&office=MVP&unit=cfs&begin=2025-05-01T00:00:00Z&end=2025-06-01T00:00:00Z&page-size=100000'
```
national 
```bash
curl -sS -w '\nstatus=%{http_code} time=%{time_total}s bytes=%{size_download}\n' \
  -H 'Accept: application/json;version=2' \
  'https://cwms-data.usace.army.mil/cwms-data/timeseries?name=Baldhill_Dam.Flow-Out.Inst.~15Minutes.0.best&office=MVP&unit=cfs&begin=2025-05-01T00:00:00Z&end=2025-06-01T00:00:00Z&page-size=100000'
```
(LRTS header not in Prod yet)

| Target | Window | Rows | Runs | Min | Median | Mean | Max | Next Page |
|---|---:|---:|---:|---:|---:|---:|---:|---|
| Local CDA | 2024-01-01 to 2026-01-01 | 70,177 | 5 | 1.815s | 1.903s | 2.075s | 2.668s | No |
| Production CDA | 2024-01-01 to 2026-01-01 | 70,177 | 3 | 29.088s | 37.805s |

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -691,12 +683,12 @@ private TimeSeries getRequestedTimeSeriesDirect(String page, int pageSize,
     private TimeSeries getRequestedTimeSeriesDirectWithOldLrtsFormatting(String page, int pageSize,
```
MikeNeilson: I'm still not sure why the "Old formatting/New Formatting" methods need to exist.

The method of formatting on output and input expectation is set at the session level and is purely visual. Regardless of visual style a time series is either: regular, irregular, psuedo regular (effective irregular for this purpose), or local regular. 

That status of the time series determine the 2 modes of data retrieval, none of which has anything to do with with an LRTS timeseries is rendered as "1DayLocal" or "~1Day" for the interval. The database output of the TS time will render that correctly.

It reads like a giant red-herring that detracts from the actual requirement of the read which boils down to the following:

`Do I need to provided predictable time placeholders (regular, local regular) or just render the rows as stored in the database  (irregular, psuedoregular).

MikeNeilson (changes requested):
Note possible conflicts with https://github.com/USACE/cwms-data-api/pull/1780 depending on which gets merged in first.

MikeNeilson:
TSV PR was merged in, conflicts need to be fixed. Also see my comments about LRTS methods and usage, feels like something is stuck down a rabbit hole of over thinking and needs to get dug out.


Ask database if this TSID is LRTS or REGULAR interval
yes -> have expected interval set in some way shape or form
no -> just return the data from the tsv query as it, the timestamps provided are by definition correct.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -240,44 +245,49 @@ private ResultQuery<Record4<Timestamp, Double, BigDecimal, Timestamp>> buildTsvD
 
         Timestamp beginTimestamp = Timestamp.from(beginTime.toInstant());
         Timestamp endTimestamp = Timestamp.from(endTime.toInstant());
+        String beginTimestampText = beginTimestamp.toLocalDateTime().format(ORACLE_DATE_FORMATTER);
```
MikeNeilson, line 344: why are we spending time formatting a timestamp that can be used as-is in the query?

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -240,44 +245,49 @@ private ResultQuery<Record4<Timestamp, Double, BigDecimal, Timestamp>> buildTsvD
...
                 .and(view.UNIT_ID.equalIgnoreCase(units))
-                .and(view.DATE_TIME.ge(beginTimestamp))
-                .and(view.DATE_TIME.le(endTimestamp))
-                .and(view.START_DATE.le(endTimestamp))
-                .and(view.END_DATE.gt(beginTimestamp));
+                .and(DSL.condition("{0} >= to_date({1}, 'yyyy-mm-dd\"T\"hh24:mi:ss')",
+                        dateTimeField, DSL.val(beginTimestampText)))
+                .and(DSL.condition("{0} <= to_date({1}, 'yyyy-mm-dd\"T\"hh24:mi:ss')",
+                        dateTimeField, DSL.val(endTimestampText)))
+                .and(view.START_DATE.isNull()
```
MikeNeilson: These (START_DATE and END_DATE) likely *can* be ignored, the query itself handles the situation better.

However, if they are to be used, their purpose is to bracket which physical at_tsv table is used.

it should be `start >= "year of data-01-01"` `end <= "year+1-01-01"`, it doesn't need the hours, minutes, or seconds.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -1148,11 +1186,17 @@ private long resolveIntervalOffset(long intervalOffset, String timeZoneId,
         return (rawRows.get(0).getDateTime().getTime() - topOfInterval.getTime()) / TimeUnit.MINUTES.toMillis(1);
     }
 
-    private boolean isRegularSeries(long intervalMinutes, String intervalPart) {
-        return intervalMinutes != 0L || isLocalRegularInterval(intervalPart);
+    private boolean isRegularSeries(long intervalMinutes, long intervalOffset, String intervalPart, boolean isLrts) {
+        return intervalOffset != UTC_OFFSET_IRREGULAR
+                && (intervalMinutes != 0L || (isLrts && isLocalRegularInterval(intervalPart)));
```
MikeNeilson: `(isLrts && isLocalRegularInterval(intervalPart)`

By definition these are either both true, or both false. You've already asked the database if the time series is local regular. We should not be trying to duplicate the logic in CDA.

MikeNeilson (changes requested):
Still some concerns with the overall logic, but some of that wasn't touched in this PR.

Actual concerns address in specific comments.

Otherwise, looks reasonable, and the tests show expected behavior correctly.

krowvin:
Spoke in CDA biweekly if RMA would also be able to review and ensure nothing was missed.

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
krowvin: When I was trying to run the test for TimeSeriesDirectReadParityIT it started failing/saying missing. 

I agree converting from one and back to the other is wasted cycles though. Any ideas on how to make it so it'll match up with `AV_TSV_DQU.DATE_TIME`?

MikeNeilson: Hmm. that's, unexpected, especially since it's a begin/end time stamp so a few seconds on either side shouldn't matter much. Might have something to do with the columns being `Date` instead of `Timestamp` in the database.

Presumably the performance is still improved overall? If so, I'd argue that challenge is reasonable cause to leave the extra conversions in. Not like it happens for every row. Just put a comment in about why for future readers.

adamkorynta: Was it failing in your dev environment, or on GitHub Actions? I'm curious if your local time zone is getting in the mix somehow

cwms-data-api/src/test/java/cwms/cda/api/TimeSeriesDirectReadParityIT.java
adamkorynta, line 210: Recommend added a separate test for the `1DayLocal` equivalent

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
```diff
@@ -999,63 +1008,80 @@ private List<TimeSeries.Record> fetchRequestedTimeSeriesRows(long tsCode, String
...
             }
-            return new TimeSeries.Record(dateTime, value, qualityCode);
+            return new TimeSeries.Record(dateTime, dataValue, quality);
         });
     }
 
+    private static int normalizeQualityCode(BigDecimal qualityCode) {
+        long quality = qualityCode == null ? 5L : qualityCode.longValue();
+        if (quality > Integer.MAX_VALUE) {
+            quality -= 4_294_967_296L;
```
adamkorynta: recommend creating a constant `private static final long UNSIGNED_INT_MODULUS = 1L << Integer.SIZE;`

adamkorynta (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
krowvin: This was failing in my local integration test run. The GitHub checks were green with the current `to_date(bindText)` comparison.

When I switched the `AV_TSV_DQU.DATE_TIME` to direct Timestamp locally `TimeSeriesDirectReadParityIT` started returning missing/zero direct read rows compared to `retrieve_ts`. I left the request conversion and added a comment explaining. Probably onto something about the local TZ though.

MikeNeilson: is your local database a docker instance or something else?

MikeNeilson (approved):

cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesDaoImpl.java
MikeNeilson: Given the conversion only happens once, it's not a huge performance hit, some I'm okay with this explanation for the time being. We'll want to look into it more, but we don't think we should hold up this change just for that.

adamkorynta (approved):

PR #1787, opened 2026-06-16
update level serialization to use Instant instead of ZonedDateTime

### LINKED ISSUE #600: Levels response datetime format doesn't contain timezone
As I mentioned in #599, the /levels/<level_id> call returns a response with a datetime that has the zone offset, but not the timezone.
This isn't useful because the offset alone doesn't uniquely identify a point in time.

```
{
  "office-id": "SPK",
  "seasonal-values": [],
  "constant-value": 568100,
  "level-units-id": "ac-ft",
  "level-date": "2024-03-28T00:00:00-07:00",
  "interval-months": 0,
  "interval-minutes": 0,
  "attribute-units-id": "ac-ft"
}
```

There's several issues with it:
1) GMT-7 can mean Mountain Standard Time, or Pacific Daylight Time. During DST, GMT-7 could be a time in California, or Arizona.

2) More relevant, is you can't perform any date arithmetic on it and assume the result is correct.
For example, say I have a datetime in California: **2023-03-11T00:00:00-0800**
I want to bump it to the next year, so add 365 days (366 in this case, since it's a leap year), that gives us **2024-03-11T00:00:00-0800**.
However, DST in 2024 switched on March 10th, so the actual time should be **2024-03-11T00:00:00-0700**.
If the timezone was present, such as **2024-03-28T00:00:00-07:00[America/Los_Angeles]** as used in the timeseries call, then date arithmetic can be adjusted correctly.

3) It's not round-trippable; you can't feed the datetime provided in the response to a query and get the same result.

It's possible this issue is present on other endpoints, but I haven't tested all of them.

adamkorynta:
this will ensure serialization to always display as UTC

## Summary

Serializes levels always to UTC using java.time.Instant

## Related Issue

Closes #600 

## Validation

Tested via unit and integration tests.

## Checklist

- [ ] AI tools used


MikeNeilson (dismissed):
Well, once the conflicts are fixed anyways.

MikeNeilson (approved):

PR #1791, opened 2026-06-20
CI Fix: Update water supply location fixtures

krowvin:
## Summary

Brings the failing water supply integration tests to match the current CWMS database image.  

## What Changed

- Updated water contract test fixtures to expect `Davis, California` instead of `Davis`.
- Updated water supply accounting test metadata:
  - county: `Yolo`
  - nearest city: `Davis, California`
  - coordinates: Davis-area latitude/longitude

## Why

The current `cwms-database` schema now gets nearest city from stored coordinates in `cwms_loc.store_location2`, returning values like `Davis, California`. The CDA tests were still expecting the older city value, and one accounting item used  city/county/coordinate data that caused `STORE_LOCATION` to fail.

## Verification

```sh
./gradlew.bat :cwms-data-api:integrationTests --tests '*WaterSupplyAccountingDaoIT' --tests '*WaterContractDaoTestIT' --no-daemon
```

## Related Issue

No issue made, tests were failing. Attempting to clean up tests so the failed tests are only related to the PR

## Checklist

- [x] AI tools used


MikeNeilson (commented):
I'm a little concerned this isn't now failing in the non-latest schemas.

adamkorynta:
> I'm a little concerned this isn't now failing in the non-latest schemas.

Looking at the db code (https://github.com/HydrologicEngineeringCenter/cwms-database/pull/151/changes#diff-325c763c7bea0e7734d1eb308eb8ab570ce4eb6b639ad175f7ca39e0888539f4R98), it should only be setting the nearest city if the value is null. Since we are storing the location withe `Davis`, it shouldn't be null. I'm also wondering if this is one of those fields that was originally inheriting from the parent base location previously.

MikeNeilson:
>  I'm also wondering if this is one of those fields that was originally inheriting from the parent base location previously.

Certainly wouldn't surprise me. But as it seems to work for everything I'm inclined to roll with it vs fretting over it.

adamkorynta (approved):

PR #1792, opened 2026-06-20
CI Fix: Avoid nested pool borrow deleting TS categories

krowvin:
## Summary

Fixes a pool exhaustion failure in time series category cascade delete that causes CI to fail

`TimeSeriesCategoryDao.cascadeDelete()` already borrows a connection before setting the session office and then calls `getDbVersion()` which could try to borrow a second connection from the same data source. `maxActive=1` can cause a deadlock on the request and return a 500 during category DELETE. This change checks the schema version using the existing `DSLContext`.

## Validation

Ran tests locally
- `:cwms-data-api:integrationTests --tests '*TimeSeriesCategoryControllerTestIT.test_create_update_ignore_nulls'`
- `:cwms-data-api:integrationTests --tests '*TimeSeriesCategoryControllerTestIT'`
- `:cwms-data-api:compileJava :cwms-data-api:checkstyleMain`

## Checklist

- [x] AI tools used

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesCategoryDao.java
+++ b/cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesCategoryDao.java
@@ -94,7 +94,8 @@ public class TimeSeriesCategoryDao extends JooqDao<TimeSeriesCategory> {
         connection(dsl, conn -> {
             DSLContext dslContext = getDslContext(conn, office);
 
-            if (getDbVersion() > Dao.CWMS_25_07_01) {
+            // Reuse the connection already borrowed above so small test pools don't deadlock.
+            if (currentDbVersion(dslContext) > Dao.CWMS_25_07_01) {
                 // With newer schema it should just work, don't need transaction
                 Configuration config = dslContext.configuration();
                 CWMS_TS_PACKAGE.call_DELETE_TS_CATEGORY(config, categoryId, formatBool(true), office);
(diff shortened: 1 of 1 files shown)


cwms-data-api/src/main/java/cwms/cda/data/dao/TimeSeriesCategoryDao.java
```diff
@@ -118,6 +119,13 @@ private void cascadeDelete(String categoryId, String office) {
         });
     }
 
+    private static int currentDbVersion(DSLContext dslContext) {
```
adamkorynta, line 122: This implementation pattern increases the amount of required boilerplate required going forward. Given `Dao::getDbVersion` caches the version for awhile, I'm thinking it might be better to set this as instance member state. If it is initialized in the Dao abstract class constructor, then the version lookup would be out of the scope of any open connection lifecycle. We might want to also increase the 5-minute cache to something like a full day too.

Thoughts? Otherwise, this specific should be replaced with parameterizing the DSLContext into `Dao::getDbVersion` instead so that all future calls benefit and this method wouldn't need to be copied around.

MikeNeilson: Either makes sense to me. I'm moving this work over to the water supply fix PR. Since both are failing in CI makes sense to combine them so we can see a clean passing of all tests.

I'll look at the other elements and make a choice. Just the parameterization is likely the easiest. but may not be the best. and yeah, the caching can definitely be a day, or more. In the future schema updates will have stopped CDA anyways, so doesn't need to be a super dynamic check.

MikeNeilson: Parameter to `Dao::getDbVersion` was the easiest. https://github.com/USACE/cwms-data-api/pull/1791/changes/a32463f3bf72050a63e76a6a727e5a87b55c1631

MikeNeilson:
Changed moved to separate PR.

PR #1797, opened 2026-06-23
Remove auth restriction on RSS, add rate limiting (1 per 10 seconds.)

MikeNeilson:
## Summary

Remove auth restriction from RSS feed, but add rate limiting to the expected usage (e.g. 1 request per client every ten seconds)

## Related Issue

N/A

## Validation

Manually, queried a bunch until it gave me the 429.

existing integration test triggered the 429 and was modified to use the provided Retry-After value.

## Checklist

- [ ] AI tools used

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
@@ -47,6 +47,7 @@ import io.javalin.plugin.openapi.annotations.OpenApi;
 import io.javalin.plugin.openapi.annotations.OpenApiContent;
 import io.javalin.plugin.openapi.annotations.OpenApiParam;
 import io.javalin.plugin.openapi.annotations.OpenApiResponse;
+
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
@@ -98,9 +99,10 @@ public final class RssHandler extends BaseHandler {
             @OpenApiResponse(status = STATUS_200, content = {
                 @OpenApiContent(from = RssFeed.class, type = Formats.RSS)
             }),
-            @OpenApiResponse(status = STATUS_404, description = "Unknown Feed")
+            @OpenApiResponse(status = STATUS_404, description = "Unknown Feed"),
+            @OpenApiResponse(status = STATUS_429, description = "Rate Limit exceeded.")            
         },
-        description = "Returns RSS feed items limited to the last week.",
+        description = "Returns RSS feed items limited to the last week. End point is limited to 1 request per 10 seconds per client per feed.",
         tags = {TAG}
     )
     @Override
--- a/cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
+++ b/cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
@@ -41,6 +41,7 @@ import cwms.cda.helpers.ReplaceUtils;
 import io.javalin.core.util.Header;
 import io.javalin.http.Context;
 import io.javalin.http.HttpCode;
+import io.javalin.http.util.NaiveRateLimit;
 import io.javalin.plugin.openapi.annotations.OpenApi;
 import io.javalin.plugin.openapi.annotations.OpenApiContent;
 import io.javalin.plugin.openapi.annotations.OpenApiParam;
@@ -50,6 +51,7 @@ import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
+import java.util.concurrent.TimeUnit;
 import java.util.function.UnaryOperator;
 import org.apache.http.client.utils.URIBuilder;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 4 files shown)


cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
```diff
@@ -120,6 +123,7 @@ public void handle(@NotNull Context ctx) throws Exception {
             MessageDao dao = new MessageDao(dsl);
             RssFeed feed = dao.retrieveFeed(cursor, pageSize, office, name, since, newLinkTemplate(ctx));
             String result = Formats.format(contentType, feed);
+            ctx.header("Retry-After", "10");
```
krowvin: What is the purpose of the retry-after header? Shouldn't the enduser decide how often they want to retry? Or is this some kind of default? 

Edit: i think I see what you're doing below with this header now

cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
```diff
@@ -102,6 +104,7 @@ public RssHandler(MetricRegistry metrics) {
     )
     @Override
     public void handle(@NotNull Context ctx) throws Exception {
+        NaiveRateLimit.requestPerTimeUnit(ctx, 6, TimeUnit.MINUTES);
```
krowvin: Is this reading as you can't make more than 6 requests per minute? (Your 10 second cap?)

cwms-data-api/src/test/java/cwms/cda/api/rss/RssHandlerIT.java
```diff
@@ -141,6 +141,13 @@ void test_rss_feed_with_pagination() {
 
             pagesVisited++;
             nextHref = nextLinkHref(nextXml);
+            String waitStr = nextPage.header("Retry-After");
+            int wait = waitStr != null && !waitStr.isEmpty() ? Integer.parseInt(waitStr) : 10;
+            try {
+                Thread.sleep(wait*1000);
```
krowvin, line 147: Is this consuming resources to keep the thread open like this? 

Instead of just immediately returning back a 429?

krowvin (changes requested):
I'm not sure on a few things, wanted to clarify first

cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
adamkorynta: I think we would want `Retry-After` to be set if the rate limit is hit instead of just on success. We could also set the cache header to something similar to avoid burst.

cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
adamkorynta: I had to reread the Javalin docs to confirm this. At first it read to me like one request per 6 minutes

cwms-data-api/src/test/java/cwms/cda/api/rss/RssHandlerIT.java
MikeNeilson: This is the test, the endpoint will return the 429 if we query too fast but this test is about making sure we get the complete set of data.

cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
MikeNeilson: Yeah, that one took me a minute as well. I'll go put a comment in on that line for future readers.

cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
MikeNeilson: Basically it's saying, we expect you to behave like "HTTP Long Polling". In long polling you hold the connection open for 10 seconds waiting for data and then retry, but this the intentional rate limit, this made sense to add to tell anyone paying attention... don't query this too fast, there's really no point.

MikeNeilson: The rate limit stops the user, this is just informational.

MikeNeilson: Agreed, I just couldn't think of a clean way... other than move that exception handling into here. which, actually fairly clean reasonable exception, will go do that.

adamkorynta (dismissed):

cwms-data-api/src/main/java/cwms/cda/api/rss/RssHandler.java
```diff
@@ -120,8 +129,21 @@ public void handle(@NotNull Context ctx) throws Exception {
...
             String result = Formats.format(contentType, feed);
+
             ctx.result(result);
             ctx.contentType(contentType.toString());
+        } catch (HttpResponseException ex) {
+            // an exception to our error handling rules. HttpResponseException, includes multiple other exception
+            // and we don't want to deal with trying to distinguish in ApiServlet. For the time being this logic will
+            // remain here.
+            if (ex.getStatus() == 429) {
+                var cdaError = new CdaError("Too many requests. Limit queue requests to no more than every 10 seconds.");
```
adamkorynta, line 142: This should probably go in the OpenAPI description

MikeNeilson: ... duh.

krowvin (dismissed):

adamkorynta (approved):

PR #1805, opened 2026-06-24
Initial batch. No functional changes other than a System.out to Logger.

MikeNeilson:
## Summary

Fix formatting according to Checkstyle results.

## Related Issue

N/A

## Validation

run `./gradlew checkstyleMain` until no results for a file in the output. (with a few exception.)

## Checklist

- [ ] AI tools used

### DIFF
--- a/cda-gui/package-lock.json
+++ b/cda-gui/package-lock.json
@@ -82,6 +82,7 @@
       "integrity": "sha512-RgHBCvtjbOK2gXSNBNIkNoEc9qoVEtau3hj8gEqKQuL3HZAibKarWFEI3Lfm6EYKkLalOh8eSrj9b+ch9H/VBA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.7",
         "@babel/generator": "^7.29.7",
@@ -1200,9 +1201,6 @@
         "arm"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1217,9 +1215,6 @@
         "arm"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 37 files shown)


adamkorynta (dismissed):

cda-gui/package-lock.json
krowvin, line 1: We should remove this because it does not appear the packages are also being updated. Even then it looks unrelated to the rest of your PR. 

I suspect you had an old lock file and have not ran `npm install` in the CDA gui dir and this is resulting in a previous lock file trying to commit?

cwms-data-api/src/main/java/cwms/cda/formatters/ContentType.java
```diff
@@ -43,17 +52,19 @@ public Map<String, String> getParameters() {
         return new LinkedHashMap<>(parameters);
     }
 
-    public String getCharset() {
-        return charset;
+    public String getInstanceCharset() {
+        return instanceCharset;
```
krowvin, line 56: Are there tests this could break renaming the getter here?

Rest looks formatting related but this is an API change

krowvin (changes requested):
I would remove this package lock. I think it's a bit of leftover.

This is minor, just mentioning 

I'm also seeing some trailing whitespace issues in

OfficeFormatV1.java
DateUtils.java
ResourceHelper.java 

I thought we had a .gitattributes to help but I'm probably thinking of another repo

cda-gui/package-lock.json
MikeNeilson: thanks, usually look for that manually and must've missed it.

As for the above Java, i'll check the files, fixing the trailing whitespace was something I was actively doing.

cwms-data-api/src/main/java/cwms/cda/formatters/ContentType.java
MikeNeilson: Yes, if the rename went wrong, the rename symbol function in VS Code search for usages and changes those as well.

MikeNeilson: Also This is purely a name style (e.g. formatting, maybe I should've called the branch formatting and style or just style) change, not a functional change.

krowvin (approved):
Thanks!

PR #1829, opened 2026-07-08
CWMS-2498: Disallow sub-minute level effective dates

### LINKED ISSUE #944: Round Level "level-date" Post to nearest Minute
The "Level" end point returns a "level-date" parameter that includes decimal seconds. When editing a level, I posted back to the database with CDA and forgot to change the "level-date" which defaulted to a very specific time (e.g. 2024-11-08T14:30:55.147439Z). CWMS-VUE cannot display this level, and I think it is related to the specific time. CWMS-VUE cannot delete the level either even when cascade delete is called, again because I think it does not expect that specific of time. Unfortunately I don't have that time stored in memory so I can't use CDA to delete either, so I will have to bug someone who can find it and delete with SQL.

Here's the CWMS-VUE cascade delete error:
`Caused by: Error : 20034, Position : 0, Sql = BEGIN "CWMS_20"."CWMS_LEVEL"."DELETE_LOCATION_LEVEL" (:1 , cast(:2  as date), :3 , :4 , :5 , :6 , :7 , :8 ) ; END;, OriginalSql = { call "CWMS_20"."CWMS_LEVEL"."DELETE_LOCATION_LEVEL" (?, cast(? as date), ?, ?, ?, ?, ?, ?) }, Error Msg = ORA-20034: ITEM_DOES_NOT_EXIST: Location level "MVP/Baldhill_Dam.Elev.Inst.0.Bottom of Flood Control@08-NOV-24" does not exist.`

zack-rma:
## Summary

Add check on level storage for level dates with sub-minute accuracy. Throws error when encountered.

## Related Issue

Closes #944

## Validation

Integration test included.

## Checklist

- [ ] AI tools used


cwms-data-api/src/main/java/cwms/cda/data/dto/locationlevel/LocationLevel.java
```diff
@@ -95,7 +97,8 @@ public abstract class LocationLevel extends CwmsDTO {
     @Schema(description = "Units the provided levels are in")
 
     private final String levelUnitsId;
-    @Schema(description = "The date/time at which this location level configuration takes effect.")
+    @Schema(description = "The date/time at which this location level configuration takes effect. "
+        + "Must be limited to minute accuracy.")
```
MikeNeilson:
```suggestion
        + "Must be limited to minute precision.")
```

MikeNeilson (changes requested):

MikeNeilson (approved):

PR #1848, opened 2026-07-24
CDA-106 - Adds UUID support for measurements

### LINKED ISSUE #1624: Changes to measurement endpoint?
USGS has changed their streamflow measurement endpoint in Feb 2026. The biggest change is:

They no longer use an integer for the streamflow measurement number. They use a UUID for a field visit ID and a UUID for a measurement ID. This allows them to store stage only measurements separately.

CWMS requires a streamflow measurement number as well as a flow and stage value. This creates complications in storing USGS measurements because the CWMS measurement number can no longer match the USGS identifiers.

They also have some channel information from the measurement at a separate endpoint.

Consider redesigning the measurement endpoint and backend cwms DB to more closely align with the USGS system. This could help this CWMS feature request as well: https://discourse.hecdev.net/t/allow-a-stage-measurement-with-no-flow-to-be-stored/4507

At a minimum, allow fields to store USGS field visit UUID and potentially measurement UUID (although for the measurement UUID there is some documentation  that it may change willy nilly?).

 We can workaround for now by hashing the USGS field visit UUID, returning an int from that for the CWMS measurement number, and storing the field visit UUID in the comments, but that's probably not the best way to do things. Example:
```
    "usgs-measurement": {
      "remarks": "field_visit_id=46dd5fe1-4036-447e-890c-ebcd6b4e285a",
      "current-rating": "",
      "control-condition": "Clear",
      "flow-adjustment": ""
    }
```

**Additional context**

It appears that the cwms DB currently aligns fairly closely with the legacy USGS measurement endpoint, but the redesigned USGS endpoint does not align that closely. 

New USGS measurement endpoints:
https://api.waterdata.usgs.gov/ogcapi/v0/collections/field-measurements/schema
https://api.waterdata.usgs.gov/ogcapi/v0/collections/channel-measurements/schema

rma-bryson:
## Summary

Adds UUID support for measurements rather than number

## Related Issue

Closes https://github.com/USACE/cwms-data-api/issues/1624

## Validation

Integration tests

## Checklist

- [ ] AI tools used


gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 jaxb-api = "2.3.1"
 jaxb-impl = "3.0.2"
-jooq-codegen = "26.02.17-RC01-oracle19c"
+jooq-codegen = "latest-dev_sha256_e7740339fdf4-oracle19c"
```
MikeNeilson: honestly can't decided if okay (I mean the sha256 is in there) or we should wait for at least the 26.07.16 RC to be out.

MikeNeilson (commented):
Looks reasonable. Does seem like a bit of a breaking change, but If I recall that's because the data itself suddenly change and so no matter what we basically have "mixed" data and just have to deal with it.

MikeNeilson (dismissed):

gradle/libs.versions.toml
rma-bryson: rebased to use latest toml

rma-bryson:
Had to abort a merge and rebase with a test fix. Should be ready to go now

MikeNeilson (approved):

PR #1852, opened 2026-07-29
Improve Data Query CSV column headers for readability

NobuNobuu:
## Summary

- Includes the timeseries location ID in query column headers
- Makes is so when you pick more than one project the csv shows location+parameter in the headers if you download it
- Example being headers such as `Elev (ft)` to `WTYT2 Elev (ft)` if you got WTYT2 selected

## Validation

-Tested in a local copy in swf's internal, csv's downloaded fine with the change there.  

## Checklist

- [ ] AI tools used


NobuNobuu:
@krowvin Can you take a peek when you got a chance, small change I tested on my local copy that I noticed we didn't do on the main one.

krowvin (approved):

PR #1862, opened 2026-08-04
Correct format lookup logic. And a few other things

### LINKED ISSUE #1861: [Bug]: Timeseries Recent call forces conflicting formats on call
### Describe the bug

So when you load up a timeseries recent call as say a json, and you can see in the network tab that it calls for a json, for some reason xml gets forced into the call too causing the formats to conflict and error out. Got some examples from dev and prod
https://water.dev.cwbi.us/cwms-data/timeseries/recent?office=SWF&ts-ids=LBWT2.Elev.Inst.1Hour.0.Decodes-Rev&_cb=1785860195842

https://cwms-data.usace.army.mil/cwms-data/timeseries/recent?office=SWF&ts-ids=LBWT2.Elev.Inst.1Hour.0.Decodes-Rev

### Expected behavior

I'm expecting just a [] from this timeseries as it's empty right now. 

### To Reproduce

In swagger when you use timeseries recent, put in office and timeseries, then grab the request url it gives and check it out. 

### Priority

Medium

### Logs/Incident Identifier

{"message":"Formatting error:No Format for this content-type and data type : (application/xml, cwms.cda.data.dto.RecentValue)","incidentIdentifier":"3dc7bf58-d804-41bd-bfbc-357badb6ae1e","source":"Parser","details":{}}

### CURL Commands

```shell

```

### CDA Version

 2026.07.31-develop and  2026.05.12-f

### Additional context

_No response_

MikeNeilson:
## Summary

Process the provided header in such a way as to not accidentally attempt to use an incorrect content type.

primary fix, always map the primary content type to itself.
Remove the logic in parseHeader that allow creating an unmapped ContentType object.
Then add appropriate annotations to deal with the fallout of doing that.

## Related Issue

Closes #1861

## Validation

Additional test for Formats, existing tests.

## Checklist

- [ ] AI tools used


cda-gui/package-lock.json
krowvin, line 1: I would remove this from this PR

cwms-data-api/src/test/java/cwms/cda/api/BasinControllerIT.java
krowvin, line 1: This looks like it's trying to accommodate the change for 
https://github.com/USACE/cwms-data-api/pull/1862/changes#r3721590710

Should we just revert this and undo that other change. Just thinking it might be out of scope for the header fix

cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
```diff
@@ -412,14 +412,6 @@ public static ContentType parseQueryParam(String queryParam, Class<? extends Cwm
...
                 logger.atFinest().log("%s converted to %s", ct, aliasType);
                 contentTypes.add(aliasType);
-            } else {
-                //If the DTO parameter is null, alias map is empty. Compare against well-known types
-                //Only use the ContentType classes initialized in contentTypeList rather than
-                //the client headers itself
-                ContentType type = new ContentType(ct);
-                if (contentTypeList.contains(type)) {
-                    contentTypes.add(type);
-                }
```
krowvin, line 422: From the CI tests this looks like it's causing the existing JSON/XML requests to return 406

i.e.

`833 tests completed, 83 failed, 20 skipped`

via
https://github.com/USACE/cwms-data-api/actions/runs/30954615142/job/92144620012

BasinControllerIT.test_get_create_delete expected HTTP 200 but received 406

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
```diff
@@ -175,11 +176,23 @@ public void beforeAll(ExtensionContext context) throws Exception {
...
-            // we only use doubles
-            RestAssured.config()
+            // actually assign the new config to the global configuration. just running this here without
+            // the assignment apparently does nothing.
+            RestAssured.config = RestAssured.config()
+                        // we only use doubles
                        .jsonConfig(
                             JsonConfig.jsonConfig()
-                                      .numberReturnType(JsonPathConfig.NumberReturnType.DOUBLE));
+                                      .numberReturnType(JsonPathConfig.NumberReturnType.DOUBLE))
```
krowvin: If we set all JSON numbers to double won't this break many of the existing integration tests?

cwms-data-api/src/main/java/cwms/cda/data/dto/rating/Ratings.java
```diff
@@ -0,0 +1,12 @@
+package cwms.cda.data.dto.rating;
+
+import cwms.cda.data.dto.CwmsDTOBase;
+import cwms.cda.formatters.Formats;
+import cwms.cda.formatters.annotations.FormattableWith;
+import cwms.cda.formatters.xml.XMLv2;
+
+@FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.DEFAULT, Formats.XML})
+@FormattableWith(contentType = Formats.JSONV2, formatter = XMLv2.class, aliases = {Formats.JSON})
```
krowvin:
```suggestion
@FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.JSON})
```

krowvin (changes requested):
Appears to be failing tests from this PR and the package lock slipped in again

cwms-data-api/src/test/java/fixtures/CwmsDataApiSetupCallback.java
MikeNeilson: It shouldn't as this was *always* supposed to be set, and we were setting it manually in the time series tests.

I might change it back though, at least for the numbers, the real change is the adding of the content type processing, only just realized we needed to actually assign the new config to the global, so it wasn't properly active before.

cda-gui/package-lock.json
MikeNeilson: yeah, that one keeps sneaking in.

cwms-data-api/src/main/java/cwms/cda/data/dto/rating/Ratings.java
MikeNeilson: Intentional, the processing actually bypasses the formatter lookup... though I should add a comment about it.

cwms-data-api/src/main/java/cwms/cda/formatters/Formats.java
MikeNeilson: Yep... *that's* the problem. The removal of this code is correct, otherwise an invalid type "mapping" gets in below instead of properly failing, so now I'm dealing with the affects of that.

cwms-data-api/src/test/java/cwms/cda/api/BasinControllerIT.java
MikeNeilson: I might, see below.

cwms-data-api/src/main/java/cwms/cda/data/dto/rating/Ratings.java
MikeNeilson: have changed it to a "DUMMY" outputformatter so the intentional is more clear.

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
```diff
@@ -33,6 +33,7 @@
 @FormattableWith(contentType = Formats.XMLV2, formatter = XMLv2.class, aliases = {Formats.XML})
 @FormattableWith(contentType = Formats.JSONV2, formatter = JsonV2.class, aliases = {Formats.DEFAULT, Formats.JSON})
 @FormattableWith(contentType = Formats.JSONV1, formatter = JsonV1.class)
+@FormattableWith(contentType = Formats.GEOJSON, formatter = JsonV2.class)
```
krowvin: Is this going to apply to every endpoint of Location.class (not just the getAll)?

If it makes a GET /locations/{id} accept geo+json and serializes JSON v2, it will label it GeoJSON no?

Might need a get all specific marker instead?

krowvin: Because later the normal formatter gets called here
https://github.com/USACE/cwms-data-api/blob/7641fe7670d9a0e1df41ee978c7c48e77c377a5e/cwms-data-api/src/main/java/cwms/cda/api/LocationController.java#L290

krowvin (changes requested):
Thanks for fixing those other concerns. 

Just had one comment about the Locaiton.java class. 

and

This PR is starting to get large because it also looks like you bumped the DB image, updated location delete error handling, and are tweaking the RSS rate limit test. (You may want to change the PR title if the scope was intended to be larger?)

This might turn into one of those endless soups!

cwms-data-api/src/main/java/cwms/cda/data/dto/Location.java
MikeNeilson: Hmm, yeah, that one should probably be the DUMMY formatter.

But to be clear the format mapping is *per dto/content-type* an end point may further constrain, or redirect if it needs to.

MikeNeilson:
> Thanks for fixing those other concerns.
> 
> Just had one comment about the Locaiton.java class.
> 
> and
> 
> This PR is starting to get large because it also looks like you bumped the DB image, updated location delete error handling, and are tweaking the RSS rate limit test. (You may want to change the PR title if the scope was intended to be larger?)
> 
> This might turn into one of those endless soups!

Thankfully, this is the end. I could've waited on the RSS limit but it was just bugging the crap out of me. The db image was a simple one, finally got release so needed an update.

That said, thanks for pushing back on scope keep, it generally is a bad idea.

MikeNeilson:
Failing test is a 429 on the RssHandler, this isn't from the current changes but a timing issue in the runners. Minor tweak was made in this PR to address but was clearly insufficient, will address more fully in a followup.

PR #1864, opened 2026-08-05
fix: return 404 for missing users

krowvin:
## Summary

- return a CDA `404 Not Found` error when `GET /users/{user-name}` cannot find the requested user
- document the 404 response in OpenAPI
- add an authenticated integration test for the missing-user case
- include the required Swagger UI screenshot

## Root cause

`UserDao.getByUniqueName(...)` returns an empty `Optional` when the user does not exist. The controller converted that to `null` and passed it to `Formats.format`, whose non-null check raised the `NullPointerException` that CDA reported as a 500. The controller now converts the empty result to `NotFoundException`, allowing CDA's existing error handling to return a structured 404 response.

## Verification

- `compileJava` and `compileTestJava`
- `checkstyleMain` and `checkstyleTest`
- focused `UserManagementTestIT.test_get_missing_user_not_found` integration test
- isolated Docker Compose end-to-end check:
  - existing user returned 200
  - missing user returned 404 with `User not found: DOES_NOT_EXIST`

image

- [x] AI tools used


MikeNeilson (approved):

krowvin:
I found this error when fishing for a stack trace for 
https://github.com/HydrologicEngineeringCenter/cwms-cli/pull/247

PR #1734, opened 2026-05-15
Add office-scoped user-list API and management UI

krowvin:
Refs #1733.

## Summary

Adds the complete, authenticated CDA user-list feature for office-scoped recipient groups:

- office-scoped user-list metadata CRUD
- list membership read and management endpoints
- candidate-member lookup
- owner and office authorization behavior
- schema/feature gating for installations without the database objects
- Groundwork-styled CDA UI for creating, editing, and deleting lists and managing members
- ADR documenting the CDA user-list context, ownership, office scope, API, and UI decisions

## Related Issue

Closes #1733

Database foundation: HydrologicEngineeringCenter/cwms-database#160. Two follow-up database commits for office scoping and CDA grants remain local for review before a separate database PR.

## Validation

- latest `origin/develop` fetched and merged: already up to date
- `:cwms-data-api:integrationTests --tests cwms.cda.api.users.UserListControllerTestIT`: 5 passed
- `:cwms-data-api:test --tests cwms.cda.features.CdaFeatureManagerProviderTest`: 6 passed with the local runtime feature override isolated
- CDA UI `npm run lint`: passed
- CDA UI source Prettier check: passed
- CDA UI `npm run build`: passed
- broad Gradle `build` completed compilation, WAR packaging, and the bundled UI build; its unit-test phase still reports unrelated existing Windows/JDK 21 fixture and Mockito failures

## Screenshots
image

image

image

image

## Checklist

- [x] AI tools used

PR #1865, opened 2026-08-06
fix: prevent CDA GUI auth bootstrap crash

krowvin:
## Summary

- Load deployed Keycloak settings at runtime from the OpenAPI document instead of eagerly reading build-time Vite variables.
- Keep the CDA GUI available when OpenID configuration is missing or invalid, and show a controlled sign-in-unavailable state.
- Add a true application-level React error boundary and correct the React Router error fallback to use `useRouteError()`.
- Share OpenID/Keycloak parsing between the application auth provider and Swagger UI, with focused regression tests.

## Related Issue

Follow-up to #1734.

The user-list UI introduced an app-wide `createKeycloakAuthMethod()` call during `main.jsx` module evaluation. The deployable production-mode bundle does not define `VITE_AUTH_HOST`, so Groundwork received `host: undefined` and failed on `host.trim()` before React or the router error boundary could render. This left the CWBI dev UI on a white error page.

## User Impact

CDA now derives authentication configuration from the deployment's own OpenAPI document. If the deployment does not advertise usable OpenID configuration, public CDA pages remain available and authenticated features explain that sign-in is unavailable instead of crashing the entire application.

## Validation

- `npm run test` — 4 tests passed
- `npm run lint`
- `npm run build` using Vite production mode
- Repeated test, lint, and production build in an isolated `npm ci --ignore-scripts` installation using the locked Groundwork 3.9.0 dependency
- Loaded the production artifact locally in a browser with unavailable OpenID configuration; verified the CDA introduction page rendered, the sign-in-unavailable state appeared, the page was not blank, and the console had no errors

## Checklist

- [x] AI tools used

### DIFF
--- a/cda-gui/src/components/AppAuthProvider.jsx
+++ b/cda-gui/src/components/AppAuthProvider.jsx
@@ -1,0 +1,121 @@
+import {
+  AuthProvider,
+  createKeycloakAuthMethod,
+} from "@usace-watermanagement/groundwork-water";
+import PropTypes from "prop-types";
+import { useEffect, useState } from "react";
+
+import { getBasePath } from "../utils/base";
+import { getKeycloakConfig, normalizeOpenIdConnectUrls } from "../utils/auth-config";
+import { AuthConfigurationContext } from "./auth-configuration-context";
+
+function createLocalAuthMethod() {
+  let token;
+  return {
+    async login() {
+      const response = await fetch(
+        `${import.meta.env.VITE_AUTH_HOST}/realms/${import.meta.env.VITE_AUTH_REALM}/protocol/openid-connect/token`,
+        {
+          method: "POST",
+          headers: { "Content-Type": "application/x-www-form-urlencoded" },
+          body: new URLSearchParams({
+            grant_type: "password",
+            client_id: "cwms",
+            username: import.meta.env.VITE_AUTH_USER,
+            password: import.meta.env.VITE_AUTH_PASSWORD,
+          }),
+        },
+      );
+      if (!response.ok) {
+        throw new Error(`Local Keycloak login failed (${response.status})`);
+      }
+      token = (await response.json()).access_token;
+    },
+    async logout() {
+      token = undefined;
+    },
+    async isAuth() {
+      return !!token;
+    },
+    get token() {
+      return token;
+    },
+  };
+}
+
+const unavailableAuthMethod = {
+  async login() {},
+  async logout() {},
+  async isAuth() {
+    return false;
+  },
+  get token() {
+    return undefined;
+  },
+};
+
+async function loadDeployedAuthMethod() {
+  const response = await fetch(`${getBasePath()}/swagger-docs`, {
+    cache: "no-store",
+    headers: { Accept: "application/json" },
+  });
+  if (!response.ok) {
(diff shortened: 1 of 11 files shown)


cda-gui/src/components/AppAuthProvider.jsx
```diff
@@ -0,0 +1,121 @@
...
+  createKeycloakAuthMethod,
+} from "@usace-watermanagement/groundwork-water";
+import PropTypes from "prop-types";
+import { useEffect, useState } from "react";
+
+import { getBasePath } from "../utils/base";
+import { getKeycloakConfig, normalizeOpenIdConnectUrls } from "../utils/auth-config";
+import { AuthConfigurationContext } from "./auth-configuration-context";
+
+function createLocalAuthMethod() {
```
MikeNeilson, line 12: So technically, this only really needs to exist if you're doing `./gradlew run`, with the docker compose keycloak is started and the runtime API spec returns the information and the user should just be redirected as normal. Also allows to logout and use a different user if desired.... but we can correct that in a follow up.

cda-gui/src/main.jsx
```diff
@@ -110,12 +62,14 @@ const router = createBrowserRouter(
...
-    <QueryClientProvider client={queryClient}>
-      <AuthProvider method={authMethod}>
-        <LinkProvider component={Link} hrefMap="to">
-          <RouterProvider router={router} />
-        </LinkProvider>
-      </AuthProvider>
-    </QueryClientProvider>
+    <GlobalErrorBoundary>
+      <QueryClientProvider client={queryClient}>
+        <AppAuthProvider>
```
krowvin, line 67: This delays auth setup until React has fully started. We need to fetch the swagger docs to know the openid host/realm/client/etc to then decide our auth methods for swagger. 

Trying to fetch this broke when the VITE_AUTH_HOST had not been set and caused a blank screen before anything had even rendered. 

Now if auth fails for some reason you will see "sign-in unavailable" instead of a blank page

cda-gui/src/components/AppAuthProvider.jsx
MikeNeilson: ... actually in the `./gradlew run` case, keycloak might not be available anyways so this should really just be more hacky with an apikey so it can actually work.

MikeNeilson (dismissed):

cda-gui/src/components/AppAuthProvider.jsx
krowvin: Tracked this in #1866.

krowvin:
I attempted to close and re-open this to retrigger the github actions that were stuck. 

Ended up sorting an import in a file to retrigger the actions (committed)

MikeNeilson:
> I attempted to close and re-open this to retrigger the github actions that were stuck.
> 
> Ended up sorting an import in a file to retrigger the actions (committed)

In the future don't do that, we can manually trigger rerunning of failed actions.

krowvin:
> in future don't do that, we can manually trigger rerunning of failed actions.


I typically wouldn't but was hoping to get around the GitHub actions issues and thought i'd try it. 

The manual retrigger was not working (due to the outage)

MikeNeilson (approved):

PR #1870, opened 2026-08-07
Harden OpenID retry initialization

krowvin:
﻿## Summary

Follow-up hardening for #1859's OpenID discovery retry and local Compose setup.

## Changes and rationale

- Treat OpenID as unable to authenticate until its configuration is ready, preventing bearer requests from dereferencing a missing parser during the retry window.
- Skip retry scheduling when no well-known URL is configured and stop scheduling after successful initialization, avoiding rejected executor work and unnecessary polling.
- Pass `APP_PORT` into the data-api container, keeping its local authentication proxy aligned with the discovery URL and Keycloak when a nondefault port is used.
- Add regression coverage for the intentionally disabled no-URL configuration.

Per review feedback, this follow-up no longer changes `Authenticator.java`.

## Validation

- `./gradlew :cwms-data-api:test --tests '*OpenIDConfigTest' :cwms-data-api:checkstyleMain :cwms-data-api:checkstyleTest`
- `docker compose config` with `APP_PORT=9090`
- `git diff --check`

The focused test and Checkstyle tasks pass. Checkstyle reports the repository's existing warning-level violations.


cwms-data-api/src/main/java/cwms/cda/security/Authenticator.java
MikeNeilson, line 1: this one is overkill, the rest do seem reasoanble.

MikeNeilson (approved):

PR #1871, opened 2026-08-07
fix: return 200 for CDA GUI client routes

### LINKED ISSUE #1687: Swagger page returns 404 but loads anyway
**Describe the bug**
The swagger page returns status code 404 on refresh, but still loads contents. This breaks reverse proxies.

**Expected behavior**
Page status code should not be 404

**To Reproduce**

1. Navigate to a CDA instance that's using vite/react, such as https://cwms-data.usace.army.mil/cwms-data
2. Open Developer console, go to Network tab
3. Click on API Docs -> Swagger UI
4. Notice page loads fine (aside from possible CORS issues.
5. Refresh the page (F5)
6. Notice the /swagger-ui page in the network tab reports 404
7. Page contents load normally.

Note: It also happens if you directly paste the swagger URL (https://cwms-data.usace.army.mil/cwms-data/swagger-ui) into the address bar, but it's more difficult to see due to needing the developer console.

At the top of any instance of CDA you will see a version. Please provide that version.

- Version 2026.03.16 

**Additional context**
When CDA is behind a reverse proxy, the proxy detects the 404 and returns a standard error page, instead of the supplied contents. So it just says "Document not found."

Image

Image

krowvin:
I had been sitting on this branch for a minute and thought I would go ahead and PR it. 

See Daniel's reported issue for details
- #1687

~~Infra was setup i believe to support having CDA's UI and API being split.~~

~~instead of~~
~~https://cwms-data.usace.army.mil/cwms-data being the UI and API~~

~~https://cwms-data.usace.army.mil -> UI~~
~~https://cwms-data.usace.army.mil/cwms-data -> API~~

Additional comment added below.

## Summary

- return HTTP 200 when known CDA GUI routes are served through the React client-side routing fallback
- preserve the existing 404 status for unknown paths and non-page requests
- support alternate CDA deployment context paths and trailing-slash navigation

## Root Cause

Tomcat routes missing static resources through the `404` error-page mapping to `index.html`. The React app therefore renders successfully, but the original 404 status remains on the response. Reverse proxies and web crawlers treat that successful page as missing.

The new error-dispatch filter changes the status only for known GUI `GET` and `HEAD` routes. API and unknown-route 404 behavior is left intact.

## Related Issue

Closes #1687

## Validation

- `./gradlew :cwms-data-api:test --tests 'cwms.cda.servlet.SpaErrorStatusFilterTest'` — 13 passed
- `./gradlew :cwms-data-api:checkstyleMain :cwms-data-api:checkstyleTest :cwms-data-api:test --tests 'cwms.cda.servlet.SpaErrorStatusFilterTest'` — passed; the new files have no Checkstyle findings
- `./gradlew build` — compilation, GUI build, WAR packaging, and 745 tests passed; the test task failed on 9 existing Mockito failures under local JDK 21 while mocking generated/final classes
- Integration tests were not run because this routing change does not require Oracle/CWMS infrastructure

## Checklist

- [x] AI tools used

### DIFF
--- a/cwms-data-api/src/main/java/cwms/cda/servlet/SpaErrorStatusFilter.java
+++ b/cwms-data-api/src/main/java/cwms/cda/servlet/SpaErrorStatusFilter.java
@@ -1,0 +1,74 @@
+package cwms.cda.servlet;
+
+import java.io.IOException;
+import java.util.Set;
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.RequestDispatcher;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Converts the error dispatch used to load known client-side routes into a successful response.
+ */
+public final class SpaErrorStatusFilter implements Filter {
+
+    // Keep these paths synchronized with cda-gui/src/route-paths.js.
+    private static final Set<String> SPA_ROUTES = Set.of(
+        "/data-query",
+        "/filter-expressions",
+        "/legacy-format",
+        "/location-search",
+        "/regexp",
+        "/swagger-ui",
+        "/timestamps",
+        "/user-lists"
+    );
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        HttpServletRequest httpRequest = (HttpServletRequest)request;
+        HttpServletResponse httpResponse = (HttpServletResponse)response;
+
+        if (isClientRoute(httpRequest)) {
+            httpResponse.setStatus(HttpServletResponse.SC_OK);
+        }
+
+        chain.doFilter(request, response);
+    }
+
+    private boolean isClientRoute(HttpServletRequest request) {
+        String method = request.getMethod();
+        if (!"GET".equalsIgnoreCase(method) && !"HEAD".equalsIgnoreCase(method)) {
+            return false;
+        }
+
+        Object errorRequestUri = request.getAttribute(RequestDispatcher.ERROR_REQUEST_URI);
+        if (!(errorRequestUri instanceof String)) {
+            return false;
+        }
+
+        String path = removeContextPath((String)errorRequestUri, request.getContextPath());
+        if (path.length() > 1 && path.endsWith("/")) {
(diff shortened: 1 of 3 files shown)


cwms-data-api/src/main/java/cwms/cda/servlet/SpaErrorStatusFilter.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Converts the error dispatch used to load known client-side routes into a successful response.
+ */
+public final class SpaErrorStatusFilter implements Filter {
```
MikeNeilson, line 17: Add `@WebFilter` (I think, look at the other filters for the right annotation) and remove the web.xml change.

Unless there's a specific reason to configure it in the web.xml instead. That recommendation is likely left over from all the examples that existed before annotations were supported in Java.

MikeNeilson (changes requested):

krowvin:
@MikeNeilson I had initially planned to go in this direction, but I pushed up this older idea for handling the routing before our group had landed on splitting the UI and the API.

Would you be okay with me updating this PR, opening a new PR, or opening a PR against this PR with a new proposed method for splitting the two? One possibility is to produce two independently deployed WARs/applications: one for the UI and one for the API.

That could require districts to install and manage two different WARs. Would that separation be a better option now that we have the new update script, or would the additional deployment be more trouble than it is worth?

MikeNeilson:
> @MikeNeilson I had initially planned to go in this direction, but I pushed up this older idea for handling the routing before our group had landed on splitting the UI and the API.
> 
> Would you be okay with me updating this PR, opening a new PR, or opening a PR against this PR with a new proposed method for splitting the two? One possibility is to produce two independently deployed WARs/applications: one for the UI and one for the API.
> 
> That could require districts to install and manage two different WARs. Would that separation be a better option now that we have the new update script, or would the additional deployment be more trouble than it is worth?

Hmm, might be more trouble than it's worth, but reasonable to investigate. Would want them both as WARs to easy the packaging but we'll see how much of a problem it is in the docker compose setup and tests, oh look at how OpenDCS does it. they are separate there. so yeah not a huge deal.

I think we'll need a WAF change in prod and test, but that's not a huge deal if we know it's coming.

MikeNeilson (approved):

PR #1872, opened 2026-08-07
CDA-98 - Creating /v2/forecast-spec endpoint using updated spec struture

### LINKED ISSUE #1648: Creating Forecast Spec with same id-designator doesn't overwrite location
**Describe the bug**
When creating a Forecast Spec with a specific id-designator and a location id, and then storing that same forecast spec but changing the location id, a new forecast spec is created, rather than overwriting the original one with an updated location id. 

**Expected behavior**
Expected the original forecast spec to be overwritten.

**To Reproduce**

1. Create a forecast spec with a specified spec-id, designator, and assign a location-id. 
2. Store forecast spec. 
3. Change the assigned location-id, leaving everything else the same and re-store. 
4. Get forecast spec with specified spec-id, and designator
5. See that 2 forecast specs are returned. 

**Logs/Incident Identifier**
See failing test that reproduces issue: https://github.com/USACE/cwms-data-api/pull/1647

**CDA Version:**
- 2026.03.19

rma-bryson:
## Summary
Established use of /v2/ with forecast-spec update to use locations with sort-order

## Related Issue

Closes https://github.com/USACE/cwms-data-api/issues/1648

## Validation

Unit tests and Integration tests

## Checklist

- [x] AI tools used


cwms-data-api/src/main/java/cwms/cda/api/v2/ForecastSpecControllerV2.java
```diff
@@ -0,0 +1,181 @@
...
+    @Override
+    protected Class<ForecastSpecV2> getDtoClass() {
+        return ForecastSpecV2.class;
+    }
+
+    @OpenApi(
+            description = "Used to create and save forecast spec data",
+            requestBody = @OpenApiRequestBody(
+                    content = {
+                            @OpenApiContent(from = ForecastSpecV2.class, type = Formats.JSONV1)
```
adamkorynta: shouldn't be JSONV1, should just be JSON

adamkorynta: should we throw an error if the version is specified? or only if version is specified and mismatched?

adamkorynta: I'd like to see a screenshot of the swagger UI

MikeNeilson (commented):
Looks reasonable, though I agree with @adamkorynta 's comments. Good to see a screenshot of the UI.

rma-bryson:
Swagger UI screenshots:

image

image

cwms-data-api/src/main/java/cwms/cda/api/v2/ForecastSpecControllerV2.java
rma-bryson: Updated to just be JSON

cwms-data-api/src/main/java/cwms/cda/api/v2/ForecastSpecControllerV2.java
```diff
@@ -0,0 +1,181 @@
+package cwms.cda.api.v2;
```
adamkorynta: Decision point: add in `forecast` to the packaging

Recommendation:

`cwms.cda.api.forecast` - includes ForecastSpecControllerV2.java‎ and ForecastSpecControllerV1.java‎ (renamed from ForecastSpecController.java‎) and make the AbstractForecastSpecController -> ForecastSpecController

extend this out to the DAO/DTO packaging



Alternatives:
add `v1.forecast` AND `v2.forecast` packaging (not very helpful if packaging is encapsulation focused)
add `forecast.v1` AND `forecast.v2` packaging (causes package explosion)

cwms-data-api/src/test/java/cwms/cda/api/ForecastSpecControllerV2TestIT.java
```diff
@@ -0,0 +1,899 @@
...
+import static cwms.cda.security.ApiKeyIdentityProvider.AUTH_HEADER;
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.isEmptyOrNullString;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+@Tag("integration")
+@MinimumSchema(260716)
+final class ForecastSpecControllerV2TestIT extends DataApiTestIT {
```
adamkorynta, line 42: @rma-bryson and I discussed the code duplication between the two integration tests. We settled that this is fine as the maintenance burden is not likely to be high, but the level of abstraction needed (or extending parameterized tests to matrices) is not necessary

MikeNeilson: Yeah, if we get to a v3 maybe. But some duplication is expected in test code.

cwms-data-api/src/main/java/cwms/cda/api/v2/ForecastSpecControllerV2.java
rma-bryson: updated controller naming and removed v2 package

cwms-data-api/src/main/java/cwms/cda/api/forecast/ForecastSpecControllerV2.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @OpenApi(
+            description = "Used to delete forecast spec data based on unique fields",
+            pathParams = {
+                    @OpenApiParam(name = NAME, required = true, description = "Specifies the "
+                            + "spec id of the forecast spec whose data is to be deleted."),
+            },
+            queryParams = {
+                    @OpenApiParam(name = OFFICE, required = true, description = "Specifies the "
```
adamkorynta: should we take the opportunity of the v2 to move office into a path param? 

`/v2/<office>/forecast/<name>`
or
`/v2/forecast/<office>/<name>`

We started doing this on some other endpoints

rma-bryson: Proposed:

Office filtering is not supported on the primary resource identified by the endpoint path, as the office context is already established by the route. If multi-office is needed (which I don't believe is often the case), then subsequent requests for each office can be made. Office-based filters are only supported for related resources and are expressed as query parameters, but not on the primary resource. 

Example:

`/v2/SWT/forecast-spec`

Example of filtering on related resource:

`/v2/SWT/location-group?category-office=CWMS`

MikeNeilson: I think we should prefer `/v2/forecast/<office>/<name>`, as it reads like "I want the forecast, for office X, named Y. where as `/v2/<office>/forecast/<name>` reads "I want, for office X, the forecast named Y.... which now that I type out I'm not actually sure which one is better as they both read reasonably well.

rma-bryson: Updated to include /v2/forecast-spec/{office}/{name}

rma-bryson:
Screenshot with office in path param: 

image

rma-bryson:
Need to fix failing OpenApi static analysis test - related to not properly detecting how office path param is being used.

cwms-data-api/src/main/java/cwms/cda/api/forecast/ForecastSpecControllerV2.java
```diff
@@ -175,6 +187,7 @@ public void getOne(@NotNull Context ctx, @NotNull String name) {
     )
     @Override
     public void update(@NotNull Context ctx, @NotNull String name) {
+        logUnusedPathParameter(ctx, OFFICE, "Body contains information");
```
adamkorynta: should this sanity check?

adamkorynta (approved):

PR #1878, opened 2026-08-12
Add reproducible development container

krowvin:
## Summary

- add a locked Dev Container configuration for Java 11, Node.js 22/npm 11, Python 3.13, and Docker Compose
- forward the CDA, Compose, Vite, and Sphinx development ports
- support Testcontainers and the existing Compose stack through the host Docker socket
- normalize Git behavior for Windows Docker Desktop bind mounts without changing the shared repository configuration
- document local VS Code and GitHub Codespaces usage

## Developer impact

A fresh clone can be reopened in a container or Codespace and use the repository's Gradle, frontend, documentation, Docker Compose, and Docker-backed test workflows without installing those toolchains directly on the host.

## Validation

- `npx @devcontainers/cli up --workspace-folder <standalone-clean-clone> --remove-existing-container`
- verified Java 11, Node.js 22, npm 11, Python 3.13, Docker client/server, and Docker Compose inside the container
- `docker run --rm alpine:3.21 echo docker-socket-ok`
- `docker compose config --quiet`
- `./gradlew :docs:buildDocs --no-daemon`
- `./gradlew build --no-daemon` compiled Java, built the production GUI and WAR, ran Checkstyle, and completed 743 tests: 741 passed, 37 skipped, and two existing `CwmsDTOValidatorTest` performance assertions exceeded their 150/200 ms thresholds on the Windows bind mount
- `./gradlew :cwms-data-api:test --tests '*CwmsDTOValidatorTest' --no-daemon` passed on the Windows host baseline; the container-only failures are limited to those timing thresholds
- verified npm 11 leaves `cda-gui/package-lock.json` unchanged after `npm install`
- `git diff --check`

Integration tests were not run because they provision the Oracle/CWMS test infrastructure and are separate from the default build.

## Checklist

- [x] AI tools used

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,58 @@
+{
+  "name": "CWMS Data API",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-24.04",
+  "features": {
+    "ghcr.io/devcontainers/features/java:1": {
+      "version": "11",
+      "jdkDistro": "ms"
+    },
+    "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {
+      "version": "latest",
+      "moby": true,
+      "dockerDashComposeVersion": "v2"
+    },
+    "ghcr.io/devcontainers/features/node:1": {
+      "version": "22"
+    },
+    "ghcr.io/devcontainers/features/python:1": {
+      "version": "3.13",
+      "installTools": false
+    }
+  },
+  "containerEnv": {
+    "TESTCONTAINERS_HOST_OVERRIDE": "host.docker.internal"
+  },
+  "forwardPorts": [7000, 8081, 5173, 8000],
+  "portsAttributes": {
+    "7000": {
+      "label": "CDA Gradle server"
+    },
+    "8081": {
+      "label": "CDA Docker Compose stack"
+    },
+    "5173": {
+      "label": "CDA GUI development server"
+    },
+    "8000": {
+      "label": "Sphinx documentation server"
+    }
+  },
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "vscjava.vscode-java-pack",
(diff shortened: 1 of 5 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,58 @@
...
+      "extensions": [
+        "vscjava.vscode-java-pack",
+        "vscjava.vscode-gradle",
+        "dbaeumer.vscode-eslint",
+        "esbenp.prettier-vscode",
+        "ms-python.python"
+      ]
+    }
+  },
+  "hostRequirements": {
```
krowvin, line 51: @dnezamfar @MikeNeilson 

Any ideas how codespaces would interact with this if a user defines a codespace smaller?

MikeNeilson: i would presume it would just fail with an error.

README.md
```diff
@@ -76,6 +76,30 @@ End user documentation available here: [📕 Read the Docs](https://cwms-data-ap
...
+2. Open the cloned repository in Visual Studio Code.
+3. Run **Dev Containers: Reopen in Container** from the command palette.
+
+GitHub Codespaces also detects the same configuration automatically. The first container creation
+downloads the development image and initializes the Gradle wrapper. After it completes, run the
+normal project commands from the container terminal, for example:
+
+```bash
+./gradlew build
+docker compose up -d --force-recreate
```
MikeNeilson, line 97: Okay, just so I understand the setup correctly, instead of going for a full "you are fully ready to poke at things "here's the URL", this basically just treats the "dev-container" as a virtual machine (and thus remote target for vs-code) and then the developer has to perform the appropriate commands as though they were on the native host?

MikeNeilson: Confirmed on a call.

MikeNeilson (approved):

PR #1884, opened 2026-08-18
Actually output the trace context in the docker image.

MikeNeilson:
## Summary

While working on https://github.com/HydrologicEngineeringCenter/cwms-python/pull/305, discovered that while we process the traceparent header, we never actually output the trace_id and span_id for viewing and analysis in the logs.

## Related Issue

Closes #

## Validation

Describe how this was tested.

## Checklist

- [ ] AI tools used

### DIFF
--- a/cwms-data-api/src/docker/logback-juli.xml
+++ b/cwms-data-api/src/docker/logback-juli.xml
@@ -19,7 +19,7 @@
     <logger name="org.apache" level="ERROR"/>
 
     <logger name="org.apache.catalina" level="INFO" additivity="false">
-        <appender-ref ref="STDERR" />
+        <appender-ref ref="OTEL-STDERR" />
     </logger>
 
     <logger name="org.apache.catalina.core.ContainerBase.[Catalina]" level="INFO" additivity="false">
--- a/cwms-data-api/src/docker/logback-juli.xml
+++ b/cwms-data-api/src/docker/logback-juli.xml
@@ -3,13 +3,17 @@
 <configuration>
     <import class="org.apache.juli.logging.ch.qos.logback.classic.encoder.JsonEncoder"/>
     <import class="org.apache.juli.logging.ch.qos.logback.core.ConsoleAppender"/>
-
     <variable name="LOG_LEVEL" value="${LOG_LEVEL:-INFO}"/>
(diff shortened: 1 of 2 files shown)


cwms-data-api/src/docker/logback-juli.xml
oskarhurst: should this also be OTEL-STDERR ?

cwms-data-api/src/docker/logback-juli.xml
```diff
@@ -3,13 +3,17 @@
...
-
     <variable name="LOG_LEVEL" value="${LOG_LEVEL:-INFO}"/>
     <contextName>CWMS-Data-API</contextName>
     <appender name="STDERR" class="ConsoleAppender">
         <encoder class="JsonEncoder"/>
         <target>System.err</target>
-    </appender>    
+    </appender>
+
+    <appender name="OTEL-STDERR" class="io.opentelemetry.instrumentation.logback.mdc.v1_0.OpenTelemetryAppender">
```
oskarhurst, line 13: Correct me if I am wrong but isn't OpenTelemetryAppender unshaded so it could never resolve and use the shaded classes provided by juli. Also isn't locgback-juli only run inside of tomcat would would never have any cda trace-ids to begin with so we only need the change in the normal logback file to add the traces ids?

oskarhurst (changes requested):
I think you changes to Juli might break tomcats logging

cwms-data-api/src/docker/logback-juli.xml
MikeNeilson: For anyone reading this PR:

The tomcat-juli jar is removed from the image and replaced with tomcat9-slf4j-logback, see https://github.com/USACE/cwms-data-api/blob/a2be20c1dc064f4b7bb35cae8ca6fd48b487a84e/Dockerfile#L43 (roughly around there, may have moved a few lines by time of reading).

Thus, no issues with any shading as that jar deals with it for us.

MikeNeilson:
> I think you changes to Juli might break tomcats logging

Sorry, missed this original comment. Doesn't so much break it as totally replaces it.

oskarhurst (approved):

PR #1891, opened 2026-08-21
Bugfix: Added defaults to optional query parameters for water contract creation

zack-rma:
## Summary

Added defaults to optional query parameters for water contract creation. Previously had optional query parameters parsed into primitive booleans, which would break when parameters were not provided.

## Checklist

- [ ] AI tools used


oskarhurst (changes requested):
Hi the fixes look good but can you also add these same fixes to some other places in the code where the same error still exists, just so we can make sure to knock them all out at once. specifically 

WaterUserCreateController.java:85
boolean failIfExists = Boolean.parseBoolean(ctx.queryParam(FAIL_IF_EXISTS));

WaterContractDeleteController.java:85
WaterUserDeleteController.java:84
DeleteMethod deleteMethod = getDeleteMethod(ctx.queryParam(METHOD));

oskarhurst (approved):

rma-bryson (approved):

Issue #1783, opened 2026-06-15
POST time series data 500 error instead of "Incoming data has multiple values for same minute"


inguyen314:
When submitting a POST request for time series data, duplicate values within the same minute currently return a generic 500 Internal Server Error. Instead, the web console should display a more descriptive message:

ERROR: Incoming data has multiple values for the same minute.

`cwms.cda.data.dao.AuthDao.prepareContextWithUser Validated Api Key for user=[CONTEXT ratelimit_period="5 SECONDS [skipped: 10]" ]
15-Jun-2026 14:59:09.840 SEVERE [ajp-nio-127.0.0.1-8009-exec-438] cwms.cda.api.TimeSeriesController.create -8388787327655537121: Internal Error
	org.jooq.exception.DataAccessException: SQL [{ call "CWMS_20"."CWMS_TS"."ZSTORE_TS" (?, ?, ?, ?, ?, cast(? as date), ?, ?) }]; ORA-20998: ERROR: ORA-06512: at "CWMS_20.CWMS_ERR", line 80
ORA-06512: at "CWMS_20.CWMS_TS", line 5808
ORA-06512: at "CWMS_20.CWMS_ERR", line 80
ORA-06512: at "CWMS_20.CWMS_TS", line 7096
ORA-20998: ERROR: Incoming data has multiple values for same minute.
ORA-06512: at "CWMS_20.CWMS_ERR", line 80
ORA-06512: at "CWMS_20.CWMS_TS", line 5808
ORA-06512: at "CWMS_20.CWMS_TS", line 5459
ORA-06512: at "CWMS_20.CWMS_TS", line 9633`

`{
  "message": "Internal Error",
  "incidentIdentifier": "89753708-3291-4680-bc3c-ea312b1a7920",
  "source": "Unknown",
  "details": {}
}`

`🔥 Error during API processing: Error: HTTP 500: {"message":"Internal Error","incidentIdentifier":"-8064202843547236245","details":{}}
    at createTS (cwms-data-upload.js:348:31)
    at async saveBtn.onclick (cwms-data-upload.js:427:21)`

krowvin:
Possibly return a 409 or 422?

MikeNeilson:
Some updates were done on error handling recently, please try it again in dev.

rma-bryson:
Confirmed this now returns a 400 error code

PR #1899, opened 2026-08-25
docs: propose duplicate timeseries value handling

krowvin:
## Summary

Adds ADR 0014 defining how CDA time-series write requests should handle multiple values that resolve to the same effective CWMS storage timestamp.

The ADR proposes a `use-if-multiple` API parameter with four strategies:

- `error` — reject the request; this is the default
- `first` — retain the first record
- `last` — retain the last record
- `average` — use the arithmetic mean of non-null values

The `cwms-cli` `csv2cwms` command already defines these options and handles them client-side. With this ADR, CDA can enforce the behavior at the backend, and `csv2cwms` can be updated to expose the methods provided by the API instead of resolving duplicates independently. 
i.e.
https://cwms-cli.readthedocs.io/en/latest/cli/csv2cwms_complete_config.html#use-if-multiple

The ADR distinguishes collisions at the effective CWMS storage timestamp from optional rounding or bucketing into a time-series interval. It also records follow-up questions about interval bucketing, averaged quality codes, future data-entry-date handling, and resolution reporting.

This is an ADR-only pull request and makes no code changes.

## Related Issue

Related to #1783

## Validation

- `./gradlew :docs:buildDocs`
- Confirmed the branch differs from `develop` only by the ADR and decision-index entry.
- Integration tests were not run because this PR contains documentation only.

## Checklist

- [x] AI tools used

### DIFF
--- a/docs/source/decisions/0014-duplicate-timeseries-values.rst
+++ b/docs/source/decisions/0014-duplicate-timeseries-values.rst
@@ -1,0 +1,124 @@
+##############################################
+Duplicate Time-Series Values in Write Requests
+##############################################
+
+
+Summary
+========
+
+This ADR proposes a consistent API policy for handling multiple time-series
+values that resolve to the same CWMS storage minute in a single write request.
+The time-series write endpoints will expose a ``use-if-multiple`` query
+parameter with four strategies: ``error``, ``first``, ``last``, and
+``average``. The default will be ``error``.
+
+
+Context
+========
+
+CWMS stores time-series timestamps at minute precision. A request can therefore
+contain records that have identical timestamps or distinct sub-minute
+timestamps that resolve to the same storage minute. Passing those records to
+the database without first resolving the collision causes the write to fail.
+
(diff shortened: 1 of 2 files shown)


MikeNeilson (commented):
This looks reasonable to me. We'll give time for others to comment.

MikeNeilson:
@rma-bryson @rma-psmorris @perrymanmd @krowvin 

Well, I thought this was merged in. but it looks like our numbering has gotten somewhat... out of hand.

rma-bryson:
> @rma-bryson @rma-psmorris @perrymanmd @krowvin
> 
> Well, I thought this was merged in. but it looks like our numbering has gotten somewhat... out of hand.

There's been a few times where ADR's are being worked on concurrently so they get the same index. That should be easily identifiable when conflicts get merged in the index.rst. Though I'm not sure exactly what purpose the enumeration in the file-name serves - is that just to see the history of when an adr was added, or it it required by some process?

PR #1919, opened 2026-09-02
Group user-list endpoints under User Management in Swagger

krowvin:
## Summary

- keep the CDA header navigation unchanged
- force every emitted `/user/list*` OpenAPI operation into the existing `User Management` tag
- add regression coverage for user-list paths and verify unrelated paths retain their existing tags

## Cause

When the `USER_LISTS` feature is disabled, CDA registers generic unsupported-feature handlers for these routes. Those fallback handlers do not carry the controllers' OpenAPI annotations, so Swagger placed the operations in `default`. Applying the tag in the existing OpenAPI response processor keeps the generated specification correct in either feature state without changing endpoint behavior.

## Validation

- `:cwms-data-api:test --tests cwms.cda.ApiServletTest` passed
- `OPEN_API_TEST=true :cwms-data-api:test --tests cwms.cda.api.OpenApiDocTest` passed
- `:cwms-data-api:checkstyleMain :cwms-data-api:checkstyleTest` completed successfully (existing repository warnings remain)
- net PR diff confirms no CDA GUI changes remain

## Checklist

- [x] AI tools used

### DIFF
--- a/cda-gui/src/links/header-links.js
+++ b/cda-gui/src/links/header-links.js
@@ -39,16 +39,9 @@ export default [
     ],
   },
   {
-    id: "user-management",
-    text: "User Management",
+    id: "user-lists",
+    text: "User Lists",
     href: "/user-lists",
-    children: [
-      {
-        id: "user-lists",
-        text: "User Lists",
-        href: "/user-lists",
-      },
-    ],
   },
   {
     id: "help",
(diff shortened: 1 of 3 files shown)


MikeNeilson (approved):
Not idea, but tying the enable/disable bits into the spec filter as a whole is a larger issue.

PR #1931, opened 2026-09-08
CDA-130 - Adds support for patial patches. Implements for text-timeseries

### LINKED ISSUE #1715: Standard Text Timeseries PATCH requires FULL write to update
There is no endpoint for PATCHING text timeseries. You would end up POST the same value. But ideally we are doing a PATCH because the field already exists and we want to distinguish at the METHOD what is being done to the data.

Create a PATCH endpoint that allows you to do things like change one ID to another and also be able to update the standard-text string optionally.

Continue to use the POST endpoint and set `fail-if-exists` to `False`

@inguyen314

rma-bryson:
## Summary

Establishes support for partial patch with text timeseries implementation.

## Related Issue

Closes https://github.com/USACE/cwms-data-api/issues/1715

## Validation

Integration and unit tests

## Checklist

- [x] AI tools used


cwms-data-api/src/main/java/cwms/cda/api/texttimeseries/TextTimeSeriesControllerV2.java
```diff
@@ -0,0 +1,225 @@
...
+            queryParams = {
+                    @OpenApiParam(name = REPLACE_ALL, type = Boolean.class, description = "Whether to "
+                            + "replace any and all existing text with the specified text for matching entries. "
+                            + "Default is " + DEFAULT_CREATE_REPLACE_ALL)},
+            method = HttpMethod.POST,
+            tags = {TAG}
+    )
+    @Override
+    public void create(@NotNull Context ctx) {
+        super.create(ctx);
```
rma-bryson, line 125: @MikeNeilson - the create for v2 does the same thing as previous behavior where it stores regardless of if the data exists or not (essentially acting as both a create and an update). Wondering if we want to start enforcing POST is only for non-existing data, PATCH should be used for data that already exists in v2 and beyond?

MikeNeilson: @krowvin what do you think?

It's certainly pedantically correct. And would save on the "fail-if-exists" query parameters.

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/RegularTextTimeSeriesRow.java
```diff
@@ -1,19 +1,28 @@
...
 
 @JsonDeserialize(builder = RegularTextTimeSeriesRow.Builder.class)
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonNaming(PropertyNamingStrategies.KebabCaseStrategy.class)
-public class RegularTextTimeSeriesRow implements TextTimeSeriesRow {
+public class RegularTextTimeSeriesRow extends CwmsDTOBase implements TextTimeSeriesRow {
 
+    // Together, date-time and data-entry-date are this row's identity (see Identifier): two rows
+    // can share the same date-time and are only distinguished by their data-entry-date, so a
+    // PATCH MERGE match requires both to agree -- not date-time alone.
```
MikeNeilson, line 21: Wouldn't that be ts_code (which yes, isn't here) dateTime and versionDate (though I'm not sure if the text time series properly has those.)

cwms-data-api/src/main/java/cwms/cda/formatters/annotations/Identifier.java
```diff
@@ -0,0 +1,11 @@
+package cwms.cda.formatters.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
```
MikeNeilson, line 7: Add some javadoc here explaining the purpose more for future developers.

cwms-data-api/src/main/java/cwms/cda/data/dto/texttimeseries/RegularTextTimeSeriesRow.java
MikeNeilson: wait, I see why you need to do this.

docs/source/decisions/0011-patch.rst
```diff
@@ -1,280 +0,0 @@
...
-     - Renames a water user.
-   * - /projects/{office}/{project-id}/water-users/{water-user}/contracts/{contract-name}
-     - WaterContractUpdateController
-     - Rename-only
-     - Renames a water contract.
-
-Decision Status
-===============
-
-(Status: accepted)
```
MikeNeilson, line 268: Don't delete decision, change the Status to superseded (or whatever reference text for the we put in the template.)

MikeNeilson (changes requested):

docs/source/decisions/0011-patch.rst
rma-bryson: I kept the decision status the same, the file just got renamed from 0011-patch.rst to 0017-patch.rst. There was already an adr at index 11, so i moved this to 17.

MikeNeilson: Okay, but 0011-patch.rst already exists and is merged into to the default branch.

If the content isn't changing, why is the file getting renamed?
If the content is changing, it's a new decision, old one states and just references the new one.

cwms-data-api/src/main/java/cwms/cda/formatters/annotations/Identifier.java
rma-bryson: Added javadoc

docs/source/decisions/0011-patch.rst
rma-bryson: Updated to keep the 0011-patch.rst with a status pointing at the updated adr. 

So for future reference - we should keep merged adr files as is (with an updated decision status pointing to the superseding adr) if there are additional details to be added?

The main reason renamed this one was because there was already a 0011-queue-messages.rst at index 11, so we had 2 different adrs at the same index.

MikeNeilson: Yeah, but now I see the issue, sorry the duplicate number didn't catch my eye.

Go ahead and change the number here so we aren't duplicating the index.

But future changes to existing decision should keep the old files in place.

If the changes are somewhat significant, create a 17 and 18 for general tracking purposes. (or even if small may make sense to only put the changes in the 2nd one.)

Not going to nitpick further given the mishap with the index (which I just say we also have for 0013) but we should all try to see that better in the future; some sort of merge gate is also in order I think... but that's future work.

rma-bryson: I see the 0013 issue as well. catalogs.rst is also missing from index.rst.

I'll leave both 0011-patch.rst and 0017-patch.rst in place. The original 0011 can remain as the historical decision record, and 0017 can represent the newer collection merge semantics and related changes.

Since index.rst only references the newer ADR, keeping 0011-patch.rst around shouldn't cause any issues and preserves the decision history.

MikeNeilson (approved):

PR #1940, opened 2026-09-10
Add optional display units to time series group assignments

krowvin:
## Summary

Add optional `units` and `unit-system` attributes to assigned time series, including a POST example. Preferences apply across groups and to EN/SI retrieval; explicit physical units retain precedence. Omitting `units` preserves a preference, explicit null clears it, and group responses include only nondefault overrides.

Group writes remain atomic when units are invalid. Existing database schemas continue to support requests without preferences; writes that specify units return HTTP 501 until the database update is installed.

## Related Issue

Related to HydrologicEngineeringCenter/cwms-database#223. Companion database PR: https://github.com/HydrologicEngineeringCenter/cwms-database/pull/227. Deploy that update before enabling preference writes.

## Validation

- Java 11 API service build passed; 744 unit tests passed and 40 skipped.
- Three HTTP tests passed on the updated schema, covering existing group behavior, numeric EN/SI and explicit-unit retrieval, shared preferences, clearing, and invalid POST/PATCH rollback.
- Three HTTP tests passed on the older schema, including rejection of unsupported writes without creating a group.
- Generated OpenAPI confirms optional nullable fields and excludes internal presence tracking. Integration tests use UTC to match the API runtime.
- Aggregate generated-client build did not complete because npm wrappers stalled; the service build and OpenAPI generation passed.

## Checklist

- [x] AI tools used

### DIFF
--- a/cwms-data-api/build.gradle
+++ b/cwms-data-api/build.gradle
@@ -309,6 +309,7 @@ task integrationTests(type: Test) {
 //    dependsOn test
     dependsOn generateConfig
     dependsOn war
+    jvmArgs += "-Duser.timezone=UTC"
 
     useJUnitPlatform() {
         includeTags "integration"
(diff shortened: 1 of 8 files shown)


cwms-data-api/build.gradle
```diff
@@ -309,6 +309,7 @@ task integrationTests(type: Test) {
 //    dependsOn test
     dependsOn generateConfig
     dependsOn war
+    jvmArgs += "-Duser.timezone=UTC"
```
krowvin, line 312: This is sneaking in, but does it mean that all hosts running it would have been local time before?

PR #1923, opened 2026-09-03
Add API key management page for signed-in users

krowvin:
Adds a signed-in API Keys page and guide for creating, viewing, rotating, and revoking keys without Swagger. Keys belong to the current user and inherit that user's permissions. Secrets are shown once and kept out of browser storage and query caches. Rotation creates a replacement before offering to revoke the original.

- Merge current `develop`, including #1935, while retaining the Swagger links to the key-management guide.
- Add dismissible toasts for create, copy, refresh, revoke, rotation, expired keys, and request failures. Toasts remain accessible inside dialogs; errors also stay beside the form. Success feedback expires after eight seconds, while errors and warnings remain dismissible.
- Translate the new endpoint validation responses into useful name/date/payload guidance; handle duplicate-name 409 responses and empty 204 revocation responses. Invalid local dates are rejected before submission.
- Keep replacement names within 64 characters and clear stale key details after refresh or a 404. Failed creation never revokes the original; failed or cancelled revocation preserves both keys.

- Show a permission warning page on denied key access, listing `CWMS Users` and a signed-in `cac_auth` session, highlighting recognized missing roles, and directing users to their CWMS Admin. Management controls are hidden until access is restored.

- Emphasize in bold that keys are shown only at creation and cannot be retrieved again, in the create/save dialogs and help/rotation instructions. Companion API response and Swagger clarification: #1941.

Validation:

- 14 adapter/auth unit tests, 10 Chromium browser tests, Storybook test, lint, formatting, and production build passed. The API-key tests now run in the web GUI CI workflow.
- Real local CDA/Oracle/Keycloak checks against the #1935 backend verified duplicate-name conflict feedback, retry in the same dialog, creation, expiration metadata refresh, one-time-secret closing, and 204 revocation feedback. Temporary keys were removed; secrets were excluded from logs and screenshots.

- Previous full Java 11 `gradlew build` passed, including 786 service unit tests and generated OpenAPI/TypeScript checks.

## Screenshots

These screenshots show the earlier layout before the latest wording and styling refinements. They are embedded from an earlier commit; screenshot files are excluded from the PR changes.

API Keys page

API key guide

The IP address shown in the curl example is generated dynamically from the current host and changes with the deployment. This screenshot was taken on a local development host.

<details>

<summary>Mobile views</summary>

API Keys mobile

API key guide mobile

</details>

One-time-key guidance validation: 14 unit tests, 10 Chromium browser tests, lint, formatting, and production build passed.

AI tools used

PR #1941, opened 2026-09-11
Omit API key secrets from GET responses and clarify one-time retrieval

krowvin:
GET `/auth/keys` and `/auth/keys/{key-name}` currently return `"api-key": null`, which suggests the secret should be available. Return only key metadata and remove `api-key` from both HTTP 200 response schemas. POST continues to return the generated secret in its HTTP 201 response.

- Add a bold notice to the POST and both GET Swagger panels: save the key at creation; it cannot be retrieved again.
- Document the creation-only secret field and preserve user, name, creation, and expiration metadata.
- Cover serialization and generated OpenAPI contracts; strengthen integration assertions to distinguish an absent property from null.

Companion UI guidance is in #1923, with bold warnings in the creation, save, help, and rotation instructions.

Validation: Java 11 service tests passed (745 passed, 40 skipped), including the three new serialization/OpenAPI regression tests; Checkstyle and WAR assembly completed. Rendered Swagger panels verified both bold GET notices and metadata-only examples. The full Gradle build reached the Docker-backed OpenAPI export, which failed because no Docker environment is available; Oracle integration and downstream generated-client validation remain unverified.

Swagger screenshots were captured outside the source tree. Attachment was blocked: the browser file chooser timed out, and native computer use stopped because it could not verify the Chrome URL.

AI tools used


cwms-data-api/src/main/java/cwms/cda/data/dto/auth/ApiKeyMetadata.java
```diff
@@ -0,0 +1,50 @@
...
+    private final String userId;
+    private final String keyName;
+    private final ZonedDateTime created;
+    private final ZonedDateTime expires;
+
+    /**
+     * Copy key information without retaining the secret.
+     * @param key key returned by the DAO
+     */
+    public ApiKeyMetadata(ApiKey key) {
```
MikeNeilson: seems like the DAO should just return ApiKeyMetaData directly instead of pulling the full key out?

MikeNeilson (changes requested):

cwms-data-api/src/main/java/cwms/cda/data/dto/auth/ApiKeyMetadata.java
krowvin: Good catch, that would be cleaner! Fixed

MikeNeilson (approved):

PR #1958, opened 2026-09-18
CDA-126 - Updates project locks to v2 using cwms-ids for office and project id.

### LINKED ISSUE #386: Fix existing useages of <data>Id
while going through various tests and code I realized a few places still use a <something>Id instead of name field.
need to review if correct for given data set (like the geo json) or if we can make them more inline with the current direction.

rma-bryson:
## Summary

Updates ProjectLock to use CwmsId in v2

## Related Issue

Closes https://github.com/USACE/cwms-data-api/issues/386

## Validation

Unit and Integration tests

## Checklist

- [ ] AI tools used


rma-bryson:
image

image

MikeNeilson (approved):

